PR #4, opened 2025-02-05
feat: Flashblock RPCs

cody-wang-cb:
Added support for 3 RPCs
- `eth_getBlockByNumber` with pending tag.
- `eth_getTransactionReceipt`, it checks if it exists in the DB first, if not, then maybe it's a flashblock receipt then it checks in the cache
- `eth_getBalance` with pending tag.

It runs a websocket client and listens to the rollup-boost web proxy. It parses the message and store the necessary information into the cache.

For message out of order scenarios, currently only handles to not updating older blocks, but doesn't handle cases like same block payload message out of order, will be TODO for future updates. 

Added unit and integration tests, supports reth 1.2.2.


src/rpc.rs
```diff
@@ -16,25 +39,179 @@ pub trait EthApiOverride {
...
+
+        return RpcBlock::<Optimism> {
+            header: Header::from_consensus(header.seal_slow(), None, None),
+            transactions: BlockTransactions::Full(converted_txs),
+            uncles: Vec::new(),
+            withdrawals: None,
+        };
+    }
+
+    pub fn transform_tx(
```
cody-wang-cb, line 102: implementation similar to https://github.com/paradigmxyz/reth/blob/main/crates/optimism/rpc/src/eth/transaction.rs#L84

src/rpc.rs
```diff
@@ -16,25 +39,179 @@ pub trait EthApiOverride {
...
+                transaction_index,
+                from,
+                effective_gas_price: Some(effective_gas_price),
+            },
+            deposit_nonce,
+            deposit_receipt_version,
+        }
+    }
+
+    pub fn transform_receipt(
```
cody-wang-cb, line 159: implementation similar to https://github.com/paradigmxyz/reth/blob/main/crates/optimism/rpc/src/eth/receipt.rs#L26

src/rpc.rs
```diff
@@ -16,25 +39,179 @@ pub trait EthApiOverride {
...
 }
 
 impl<E> EthApiExt<E> {
-    pub const fn new(eth_api: E) -> Self {
-        Self { eth_api }
+    pub const fn new(eth_api: E, cache: Arc<Cache>) -> Self {
+        Self { eth_api, cache }
+    }
+
+    pub fn transform_block(&self, block: OpBlock) -> RpcBlock<Optimism> {
```
cody-wang-cb: implementation similar to https://github.com/paradigmxyz/reth/blob/v1.1.5/crates/rpc/rpc-types-compat/src/block.rs#L84

src/flashblocks.rs
```diff
@@ -0,0 +1,163 @@
+use crate::cache::Cache;
```
cody-wang-cb, line 1: websocket implementation similar to https://github.com/flashbots/rollup-boost/blob/flashblocks/src/flashblocks.rs

src/main.rs
```diff
@@ -34,14 +55,32 @@ fn main() {
...
-                        .with_memory_block_buffer_target(rollup_args.memory_block_buffer_target);
+                    let engine_tree_config = TreeConfig::default();
+                    // .with_persistence_threshold(
+                    //     flashblocks_rollup_args.rollup_args.persistence_threshold,
+                    // )
+                    // .with_memory_block_buffer_target(
+                    //     flashblocks_rollup_args
+                    //         .rollup_args
+                    //         .memory_block_buffer_target,
+                    // );
```
danyalprout: Why did we remove this?

src/flashblocks.rs
```diff
@@ -0,0 +1,304 @@
...
+                                _ => {} // Handle other message types if needed
+                            }
+                        }
+                    }
+                    Err(e) => {
+                        error!(
+                            message = "WebSocket connection error, retrying in 5 seconds",
+                            error = %e
+                        );
+                        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
danyalprout: nit: we should probably do a dynamic backoff vs. 5s everytime

src/flashblocks.rs
```diff
@@ -0,0 +1,304 @@
...
+                                    let text = match String::from_utf8(bytes.to_vec()) {
+                                        Ok(text) => text,
+                                        Err(e) => {
+                                            error!("Failed to decode binary message: {}", e);
+                                            continue;
+                                        }
+                                    };
+
+                                    // Then parse JSON
+                                    let payload: FlashblocksPayloadV1 =
```
danyalprout, line 93: I wonder if we should move the parsing logic off the receiving worker. Feels like this could potentially block new messages.

src/flashblocks.rs
```diff
@@ -0,0 +1,304 @@
...
+
+                        let block: OpBlock = execution_payload
+                            .try_into_block()
+                            .expect("failed to convert execution payload to block");
+
+                        // "pending" because users query the block using "pending" tag
+                        // This is an optimistic update will likely need to tweak in the future
+                        cache_clone
+                            .set("pending", &block, Some(10))
+                            .expect("failed to set block in cache");
```
danyalprout: I feel like panicking in all these cases may not be ideal, as it'll crash the program I believe.

src/cache.rs
```diff
@@ -0,0 +1,61 @@
...
+            store: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    pub fn set<T: Serialize>(
+        &self,
+        key: &str,
+        value: &T,
+        ttl_secs: Option<u64>,
+    ) -> Result<(), Box<dyn std::error::Error>> {
```
danyalprout, line 29: I wonder if instead of structuring this as a generic K/V store if we could have a Flashblock object that the data gets added to and methods like `get_balance` etc.

Feels like it would remove the potential of errors around building the keys.

src/flashblocks.rs
cody-wang-cb: I feel like we need to be a bit cautious of making this process entirely async, given there's order requirements (payload with lower index should be processed first).
Which means even if it ends up blocking i think it's ok, but the most important thing is to make the parsing logic fast (which it is now, currently it's < 1ms for sepolia payload). Wdyt?

src/main.rs
cody-wang-cb: Fixed, these configs went to engine_config

src/cache.rs
cody-wang-cb: that's a good point. I think im going to leave it as is for now and create a ticket for this improvement later

src/flashblocks.rs
danyalprout: We can still do ordering and move it off the worker for receiving (with a queue of some kind). Though I'm happy for us to address this in a follow up.

cody-wang-cb: Sounds good, created a ticket for this

danyalprout (approved):

PR #26, opened 2025-05-09
(fix): Preload txn receipt from eth_api for historical deposit txns

haardikk21:
The Flashblocks RPC had a bug where it would panic if `getTransactionByHash` was used to load  historical deposit transactions.

These transactions did not exist in the mempool (system transaction) and did not exist in the cache (historical transaction) but in the case of deposit transactions we would always try to fetch a receipt from the cache to populate `deposit_receipt_version` and `deposit_nonce`.

Since we did `unwrap()` without an error check in that path, it would cause a panic.

This PR attempts to fix that by pre-loading the deposit transaction receipt from `eth_api` in case a historical deposit transaction is attempted to be fetched via `getTransactionByHash`.

Edit; ignore the branch name. i thought the root cause was a race condition, it was not.

### DIFF
--- a/crates/flashblocks-rpc/src/rpc.rs
+++ b/crates/flashblocks-rpc/src/rpc.rs
@@ -13,6 +13,7 @@ use jsonrpsee::{
     proc_macros::rpc,
 };
 use op_alloy_consensus::OpTxEnvelope;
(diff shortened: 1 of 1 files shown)


haardikk21:
Closing in favor of https://github.com/base/node-reth/pull/28

PR #27, opened 2025-05-10
feat: update cache to use structured keys

haardikk21:
Our cache previously used raw `String` types everywhere for the keys. This was messy, error-prone, and also a bit more memory-heavy.

This PR introduces a `CacheKey` enum that includes variants for all currently used keys in a structured type-safe manner. It can be extended easily later if new key types need to be added. It also implements the `Display` trait for the keys so they can be logged if necessary similar to the previous format.


cody-wang-cb (approved):

danyalprout (approved):
thanks -- as a follow up I wonder if we could structure the in-memory store in a way that we don't need all the casts?

Maybe we should have a cache per type or something?

PR #28, opened 2025-05-12
fix: Preload txn receipt from eth_api for historical deposit txns

haardikk21:
The Flashblocks RPC had a bug where it would panic if `getTransactionByHash` was used to load  historical deposit transactions.

These transactions did not exist in the mempool (system transaction) and did not exist in the cache (historical transaction) but in the case of deposit transactions we would always try to fetch a receipt from the cache to populate `deposit_receipt_version` and `deposit_nonce`.

Since we did `unwrap()` without an error check in that path, it would cause a panic.

This PR attempts to fix that by pre-loading the deposit transaction receipt from `eth_api` in case a historical deposit transaction is attempted to be fetched via `getTransactionByHash`.

---

P.S. Opened this PR as the forked repo PR did not allow me to trigger a `base-builds` sepolia alpha deployment to test without more changes than necessary to those Dockerfiles.

Edit; ignore the branch name. i thought the root cause was a race condition, it was not.

### DIFF
--- a/crates/flashblocks-rpc/src/rpc.rs
+++ b/crates/flashblocks-rpc/src/rpc.rs
@@ -13,6 +13,7 @@ use jsonrpsee::{
     proc_macros::rpc,
 };
 use op_alloy_consensus::OpTxEnvelope;
+use op_alloy_consensus::{OpDepositReceipt, OpReceiptEnvelope};
 use op_alloy_network::Optimism;
 use op_alloy_rpc_types::Transaction;
 use reth::providers::TransactionsProvider;
@@ -102,7 +103,7 @@ impl<E> EthApiExt<E> {
                         index: Some(idx as u64),
                         base_fee: block.base_fee_per_gas,
                     };
-                    self.transform_tx(signed_tx_ec_recovered, tx_info)
+                    self.transform_tx(signed_tx_ec_recovered, tx_info, None)
                 })
                 .collect();
             RpcBlock::<Optimism> {
@@ -126,19 +127,26 @@ impl<E> EthApiExt<E> {
         &self,
         tx: Recovered<OpTransactionSigned>,
         tx_info: TransactionInfo,
+        deposit_receipt: Option<OpDepositReceipt>,
     ) -> Transaction {
         let tx = tx.convert::<OpTxEnvelope>();
         let mut deposit_receipt_version = None;
         let mut deposit_nonce = None;
 
         if tx.is_deposit() {
-            let receipt = self
-                .cache
-                .get::<OpReceipt>(&CacheKey::Receipt(tx_info.hash.unwrap()))
-                .unwrap();
-            if let OpReceipt::Deposit(receipt) = receipt {
+            if let Some(receipt) = deposit_receipt {
                 deposit_receipt_version = receipt.deposit_receipt_version;
                 deposit_nonce = receipt.deposit_nonce;
+            } else {
+                let cached_receipt = self
+                    .cache
+                    .get::<OpReceipt>(&format!("receipt:{:?}", tx_info.hash.unwrap().to_string()))
+                    .unwrap();
+
+                if let OpReceipt::Deposit(receipt) = cached_receipt {
+                    deposit_receipt_version = receipt.deposit_receipt_version;
+                    deposit_nonce = receipt.deposit_nonce;
+                }
             }
         }
 
@@ -370,7 +378,7 @@ where
                 TransactionSource::Pool(tx) => {
                     // Convert the pool transaction
                     let tx_info = TransactionInfo::default();
-                    Ok(Some(self.transform_tx(tx, tx_info)))
+                    Ok(Some(self.transform_tx(tx, tx_info, None)))
                 }
                 TransactionSource::Block {
                     transaction,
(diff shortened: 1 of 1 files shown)


crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -126,19 +127,26 @@ impl<E> EthApiExt<E> {
...
-                .unwrap();
-            if let OpReceipt::Deposit(receipt) = receipt {
+            if let Some(receipt) = deposit_receipt {
                 deposit_receipt_version = receipt.deposit_receipt_version;
                 deposit_nonce = receipt.deposit_nonce;
+            } else {
+                let cached_receipt = self
+                    .cache
+                    .get::<OpReceipt>(&format!("receipt:{:?}", tx_info.hash.unwrap().to_string()))
+                    .unwrap();
```
jowparks, line 144: Is it possible this cache grab can be null? If it is then this will panic.

haardikk21: This is where the panic used to be before the change, but not anymore. Deposit transactions only have two pathways to this function from the caller - historical come in via TransactionSource::Block, which will reach out to the Eth API and not attempt loading from cache, or they exist purely in cache and havent been included in a block yet so they will exist in cache.

TransactionSource::Pool isnt a valid pathway for deposits, which would've been a case where this panics, because system transactions dont go in the mempool.

crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -387,7 +395,23 @@ where
...
                         base_fee,
                     };
-                    Ok(Some(self.transform_tx(transaction, tx_info)))
+                    // preload transaction receipt if it's a deposit transaction
+                    if transaction.is_deposit() {
+                        let receipt =
+                            EthTransactions::transaction_receipt(&self.eth_api, tx_hash).await;
+
+                        // since this txn hash exists in a block, we should always be able to fetch a receipt for it
+                        let envelope: OpReceiptEnvelope = receipt.unwrap().unwrap().into();
```
jowparks: Generally the semantics in Rust, you avoid unwraps whenever possible, you can use matching and allow the calling method to handle the errors.

haardikk21: I'm slightly unsure here what we should do if this actually errors. Typically that should never happen given the context we're operating in - but in case our state is messed up for some reason and it does happen, do we just return `None` as the result for this `getTransactionByHash` call? cc @danyalprout as well

danyalprout: Agree with Joe, would be great if we could propagate this error back to the client (& error log on our side) vs. panicking

haardikk21: https://github.com/base/node-reth/pull/28/commits/c231313c0f86ac03784e36f3d0ea2f98b53a5bd6

haardikk21:
The codebase has several other places where unsafe `.unwrap()` usage exists. Those broader fixes are out of scope for this PR as this focuses solely on fixing the reported bug of fetching historical deposit transactions. I will do a separate PR to remove unsafe unwraps everywhere else later.

crates/flashblocks-rpc/src/rpc.rs
jowparks: Based on your explanation seems fine for now. I would say in a future PR we should change this to return a result type instead so we can avoid panics and handle the conversion error gracefully.

crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -387,7 +395,41 @@ where
...
+                                    tx_hash
+                                );
+                                return Ok(None);
+                            }
+                            Err(e) => {
+                                error!(
+                                    "unable to fetch receipt for block transaction: {:?} error: {:?}",
+                                    tx_hash, e
+                                );
+                                return Err(e.into());
```
jowparks: You can remove this if you just use a `?;` on line 400

crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -387,7 +395,41 @@ where
                         block_number: Some(block_number),
                         base_fee,
                     };
-                    Ok(Some(self.transform_tx(transaction, tx_info)))
+                    // preload transaction receipt if it's a deposit transaction
+                    if transaction.is_deposit() {
```
jowparks, line 399: Why did all this move to this method and pass in the receipt? Why not keep this logic in the `transform_tx` and prevent the addition of the receipt param? That way if the method is used anywhere else we won't have to replicate this logic. Otherwise looks good

haardikk21: pretty much to avoid marking a bunch of other functions async because rust has colored functions

crates/flashblocks-rpc/src/rpc.rs
haardikk21: doesn't that get rid of the error log we wish to have?

jowparks:
The diff looks crazy, does this need a rebase or something? I came back to approve this morning and looks like a bunch of cache stuff is in there.

haardikk21:
> The diff looks crazy, does this need a rebase or something? I came back to approve this morning and looks like a bunch of cache stuff is in there.

messed up previous rebase a bit. should be fine now.

cody-wang-cb (approved):

PR #40, opened 2025-07-22
Add eth_call support 

SozinM:
I've added cache for pending state that i refresh on base flashblocks and append on proper flashblocks
Then we will serve eth_call from this state


jowparks (commented):
Is there a reason we can't use a strategy where we rely on the api instead of adding this new dependency and computation here? I had a POC in this PR, I don't think I ever tested it though.

https://github.com/base/node-reth/pull/17

haardikk21:
Closing this in favour of #43

PR #42, opened 2025-07-23
Support reth 1.6.0

cody-wang-cb:
Update to reth 1.6.0

### DIFF
--- a/crates/flashblocks-rpc/src/rpc.rs
+++ b/crates/flashblocks-rpc/src/rpc.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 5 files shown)


crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -246,16 +244,29 @@ where
...
 
-        OpReceiptBuilder::new(
-            chain_spec,
-            &tx,
+        let tx = get_recovered_tx(&self.cache, tx_hash);
+
+        let mut gas_used = 0;
+        let mut next_log_index = 0;
+
+        if meta.index > 0 {
```
cody-wang-cb, line 252: These implementations copied from
https://github.com/paradigmxyz/reth/blob/8bd6bf5dc10edeb3a921e4797e5b3224859da923/crates/rpc/rpc-eth-api/src/helpers/receipt.rs#L48

danyalprout (approved):

PR #43, opened 2025-07-24
Implement eth_call via overrides

SozinM:
Implement eth_call
We build up EthOverride structure from execution flashblocks transaction that we are using when execution eth_call.


danyalprout (approved):
lgtm, I'll rebase https://github.com/base/node-reth/pull/46 once this is merged

PR #46, opened 2025-07-28
chore: refactor flashblocks-rpc

danyalprout:
### Description
* Updated the existing integration tests to use in-process Reth node, vs. relying on an external binary 
* Removed the hashmap cache and replaced it with a PendingBlock. The PendingBlock is typed and setup once per Flashblock and can serve all requests
    * The processing logic is in `Pending::insert_data`
* We don't cache historical pending blocks, only the current one
* Updated response information from send raw transaction sync to include timeout errors
* Remove params for send raw txn sync
* Updated rollup-boost to update transitive reth deps to 1.6.0

### Future TODOs

- Run on devnet, verify
- Add integration test (planning to ingest a few hundred Flashblocks from Base Mainnet and verify the data/rpc responses)
- Clear pending data on new canonical block. If P2P block of 10 arrives, clear pending data for any blocks with block number <= 10
- Add send sync RPC tests
- Incremental execution of Flashbot transactions, right now each Flashblock causes a full rexecution
- Send raw transaction API should support client side timeout


crates/flashblocks-rpc/src/metrics.rs
```diff
@@ -12,8 +12,8 @@ pub struct Metrics {
     #[metric(describe = "Time taken to process a message")]
     pub block_processing_duration: Histogram,
 
-    #[metric(describe = "Time taken to process a websocket message")]
-    pub websocket_processing_duration: Histogram,
+    #[metric(describe = "Time taken to process a message")]
```
haardikk21: description doesn't match the metric

crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -419,179 +207,87 @@ where
...
-                );
-                Ok(Some(receipt))
-            }
-            None => {
-                debug!(
-                    message = "timeout waiting for receipt",
-                    tx_hash = %tx_hash
-                );
-                Ok(None)
+        const TIMEOUT_DURATION: Duration = Duration::from_secs(6);
```
haardikk21, line 253: I wonder if this should be client-side configurable, with a reasonable max limit so we don't enter a deadlock

crates/flashblocks-rpc/src/state.rs
```diff
@@ -0,0 +1,519 @@
...
+        flashblock.metadata.block_number == pending_block.block_number
+            && flashblock.index == pending_block.flashblock_idx + 1
+    }
+
+    fn update_block(&self, block: eyre::Result<PendingBlock>) {
+        let start_time = Instant::now();
+        match block {
+            Ok(block) => {
+                self.pending_block.swap(Some(Arc::new(block)));
+                self.metrics
```
haardikk21, line 81: im slightly confused about this. its just tracking how much time it takes to do an arc swap? i think this may be misplaced

haardikk21 (approved):
mostly LGTM, minor notes + awaiting #43

crates/flashblocks-rpc/src/state.rs
danyalprout: This is a bug, will update -- thanks!

danyalprout: This is a bug, will update -- thanks!

crates/flashblocks-rpc/src/rpc.rs
danyalprout: Unfortunately the API doesn't have this, my 2c is lets start w/ 3 blocks and if there's a demand for more configuration we can setup a flag for it

crates/flashblocks-rpc/src/pending.rs
```diff
@@ -0,0 +1,171 @@
...
+            .cloned()
+            .unwrap_or(U256::from(0))
+    }
+
+    pub fn get_balance(&self, address: Address) -> Option<U256> {
+        self.account_balances.get(&address).cloned()
+    }
+
+    pub fn get_state_overrides(&self) -> Option<StateOverride> {
+        self.state_overrides.clone()
```
haardikk21, line 169: not sure if this should be `clone` or `cloned`. idt it makes a difference its just that you're doing `cloned` everywhere else

haardikk21 (approved):

crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -426,129 +218,61 @@ where
...
         debug!(
-            message = "processing transaction_by_hash request",
+            message = "rpc::transaction_by_hash",
             tx_hash = %tx_hash
         );
-        let tx = EthTransactions::transaction_by_hash(&self.eth_api, tx_hash)
-            .await
-            .map_err(Into::into)?;
-
-        // Process the result without using map() and transpose()
```
danyalprout, line 436: @cody-wang-cb curious if there was any reason we weren't using map / transpose for this?

crates/flashblocks-rpc/src/pending.rs
danyalprout: It should be `clone()` I looked into it and cloned is for an ref to value

PR #48, opened 2025-07-29
feat: `eth_simulateV1`, fix `eth_call` user-specified state overrides

haardikk21:
Merge #78 before this, this needs to be updated for the new state management structure

- enables passing overrides via `eth_call` to match json rpc
- adds `eth_simulateV1` support
- adds `eth_estimateGas` support


SozinM:
Should we take base flashblock into the account here?
Overrides won't contain changes to base fee for example

I thought that we would need to following:
Build up state (CacheDB) based on flashblock + execute transactions in it
Then execute estimateGas on this state

crates/flashblocks-rpc/src/tests/rpc.rs
```diff
@@ -526,6 +527,115 @@ mod tests {
         Ok(())
     }
 
+    // #[tokio::test]
+    // async fn test_eth_estimate_gas() -> eyre::Result<()> {
+    //     reth_tracing::init_test_tracing();
```
danyalprout: Shall we move the commented out implementation into a separate PR?

haardikk21: Yeah ill make a separate PR that just uncomments it all lol

paranoiacblack:
Now that Reth 1.7.0 has flashblocks support, what is the intention with this issue and repo? For the time being, it doesn't look like the new Reth node has `eth_simulateV1`, `eth_call`, etc. So I imagine you will continue updating this repo for those features. Is there also an intention to merge this into Reth eventually?

haardikk21:
> Now that Reth 1.7.0 has flashblocks support, what is the intention with this issue and repo? For the time being, it doesn't look like the new Reth node has `eth_simulateV1`, `eth_call`, etc. So I imagine you will continue updating this repo for those features. Is there also an intention to merge this into Reth eventually?

Reth 1.7.0 flashblocks support is fairly vanilla right now. Eventually we'd like to drop this repo in favour of something built into Reth, at least for the flashblock overlay part of things, but for now their implementation does not work for Base for a few reasons:
- RPC methods where you do not explicitly specify a block tag will not read from flashblocks state (e.g. `get_transaction_receipt`)
- ~They do not support state-root-less flashblocks, and assume `state_root` field will be present, as they attempt to progress the underlying chain based on flashblock data. This optimization is crucial to support for Base as we simply cannot produce state roots within 200ms reliably given our transaction volume and large state trie~ This turned out to be false, they do support it
- Their code has an issue we found earlier with `eth_call` and similar methods that depend on state overrides when flashblocks for block `n+1` start coming in before a canonical block `n` has been gossiped to the node over P2P, which happens a majority of the time. This causes the current implementation to drop all pending state to start building new pending state for block `n+1`, and breaks `eth_call` because the underlying chain hasn't actually processed block `n` yet so `eth_call` (and other) queries return incorrect data since effectively all of block `n` is now missing. We are almost ready to merge #78 which addresses that bug for our node by building pending state for multiple blocks if necessary 


---

NOTE: #78 + #48 have to undergo testing on Base Sepolia still. I would not recommend running off the tip of this repo assuming the bugs are fixed. Once Sepolia testing proves it's good, we will cut a new release and update https://github.com/base/node and ship to mainnet.

danyalprout (approved):

PR #56, opened 2025-08-05
added test for send raw txn sync

### LINKED ISSUE #52: Add tests for send transaction sync
Currently there are no tests for [send raw txn sync](https://github.com/base/node-reth/blob/main/crates/flashblocks-rpc/src/rpc.rs#L236-L277) , we should add some tests for this to the integration [setup](https://github.com/base/node-reth/blob/main/crates/flashblocks-rpc/src/tests/rpc.rs)

DeluxeRaph:
This adds an async integration test for  ```test_send_raw_transaction_sync```, to verify the ```sendRawTransactionSync```  request. A new helper method on ```NodeContext``` submits a raw signed tx and awaits until a receipt is returned; meanwhile the test feeds a payload containing that tx, then asserts the receipt’s hash matches. ```setup_node``` now sets the chain ID to Base Sepolia (84532) so the raw byte string ```TRANSFER_ETH_TX``` transaction is accepted.

### DIFF
--- a/crates/flashblocks-rpc/src/tests/rpc.rs
+++ b/crates/flashblocks-rpc/src/tests/rpc.rs
@@ -74,15 +74,12 @@ mod tests {
             &self,
             tx: Bytes,
         ) -> eyre::Result<RpcReceipt<Optimism>> {
-            // ② build an Alloy RpcClient that speaks HTTP
             let url     = format!("http://{}", self.http_api_addr);
             let client  = RpcClient::new_http(url.parse()?);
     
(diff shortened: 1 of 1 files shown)


danyalprout:
@DeluxeRaph thank you for working on this, can you run cargo fmt and repush -- thanks!

DeluxeRaph:
no problem, @danyalprout. Formatted

danyalprout (approved):

danyalprout:
Thanks for the contribution 🙏

PR #65, opened 2025-08-14
Clear pending Flashblocks on canonical catch-up and new metrics

DeluxeRaph:
Connects ```FlashblocksState``` to the canonical state stream so pending flashblocks are cleared when a canonical block with an equal or higher number is received. This uses a shared ```Arc<OnceCell<Arc<FlashblocksState>>>``` so both the ExEx listener and RPC module operate on the same state instance. Adds metrics to track pending snapshot height, flashblock index, and clear-on-catchup events.


crates/flashblocks-rpc/src/state.rs
```diff
@@ -64,6 +64,35 @@ where
...
+            self.metrics
+                .pending_snapshot_height
+                .set(prev.block_number() as f64);
+            self.metrics
+                .pending_snapshot_fb_index
+                .set(prev.flashblock_index() as f64);
+        }
+    }
+
+    pub fn clear_on_canonical_catchup(&self, canonical_number: u64) {
```
danyalprout, line 79: Can we call this on_canonical_block_received and update the params to take the full block, this:

- Is consistent with the Flashblocks [method](https://github.com/base/node-reth/blob/main/crates/flashblocks-rpc/src/state.rs#L324)
- We're likely going to extend this method to do more than clear to fix https://github.com/base/node-reth/issues/59

crates/node/src/main.rs
```diff
@@ -43,11 +46,37 @@ fn main() {
             let flashblocks_enabled = flashblocks_rollup_args.flashblocks_enabled();
             let op_node = OpNode::new(flashblocks_rollup_args.rollup_args.clone());
 
+            // Shared cell for a single FlashblocksState instance
```
danyalprout, line 49: nit: redundant comment

crates/node/src/main.rs
```diff
@@ -43,11 +46,37 @@ fn main() {
...
+
             let handle = builder
                 .with_types_and_provider::<OpNode, BlockchainProvider<_>>()
                 .with_components(op_node.components())
                 .with_add_ons(op_node.add_ons())
                 .on_component_initialized(move |_ctx| Ok(()))
+                .install_exex_if(flashblocks_enabled, "flashblocks-canon", {
+                    let fb_cell = fb_cell.clone();
+                    move |mut ctx| async move {
+                        // Initialize or reuse the shared state (created with ctx.provider())
```
danyalprout, line 60: nit: redundant comment

danyalprout (changes requested):
This looks great thanks, can we tweak the method name/signature. Regarding tests I wouldn't worry about it for now, we can handle that separately (I would like to make the state.rs stuff unit testable, but don't want to bloat this PR)

danyalprout:
I'm going to make these changes in a follow up PR

PR #66, opened 2025-08-14
Improve flashblocks transaction execution

SozinM:
This pr fixes case when transaction were reexecuting on every flashblock.
Now we execute transaction from latest flashblock and transition state of previous executed transactions

### DIFF
--- a/crates/flashblocks-rpc/src/state.rs
+++ b/crates/flashblocks-rpc/src/state.rs
@@ -95,6 +95,8 @@ where
 
     fn process_flashblock(&self, flashblocks: Vec<Flashblock>) -> eyre::Result<PendingBlock> {
         let mut pending_block_builder = PendingBlockBuilder::new();
+        // Number of txs in the block until last flashblock txs start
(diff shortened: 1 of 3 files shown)


crates/flashblocks-rpc/src/state.rs
```diff
@@ -280,9 +291,16 @@ where
...
             gas_used = receipt.cumulative_gas_used();
             next_log_index += receipt.logs().len();
-
+        }
+        let mut state_cache_builder = StateOverridesBuilder::default();
+        // Execute recovered transaction that belongs to the last flashblocks
+        for tx in recovered_transactions
+            .into_iter()
+            .rev()
+            .take(latest_flashblock.diff.transactions.len())
```
danyalprout: Is this going to execute them back to front for the last flashblock? i.e. if the txns in order are:

fb1-tx1, fb1-tx2, fb2-tx3, fb2-tx4

Will this execute tx4, then tx3 when it should be the other way around? I think we may need to do rev/take/rev

SozinM: Ohhh, my bad, though im smart by doing this

SozinM: Fixed

danyalprout (approved):

PR #75, opened 2025-09-02
[metrics] track tx lifecycle time

wlawt:
This PR adds an ExEx to track the lifecycle of a transaction. 

Specifically, it tracks the time to inclusion when a transaction is first detected in the mempool to when it's included in a block. It also tracks the number of transactions that are dropped/replaced. These are recorded as Histogram metrics.

TODO:
- [x] Time to inclusion
- [x] # of transactions dropped
- [x] # of transactions replaced


crates/transaction-tracing/src/tracing.rs
```diff
@@ -0,0 +1,159 @@
...
+                tx_hash = ?tx_hash,
+                event = ?event,
+                time_in_mempool = ?time_in_mempool.as_millis(),
+                "Transaction event",
+            );
+        }
+    }
+
+    // Cleanup old entries in our map that have been sitting in the mempool for too long
+    fn cleanup(&mut self) {
```
wlawt: have a queue along with the hashmap and pop off oldest to avoid using this `cleanup`

wlawt: or better yet: use `lru` instead of hashmap/queue duo

wlawt: size: 30k

crates/transaction-tracing/src/tracing.rs
```diff
@@ -0,0 +1,159 @@
...
+                "Transaction included in block (not tracked by ExEx)",
+            );
+        }
+    }
+
+    // Handle transaction event (drop, invalid, etc.)
+    fn handle_transaction_event(&mut self, tx_hash: TxHash, event: &str) {
+        if let Some(mempool_time) = self.txs.remove(&tx_hash) {
+            let time_in_mempool = Instant::now().duration_since(mempool_time);
+            info!(
```
wlawt: make these debug logs

crates/transaction-tracing/src/tracing.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+impl Tracker {
+    fn new() -> Self {
+        Self {
+            txs: HashMap::new(),
+        }
+    }
+
+    fn mempool_tx(&mut self, tx_hash: TxHash) {
```
danyalprout: nit: lets call this `transaction_inserted`

crates/transaction-tracing/src/tracing.rs
```diff
@@ -0,0 +1,159 @@
...
+        let now = Instant::now();
+        self.txs.entry(tx_hash).or_insert(now);
+        info!(target: "transaction-tracing", "Transaction {} added to mempool", tx_hash);
```
danyalprout: nit: can we use structured logging for this, it'll make it easier to query by txhash externally, e.g.

```
            debug!(
                target: "transaction-tracing",
                tx_hash = ?tx_hash,
                message = "transaction added to mempool",
            );
```

crates/transaction-tracing/src/tracing.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    fn block_inclusion(&mut self, tx_hash: TxHash, block_number: u64) {
```
danyalprout: nit: `transaction_included_in_block`

crates/transaction-tracing/src/tracing.rs
```diff
@@ -0,0 +1,159 @@
...
+        let inclusion_time = Instant::now();
+
+        if let Some(mempool_time) = self.txs.remove(&tx_hash) {
+            let time_in_mempool = inclusion_time.duration_since(mempool_time);
+
+            info!(
```
danyalprout: Lets downgrade the logging to debug (to save on logging costs). I would recommend that we have a histogram [metric](https://docs.rs/metrics/latest/metrics/macro.histogram.html) here. 

This will let us track the distribution of inclusion times for transactions.

crates/transaction-tracing/src/tracing.rs
```diff
@@ -0,0 +1,159 @@
...
+                target: "transaction-tracing",
+                tx_hash = ?tx_hash,
+                block_number = ?block_number,
+                "Transaction included in block (not tracked by ExEx)",
+            );
+        }
+    }
+
+    // Handle transaction event (drop, invalid, etc.)
+    fn handle_transaction_event(&mut self, tx_hash: TxHash, event: &str) {
```
danyalprout: nit: `transaction_event` and lets swap event to be an enum instead of a string

crates/transaction-tracing/src/tracing.rs
```diff
@@ -0,0 +1,159 @@
...
+        if let Some(mempool_time) = self.txs.remove(&tx_hash) {
+            let time_in_mempool = Instant::now().duration_since(mempool_time);
```
danyalprout: We could potentially log out a histogram here too and set a label on it to event (so we would have a metric for drop/replace etc).

If you do that you can then combine this method with transaction_included_in_block and just have an event for "BlockInclusion"

danyalprout (commented):
Looks great, left a few comments. Can we also downgrade all logs to debug from info. Thanks

crates/transaction-tracing/src/tracing.rs
```diff
@@ -0,0 +1,150 @@
...
+            self.transaction_event(tx_hash, TxEvent::BlockInclusion);
+        } else {
+            debug!(
+                target: "transaction-tracing",
+                tx_hash = ?tx_hash,
+                block_number = ?block_number,
+                "Transaction included in block (not tracked by ExEx)",
+            );
+        }
+    }
```
danyalprout: nit: personally I would combine this with transaction event, you could move the else clause there (drop the block number param) and delete this function

danyalprout (approved):

crates/node/src/main.rs
```diff
@@ -53,6 +58,9 @@ fn main() {
                 .with_components(op_node.components())
                 .with_add_ons(op_node.add_ons())
                 .on_component_initialized(move |_ctx| Ok(()))
+                .install_exex("transaction-tracing", |ctx| async move {
```
danyalprout: It maybe worth doing this as install if (the flag is set)

PR #78, opened 2025-09-05
feat: cross-block state management

haardikk21:
Cross block state management for flashblocks

Optimistically queues flashblocks state updates and overrides, even across block boundaries. Will clear pending state if canonical block signals a reorg occured.


crates/flashblocks-rpc/src/state.rs
```diff
@@ -60,44 +59,55 @@ where
...
                     self.metrics
                         .pending_snapshot_fb_index
-                        .set(prev.flashblock_index() as f64);
+                        .set(prev.latest_flashblock_index() as f64);
                 }
+            } else {
+                // Clear everything not necessary
+                // and re-process flashblocks
+                let mut flashblocks = pending_blocks.get_flashblocks();
+                flashblocks.retain(|flashblock| flashblock.metadata.block_number > block.number);
```
danyalprout: Can do in a follow up, but I think we should check if there's a flashblock/canonical reorg here and if so clear everything (as the flashblocks are built on bad state)

crates/flashblocks-rpc/src/state.rs
```diff
@@ -109,258 +119,249 @@ where
...
-                    block_hash: latest_flashblock.diff.block_hash,
-                    transactions,
-                },
-            },
-        };
-
-        let block: OpBlock = execution_payload.try_into_block()?;
-        let mut l1_block_info = reth_optimism_evm::extract_l1_info(&block.body)?;
+    fn process_flashblocks(&self, flashblocks: &Vec<Flashblock>) -> eyre::Result<PendingBlocks> {
+        let mut flashblocks_per_block = HashMap::<BlockNumber, Vec<Flashblock>>::new();
```
danyalprout: As you're sorting by key below, I wonder if there's a btree map implementation

danyalprout (commented):
lgtm (assuming we optimize the perf in a follow up)

To test this, can we clone this test, but make the flashblocks across block:

https://github.com/base/node-reth/blob/main/crates/flashblocks-rpc/src/tests/state.rs#L289

crates/flashblocks-rpc/src/state.rs
haardikk21: https://github.com/base/node-reth/pull/78/commits/74e547376159c0504321aa4da213c90a210b9693

haardikk21: wait nvm this wont work since block hash will change with presence of state root

danyalprout (approved):

cody-wang-cb (approved):

crates/flashblocks-rpc/src/state.rs
```diff
@@ -59,402 +66,564 @@ where
...
+            .load_full()
+            .and_then(|pb| pb.get_transaction_by_hash(tx_hash))
+    }
 
-            gas_used = receipt.cumulative_gas_used();
-            next_log_index += receipt.logs().len();
-        }
+    fn get_balance(&self, address: Address) -> Option<U256> {
+        self.pending_blocks
+            .load_full()
```
SozinM, line 151: better to use .load if possible (you will need deref it), because you have 5 premade instances that could be used instantly (and if it exhausted it will fallback to load_full)

haardikk21: will do this in a follow-up.

crates/flashblocks-rpc/src/state.rs
```diff
@@ -59,402 +66,564 @@ where
...
+                    &mut l1_block_info,
+                )?
+                .build();
+
+                pending_blocks_builder.with_receipt(transaction.tx_hash(), op_receipt);
+                gas_used = receipt.cumulative_gas_used();
+                next_log_index += receipt.logs().len();
+
+                let mut should_execute_transaction = false;
+                match &prev_pending_blocks {
```
SozinM, line 561: This is interesting part, would we have state at all, i assume we receive only new transaction in flashblocks
And will this be cleaned on reorgs?

SozinM: Ah, because we have vec of flashblocks
This part could be speed up, because we statically know which transactions should we execute. Only last flashblock should have unexecuted transactions

SozinM:
Looks great IMO, only 2 places when i see simple optimisations

crates/flashblocks-rpc/src/state.rs
haardikk21: technically yeah, can probably remove that check in exchange for extracting last flashblock txns separately and checking if current tx is in the set of the last fb's txns. not sure will be much of a difference in practice but can be done as a small follow up

PR #87, opened 2025-09-15
[fix][metrics] use `all_transactions` listener and add pending/queue metrics

wlawt:
Changes:
- Uses `all_transactions_event_listener` instead of `pending_transactions_listener`
- Record metrics for transactions moving between the `Pending` and `Queued` pools

### DIFF
--- a/crates/transaction-tracing/src/tracing.rs
+++ b/crates/transaction-tracing/src/tracing.rs
@@ -22,7 +22,7 @@ enum TxEvent {
 
 /// Types of pools a transaction can be in
 #[derive(Debug, Clone, PartialEq)]
-enum Pools {
+enum Pool {
     Pending,
     Queued,
 }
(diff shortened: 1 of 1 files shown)


crates/transaction-tracing/src/tracing.rs
```diff
@@ -9,36 +9,76 @@ use reth_exex::{ExExContext, ExExEvent, ExExNotification};
...
     Dropped,
     Replaced,
     BlockInclusion,
+    PendingToQueued,
+    QueuedToPending,
+}
+
+/// Types of pools a transaction can be in
+#[derive(Debug, Clone, PartialEq)]
+enum Pools {
```
danyalprout: nit: Pool

danyalprout (approved):
lgtm -- thank you

PR #88, opened 2025-09-18
[logs] emit `info!` on a tx's `TxEvents`

wlawt:
On cache eviction in the ExEx's LRU cache, we should log the `TxEvents` a transaction went through. 

This could provide insight in the lifecycle state a transaction is in after being submitted successfully. 

The eviction should happen on a `block_inclusion` event since that would be last event a transaction would experience and it would eventually become the oldest entry. There might be other cases where it got moved from `Pending` to `Queued` and then never made it to inclusion in the time newer entries were added to the LRU cache, so it would be good here to record this event log.

### DIFF
--- a/crates/transaction-tracing/src/tracing.rs
+++ b/crates/transaction-tracing/src/tracing.rs
@@ -118,6 +118,32 @@ impl Tracker {
         debug!(target: "transaction-tracing", tx_hash = ?tx_hash, state = ?pool, "Transaction moved pools");
     }
 
+    /// Track a transaction being included in a block or dropped.
+    fn transaction_completed(&mut self, tx_hash: TxHash, event: TxEvent) {
+        // if a tx is included/dropped, log it and pop it so that we keep the LRU cache size small
+        // which will help longer-lived txs.
+        if let Some(event_log) = self.txs.pop(&tx_hash) {
+            let mempool_time = event_log.mempool_time;
+            let time_in_mempool = Instant::now().duration_since(mempool_time);
+
+            self.log(
+                &tx_hash,
+                &event_log,
+                match event {
+                    TxEvent::BlockInclusion => "Transaction included in block",
+                    TxEvent::Dropped => "Transaction dropped",
+                    _ => "",
+                },
+            );
+            metrics::histogram!("reth_transaction_tracing_tx_event", "event" => match event {
+                TxEvent::BlockInclusion => "block_inclusion",
+                TxEvent::Dropped => "dropped",
+                _ => "",
+            })
+            .record(time_in_mempool.as_millis() as f64);
+        }
+    }
+
     /// Track a transaction event
     fn transaction_event(&mut self, tx_hash: TxHash, event: TxEvent) {
         if let Some(mut event_log) = self.txs.pop(&tx_hash) {
@@ -141,13 +167,12 @@ impl Tracker {
 
             // Record histogram with event label
             let event_label = match event {
-                TxEvent::Dropped => "dropped",
                 TxEvent::Replaced => "replaced",
                 TxEvent::Pending => "pending",
                 TxEvent::Queued => "queued",
-                TxEvent::BlockInclusion => "block_inclusion",
                 TxEvent::PendingToQueued => "pending_to_queued",
                 TxEvent::QueuedToPending => "queued_to_pending",
+                _ => "",
             };
             metrics::histogram!("reth_transaction_tracing_tx_event", "event" => event_label)
                 .record(time_in_mempool.as_millis() as f64);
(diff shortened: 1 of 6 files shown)


crates/transaction-tracing/src/tracing.rs
```diff
@@ -27,10 +32,13 @@ enum Pool {
     Queued,
 }
 
+/// History of events for a transaction
+type EventLog = (Instant, Vec<TxEvent>);
```
danyalprout: nit: I would make this a struct (less confusing than the .0)

danyalprout: nit: add a limit to the size of the vec (maybe use a queue, max size 10 or something)

crates/transaction-tracing/src/tracing.rs
```diff
@@ -80,13 +99,20 @@ impl Tracker {
...
             let time_in_mempool = Instant::now().duration_since(mempool_time);
 
+            // Update the change log with the new event
+            event_log.1.push(event);
+            self.txs.put(tx_hash, event_log);
+
             // Record histogram with event label
             let event_label = match event {
                 TxEvent::Dropped => "dropped",
                 TxEvent::Replaced => "replaced",
```
danyalprout: Replacement, we'll want to to get old tx event data, remove it from the map & add the new data

crates/transaction-tracing/src/tracing.rs
```diff
@@ -80,13 +99,20 @@ impl Tracker {
...
     fn transaction_event(&mut self, tx_hash: TxHash, event: TxEvent) {
-        if let Some(mempool_time) = self.txs.pop(&tx_hash) {
+        if let Some(mut event_log) = self.txs.pop(&tx_hash) {
+            let mempool_time = event_log.0;
             let time_in_mempool = Instant::now().duration_since(mempool_time);
 
+            // Update the change log with the new event
+            event_log.1.push(event);
+            self.txs.put(tx_hash, event_log);
+
```
danyalprout: Log here for included/discarded so we keep the LRU less full and it works better for long lived txns

danyalprout (commented):
We're going to want specialized behavior for these:

- Replaced: Get old txn data, remove from map, use audit log of old as start for new
- Discarded: Block Included: Log the event log, remove from map

crates/transaction-tracing/src/tracing.rs
```diff
@@ -78,44 +96,69 @@ impl Tracker {
         debug!(target: "transaction-tracing", tx_hash = ?tx_hash, state = ?pool, "Transaction moved pools");
     }
 
-    /// Track a transaction event
-    fn transaction_event(&mut self, tx_hash: TxHash, event: TxEvent) {
-        if let Some(mempool_time) = self.txs.pop(&tx_hash) {
+    /// Track a transaction being included in a block or dropped.
+    fn transaction_completed(&mut self, tx_hash: TxHash, event: TxEvent) {
```
wlawt, line 100: nit: any better name other than `completed`?

crates/transaction-tracing/src/tracing.rs
```diff
@@ -78,44 +96,77 @@ impl Tracker {
...
     }
+
+    fn log(&self, tx_hash: &TxHash, event_log: &EventLog, msg: &str) {
+        if !self.enable_logs {
+            return;
+        }
+
+        let events = event_log.to_vec();
+        if !events.is_empty() {
+            info!(target: "transaction-tracing", tx_hash = ?tx_hash, events = ?events, %msg);
```
danyalprout, line 143: I think you may need to format events as an array of strings for it to be parsed ok by datadog (I could be wrong on this)

I think you could use to_vec for it and it should work?

crates/transaction-tracing/src/tracing.rs
```diff
@@ -78,44 +96,77 @@ impl Tracker {
...
+            return false;
+        }
+
+        self.log(
+            tx_hash,
+            event_log,
+            "Transaction removed from cache due to limit",
+        );
+        record_histogram(event_log.mempool_time.elapsed(), TxEvent::Overflowed);
+        true
```
danyalprout, line 160: any reason to remove it from the cache vs. popping the oldest event?

crates/transaction-tracing/src/tracing.rs
danyalprout: Ah sorry, didn't see 141

crates/transaction-tracing/src/tracing.rs
danyalprout: I think completed is ok for now 👍

crates/transaction-tracing/src/tracing.rs
wlawt: my thinking was that since we set the threshold to be decently high like 10, if a tx went over this then likely someone is trying to abuse this and it's probably noise we don't even want to record

i can't think of a scenario where a good-intent tx will ever cycle through 10 different events before included (maybe im wrong), so that's why we should remove from the cache and not care about it

danyalprout: Fair enough 👍

danyalprout (approved):

PR #99, opened 2025-09-25
feat: implement eth_getLogs RPC method for flashblocks

cody-wang-cb:
## Summary
  - Implements `eth_getLogs` RPC method for flashblocks with pure pending query support
  - Adds comprehensive test coverage with realistic Transfer event logs
  - Delegates historical queries to underlying eth API for backwards compatibility

  ## Changes Made
  - **RPC Method**: Added `get_logs` to `EthApiOverride` trait with pure pending logic
  - **Pending Logic**: when `toBlock:"pending"`, queries use flashblocks state
  - **Log Filtering**: Uses alloy's built-in `filter.matches()` for efficient address/topic filtering
  - **API Extension**: Added `get_pending_logs` method to `FlashblocksAPI` trait
  - **Test Coverage**: Enhanced tests with realistic Transfer event logs and comprehensive filtering scenarios

  ## Test Plan
  - [x] Pure pending queries return logs from flashblocks state
  - [x] Historical queries delegate to underlying eth API
  - [x] Address filtering works correctly
  - [x] Topic filtering supports event signatures and indexed parameters
  - [x] Mixed block range queries (e.g., block 0 to pending) delegate to eth API
  - [x] All existing tests continue to pass

  🤖 Generated with [Claude Code](https://claude.ai/code)


crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -430,13 +420,50 @@ where
...
+        // Only handle pure pending queries: fromBlock="pending" and toBlock="pending"
+        // Everything else goes to regular eth API
+        if self.is_pending_query(&filter) {
+            self.metrics.get_logs.increment(1);
+            let pending_logs = self.flashblocks_state.get_pending_logs(&filter);
+            Ok(pending_logs)
+        } else {
+            // All other queries - delegate to underlying reth node
+            self.eth_filter.logs(filter).await
+        }
```
danyalprout: Should it be possible for someone to specify:

fromBlock: latest, toBlock: pending
or
fromBlock: num, toBlock: pending

Should that return:

```
self.eth_filter.logs((latest, latest)) + self.flashblocks_state.get_pending_logs(filter) 
```

danyalprout (commented):
lgtm, curious if we should/can support more than just (pending, pending) and combine with other cases

crates/flashblocks-rpc/src/rpc.rs
haardikk21: +1 this feels infinitely more useful than "pending" to "pending" - because at that point they may as well use `eth_subscribe` instead of `eth_getLogs`?

cody-wang-cb: Updated, can I get another round of reviews again @danyalprout @haardikk21 

I couldn't get historical logs to work in test though, but tested in dev, it's able to fetch historical logs properly when the toBlock is `pending`

haardikk21: Curious what the issue is re: historical logs not working in tests?

cody-wang-cb: it's just that I can't figure out a way to advance the chain for the node in unit test to include some mock historical logs
probably a skills issue..

haardikk21: it's not too important but the way to do it is to build a payload and trigger `newPayload` on the engine API - https://github.com/paradigmxyz/reth/blob/80b4d26b9d79bb71ac2e3d90d2ad8d4068859e63/crates/e2e-test-utils/src/node.rs

but its definitely non-trivial to update the test framework to do this

cody-wang-cb: yeah I was trying `NodeTestContext` but it kind of turned into a rabbit hole...
might give another attempt later on, but don't think it's necessary to block this PR because of that

danyalprout (approved):

PR #101, opened 2025-09-26
fix: use canon block instead of latest in rpcs

haardikk21:
Solves the race condition which could result in double-counting when fetching nonce or apply unnecessary overrides in `eth_call`, `eth_estimateGas` or `eth_simluateV1`

### DIFF
--- a/crates/flashblocks-rpc/src/pending_blocks.rs
+++ b/crates/flashblocks-rpc/src/pending_blocks.rs
@@ -1,4 +1,5 @@
 use alloy_consensus::{Header, Sealed};
+use alloy_eips::BlockNumberOrTag;
 use alloy_primitives::{
     map::foldhash::{HashMap, HashMapExt},
     Address, BlockNumber, TxHash, B256, U256,
@@ -149,6 +150,10 @@ pub struct PendingBlocks {
 impl PendingBlocks {
     pub fn latest_block_number(&self) -> BlockNumber {
         self.headers.last().unwrap().number
+    }
+
+    pub fn canonical_block_number(&self) -> BlockNumberOrTag {
+        BlockNumberOrTag::Number(self.headers.first().unwrap().number - 1)
     }
 
     pub fn latest_flashblock_index(&self) -> u64 {
--- a/crates/flashblocks-rpc/src/rpc.rs
+++ b/crates/flashblocks-rpc/src/rpc.rs
@@ -35,6 +35,9 @@ pub const MAX_TIMEOUT_SEND_RAW_TX_SYNC_MS: u64 = 6_000;
 
 /// Core API for accessing flashblock state and data.
 pub trait FlashblocksAPI {
+    /// Retrieves the canonical block number.
+    fn get_canonical_block_number(&self) -> BlockNumberOrTag;
+
     /// Retrieves the current block. If `full` is true, includes full transaction details.
     fn get_block(&self, full: bool) -> Option<RpcBlock<Optimism>>;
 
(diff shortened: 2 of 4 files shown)


crates/flashblocks-rpc/src/pending_blocks.rs
```diff
@@ -151,6 +152,10 @@ impl PendingBlocks {
         self.headers.last().unwrap().number
     }
 
+    pub fn canonical_block_number(&self) -> BlockNumberOrTag {
+        BlockNumberOrTag::Number(self.headers.first().unwrap().number - 1)
```
cody-wang-cb, line 156: maybe a dumb question, why is -1 always the canonical?

haardikk21: The first header is for the first pending block we are tracking, so the one before it is the Canon state all the pending state is built on top of

crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -219,16 +222,15 @@ where
...
-            let latest_count = EthState::transaction_count(
-                &self.eth_api,
-                address,
-                Some(BlockId::Number(BlockNumberOrTag::Latest)),
-            )
-            .await
-            .map_err(Into::into)?;
-
+            let canon_block = self.flashblocks_state.get_canonical_block_number();
             let fb_count = self.flashblocks_state.get_transaction_count(address);
```
danyalprout: I believe this will still have a race condition (though much shorter than the current setup).

routine 1:
            let canon_block = self.flashblocks_state.get_canonical_block_number();

routine 2:
           // new flashblock comes up that changes the base canonical block

routine 1:
            let fb_count = self.flashblocks_state.get_transaction_count(address);

Whenever we do multiple lookups, I think we need to get a copy of the arcswap pointer

crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -329,11 +331,12 @@ where
...
         );
 
-        let block_id = block_number.unwrap_or_default();
+        let mut block_id = block_number.unwrap_or_default();
         let mut pending_overrides = EvmOverrides::default();
         // If the call is to pending block use cached override (if they exist)
         if block_id.is_pending() {
             self.metrics.call.increment(1);
+            block_id = self.flashblocks_state.get_canonical_block_number().into();
             pending_overrides.state = self.flashblocks_state.get_state_overrides();
```
danyalprout: Same here

crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -368,11 +371,12 @@ where
...
         );
 
-        let block_id = block_number.unwrap_or_default();
+        let mut block_id = block_number.unwrap_or_default();
         let mut pending_overrides = EvmOverrides::default();
         // If the call is to pending block use cached override (if they exist)
         if block_id.is_pending() {
             self.metrics.estimate_gas.increment(1);
+            block_id = self.flashblocks_state.get_canonical_block_number().into();
             pending_overrides.state = self.flashblocks_state.get_state_overrides();
```
danyalprout: And here

crates/flashblocks-rpc/src/rpc.rs
```diff
@@ -396,12 +400,13 @@ where
...
 
-        let block_id = block_number.unwrap_or_default();
+        let mut block_id = block_number.unwrap_or_default();
         let mut pending_overrides = EvmOverrides::default();
 
         // If the call is to pending block use cached override (if they exist)
         if block_id.is_pending() {
             self.metrics.simulate_v1.increment(1);
+            block_id = self.flashblocks_state.get_canonical_block_number().into();
             pending_overrides.state = self.flashblocks_state.get_state_overrides();
```
danyalprout: And here

danyalprout (approved):

danyalprout (approved):

PR #123, opened 2025-10-06
feat: implement Web3ApiExt to override `web3_clientVersion`

### LINKED ISSUE #122: Override version reported by Reth to include node-reth version as well
Currently calling `web3_clientVersion` RPC on this node returns the version of Reth being used (currently `1.8.1`). This is confusing for third parties running this node who need to differentiate it from vanilla Reth and also understand what version of the reth extension software they're running.

Introduce a new RPC override for it such that it returns `{RETH_VERSION]-{EXTENSION_VERSION}` i.e. currently it would be `1.8.1-0.1.9` instead of just `1.8.1`

haardikk21:
Closes https://github.com/base/node-reth/issues/122

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1521,7 +1521,7 @@ dependencies = [
 
 [[package]]
 name = "base-reth-flashblocks-rpc"
-version = "0.1.0"
+version = "0.1.10"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -1575,7 +1575,7 @@ dependencies = [
 
 [[package]]
 name = "base-reth-node"
-version = "0.1.0"
+version = "0.1.10"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -1633,7 +1633,7 @@ dependencies = [
 
 [[package]]
 name = "base-reth-transaction-tracing"
-version = "0.1.0"
+version = "0.1.10"
 dependencies = [
  "alloy-primitives",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "0.1.0"
+version = "0.1.10"
 edition = "2021"
 rust-version = "1.85"
 license = "MIT"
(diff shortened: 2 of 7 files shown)


danyalprout:
I haven't tested this out (so maybe wrong), but I wonder if we should configure this via these [env vars](https://github.com/paradigmxyz/reth/blob/94c93583af801a75ae5bb96080d21dc1851325fa/crates/node/core/src/version.rs#L109).

This looks like it'll also cause it to be exposed elsewhere, e.g. [metrics](https://github.com/paradigmxyz/reth/blob/0a4a8839235454da1b7e43e057ad9cd275ec5a30/crates/node/metrics/src/version.rs#L23) ([setup](https://github.com/paradigmxyz/reth/blob/94c93583af801a75ae5bb96080d21dc1851325fa/crates/cli/commands/src/stage/run.rs#L122)).

haardikk21:
> I haven't tested this out (so maybe wrong), but I wonder if we should configure this via these [env vars](https://github.com/paradigmxyz/reth/blob/94c93583af801a75ae5bb96080d21dc1851325fa/crates/node/core/src/version.rs#L109).
> 
> This looks like it'll also cause it to be exposed elsewhere, e.g. [metrics](https://github.com/paradigmxyz/reth/blob/0a4a8839235454da1b7e43e057ad9cd275ec5a30/crates/node/metrics/src/version.rs#L23) ([setup](https://github.com/paradigmxyz/reth/blob/94c93583af801a75ae5bb96080d21dc1851325fa/crates/cli/commands/src/stage/run.rs#L122)).

These are compile time env vars (env! Macro) not runtime so i dont think those will work. Can test tmr

danyalprout (approved):
Would be great to figure out how to override the version tag on metrics, as that's what our internal dashboards use

haardikk21:
> Would be great to figure out how to override the version tag on metrics, as that's what our internal dashboards use

Ah I think we may actually be able to use https://github.com/paradigmxyz/reth/blob/6f96a328128eb79a3009669b0a32f232971a99c9/crates/node/core/src/version.rs#L15 to override it entirely on our own. I will do a separate PR and close this one once I can test it works.

PR #129, opened 2025-10-10
chore: add github action to build docker image on PRs and merges to master

danyalprout:
### Description
We broke the docker builds (fixed in https://github.com/base/node-reth/pull/128), we should check this in CI

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,9 +2,9 @@ name: Docker Build
 
 on:
   push:
-    branches: [master]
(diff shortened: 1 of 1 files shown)


haardikk21 (approved):
Approved, but alternatively can get rid of this dockerfile too since we dont use it. We use the one in base/node

danyalprout:
My thinking is base/node contains Base specific releases (and configs), we should keep this dockerfile and packages in this repo that can be shared by other teams using the ExExs

PR #131, opened 2025-10-10
fix(rpc): remove incorrect metrics increment in transaction_by_hash

sashaodessa:
### Problem
The `getTransactionByHash` RPC method was incorrectly incrementing the `get_transaction_receipt` metric when returning transactions from flashblocks cache. This caused metric pollution and reduced observability accuracy.

### Solution
- Removed the erroneous `get_transaction_receipt.increment(1)` call from the flashblocks cache path in `transaction_by_hash`
- This ensures metrics accurately reflect actual RPC method usage


haardikk21:
This is not erroneous, since `transaction_by_hash` does rely on `get_transaction_receipt`

PR #136, opened 2025-10-12
docs(README): reference Cargo.toml version

thinhnc1975:
README referenced Reth v1.8.1 while the workspace pins Reth to v1.8.2 in Cargo.toml so I replaced the two README occurrences of v1.8.1 with v1.8.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 > [!WARNING]
 > This repository is for development purposes. For production deployments, please use the releases referenced in [base/node](https://github.com/base/node/releases).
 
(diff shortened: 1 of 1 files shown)


danyalprout (changes requested):
Instead of doing this, can you update the readme to point to the version in the cargo.toml? That way we don't have to update this every time we update Rath.

thinhnc1975:
Hi @danyalprout, done as request, pls check, thanks!

danyalprout (approved):

PR #137, opened 2025-10-20
chore: release optimizations

AnDr1k:
perf: optimize build configuration and memory allocation

- Enable jemalloc by default for improved memory management
- Add aggressive release profile with LTO and single codegen unit
- Remove redundant time crate dependency (chrono sufficient)
- Configure strip and panic=abort for smaller binary size
- Add production profile for deployment builds

Performance improvements:
- Runtime: +10-20% through fat LTO
- Memory usage: -10-15% with jemalloc allocator
- Binary size: -20-30% through stripping and optimizations

Breaking: jemalloc now enabled by default. Disable with --no-default-features if needed.

### DIFF
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -82,10 +82,21 @@ metrics-derive.workspace = true
 itertools.workspace = true
 eyre.workspace = true
 uuid.workspace = true
-time.workspace = true
 chrono.workspace = true
 once_cell.workspace = true
 
 [features]
-default = []
+default = ["jemalloc"]
 jemalloc = ["reth-cli-util/jemalloc", "reth-optimism-cli/jemalloc"]
+
+[profile.release]
+opt-level = 3
+lto = "fat"
+codegen-units = 1
+strip = true
+panic = "abort"
+
+[profile.production]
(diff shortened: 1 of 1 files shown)


danyalprout (approved):

0xsimulacra:
@danyalprout 

Rust is new to me, but it looks like this PR introduces some optimizations to the `node` crate under the wrong profile name.

In the workspace root `Cargo.toml`, the profile is named `maxperf`, whereas in `crates/node/Cargo.toml` it’s named `production`.

Also, per the Cargo docs:

> Cargo only looks at the profile settings in the Cargo.toml manifest at the root of the workspace. Profile settings defined in dependencies will be ignored.

Source: [https://doc.rust-lang.org/cargo/reference/profiles.html](https://doc.rust-lang.org/cargo/reference/profiles.html)

Because of that, any `[profile.*]` sections in the `node` crate won’t have any effect when building the workspace. These profiles should be defined in the root `Cargo.toml` (and per-package overrides can be used there if needed). 

Am I wrong ?

danyalprout:
@0xsimulacra, you're right. Thanks will revert this

PR #142, opened 2025-10-24
Add base_meterBundle RPC for transaction bundle metering

niran:
Implements JSON-RPC endpoint to simulate and meter Optimism transaction bundles, returning detailed gas usage, execution time, and bundle hash using Flashbots methodology.

Includes reusable meter_bundle() function for non-RPC contexts.


crates/metering/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+
+# async
+tokio.workspace = true
+
+# misc
+tracing.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+eyre.workspace = true
+hex = "0.4"
```
danyalprout: nit: can we move the version for this into the root Cargo.toml?

crates/metering/src/rpc.rs
```diff
@@ -0,0 +1,207 @@
...
+            let mut reader = tx_bytes.as_ref();
+            let tx = op_alloy_consensus::OpTxEnvelope::decode_2718(&mut reader).map_err(|e| {
+                jsonrpsee::types::ErrorObjectOwned::owned(
+                    jsonrpsee::types::ErrorCode::InvalidParams.code(),
+                    format!("Failed to decode transaction: {}", e),
+                    None::<()>,
+                )
+            })?;
+            decoded_txs.push(tx);
+        }
```
danyalprout: Do we want to include deserialization/block lookup/header lookup as part of the total execution time? Feels like it's very likely an ~static overhead (maybe except for deserialization, but we'll already have done that in the ingress-rpc).

crates/metering/src/meter.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// Reference: <https://github.com/flashbots/ethers-provider-flashbots-bundle/blob/0d404bb041b82c12789bd62b18e218304a095b6f/src/index.ts#L266-L269>
+fn calculate_bundle_hash(tx_hashes: &[B256]) -> B256 {
+    use alloy_primitives::keccak256;
+    let mut combined = Vec::with_capacity(tx_hashes.len() * 32);
+    for hash in tx_hashes {
+        combined.extend_from_slice(hash.as_slice());
+    }
+    keccak256(&combined)
+}
```
danyalprout: We don't need to change this, but I've created our own Bundle/BundleWithMetadata types that have helper methods for this too (see [here](https://github.com/base/tips/pull/39/files#diff-e406882a4d69985d1e0b0012a425d21efb6010622c70f2fd88d31c3b0a3e7208))

crates/metering/src/rpc.rs
```diff
@@ -0,0 +1,207 @@
...
+use serde::{Deserialize, Serialize};
+use std::time::Instant;
+use tracing::{error, info};
+
+use crate::meter_bundle;
+
+/// Request payload for base_meterBundle
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MeterBundleRequest {
```
danyalprout: I wonder if we should have this be something like:

```
MeterBundleRequest {
   bundle: Bundle (see [here](https://github.com/base/tips/pull/39/files#diff-e406882a4d69985d1e0b0012a425d21efb6010622c70f2fd88d31c3b0a3e7208R10-R52))
   ...meter params like state block etc
}
```

I worry once we need to support Flashblock simulation/dropped transactions if we're just going to duplicate types

danyalprout (commented):
Looks good to me, left some comments on things that stood out but aren't blockers to merging.

Can we add a sanity check e2e test to this? Happy to do it in a follow up PR

crates/metering/Cargo.toml
niran: removed in a9574f071702ef142a2860875e4c21dc6a3f8400

crates/metering/src/rpc.rs
niran: fixed in b33e325532ae3b550ff68e3705f6d70d9263257c

crates/metering/src/rpc.rs
niran: Replaced `MeterBundleRequest` with `tips_core::Bundle` in 0fdc2d56ce9081185d93b235576d4f30772ccc4d.

crates/metering/src/meter.rs
niran: replaced in 0fdc2d56ce9081185d93b235576d4f30772ccc4d

crates/metering/src/rpc.rs
```diff
@@ -0,0 +1,194 @@
...
+        info!(
+            num_transactions = bundle.txs.len(),
+            block_number = bundle.block_number,
+            "Starting bundle metering"
+        );
+
+        // Get the header
+        let header = self
+            .provider
+            .sealed_header(bundle.block_number)
```
niran: The biggest wart from moving to `tips_core::Bundle` is that specific block numbers now need to be provided in the request. The ability to specify `"pending"` or `"latest"` has been lost.

crates/metering/Cargo.toml
```diff
@@ -0,0 +1,63 @@
...
+reth.workspace = true
+reth-provider.workspace = true
+reth-primitives.workspace = true
+reth-primitives-traits.workspace = true
+reth-evm.workspace = true
+reth-optimism-evm.workspace = true
+reth-optimism-chainspec.workspace = true
+reth-optimism-primitives.workspace = true
+reth-transaction-pool.workspace = true
+reth-optimism-cli.workspace = true  # Enables serde & codec traits for OpReceipt/OpTxEnvelope
```
niran, line 28: Without this dependency, the traits needed for de/serializing Optimism transactions weren't available. Including it allows the code to compile without changing imports in any code. This was surprising to me.

crates/metering/src/rpc.rs
wlawt: @niran i think we could do `BlockNumberOrTag::Latest.as_number()`?

crates/metering/src/rpc.rs
```diff
@@ -0,0 +1,194 @@
...
+    pub bundle_hash: B256,
+    pub coinbase_diff: String,
+    pub eth_sent_to_coinbase: String,
+    pub gas_fees: String,
+    pub results: Vec<TransactionResult>,
+    pub state_block_number: u64,
+    pub total_gas_used: u64,
+    /// Resource metering: total execution time in microseconds
+    pub total_execution_time_us: u128,
+}
```
wlawt: not sure, but i wonder if these types should also be in `tips-core`?

crates/metering/src/rpc.rs
niran: Yep, that's what the code was doing before. But since the `Bundle` type has its own block number field, we can either just use `Bundle`'s field, refactor the core types to make it easier to provide the block reference in the right way, or just avoid the core types altogether.

The types refactoring approach seems like the best option to me, but it's a little messy if we want to be able to maintain the main benefit that changes to bundle features are reflected in every crate that depends on tips-core. I think it'd be something like this in tips-core, which implements conversions between types to ensure they stay compatible. (Unfortunately, for other crates to inherit this, they'll have to remember to add similar conversion functions. Otherwise conversions will silently break.)

```
  // Core bundle data without timing/targeting
  pub struct BundleCore {
      pub txs: Vec<Bytes>,
      pub reverting_tx_hashes: Vec<TxHash>,
      pub replacement_uuid: Option<String>,
      pub dropping_tx_hashes: Vec<TxHash>,
  }

  // Timing/targeting metadata separate from core
  pub struct TimingMetadata {
      pub block_number: u64,
      pub flashblock_number_min: Option<u64>,
      pub flashblock_number_max: Option<u64>,
      pub min_timestamp: Option<u64>,
      pub max_timestamp: Option<u64>,
  }

  // Original Bundle type (unchanged for backward compatibility)
  pub struct Bundle {
      pub txs: Vec<Bytes>,
      pub block_number: u64,
      pub flashblock_number_min: Option<u64>,
      pub flashblock_number_max: Option<u64>,
      pub min_timestamp: Option<u64>,
      pub max_timestamp: Option<u64>,
      pub reverting_tx_hashes: Vec<TxHash>,
      pub replacement_uuid: Option<String>,
      pub dropping_tx_hashes: Vec<TxHash>,
  }

  // Decompose Bundle into parts
  impl Bundle {
      pub fn into_parts(self) -> (BundleCore, TimingMetadata) {
          let core = BundleCore {
              txs: self.txs,
              reverting_tx_hashes: self.reverting_tx_hashes,
              replacement_uuid: self.replacement_uuid,
              dropping_tx_hashes: self.dropping_tx_hashes,
          };

          let timing = TimingMetadata {
              block_number: self.block_number,
              flashblock_number_min: self.flashblock_number_min,
              flashblock_number_max: self.flashblock_number_max,
              min_timestamp: self.min_timestamp,
              max_timestamp: self.max_timestamp,
          };

          (core, timing)
      }

      pub fn from_parts(core: BundleCore, timing: TimingMetadata) -> Self {
          Bundle {
              txs: core.txs,
              block_number: timing.block_number,
              flashblock_number_min: timing.flashblock_number_min,
              flashblock_number_max: timing.flashblock_number_max,
              min_timestamp: timing.min_timestamp,
              max_timestamp: timing.max_timestamp,
              reverting_tx_hashes: core.reverting_tx_hashes,
              replacement_uuid: core.replacement_uuid,
              dropping_tx_hashes: core.dropping_tx_hashes,
          }
      }

      pub fn core(&self) -> &BundleCore {
          // For zero-copy access
      }
  }

  // Standard conversions
  impl From<Bundle> for BundleCore {
      fn from(bundle: Bundle) -> Self {
          bundle.into_parts().0
      }
  }

  impl From<&Bundle> for BundleCore {
      fn from(bundle: &Bundle) -> Self {
          BundleCore {
              txs: bundle.txs.clone(),
              reverting_tx_hashes: bundle.reverting_tx_hashes.clone(),
              replacement_uuid: bundle.replacement_uuid.clone(),
              dropping_tx_hashes: bundle.dropping_tx_hashes.clone(),
          }
      }
  }
```

Then the `MeterBundleRequest` would have its own handling for `BlockNumberOrTag::Latest.as_number()` with a method that extracts a `TimingMetadata` to maintain compatibility over time.

Whatever we decide here, I think we should do it as its own feature later.

crates/metering/src/rpc.rs
niran: That's probably true since ingress-rpc will need it. Will move.

niran:
Once this is reviewed, I'm going to ~re-open this PR against main~ to merge it in now that there's a feature flag for it. (thanks @wlawt)

EDIT: turns out the merge base is editable! It's now based on main.

crates/metering/src/rpc.rs
niran: base/tips#45

crates/metering/src/rpc.rs
danyalprout: > The biggest wart from moving to tips_core::Bundle is that specific block numbers now need to be provided in the request. The ability to specify "pending" or "latest" has been lost.

Any reason to not always sim against latest (or ideally pending/flashblock state)?

niran: Yeah, I guess we don't really need to be able to specify the state for `base_meterBundle` since specifying an older state to generate metering data is really only useful for debugging. Updated in 3deb676cd612f9a9af87dceb93fbb87db565b175.

I'll modify this to use the flashblock state in another PR since we'll need to change the response format to indicate that state that was used more granularly (i.e. not just `state_block_number` but flashblock number too).

danyalprout: Makes sense 👍

danyalprout (approved):

PR #147, opened 2025-10-28
feat: Add state root calculation time from latest flashblock state to metering RPC

niran:
Adds state root calculation timing with flashblock state support to the metering RPC to provide visibility into state root computation performance as an independent metric.

### TODO

- [X] Add tests with state changes across blocks and flashblocks based on the upcoming refactored flashblocks-rpc test suite

### Key Changes

**1. Flashblock State Integration**
- Bundles now execute on top of pending flashblock state (not just canonical block)
- Accurately accounts for nonces, balances, storage, and code changes from pending transactions
- Adds `state_flashblock_index` to response

**2. State Root Optimization**
- New `FlashblockTrieCache`: caches latest flashblock's trie nodes
- Computes flashblock state root once, reuses for all subsequent bundle metering calls
- Cache keyed by `(block_hash, flashblock_index)`
- Uses `TrieInput.prepend_cached()` to combine cached flashblock trie with bundle changes
- Significant performance improvement for repeated metering on same flashblock

**3. State Root Metrics**
- Separately tracks state root calculation time
- Flashblock trie computation excluded from timing metrics

**4. Flashblocks-RPC Fixes (discovered during implementation)**
- **Fixed incorrect database layering:** Changed from `CacheDB → State` to `State → CacheDB → StateProviderDatabase`
- **Added bundle state tracking:** `PendingBlocks` now tracks accumulated `BundleState` alongside cache
- **Added state accumulation:** Uses `with_bundle_prestate()` to accumulate state changes across flashblocks
- These fixes enable accurate state root calculations for metering and potentially benefit other flashblock state consumers


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/flashblocks-rpc/src/pending_blocks.rs
```diff
@@ -195,6 +208,10 @@ impl PendingBlocks {
         self.db_cache.clone()
     }
 
+    pub fn get_bundle_state(&self) -> BundleState {
+        self.bundle_state.clone()
+    }
```
haardikk21, line 241: Do we have any intuition for how expensive this clone can get? Slightly concerned about introducing a perf regression here. A not too older version here had a significant perf regression because of some extremely large clones that started happening. Im unsure how fast bundlestate can grow.

This may be worth testing on mainnet dev node before merging to main

crates/metering/Cargo.toml
```diff
@@ -26,17 +26,25 @@ reth-optimism-chainspec.workspace = true
...
 alloy-eips.workspace = true
+alloy-sol-types.workspace = true
+alloy-rpc-types-eth.workspace = true
 
 # op-alloy
 op-alloy-consensus.workspace = true
+op-alloy-rpc-types.workspace = true
+
+# base
+base-reth-flashblocks-rpc = { path = "../flashblocks-rpc" }
```
danyalprout: Can we use a workspace setting for this?

haardikk21:
Could you also expand on this?

> Fixed incorrect database layering: Changed from CacheDB → State to State → CacheDB → StateProviderDatabase

What issues existed previously and what implications does this have? i'm concerned about changes to the core processing pipeline introducing regressions not caught in the current tests. Definitely think this needs to go on mainnet dev node before being merged

niran:
> > Fixed incorrect database layering: Changed from CacheDB → State to State → CacheDB → StateProviderDatabase
> 
> What issues existed previously and what implications does this have?

`CacheDB` is a read-through cache. Writes are not propagated. As a result, I believe the `State` layer in the previous implementation wasn't doing anything. `State` is for producing `BundleStates` that the rest of the reth machinery can use to persist the changes, or in our case, calculate the state root. If a `CacheDB` is in front of it, no changes will be written to the `State`.

We can either remove the `CacheDB` entirely, or we can keep it around like the current PR does to maintain a cache of reads.

~I could be wrong about this since I haven't actually written code that compares both outcomes. But that's my reading of the docs.~ (Tests added in 1942db86dd98561d60fd72008801ebf44941ed82.)

> Definitely think this needs to go on mainnet dev node before being merged

Makes sense.

crates/flashblocks-rpc/src/pending_blocks.rs
niran: Added metrics for this in 9752d0ab008d02fa996e82dc12e127dc44d9c89e.

niran:
There are now tests in 1942db86dd98561d60fd72008801ebf44941ed82 that prove that the new `State<CacheDB<StateProviderDatabase>>` layering is correct and necessary for state root calculation.

danyalprout:
Closing temp until after rebase

PR #166, opened 2025-11-03
feat: refactor integration testing setup

### LINKED ISSUE #162: implement support for engine API in the test framework
Currently the RPC test framework does not include CL engine api calls. This means the test chain we spin up in tests don't actually produce canonical blocks and progress the l2 head. The flashblocks state processor can "pretend" a canonical block has come in, as we currently do, but this limits the full scenarios we can run and prevents writing certain types of tests that are becoming necessary to write.

For a better end to end test setup, we should have a proper e2e setup where the chain also can produce real canonical blocks and progress the L2 head. It's okay if these need to be manually triggered rather than producing a new canonical block at a fixed interval.

### LINKED ISSUE #163: use Foundry + `sol!` macros for contract tests
the way the tests currently include dummy contracts is pretty terrible and major tech debt. There's a whole dance to be done manually around constructing a raw transaction, figuring out the txn hash, figuring out the contract address, etc. that needs to happen outside the test files manually and then the test files just have magic raw hex data.

This makes it extremely hard to add new tests or modify existing ones.

We need to significantly improve this situation. We should have an embedded foundry project with test contracts that can be compiled and their artifacts included in the test files with alloy's `sol!` macro. Then, we can programatically easily construct their deployments and function calls which is significantly more human readable and 100x easier to add/modify new contracts and new tests to the setup.

### LINKED ISSUE #164: Move common test framework setup to it's own crate
As other crates start making their way into this repo (re: resource-metering, and access-lists), they all need a common shared test setup. Set up a dummy EL client, CL client, have a way to progress the chain, have a way to make transactions, etc.

Instead of copy-pasting initialization code across all these crates, we should create a new `test-utils` crate or similar to have a common shared framework that can be reused across the board.

haardikk21:
Introduces a shared integration test framework in `crates/test-utils` and migrates the existing flashblocks RPC tests to use this new infrastructure. This framework supports tests that need to advance the state of the chain via flashblocks or full blocks to verify behavior.

- **`TestHarness`**: The main orchestration layer that combines node management, Engine API interactions, and account management. It handles block building, chain advancement, and flashblocks submission.
- **`LocalNode`**: An in-process Optimism node wrapper with Base Sepolia configuration, isolated from P2P discovery for parallel testing.
- **`EngineApi`**: A wrapper around the Consensus Layer (CL) Engine API for forkchoice updates and payload submission.
- **`TestAccounts`**: Pre-funded, deterministic accounts (Alice, Bob, Charlie, Deployer) compatible with Anvil defaults.

Closes #162 and #164.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/test-utils/src/node.rs
```diff
@@ -0,0 +1,278 @@
...
+    ///
+    /// Unlike `NodeBuilder::testing_node()` which hardcodes an 8 TB map size,
+    /// this method configures the database with a 100 MB map size. This prevents
+    /// `ENOMEM` errors when running parallel tests with `cargo test`, as the
+    /// default 8 TB size can cause memory exhaustion when multiple test processes
+    /// run concurrently.
+    fn create_test_database() -> Result<Arc<TempDatabase<DatabaseEnv>>> {
+        let default_size = 100 * 1024 * 1024; // 100 MB
+        Self::create_test_database_with_size(default_size)
+    }
```
niran, line 229: The reth tests don't seem to need this workaround, but it might be because they use `cargo nextest`.

haardikk21: See https://github.com/base/node-reth/blob/main/crates/flashblocks-rpc/src/tests/utils.rs to also help with the `global executor already set` issues. doesnt entirely get rid of them but doesnt cause tests to fail and allows to run at least # of CPU cores worth of tests in parallel afaik

crates/test-utils/assets/genesis.json
```diff
@@ -0,0 +1,106 @@
...
+        "mergeNetsplitBlock": 0,
+        "bedrockBlock": 0,
+        "regolithTime": 0,
+        "canyonTime": 0,
+        "ecotoneTime": 0,
+        "fjordTime": 0,
+        "graniteTime": 0,
+        "isthmusTime": 0,
+        "pragueTime": 0,
+        "terminalTotalDifficulty": 0,
```
haardikk21, line 27: add `jovianTime`

crates/test-utils/src/node.rs
```diff
@@ -0,0 +1,280 @@
...
+        let provider_cell: Arc<OnceCell<LocalNodeProvider>> = Arc::new(OnceCell::new());
+
+        let NodeHandle { node: node_handle, node_exit_future } =
+            NodeBuilder::new(node_config.clone())
+                .with_database(temp_db)
+                .with_launch_context(exec.clone())
+                .with_types_and_provider::<OpNode, BlockchainProvider<_>>()
+                .with_components(node.components_builder())
+                .with_add_ons(node.add_ons())
+                .install_exex("flashblocks-canon", {
```
haardikk21: The goal with the custom launcher was to not be required to install this ExEx and the Flashblocks State provider/subscriber on every instance of the node, and delegate those responsibilities to the callsite where the tests are being set up. 

Particularly not every test - e.g. transaction tracing crate, or metering stuff - requires Flashblocks integration to be set up. I'm curious why this was re-added here when https://github.com/base/node-reth/pull/166/files#diff-8be6048eaf2268aa2590e45c82d03324ec4a4727ab76b1d2d3447c1174fe3e8f for example has a custom launcher function that installs it where it is required.

crates/test-utils/src/node.rs
```diff
@@ -0,0 +1,280 @@
...
+                            Ok(())
+                        })
+                    }
+                })
+                .extend_rpc_modules({
+                    let fb_cell = fb_cell.clone();
+                    let provider_cell = provider_cell.clone();
+                    let mut receiver = Some(receiver);
+                    move |ctx| {
+                        let provider = provider_cell.get_or_init(|| ctx.provider().clone()).clone();
```
haardikk21: same as above

crates/test-utils/src/node.rs
niran: Ah. This is extracted from a branch where I was working on metering tests, so I think I just didn't notice the expected pattern. I'll back this out.

niran: There's now a `FlashblocksHarness` separate from the base `TestHarness`.

haardikk21:
LGTM - thanks a lot! I can't do an approval since I opened this PR

cc @danyalprout

danyalprout (approved):

PR #215, opened 2025-12-02
feat: add transaction status API

danyalprout:
### Description
Add transaction status API, for Base nodes this will use the local mempool as the source of truth. External nodes will forward requests to the sequencer URL.

### Follow ups
- We should combine all of the RPC extensions into a single crate


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/transaction-status/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+pub mod rpc;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
+pub enum Status {
+    Unknown,
+    Known,
+}
```
wlawt, line 9: is it correct to assume that in later iterations we will give more granular insights (i.e., mempoolReceiver, BuilderIncluded, BlockIncluded, etc.)?

crates/transaction-status/src/rpc.rs
```diff
@@ -0,0 +1,197 @@
...
+        test_utils::{MockTransaction, testing_pool},
+    };
+    use serde_json::{self, json};
+
+    use super::*;
+    use crate::Status::{Known, Unknown};
+
+    #[tokio::test]
+    async fn test_transaction_status() -> eyre::Result<()> {
+        let pool = testing_pool();
```
wlawt, line 87: neat! didn't know this existed

crates/node/src/main.rs
```diff
@@ -141,6 +142,13 @@ fn main() {
                         ctx.modules.merge_configured(metering_api.into_rpc())?;
                     }
 
+                    let proxy_api = TransactionStatusApiImpl::new(
```
wlawt, line 145: do we want this to be feature flagged so we can easily toggle it?

crates/transaction-status/src/lib.rs
danyalprout: Probably yes, my thinking was if we start as coarse as possible we can narrow it in future releases.

crates/node/src/main.rs
danyalprout: I don't think we need to feature flag it (though I maybe missing some case). My thinking is:

* External nodes/RPC nodes will have the sequencer URL set which will forward to us
* Mempool nodes will receive traffic and check their local mempool

wlawt (approved):

PR #278, opened 2025-12-12
test(rpc): add ERC-20 eth_call tests with proxy support

### LINKED ISSUE #151: improve `eth_call` tests to have more real-world scenarios
our `eth_call` tests are fairly basic right now, simply incrementing a counter contract

in the real world the use cases are usually a bit more complex, and have led to tests not being able to catch issues that come up post-deploy

at the very least, we should have tests for:
- basic ERC-20 functionality
- ERC-20 with a TransparentProxy wrapper (e.g. USDC token)

the proxy contract with it's `delegatecall` behavior has some edge cases that broke this behavior on mainnet today. tests can be the exact same between both with the exception of the `to` field

Himess:
Closes #151
## Summary

Adds integration tests for `eth_call` with ERC-20 token operations, addressing the gap identified after the v13.3 bug where proxy-based tokens broke on mainnet.

## Changes

**New Contracts:**
- `TestERC20.sol` - Simple ERC-20 with mint/burn for testing
- `TransparentProxy.sol` - EIP-1967 compliant proxy (USDC-style)

**New Tests (8 total):**
- `test_erc20_transfer` - basic transfer
- `test_erc20_mint` - minting tokens
- `test_erc20_burn` - burning tokens  
- `test_erc20_approve_transfer_from` - approval flow
- `test_transparent_proxy_erc20_transfer` - transfer via proxy
- `test_transparent_proxy_erc20_mint` - mint via proxy
- `test_transparent_proxy_erc20_burn` - burn via proxy
- `test_transparent_proxy_erc20_approve_transfer_from` - approval via proxy

Each test verifies state changes through `eth_call` after executing transactions, catching issues with delegatecall behavior that caused the mainnet bug.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Himess:
Quick update on what I tried:

1. Started with TestHarness + Engine API - blocks commit fine but eth_call returns 0 for contract state. Tried sleep delays, explicit block numbers, nothing worked.

2. Then marked tests as `#[ignore]` but that defeats the purpose.

3. Now trying FlashblocksHarness approach like flashblocks_rpc.rs does - manually building payloads. Waiting on CI.

Basically TestHarness commits to canonical chain but RPC doesn't see it, while FlashblocksHarness exposes state as pending block which eth_call can read.

crates/rpc/tests/eth_call_erc20.rs
```diff
@@ -0,0 +1,387 @@
...
+);
+
+sol!(
+    #[sol(rpc)]
+    TransparentProxy,
+    concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/../test-utils/contracts/out/TransparentProxy.sol/TransparentProxy.json"
+    )
+);
```
refcell: Since we have `base_reth_test_utils` as a dependency here, could we localize these into the test-utils crate and then export whatever is needed? That way we don't need to use an external path like `"/../test-utils/"` which is subject to change + breakage.

refcell (commented):
All the tests look good, but I think we can move a few things to `test-utils` that don't need to be in the rpc tests module so we're not relying on external files from inside the rpc crate.

crates/rpc/tests/eth_call_erc20.rs
Himess: Okey noted , i will do it

Himess:
@refcell i am still working to solve CI/Test .

Himess:
All 4 ERC-20 tests now pass locally:
- `test_erc20_deployment` ✅
- `test_proxy_erc20_deployment` ✅  
- `test_erc20_mint` ✅
- `test_proxy_erc20_mint` ✅

Existing `flashblocks_rpc` tests (19) still pass.

haardikk21:
Can you replace the contract implementations with `ERC20` or `MockERC20` from `solmate` as that's already a dependency and perhaps the upgradeable proxy with the one from OpenZeppelin instead? 

just to ensure it works with widely used things instead of needing to maintain these handwritten contracts in the repo. To have them compiled and outputted as artifacts you can do a simple script similar to the `DeployCounter.s.sol` which will ensure that running `just build-contracts` builds those two contracts as well

Himess:
> Can you replace the contract implementations with `ERC20` or `MockERC20` from `solmate` as that's already a dependency and perhaps the upgradeable proxy with the one from OpenZeppelin instead?
> 
> just to ensure it works with widely used things instead of needing to maintain these handwritten contracts in the repo. To have them compiled and outputted as artifacts you can do a simple script similar to the `DeployCounter.s.sol` which will ensure that running `just build-contracts` builds those two contracts as well

Okey okey it make sense.
I will update shortly.
Thanks for feedback btw

Himess:
Done. Updated in commit 9a9252e:
@haardikk21

haardikk21 (approved):

PR #300, opened 2026-01-06
feat: remove metadata receipts and account balances

meyer9:
The metadata field used to contain receipts and changed account balances as a performance optimization to be able to serve RPC requests without actually executing the block, but we still need to execute the block in most cases anyway. This removes the `receipts` and `changed_account_balances` fields from the response.

These are generated during flashblock execution instead.

This still preserves skipping execution of transactions that were executed in previous flashblocks as long as we have the receipt and state from the previous execution. In either the cached or executed path, `cumulative_gas` and `next_log_index` are updated.

The tests also needed to be fixed since they relied on the receipts for logs and the changed balances. Now, since the transactions are actually executed, we need to actually emit logs from the contracts when executed and send balance to the address rather than relying on `changed_account_balances`.

Fixes CHAIN-2657

### DIFF
--- a/crates/flashblocks/src/processor.rs
+++ b/crates/flashblocks/src/processor.rs
@@ -7,18 +7,19 @@ use std::{
 };
 
 use alloy_consensus::{
-    Header, TxReceipt,
+    Eip658Value, Header, TxReceipt,
     transaction::{Recovered, SignerRecoverable, TransactionMeta},
 };
 use alloy_eips::BlockNumberOrTag;
-use alloy_primitives::{Address, B256, BlockNumber, Bytes, Sealable, map::foldhash::HashMap};
+use alloy_op_evm::block::receipt_builder::OpReceiptBuilder;
+use alloy_primitives::{Address, B256, BlockNumber, Bytes, Sealable};
 use alloy_rpc_types::{TransactionTrait, Withdrawal};
 use alloy_rpc_types_engine::{ExecutionPayloadV1, ExecutionPayloadV2, ExecutionPayloadV3};
 use alloy_rpc_types_eth::state::StateOverride;
 use arc_swap::ArcSwapOption;
 use base_flashtypes::Flashblock;
 use eyre::eyre;
-use op_alloy_consensus::OpTxEnvelope;
+use op_alloy_consensus::{OpDepositReceipt, OpTxEnvelope};
 use op_alloy_network::TransactionResponse;
 use op_alloy_rpc_types::Transaction;
 use rayon::prelude::*;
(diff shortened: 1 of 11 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/flashblocks/src/processor.rs
```diff
@@ -333,7 +333,9 @@ where
                 .iter()
                 .map(|flashblock| flashblock.metadata.receipts.clone())
                 .fold(HashMap::default(), |mut acc, receipts| {
-                    acc.extend(receipts);
+                    if let Some(receipts) = receipts {
+                        acc.extend(receipts);
+                    }
```
refcell: Should we have/add a test for this?

refcell (dismissed):
Non-blocking nit: maybe add a test

crates/flashblocks/src/processor.rs
meyer9: yeah good call! Would be good to verify the behavior of missing deposit receipts on the returned nonce.

danyalprout:
Instead of making these optional, can we just remove the account balance/receipts completely? These are a holdover from the time when we didn't execute Flashblocks, but rather just stored them in a cache.

meyer9:
I think most of the failing tests are due to relying on passed in receipts rather than actually executing. I'll need to update the EVM bytecode to actually emit logs instead of assuming the receipts are correct.

meyer9:
Updated the PR to remove account balances as well (these should still serialize fine if they exist due to not using strict mode serde). Also, updated tests to actually emit logs instead of depending on receipts/balances passed in from the flashblock.

refcell (approved):
Very nice!

cb-heimdall:
Review Error for refcell @ 2026-01-07 16:56:18 UTC

refcell (approved):
oops forgot mfa

meyer9:
resolved branch conflicts ^

refcell (approved):

PR #308, opened 2026-01-07
feat: move state/receipt building to new method

meyer9:
Moves transaction processing logic of build_pending_state to a new struct. Also separates out reusing a cached execution from executing using the EVM.

This also removes block_hash from receipts and transactions which matches the spec and removes the need for calculating a useless block hash for every flashblock.

The next step is to move the for loop into this struct which will return an iterator of results.

Using this, we'll be able to prewarm accessed addresses using FALs and eventually process transactions in parallel. Moving execution to a new struct allows us to implement a trait for building flashblocks and switch out different implementations (prewarming, parallel, sequential).

### DIFF
--- a/crates/flashblocks/src/processor.rs
+++ b/crates/flashblocks/src/processor.rs
@@ -200,7 +200,7 @@ where
     ) -> eyre::Result<ExecutedPendingTransaction> {
         let tx_hash = transaction.tx_hash();
 
-        match self.evm.transact(transaction.clone()) {
+        match self.evm.transact(&transaction) {
             Ok(ResultAndState { state, result }) => {
                 let gas_used = result.gas_used();
                 let mut touched_balances = HashMap::default();
--- a/crates/flashblocks/src/processor.rs
+++ b/crates/flashblocks/src/processor.rs
@@ -7,12 +7,12 @@ use std::{
 };
 
 use alloy_consensus::{
-    Eip658Value, Header, TxReceipt,
+    Block, Eip658Value, Header, TxReceipt,
     transaction::{Recovered, SignerRecoverable, TransactionMeta},
 };
 use alloy_eips::BlockNumberOrTag;
 use alloy_op_evm::block::receipt_builder::OpReceiptBuilder;
-use alloy_primitives::{Address, B256, BlockNumber, Bytes, Sealable};
+use alloy_primitives::{Address, B256, BlockNumber, Bytes, U256, map::HashMap};
 use alloy_rpc_types::{TransactionTrait, Withdrawal};
 use alloy_rpc_types_engine::{ExecutionPayloadV1, ExecutionPayloadV2, ExecutionPayloadV3};
 use alloy_rpc_types_eth::state::StateOverride;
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

meyer9:
This needs a rebase on main once #307 is merged

refcell (commented):
If we're going to refactor the pending logic into a struct, could we take the extra step and move it into a new module to try to make this code a bit easier to read?

crates/flashblocks/src/processor.rs
```diff
@@ -52,6 +55,286 @@ pub enum StateUpdate {
...
+        Ok(ExecutedPendingTransaction { rpc_transaction, receipt, state, touched_balances })
+    }
+
+    /// Executes the transaction through the EVM and builds the result from scratch.
+    fn execute_with_evm(
+        &mut self,
+        transaction: Recovered<OpTxEnvelope>,
+        idx: usize,
+        effective_gas_price: u128,
+    ) -> eyre::Result<ExecutedPendingTransaction> {
```
refcell: This is a nice refactor!

crates/flashblocks/src/processor.rs
```diff
@@ -52,6 +55,286 @@ pub enum StateUpdate {
...
+        let tx_hash = transaction.tx_hash();
+
+        match self.evm.transact(transaction.clone()) {
```
refcell: I wish there was a nicer way to transact without consuming the `Recovered<OpTxEnvelope>` but maybe it's unavoidable.

crates/flashblocks/src/processor.rs
```diff
@@ -52,6 +55,286 @@ pub enum StateUpdate {
...
+                        .iter()
+                        .map(|(&key, slot)| (B256::from(key), B256::from(slot.present_value)));
+
+                    existing.extend(changed_slots);
+                }
+
+                self.cumulative_gas_used = self
+                    .cumulative_gas_used
+                    .checked_add(gas_used)
+                    .ok_or(eyre!("cumulative gas used overflow"))?;
```
refcell: Eventually we should move toward strongly typed errors so we can test more explicitly.

asserting against an enum variant is a lot easier than an eyre report and has compile-time type guarantees

crates/flashblocks/src/processor.rs
```diff
@@ -52,6 +55,286 @@ pub enum StateUpdate {
...
+    }
+
+    /// Executes the transaction through the EVM and builds the result from scratch.
+    fn execute_with_evm(
+        &mut self,
+        transaction: Recovered<OpTxEnvelope>,
+        idx: usize,
+        effective_gas_price: u128,
+    ) -> eyre::Result<ExecutedPendingTransaction> {
+        let tx_hash = transaction.tx_hash();
```
refcell: We compute the tx hash in `execute_transaction` which calls this method, maybe the `tx_hash` should just be passed in as an argument?

crates/flashblocks/src/processor.rs
```diff
@@ -52,6 +55,286 @@ pub enum StateUpdate {
...
+                            .transpose()
+                            .map_err(|_| eyre!("failed to load cache account for depositor"))?;
+
+                        self.receipt_builder.build_deposit_receipt(OpDepositReceipt {
+                            inner: receipt,
+                            deposit_nonce,
+                            deposit_receipt_version: is_canyon_active.then_some(1),
+                        })
+                    }
+                };
```
refcell: It feels strange to me that we have a receipt builder that then needs a lot more inline logic to build the receipt after the `build_receipt` method is called. Happy to address this in a follow-up pr or in the future, but wanted to call that out.

crates/flashblocks/src/processor.rs
meyer9: ah, good call! It accepts a reference, so no need to clone.

crates/flashblocks/src/processor.rs
meyer9: yup totally agree! I'll make this change in a separate PR tomorrow since there are a few errors we can update (plus wrapping EVMError).

crates/flashblocks/src/processor.rs
meyer9: Yeah it's definitely a clunky API. This also appears in rbuilder, so would probably make sense to fix there as well. It seems like receipt_builder should have everything it needs already to return an actual deposit receipt (chainspec and access to the DB). Created an issue here to track: https://github.com/base/node-reth/issues/309

crates/flashblocks/src/processor.rs
meyer9: I think this is cached by the transaction object, so shouldn't be any extra work: https://github.com/alloy-rs/alloy/blob/main/crates/consensus/src/signed.rs#L127

meyer9:
Yup, good idea! Made a few changes: moved the state builder it to a new `state_builder` module, passed tx reference to `transact` to avoid cloning, and removed `touched_account_balances` since that's duplicating `EvmState` that's already returned.

haardikk21 (approved):

crates/flashblocks/src/processor.rs
refcell: Great thank you!

refcell (approved):
Looks great thanks @meyer9

PR #315, opened 2026-01-08
chore(flashblocks): Strong, Domain Error Typing

### LINKED ISSUE #312: feat(flashblocks): Replace `eyre::Result` with Domain-Specific Error Types
### Summary

Introduce `StateProcessorError` enum to replace generic `eyre::Result` throughout the flashblocks crate.

### Problem

Current error handling uses `eyre::Result` everywhere, which:
- Loses semantic information (was it validation? execution? storage?)
- Makes error matching in tests difficult and fickle
- Provides no structured context for debugging
- Can't distinguish recoverable vs unrecoverable errors

Example of current ambiguous errors:
```rust
Err(eyre!("missing headers"))
Err(eyre!("no flashblocks"))
Err(eyre!("first flashblock missing base"))
```

### Solution

**New file:** `src/error.rs`

```rust
#[derive(Debug, thiserror::Error)]
pub enum StateProcessorError {
    // Protocol errors (upstream issues)
    #[error("invalid flashblock sequence: expected {expected}, got {got}")]
    InvalidSequence { expected: u64, got: u64 },

    #[error("flashblock missing base payload")]
    MissingBase,

    #[error("cannot assemble block from empty flashblocks")]
    EmptyFlashblocks,

    // Operational errors (sync issues)
    #[error("missing canonical header for block {block_number}")]
    MissingCanonicalHeader { block_number: u64 },

    #[error("state provider error: {0}")]
    StateProvider(String),

    // Execution errors
    #[error("transaction execution failed: {tx_hash}")]
    TransactionExecution { tx_hash: B256 },

    #[error("sender recovery failed: {0}")]
    SenderRecovery(String),

    // Build errors
    #[error("missing headers in pending blocks")]
    MissingHeaders,

    #[error("missing flashblocks in pending blocks")]
    MissingFlashblocks,
}
```

### Benefits

```rust
// Before: opaque error
Err(eyre!("first flashblock missing base"))

// After: matchable, structured
Err(StateProcessorError::MissingBase)

// Enables test assertions
assert!(matches!(result, Err(StateProcessorError::MissingBase)));

// Enables targeted error handling
match error {
    StateProcessorError::MissingCanonicalHeader { .. } => /* retry sync */,
    StateProcessorError::InvalidSequence { .. } => /* wait for recovery */,
    _ => /* log and continue */,
}
```

### Acceptance

- [ ] All `eyre!()` calls replaced with specific variants
- [ ] Error types exported from `lib.rs`
- [ ] Unit tests can match on specific errors
- [ ] Integration tests still pass

refcell:
### Description

Fixes #312

Introduces a new `StateProcessorError` type that strongly types errors in the flashblocks crate.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/flashblocks/src/error.rs
```diff
@@ -0,0 +1,247 @@
...
+use alloy_consensus::crypto::RecoveryError;
+use alloy_primitives::{Address, B256};
+use thiserror::Error;
+
+/// A type alias for `Result<T, StateProcessorError>`.
+pub type Result<T> = std::result::Result<T, StateProcessorError>;
+
+/// Errors that can occur during flashblock state processing.
+#[derive(Debug, Error)]
+pub enum StateProcessorError {
```
refcell: I don't particularly like having this enum contain the aggregate of all error variants. Will attempt to break it up locally and see what that looks like.

crates/flashblocks/src/error.rs
```diff
@@ -0,0 +1,297 @@
...
+
+/// Errors related to state provider and infrastructure operations.
+#[derive(Debug, Clone, Eq, PartialEq, Error)]
+pub enum ProviderError {
+    /// Missing canonical header for a given block number.
+    #[error("missing canonical header for block {block_number}")]
+    MissingCanonicalHeader {
+        /// The block number for which the header is missing.
+        block_number: u64,
+    },
```
haardikk21: This substring of the original error
> This can be ignored if the node has recently restarted, restored from a snapshot or is still syncing.

should be added back imo. 

When a reth node is restarted, it has some behaviour where it doesn't commit data to the underlying database immediately after catching up to tip. Previously before the explanation that it is okay under those cases as part of the error message, external folks regularly complained that fetching latest block number from the Reth node shows it's at tip but Flashblocks refuse to be processed so something is wrong with our implementation - when it's just the behavior of how Reth chooses to commit to the DB

Anecdotally, it appears to take ~10 minutes after catching up to tip before this error goes away but that level of specificity is probably unnecessary

crates/flashblocks/src/error.rs
```diff
@@ -0,0 +1,297 @@
...
+    BlockConversion(String),
+
+    /// Failed to load cache account for depositor.
+    #[error("failed to load cache account for deposit transaction sender")]
+    DepositAccountLoad,
+}
+
+impl From<RecoveryError> for ExecutionError {
+    fn from(err: RecoveryError) -> Self {
+        Self::SenderRecovery(err.to_string())
```
meyer9, line 85: non-blocking, but just for my understanding, when is it a good idea to wrap the error vs convert to string?

meyer9 (approved):

cb-heimdall:
Review Error for meyer9 @ 2026-01-08 20:06:28 UTC

crates/flashblocks/src/error.rs
haardikk21: if you want to handle errors higher up in the callsite, it's easier and safer to pattern-match with strongly typed variants rather than string matching which can break easily and unknowingly. presumably as this codebase gets more modular and abstracted, things like that will be required so this is setting the stage

meyer9: yup, I just mean if there's an inner error that we're wrapping. For example, we couldn't pattern match on the inner error if we convert it to a string, but that's probably fine for super low-level errors like RecoveryError.

refcell: Apologize for not being clear why we're converting this to a string.

The [`RecoveryError`](https://docs.rs/alloy-consensus/1.3.0/alloy_consensus/crypto/struct.RecoveryError.html) type here doesn't implement `Clone + PartialEq + Eq`, of which `PartialEq` / `Eq` is required for a comparison.

I figured the most complete thing to do here is just to grab the error in string format rather than manually stubbing out a wrapper type or just not bubbling up the `RecoveryError` at all.

meyer9: no worries! It was mainly just out of curiosity. Makes sense - so most errors can be wrapped as long as there are no required traits missing.

meyer9 (dismissed):

crates/flashblocks/src/error.rs
refcell: Got it - that makes sense to me. Added back.

haardikk21 (approved):

PR #322, opened 2026-01-09
feat(fbal): FBAL Builder database wrapper

haardikk21:
Alternative implementation for a FBAL Builder instead of #273. Contains all the same test cases (+ improved with having stronger assertions). Retains all the same functionality so far. 

Makes it easier to build custom Payload Builders and Block Executors with this pattern rather than the `revm` inspector approach used previously.

Per the spec, there are some minor bugs that remain to be fixed that were also bugs in the older PR #273. For now, this PR is 1:1 compatible with the older PR. Additional identified bugs will be resolved in a followup.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/fbal/src/db.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /// Attempts to commit the changes to the underlying database
+    /// as well as applies account/storage changes to the access list builder
+    fn try_commit(
+        &mut self,
+        changes: HashMap<Address, Account>,
+    ) -> Result<(), <DB as Database>::Error> {
+        for (address, account) in changes.iter() {
+            let changes = self.access_list.changes.entry(*address).or_default();
```
refcell: I'm a bit confused here how this works -- it looks like we're shadowing the `changes` variable that's passed in as an argument. Then after the loop, we commit the changes to the db, but this `changes` would goes out of scope each for loop iteration? Maybe I'm missing something here.

haardikk21: I can adjust it to not be a shadowed variable, that was unintentional and didn't realize I had done that.

I'm not sure I understand the rest of your question. If you mean what happens to the `let changes = ...` variable, nothing. It's a mutable reference we are using to update stuff in `self.access_list`. It going out of scope after the iteration is fine

crates/fbal/README.md
refcell, line 1: Very nice `README.md` -- thank you for writing this!

crates/fbal/tests/builder/main.rs
```diff
@@ -0,0 +1,127 @@
...
+mod transfers;
+
+sol!(
+    #[sol(rpc)]
+    AccessListContract,
+    concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/../test-utils/contracts/out/AccessList.sol/AccessList.json"
+    )
+);
```
refcell, line 36: I'm not sure of the best way to organize this - we should brainstorm a bit - but I don't think it generally makes sense to have individual crates depend on `test-utils` in any way. If we need to use the accesslist.json here, we should probably localize it and then re-export the type under the `test-utils` feature flag so the `test-utils` crate can re-use this type.

Not a blocker for this PR, but worth investigation and coming to a consensus

haardikk21: 😅 hahah. yeah this repo used to have localized testing setups not too long ago, the `test-utils` consolidation is fairly recent. 

note however the JSON is a build artifact of contracts that live within `test-utils`. I originally did have contracts at the repo root which tbh i preferred, but those got shifted to the `test-utils` crate so now we have this situation. I don't think we need to re-export the type in hopes `test-utils` will ever use it - the dependency relationship is almost always `X (depends on) test-utils` (for harnesses, fixtures, etc) and very rarely the other way around

crates/fbal/src/db.rs
refcell: Got it thanks for the clarity. I was mostly confused because it's not clear it's a mutable reference and it looked like maybe those changes were not being committed since later on the `changes` argument was being committed to the db.

refcell: This looks good now!

crates/fbal/src/db.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// as well as applies account/storage changes to the access list builder
+    fn try_commit(
+        &mut self,
+        changes: HashMap<Address, Account>,
+    ) -> Result<(), <DB as Database>::Error> {
+        for (address, account) in changes.iter() {
+            let account_changes = self.access_list.changes.entry(*address).or_default();
+
+            // Update balance, nonce, and code
+            match self.db.basic(*address)? {
```
meyer9, line 65: Could also check `account.is_touched()` before running this DB lookup (very small perf optimization)

crates/fbal/src/types.rs
```diff
@@ -0,0 +1,36 @@
...
+/// FlashblockAccessList represents the access list for a single flashblock
+pub struct FlashblockAccessList {
+    /// All the account changes in this access list
+    pub account_changes: Vec<AccountChanges>,
+    /// Minimum txn index from the full block that's included in this access list
+    pub min_tx_index: u64,
+    /// Maximum txn index from the full block that's included in this access list
+    pub max_tx_index: u64,
+    /// keccak256 hash of the RLP-encoded account changes list
+    pub fal_hash: B256,
```
meyer9, line 16: I know the original EIP had access list hash stored in the header to ensure the access list was determined by the time the block was built, but I don't necessarily see the point of including the hash here - do you have a specific use case in mind? Can also address in a future PR!

refcell (approved):
Looks great!

crates/fbal/tests/builder/transfers.rs
```diff
@@ -0,0 +1,168 @@
...
+    let access_list = execute_txns_build_access_list(vec![tx], None, None)
+        .expect("access list build should succeed");
+
+    // Verify we got an access list (precompiles/system contracts should be touched)
+    assert!(!access_list.account_changes.is_empty(), "Access list should not be empty");
+}
+
+#[test]
+/// Tests that a single ETH transfer is included in the access list
+fn test_single_transfer() {
```
meyer9, line 25: can you add a test for sending 0 balance? In this case, the receiving account shouldn't be included in the access list since it hasn't changed, but the account will be considered touched.

meyer9 (approved):
LGTM! Left a few non-blocking comments.

crates/fbal/tests/builder/transfers.rs
meyer9: feel free to take or leave since we're not even using `account.is_touched` right now.

cb-heimdall:
Review Error for meyer9 @ 2026-01-09 15:38:44 UTC

crates/fbal/src/db.rs
haardikk21: I'll try it out.

Per the spec we need to have an entry in AccountChanges even for addresses that were simply loaded/read without having any modifications made (just w/ empty changes though). I'm wondering if there's any case where is_touched is false but we also haven't previously filed an entry for that account via our wrapper around db.basic(...)?

meyer9 (approved):

crates/fbal/src/types.rs
haardikk21: not as of today, think it's an artifact of having modified the BAL spec. if necessary we can remove this later, but will think about it a bit more

crates/fbal/tests/builder/transfers.rs
haardikk21: that's interesting, i'll add it in a followup alongside some other fixes that need to be made

PR #387, opened 2026-01-12
feat(metering): use pending flashblocks state for bundle metering

niran:
## Summary

Enable bundle metering to use the current pending flashblocks state instead of just the canonical block state. This ensures that bundle simulations see the same state that will be used when the bundle is actually included in a flashblock.

### Metering Crate
- **meter.rs**: Add `PendingState` struct containing `BundleState` and optional `PendingTrieInput`. Update `meter_bundle` to accept an optional `PendingState` parameter. Use `with_bundle_prestate()` to layer pending flashblock state on top of canonical state.
- **rpc.rs**: Update `MeteringApiImpl` to integrate with flashblocks API. Get pending blocks state and pass it to `meter_bundle`. Include `state_flashblock_index` in response.
- **extension.rs**: Update `MeteringExtension` to accept `FlashblocksConfig` for state access.
- **lib.rs**: Export `PendingState` and `PendingTrieInput`.

### Flashblocks Crate
- **state.rs**: Add `Default` impl for `FlashblocksState`.
- **extension.rs**: Add `Clone` derive to `FlashblocksConfig`.
- **tests/state.rs**: Add integration tests verifying that bundle metering correctly sees pending flashblock state changes.

## Architecture

```
Bundle Metering with Pending State:

┌─────────────────────────────────┐
│  State (bundle tracking)        │  ← Tracks bundle writes
│  with_bundle_prestate()         │  ← Includes flashblock state as prestate
├─────────────────────────────────┤
│  StateProviderDatabase          │  ← Reads from canonical chain state
└─────────────────────────────────┘
```

## Test Plan

- [x] Bundle metering correctly fails for nonce=1 tx without flashblock state (canonical nonce is 0)
- [x] Bundle metering succeeds for nonce=1 tx with flashblock state showing pending nonce=1
- [x] State visibility tests pass
- [x] `cargo +nightly clippy` passes
- [x] `cargo +nightly fmt` passes

### DIFF
--- a/crates/client/metering/src/extension.rs
+++ b/crates/client/metering/src/extension.rs
@@ -1,22 +1,27 @@
 //! Contains the [`MeteringExtension`] which wires up the metering RPC surface
 //! on the Base node builder.
 
+use std::sync::Arc;
+
 use base_client_node::{BaseNodeExtension, FromExtensionConfig, OpBuilder};
+use base_flashblocks::{FlashblocksConfig, FlashblocksState};
 use tracing::info;
 
 use crate::{MeteringApiImpl, MeteringApiServer};
 
 /// Helper struct that wires the metering RPC into the node builder.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug)]
 pub struct MeteringExtension {
     /// Whether metering is enabled.
     pub enabled: bool,
+    /// Optional Flashblocks configuration (includes state).
+    pub flashblocks_config: Option<FlashblocksConfig>,
 }
 
 impl MeteringExtension {
     /// Creates a new metering extension.
-    pub const fn new(enabled: bool) -> Self {
-        Self { enabled }
+    pub const fn new(enabled: bool, flashblocks_config: Option<FlashblocksConfig>) -> Self {
+        Self { enabled, flashblocks_config }
     }
 }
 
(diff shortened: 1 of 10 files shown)


crates/client/metering/src/extension.rs
```diff
@@ -27,19 +39,66 @@ impl BaseNodeExtension for MeteringExtension {
...
+            let flashblocks_state = flashblocks_config.as_ref().map(|cfg| {
+                shared_state
+                    .get_or_init(|| {
+                        Arc::new(FlashblocksState::new(
+                            ctx.provider().clone(),
+                            cfg.max_pending_blocks_depth,
+                        ))
+                    })
+                    .clone()
+            });
```
danyalprout: I think the metering crate needing the FlashblocksConfig to init this feels like a design flaw in our node setup.

I would suggest we make the following change in `main.rs`

```
let fb_state = Arc::new(FlashblocksState::new())

install_ext::<Flashblocks>({ fb_state.clone() })
install_ext::<Metering>({ fb_state })
```

We should be able to just totally remove the oncecell setup and dependency. This was the symptom of a time where we had two tokio routines racing to initialize it, but I suspect we can get rid of it now.

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/client/metering/src/extension.rs
niran: FlashblocksState needs a `Client`/`Provider` object, which we currently get from the context passed to `extend_rpc_modules`. Unless the `builder` provides a way to get a reference to something that will eventually give us a provider, we're kind of stuck. One option we still have is to move 
the cell inside `FlashblocksState`, which should be able to give us your code snippet above. But it just moves the mess somewhere else, and probably adds more of it since FlashblocksState would be full of checks for an empty cell.

danyalprout: Gotcha, lets pair tomorrow morning, I have an idea

refcell (approved):
This looks good to me pending @danyalprout

danyalprout (approved):
thank you! this looks great

PR #388, opened 2026-01-12
feat(metering): cache pending trie nodes for faster state root calculation

niran:
## Summary

Cache trie nodes computed during pending flashblock state root calculations and reuse them across bundle metering simulations. This optimization ensures that each bundle simulation only pays the I/O cost for its own state changes, not for all previous flashblock state changes.

## Problem

Without caching, every bundle simulation must:
1. Compute the hashed post-state for all flashblock changes
2. Read trie nodes from disk for all accounts/storage touched by flashblocks
3. Then compute the bundle's incremental changes on top

With many bundles being simulated per flashblock, this redundant I/O becomes a bottleneck.

## Solution

Introduce `PendingTrieCache` that:
1. Caches `TrieUpdates` and `HashedPostState` from pending state root computation
2. Uses `arc_swap::ArcSwap` for lock-free reads and atomic updates
3. Provides `ensure_cached()` for lazy computation and caching
4. Prepends cached data to bundle's `TrieInput` via `prepend_cached()`

## Changes

- **metrics.rs**: New module with metrics for cache performance tracking.
- **trie_cache.rs**: New module implementing `PendingTrieCache` with `ensure_cached()` method that computes and caches the pending trie input, or returns the cached entry if it matches the current flashblock.
- **meter.rs**: Add `PendingTrieInput` struct containing `TrieUpdates` and `HashedPostState`. Add shared `compute_pending_trie_input()` helper with metrics. Update `meter_bundle` to use cached trie input when available. Use `TrieInput::prepend_cached()` to prepend pending trie for incremental state root calculation.
- **rpc.rs**: Integrate with `PendingTrieCache` to compute/retrieve cached trie data for bundle simulations.
- **lib.rs**: Export `PendingTrieCache` and `PendingTrieInput`.
- **Cargo.toml**: Add `arc-swap`, `metrics`, `metrics-derive`, and `reth-trie-common` dependencies.

## Metrics

New Prometheus metrics for monitoring cache performance:
- `reth_metering_pending_trie_cache_hits` - counter for cache hits
- `reth_metering_pending_trie_cache_misses` - counter for cache misses (trie computed)
- `reth_metering_pending_trie_compute_duration` - histogram of compute time in seconds

## Performance Impact

Before: Each bundle simulation reads trie nodes for all flashblock changes
After: Each bundle simulation only reads trie nodes for its own changes

The improvement scales with:
- Number of accounts/storage slots modified by flashblocks
- Number of bundles simulated per flashblock

## Test Plan

- [x] Existing metering tests pass (20 tests)
- [x] State root calculation produces correct results with cached data
- [x] `cargo +nightly clippy` passes
- [x] `cargo +nightly fmt` passes

### DIFF
--- a/crates/client/metering/Cargo.toml
+++ b/crates/client/metering/Cargo.toml
@@ -25,6 +25,7 @@ reth-optimism-evm.workspace = true
 reth-optimism-chainspec.workspace = true
 reth-optimism-primitives.workspace = true
 reth-revm.workspace = true
+reth-trie-common.workspace = true
 
 # revm
 revm-database.workspace = true
@@ -38,6 +39,9 @@ alloy-eips.workspace = true
 jsonrpsee.workspace = true
 
 # misc
+arc-swap.workspace = true
+metrics.workspace = true
+metrics-derive.workspace = true
 tracing.workspace = true
 eyre.workspace = true
 serde.workspace = true
--- a/crates/client/metering/src/meter.rs
+++ b/crates/client/metering/src/meter.rs
@@ -11,13 +11,58 @@ use reth_optimism_chainspec::OpChainSpec;
 use reth_optimism_evm::{OpEvmConfig, OpNextBlockEnvAttributes};
 use reth_primitives_traits::SealedHeader;
 use reth_revm::{database::StateProviderDatabase, db::State};
+use reth_trie_common::TrieInput;
 use revm_database::states::{BundleState, bundle_state::BundleRetention};
 
-/// State from pending flashblocks that is used as a base for metering
+use crate::metrics::Metrics;
+
+/// Computes the pending trie input from the bundle state.
+///
+/// This function records metrics for cache misses and compute duration.
+pub(crate) fn compute_pending_trie_input<SP>(
+    state_provider: &SP,
+    bundle_state: &BundleState,
+    metrics: &Metrics,
+) -> EyreResult<PendingTrieInput>
+where
+    SP: reth_provider::StateProvider + ?Sized,
+{
+    metrics.pending_trie_cache_misses.increment(1);
+    let start = Instant::now();
+
+    let hashed = state_provider.hashed_post_state(bundle_state);
+    let (_state_root, trie_updates) = state_provider.state_root_with_updates(hashed.clone())?;
+
+    let elapsed = start.elapsed();
+    metrics.pending_trie_compute_duration.record(elapsed.as_secs_f64());
+
+    Ok(PendingTrieInput { trie_updates, hashed_state: hashed })
+}
+
+/// Pre-computed trie input from pending state for efficient state root calculation.
+///
+/// When metering bundles on top of pending flashblocks, we first compute the trie updates
+/// and hashed state for the pending state. This can then be prepended to the bundle's
+/// trie input, so state root calculation only performs I/O for the bundle's changes.
 #[derive(Debug, Clone)]
-pub struct FlashblocksState {
-    /// The accumulated bundle of state changes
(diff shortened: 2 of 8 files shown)


crates/client/metering/src/meter.rs
```diff
@@ -50,25 +70,43 @@ pub fn meter_bundle<SP>(
...
+            if let Some(ref cached) = ps.trie_input {
+                Ok(cached.clone())
+            } else {
+                let hashed = state_provider.hashed_post_state(&ps.bundle_state);
+                let (_state_root, trie_updates) =
+                    state_provider.state_root_with_updates(hashed.clone())?;
+                Ok(PendingTrieInput { trie_updates, hashed_state: hashed })
+            }
+        })
+        .transpose()?;
```
refcell, line 120: Is it worth placing this behind a separate metric or computing the trie updates isn't a concern?

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (commented):
Looks like this needs a rebase, but otherwise, core triedb cache changes look solid

crates/client/metering/src/meter.rs
niran: Added metrics to track cache hits/misses and compute duration:

- `reth_metering_pending_trie_cache_hits` - counter for cache hits
- `reth_metering_pending_trie_cache_misses` - counter for cache misses (trie computed)
- `reth_metering_pending_trie_compute_duration` - histogram of compute time in seconds

The compute logic is now in a shared `compute_pending_trie_input()` helper so both the cache and inline paths use the same code with metrics.

crates/client/metering/Cargo.toml
```diff
@@ -51,5 +55,6 @@ reth-transaction-pool = { workspace = true, features = ["test-utils"] }
 alloy-genesis.workspace = true
 alloy-rpc-client.workspace = true
 op-alloy-consensus.workspace = true
+once_cell.workspace = true
```
danyalprout: out of curiosity, why do we need to add this?

danyalprout (dismissed):

danyalprout (approved):

crates/client/metering/Cargo.toml
niran: We don't - removed. Must have been a rebase artifact that got picked up from an earlier version of the branch.

refcell (approved):

PR #402, opened 2026-01-13
feat: copy Optimism ExEx crates

meyer9:
:shipit:


refcell:
Screenshot 2026-01-12 at 7 39 44 PM
image

haardikk21:
who would be the best person to review this? i know you've been working on this stuff in diff repos. has anyone else from our team been involved or mostly just OP people?

meyer9:
unfortunately nobody else from Base was involved. This is dependent on v1.10.0 anyway, so we can figure it out once that's merged. I think I plan on doing a few somewhat reviewable commits:

1. Copy in code from their repo (maybe possible to preserve Git history)
  a. This is all well-reviewed and tested.
2. Update their code with a few small changes required to work with upstream Reth (should be reviewed by us)
3. Glue code to work with our extension system (should be reviewed by us)
4. Remove unnecessary code like some of their RPC/ExEx stuff (very straightforward - check unused and remove)
5. Formatting fixes

I don't think submodules will work since we have to apply some fixes to get it working with upstream Reth. I don't think importing as a crate will work because then we're stuck with their version of Reth and dependent on them updating.

refcell:
I'm happy to give it thorough review when desired.

I would also +1 on advising against submodules.

meyer9:
Superseded by #579

PR #446, opened 2026-01-14
fix(client): accumulate RPC extension hooks instead of replacing

### LINKED ISSUE #438: fix(client): RPC extensions overwrite each other due to reth hook replacement model
## Problem

Reth's `extend_rpc_modules` and `on_node_started` use a **replacement model** - each call replaces the previous hook rather than accumulating. This causes only the last extension's hooks to execute.

Current behavior in `bin/node/src/main.rs`:
1. `TxPoolExtension` calls `extend_rpc_modules` → sets hook  1
2. `MeteringExtension` calls `extend_rpc_modules` → **replaces** with hook 2  
3. `FlashblocksExtension` calls `extend_rpc_modules` → **replaces** with hook 3

**Result:** Only FlashblocksExtension's RPC methods are registered. `base_meterBundle` and `base_transactionStatus` endpoints are missing even when `--enable-metering` is passed.

## Symptoms

- "Starting Metering RPC" log message does NOT appear
- "Starting Transaction Status RPC" log message does NOT appear
- "Starting Flashblocks RPC" log message DOES appear (last extension wins)
- `base_meterBundle` RPC endpoint returns "method not found"

## Root Cause

Confirmed by reading reth v1.9.3 source: `RpcHooks` struct stores hooks as single `Box` fields with `set_*` methods, not as vectors with `push` methods. Each `extend_rpc_modules` call overwrites the previous hook.

## Solution

Create a `BaseBuilder` wrapper that accumulates `extend_rpc_modules` and `on_node_started` hooks in `Vec`s, then applies them all in a single reth call when launching. This maintains the existing extension API while fixing the accumulation issue.

### Implementation

1. Create `BaseBuilder` in `crates/client/node/src/builder.rs` with:
   - `add_rpc_module()` - accumulates RPC hooks
   - `add_node_started_hook()` - accumulates node-started hooks
   - `finalize()` - combines all hooks into single reth calls

2. Update `BaseNodeExtension` trait to use `BaseBuilder` instead of `OpBuilder`

3. Update all extensions to use new method names

## Affected Files

- `crates/client/node/src/builder.rs` (NEW)
- `crates/client/node/src/lib.rs`
- `crates/client/node/src/extension.rs`
- `crates/client/node/src/runner.rs`
- `crates/client/txpool/src/extension.rs`
- `crates/client/metering/src/extension.rs`
- `crates/client/flashblocks/src/extension.rs`
- `crates/client/node/src/test_utils/node.rs`

niran:
## Summary

- Introduces `BaseBuilder` wrapper that accumulates `extend_rpc_modules` and `on_node_started` hooks in vectors instead of replacing them
- Updates all extensions (`TxPoolExtension`, `MeteringExtension`, `FlashblocksExtension`) to use new `add_rpc_module()` and `add_node_started_hook()` methods
- Adds integration test verifying multiple RPC extensions are accessible simultaneously

## Problem

Reth's `extend_rpc_modules` and `on_node_started` use a replacement model—each call overwrites the previous hook. This caused only the last extension's hooks to execute, making endpoints like `base_meterBundle` and `base_transactionStatus` unavailable even when their respective extensions were enabled.

## Solution

The new `BaseBuilder` wrapper collects all hooks during extension setup, then applies them all in a single reth call when launching. This maintains the existing extension API while fixing the accumulation issue.

Closes #438

## Test plan

- [ ] Verify `cargo test --package base-client-node --features test-utils` passes
- [ ] Confirm all RPC endpoints (`base_transactionStatus`, `base_meterBundle`, flashblocks) are accessible when multiple extensions are enabled

### DIFF
--- a/crates/client/flashblocks/src/extension.rs
+++ b/crates/client/flashblocks/src/extension.rs
@@ -3,7 +3,7 @@
 
 use std::sync::Arc;
 
-use base_client_node::{BaseNodeExtension, FromExtensionConfig, OpBuilder};
+use base_client_node::{BaseBuilder, BaseNodeExtension, FromExtensionConfig};
 use reth_chain_state::CanonStateSubscriptions;
 use tokio_stream::{StreamExt, wrappers::BroadcastStream};
 use tracing::info;
@@ -50,7 +50,7 @@ impl FlashblocksExtension {
 
 impl BaseNodeExtension for FlashblocksExtension {
     /// Applies the extension to the supplied builder.
-    fn apply(self: Box<Self>, builder: OpBuilder) -> OpBuilder {
+    fn apply(self: Box<Self>, builder: BaseBuilder) -> BaseBuilder {
         let Some(cfg) = self.config else {
             info!(message = "flashblocks integration is disabled");
             return builder;
@@ -64,7 +64,7 @@ impl BaseNodeExtension for FlashblocksExtension {
         let state_for_start = state;
 
         // Start state processor, subscriber, and canonical subscription after node is started
-        let builder = builder.on_node_started(move |ctx| {
+        let builder = builder.add_node_started_hook(move |ctx| {
             info!(message = "Starting Flashblocks state processor");
             state_for_start.start(ctx.provider().clone());
             subscriber.start();
@@ -84,7 +84,7 @@ impl BaseNodeExtension for FlashblocksExtension {
         });
 
         // Extend with RPC modules
-        builder.extend_rpc_modules(move |ctx| {
+        builder.add_rpc_module(move |ctx| {
             info!(message = "Starting Flashblocks RPC");
 
             let api_ext = EthApiExt::new(
--- a/crates/client/flashblocks/src/test_harness.rs
+++ b/crates/client/flashblocks/src/test_harness.rs
@@ -13,7 +13,7 @@ use std::{
 };
 
 use base_client_node::{
-    BaseNodeExtension,
+    BaseBuilder, BaseNodeExtension,
     test_utils::{
         LocalNode, NODE_STARTUP_DELAY_MS, TestHarness, build_test_genesis, init_silenced_tracing,
     },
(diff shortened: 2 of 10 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/client/node/Cargo.toml
```diff
@@ -48,6 +48,13 @@ test-utils = [
 	"reth-provider/test-utils",
 ]
 
+[dev-dependencies]
+base-metering = { path = "../metering" }
+base-txpool = { path = "../txpool" }
+jsonrpsee = { workspace = true, features = ["http-client"] }
+serde_json = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
```
danyalprout: We don't want to have client-node depend on base-metering (as it introduces a circular dependency).

This is a very useful test (and one that was missing!), but for now lets just add an issue for adding this. We can either add it to our system test framework or have a top level client client

danyalprout (changes requested):
looks great to me, would like to avoid the dependency loop on `client-node` though

crates/client/node/Cargo.toml
haardikk21: Sidenote, I wonder where tests like these should live. They're not crossing top-level boundaries (i.e. need both builder + client) - but also are not limited to a single crate. 

Do we introduce some tests within the `bin/...` crates to have tests that are constrained to only builder or client - but work across multiple sub-packages? 

Or should they just go alongside the system tests when we do those even though they dont require the builder particularly

niran: Removed and added #458.

danyalprout (approved):
🚢

danyalprout:
@niran we'll need to backport this commit onto `releases/v0.3.0`.

Issue #444, opened 2026-01-14
chore(ci): consolidate CI checks for builder and client


haardikk21:
Things like `clippy` and `cargo fmt` etc currently have different recipes in the Justfile for builder vs. client

These should be consolidated. Specifically, the builder specific ones be removed and just run as part of the general `ci` command and dependencies

Farukest:
Hi @danyalprout, opened a PR for this. Let me know if any changes are needed.

PR #452, opened 2026-01-14
chore(workspace): Justfile and Workspace Cleanup

refcell:
### Description

This PR now does a few things.

- Adds the `--workspace` cli flag on linting, formatting, and check cargo commands in the Justfile.
- Removes the now redundant custom builder Justfile targets.
- Fixes lint errors that result from the new `--workspace` cli flag across the workspace.
- Removes the redundant `builder.yml` github action workflow since the builder is now covered in the regular ci workflow.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Justfile
haardikk21: I ran into this earlier. These *do not* cover the builder because it doesn't mention `--workspace` and therefore just targets default members - of which the builder is not a part. This is because we explicitly define `default-members` in our Cargo.toml. If we had not defined it, it would run on all packages. However when it's explicitly defined, it only covers those unless we add `--workspace`

I was getting confused why after running `just fix` multiple times, when I opened the PR, the CI failed with builder clippy check until I ran `just fix-clippy-builder`. It's because the latter explicitly targets that package - whereas the general ones currently *do not* hit anything except default-members.

Heck we might even be not running them on other crates except the top-level `bin` ones right now.

haardikk21 (dismissed):
Don't think it works in the current state. Also need to adjust Github workflows to match - they're attempting to run recipes that don't exist anymore.

Also, closes #444

cb-heimdall:
Review Error for haardikk21 @ 2026-01-14 13:52:26 UTC

Justfile
refcell: I think you're right here. I've modified the justfile targets to use the `--workspace` flag.

danyalprout (approved):

PR #464, opened 2026-01-14
feat(node): extract base node to customize validator

meyer9:
This will allow customizing parts of the node further. The main goal is to customize the engine validator which handles executing blocks which need to be validated. (#465 builds on this to implement this)

I copied the `OpNode` implementation and removed all of the components we're not changing. This left me with just the node, addons and addon builder.

We can also probably remove some features we're not using from here, but to keep this PR straightforward, I haven't changed it from the Reth version other than removing code that isn't going to be customized.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

haardikk21:
<s>Since the goal is specifically to be able to add some custom addons/components to the node, I wonder if this could be further simplified and we can just install the addon directly either in:
- https://github.com/base/node-reth/blob/main/bin/node/src/main.rs
- or, https://github.com/base/node-reth/blob/fcb2d34e380ab4b11734ddd1be7a4a0d8e351fd2/crates/client/node/src/runner.rs#L47

Is it not possible to just add an additional component without needing to rewire the whole `OpNode` setup? Based on Reth's examples of adding custom components, I haven't seen them needing to rewire `OpNode` or `EthereumNode` to be able to add a custom component anywhere</s>

ignore

crates/client/node/src/node.rs
```diff
@@ -0,0 +1,790 @@
...
+            })
+            .map(|(historical_rpc, bedrock_block)| -> eyre::Result<_> {
+                info!(target: "reth::cli", %bedrock_block, ?historical_rpc, "Using historical RPC endpoint pre bedrock");
+                let provider = ctx.node.provider().clone();
+                let client = HistoricalRpcClient::new(&historical_rpc)?;
+                let layer = HistoricalRpc::new(provider, client, bedrock_block);
+                Ok(layer)
+            })
+            .transpose()?
+            ;
```
refcell, line 489: Ah yea this can go

meyer9:
Opened an issue here for cleaning up some OP specific stuff we don't need: https://github.com/base/node-reth/issues/467

crates/client/node/src/node.rs
```diff
@@ -0,0 +1,790 @@
...
+    RpcMiddleware: Send,
+{
+    type ValidatorBuilder = EVB;
+
+    fn engine_validator_builder(&self) -> Self::ValidatorBuilder {
+        EngineValidatorAddOn::engine_validator_builder(&self.rpc_add_ons)
+    }
+}
+
+/// A regular optimism evm and executor builder.
```
refcell, line 607: Maybe we could split everything here and below into a new builder.rs module just so this file doesn't bloat

refcell (approved):
As discussed offline, other cleanup can be done in followups :)

crates/client/node/src/node.rs
meyer9: sounds good, yeah maybe the Addons and AddonsBuilder can be moved. I'll add that to the next PR and take a swing at it today.

PR #493, opened 2026-01-15
feat(txpool/tracker): make tracker fb aware

wlawt:
Changes
- Introduces a new metric `fb_inclusion_duration`
- Records the metric when it sees the txhash in a `PendingBlock`. NOTE: on FB inclusion, it does not remove the TxHash in the Tracker's cache. This is to let us record when its included in the canonical block for the other existing metrics.
- Made `FlashblocksBuilder` public

TODO
- [x] Add a unit test that invokes `track_flashblock_transactions`


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/client/txpool/src/tracker.rs
```diff
@@ -83,6 +90,13 @@ impl Tracker {
...
 
+    fn track_flashblock_transactions(&mut self, pending_blocks: &PendingBlocks) {
+        // Get all transaction hashes from pending blocks
+        for tx_hash in pending_blocks.get_pending_transaction_hashes() {
+            self.transaction_fb_included(tx_hash);
+        }
+    }
+
     /// Track the first time we see a transaction in the mempool.
     pub fn transaction_inserted(&mut self, tx_hash: TxHash, event: TxEvent) {
```
wlawt, line 101: AFAIK, a transaction sent to us is sent to both the builder and the mpool, at the same time. Therefore, this should be accurate, as we'll start tracking right when we see it in the mpool to when it's in a `PendingBlock`

bin/node/src/cli.rs
```diff
@@ -60,6 +60,9 @@ impl From<Args> for TxpoolConfig {
             tracing_enabled: args.enable_transaction_tracing,
             tracing_logs_enabled: args.enable_transaction_tracing_logs,
             sequencer_rpc: args.rollup_args.sequencer,
+            flashblocks_config: args
+                .websocket_url
+                .map(|url| FlashblocksConfig::new(url, args.max_pending_blocks_depth)),
```
danyalprout: I don't think we're going to want to create a new Flashblocks config here, we can just pass through the shared state. 

See how the bundle does this here:
https://github.com/base/node-reth/blob/main/bin/node/src/main.rs#L36

wlawt: we can probably remove the `From` then and do something similar to this? https://github.com/base/node-reth/blob/bdbba6a96a24311f0c34b816de45f19fcf551faa/bin/node/src/main.rs#L39

crates/shared/primitives/src/lib.rs
```diff
@@ -6,5 +6,10 @@
 #[cfg(any(test, feature = "test-utils"))]
 pub mod test_utils;
 
+// These dependencies are only used in test-utils feature
+#[cfg(any(test, feature = "test-utils"))]
+use alloy_contract as _;
+#[cfg(any(test, feature = "test-utils"))]
+use alloy_sol_types as _;
```
danyalprout: You can mark these as optional in the cargo.toml, vs. doing this

crates/client/txpool/src/tracker.rs
```diff
@@ -527,4 +575,140 @@ mod tests {
...
+                extra_data: Bytes::new(),
+                base_fee_per_gas: U256::ZERO,
+            }),
+            diff: ExecutionPayloadFlashblockDeltaV1 {
+                blob_gas_used: Some(0),
+                transactions: vec![L1_BLOCK_INFO_DEPOSIT_TX, tx],
+                ..Default::default()
+            },
+            metadata: Metadata { block_number: 1 },
+        };
```
danyalprout: Can we use the Flashblock builder for this?

danyalprout (dismissed):

crates/shared/primitives/src/lib.rs
wlawt: i think it was marked optional before, but im unsure why its failing the CI Docker check
https://github.com/base/node-reth/blob/85d90b88ab808084f395683a9c321d27521a1bbb/crates/shared/primitives/Cargo.toml#L20

danyalprout (dismissed):
I think this is neater, lets get @refcell to take a sanity check too

crates/shared/primitives/src/lib.rs
```diff
@@ -1,10 +1,11 @@
...
+#![doc(issue_tracker_base_url = "https://github.com/base/node-reth/issues/")]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 
 #[cfg(any(test, feature = "test-utils"))]
 pub mod test_utils;
 
+// These dependencies are only used in test-utils feature
 #[cfg(any(test, feature = "test-utils"))]
 pub use test_utils::*;
```
refcell, line 11: Ah right so since we don't have any tests in primitives inside `test_utils`, it's not used by `cfg(test)` which is causing a warning I believe. We should only use `#[cfg(feature="test-utils")]` for this and the `pub mod test_utils;` line above. Then dependencies should just be optional and listed under the `test_utils` feature flag.

refcell (approved):
A follow on pr can fix the primitives feature flag issue

PR #561, opened 2026-01-23
chore(repo): bump reth to v1.10.2

meyer9:


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

bin/node/src/cli.rs
```diff
@@ -11,10 +11,6 @@ pub struct Args {
     #[command(flatten)]
     pub rollup_args: RollupArgs,
 
-    /// The websocket url used for flashblocks.
-    #[arg(long = "websocket-url", value_name = "WEBSOCKET_URL")]
-    pub websocket_url: Option<String>,
```
meyer9, line 16: This is now overshadowed by `rollup_args.flashblocks_url`. It never gets set, so I removed this and used the rollup_args version instead.

deny.toml
```diff
@@ -125,28 +125,30 @@ skip = [
     "base64",
     "bindgen",
     "cargo_metadata",
+    "cargo-platform",
     "core-foundation",
     "crossterm",
     "gloo-timers",
     "indexmap",
-    "opentelemetry",
+    "miniz_oxide",
```
meyer9, line 133: added miniz_oxide and cargo-platform then sorted alphabetically

bin/node/src/cli.rs
danyalprout: Can you explain how this is overshadowed?

```
--flashblocks-url <FLASHBLOCKS_URL>
          A URL pointing to a secure websocket subscription that streams out flashblocks.

          If given, the flashblocks are received to build pending block. All request with "pending" block tag will use the pending state based on flashblocks.

      --websocket-url <WEBSOCKET_URL>
          The websocket url used for flashblocks
```

I maybe missing something in the code.  If this isn't can you please undo this as it's a breaking change for node operators.

FWIW in a couple of releases we're going to completely abstract the arguments and we'll remove this duplication.

danyalprout (commented):
lgtm aside from a question re: the websocket-url param

haardikk21:
> lgtm aside from a question re: the websocket-url param

They aliased the `--flashblocks-url` arg to `--websocket-url` upstream here: https://github.com/paradigmxyz/reth/commit/4adf163fdcbd77009ab7413e26e866f16f242251

Locally, trying to run `cargo run -p base-reth-node -- node --help` from this branch doesn't display `--websocket-url` as an argument in the listed arguments *BUT* if you do try to pass it by running `cargo run -p base-reth-node -- node --websocket-url <smth>` it does work and gets picked up, with logs showing it's running our flashblocks extension (I added some random logs there to test)

I am, however, unsure what the behavior would be if both our `Cli` struct and the `RollupArgs` both contain a `websocket-url` flag and which one will get set and which will be unset.

meyer9:
Ours does not get set if you provide websocket-url. Only the one in RollupArgs gets set. I ran into this while testing.

meyer9:
I think `flashblocks-url` is probably a better name anyway given authrpc and regular rpc can work over websocket.

haardikk21:
> I think `flashblocks-url` is probably a better name anyway given authrpc and regular rpc can work over websocket.
one 

<del>One important thing to note here is while `--websocket-url` works fine but setting the `WEBSOCKET_URL` env var - which previously worked - no longer works. The env var now needs to be `FLASHBLOCKS_URL` for it to work</del>

<del>Node operators passing URLs based on a `WEBSOCKET_URL` env var will need to update</del>

danyalprout:
if someone sets `--websocket-url`, does that now result in our node running both our Flashblocks implementation and the internal one?

Should we zero it out after we read it?

meyer9:
Does Reth handle env vars through `value_name`? I was under the impression that only affects what's shown in the help docs (the `WEBSOCKET_URL` in `--websocket-url=<WEBSOCKET_URL>`). I think we handle envs in base/node separately, but I'm pretty sure Reth doesn't pull configs from env on most args.

meyer9:
> if someone setups `--websocket-url`, does that now result in our node running both our Flashblocks implementation and the internal one?
> 
> Should we zero it out after we read it?

It's only used in this which we already removed: https://github.com/base/base/blob/main/crates/client/node/src/node.rs

haardikk21:
> if someone sets `--websocket-url`, does that now result in our node running both our Flashblocks implementation and the internal one?
> 
> Should we zero it out after we read it?

https://github.com/paradigmxyz/reth/blob/768a687189d304355a68f1271f17c1a77011f2d6/crates/optimism/rpc/src/eth/mod.rs#L577 This does not get emitted so I believe it doesn't start the internal one 

I dont think that will start unless we add `.with_flashblocks` in the `BaseAddOnsBuilder` when we are configuring `OpAddOnsBuilder`

danyalprout (approved):

haardikk21:
> Does Reth handle env vars through `value_name`? I was under the impression that only affects what's shown in the help docs (the `WEBSOCKET_URL` in `--websocket-url=<WEBSOCKET_URL>`). I think we handle envs in base/node separately, but I'm pretty sure Reth doesn't pull configs from env on most args.

Yes nvm you're right

PR #562, opened 2026-01-23
feat: introduce `base_transactionLifecycle` API

wlawt:
## Motivation

- We currently have a `Tracker` ExEx that measures the inclusion time of a transaction and also the different states that it moves between (i.e., Pending, Queued, Included, etc.)
- This `Tracker` emits an `info!` log, which we can see in DataDog

## Problem
- We have been and will experience more sampling of logs on DataDog
- Sometimes a `TxHash` is not there to look up on the DD widget

## Solution
- Provide a self-served API for us (Base folks) and for third-parties to use

## Changes
- Introduces a new RPC called `base_transactionLifecycle`
- The RPC contains a TTL cache (2min expiry)
- The cache is updated when it receives a tx update from the Tracker
- Added a unit test to verify behaviour

### DIFF
--- a/crates/client/txpool/src/extension.rs
+++ b/crates/client/txpool/src/extension.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 
 use base_client_node::{BaseBuilder, BaseNodeExtension, FromExtensionConfig};
 use base_flashblocks::{FlashblocksConfig, FlashblocksState};
+use alloy_primitives::TxHash;
 use reth_provider::CanonStateSubscriptions;
 use tokio_stream::wrappers::BroadcastStream;
 use tracing::info;
@@ -51,7 +52,9 @@ impl BaseNodeExtension for TxPoolExtension {
 
         builder.add_rpc_module(move |ctx| {
             info!(message = "Starting Transaction Status RPC");
-            let proxy_api = TransactionStatusApiImpl::new(sequencer_rpc, ctx.pool().clone())
+            let (tx_send, tx_recv) = tokio::sync::mpsc::channel::<(TxHash, Vec<String>)>(1000);
+
+            let proxy_api = TransactionStatusApiImpl::new(sequencer_rpc, ctx.pool().clone(), tx_recv)
                 .expect("Failed to create transaction status proxy");
             ctx.modules.merge_configured(proxy_api.into_rpc())?;
 
@@ -65,7 +68,7 @@ impl BaseNodeExtension for TxPoolExtension {
                 let fb_state: Arc<FlashblocksState> =
                     flashblocks_config.as_ref().map(|cfg| cfg.state.clone()).unwrap_or_default();
 
-                tokio::spawn(tracex_subscription(canonical_stream, fb_state, pool, logs_enabled));
+                tokio::spawn(tracex_subscription(canonical_stream, fb_state, pool, logs_enabled, tx_send));
             }
 
             Ok(())
--- a/crates/client/txpool/src/extension.rs
+++ b/crates/client/txpool/src/extension.rs
@@ -3,9 +3,9 @@
 
 use std::sync::Arc;
 
+use alloy_primitives::TxHash;
 use base_client_node::{BaseBuilder, BaseNodeExtension, FromExtensionConfig};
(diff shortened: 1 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/client/txpool/src/extension.rs
```diff
@@ -51,8 +52,11 @@ impl BaseNodeExtension for TxPoolExtension {
 
         builder.add_rpc_module(move |ctx| {
             info!(message = "Starting Transaction Status RPC");
-            let proxy_api = TransactionStatusApiImpl::new(sequencer_rpc, ctx.pool().clone())
-                .expect("Failed to create transaction status proxy");
+            let (tx_send, tx_recv) = tokio::sync::mpsc::channel::<(TxHash, Vec<String>)>(1000);
```
haardikk21, line 55: The `1000` should be a constant at the top of the file or something

crates/client/txpool/src/rpc.rs
```diff
@@ -60,7 +82,17 @@ impl<Pool: TransactionPool + 'static> TransactionStatusApiImpl<Pool> {
             None
         };
 
-        Ok(Self { sequencer_client, pool })
+        // 2 mins = 60 blocks (120s/2s), with each block having ~300 txs, so 18000 txs, round up for some buffer
+        // TODO: make this configurable?
+        let tx_events_cache = Arc::new(Cache::new(20_000));
```
haardikk21, line 87: In lieu of being configurable, should be made a constant at least

crates/client/txpool/src/rpc.rs
```diff
@@ -28,19 +32,36 @@ pub struct TransactionStatusResponse {
     pub status: Status,
 }
 
+/// Response containing the lifecycle events of a transaction.
+#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
+pub enum TransactionLifecycleResponse {
+    /// The lifecycle events of the queried transaction.
+    Events(Vec<String>),
```
haardikk21, line 39: Can we make this work with `[str]` (str slice) rather than `Vec<String>`? `Vec<String>` is a doubly-indirected pointer (Vec is a ptr, and String is a ptr) - it can be slow to read all the data from it depending on how many elements are here

Since we are not mutating this data here, and it's constructed in `EventLog::to_vec` - unless necessary, I would urge to reconsider and do smth like `EventLog::to_str_slice` or smth and use that. There will be performance benefits, stronger mutability guarantees, and less memory allocations I think

Not a blocker for this PR though, especially cuz this isn't a particularly hot-path RPC anyway

danyalprout:
I think we should do this in the base/infra repo and not the node.

Ideally this would work reliably over time and not just based on an in memory cache

wlawt:
As discussed offline, this will be part of `base/infra` to avoid having the node be infra-aware. Will revisit after migrating `base/tips` and `base/infra-routing` into `base/infra`.

PR #579, opened 2026-01-27
feat: integrate Optimism proofs ExEx

meyer9:
Blocked by https://github.com/op-rs/op-reth/pull/648 but once that's merged and we update the optimism deps, this should be good to go!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ base-engine-ext = { path = "crates/shared/engine-ext" }
 base-access-lists = { path = "crates/shared/access-lists" }
 base-reth-rpc-types = { path = "crates/shared/reth-rpc-types" }
 base-flashblocks-node = { path = "crates/client/flashblocks-node" }
+base-proofs-extension = { path = "crates/client/proofs" }
 
 # Client
 base-txpool = { path = "crates/client/txpool" }
@@ -130,12 +131,9 @@ reth-rpc-eth-types = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10
 reth-trie-parallel = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
 reth-network-peers = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
 reth-tracing-otlp = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
-reth-optimism-cli = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
 reth-optimism-evm = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
 reth-rpc-engine-api = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
 reth-node-ethereum = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
-reth-optimism-rpc = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
-reth-optimism-node = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
 reth-payload-builder = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
 reth-execution-types = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
 reth-e2e-test-utils = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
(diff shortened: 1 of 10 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/client/proofs/src/proofs.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// Creates a new transaction pool extension helper.
+    pub const fn new(config: RollupArgs) -> Self {
+        Self { config }
+    }
+}
+
+impl BaseNodeExtension for ProofsHistoryExtension {
+    /// Applies the extension to the supplied builder.
+    fn apply(self: Box<Self>, mut builder: BaseBuilder) -> BaseBuilder {
+        // TODO: if BaseBuilder exposes the underlying OpBuilder, we can call launch_node_with_proof_history
```
refcell, line 37: Is this still in play?

Cargo.toml
```diff
@@ -154,6 +152,14 @@ reth-optimism-primitives = { git = "https://github.com/paradigmxyz/reth", tag =
...
 reth-payload-builder-primitives = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
 reth-optimism-payload-builder = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
+reth-exex = { git = "https://github.com/paradigmxyz/reth", tag = "v1.10.2" }
+
+# op-reth
+reth-optimism-rpc = { git = "https://github.com/op-rs/op-reth", rev = "8746d94ea4c6d2743ce5fe8e9f7c9301af02cd66" }
+reth-optimism-node = { git = "https://github.com/op-rs/op-reth", rev = "8746d94ea4c6d2743ce5fe8e9f7c9301af02cd66" }
+reth-optimism-cli = { git = "https://github.com/op-rs/op-reth", rev = "8746d94ea4c6d2743ce5fe8e9f7c9301af02cd66" }
+reth-optimism-trie = { git = "https://github.com/op-rs/op-reth", rev = "8746d94ea4c6d2743ce5fe8e9f7c9301af02cd66" }
+reth-optimism-exex = { git = "https://github.com/op-rs/op-reth", rev = "8746d94ea4c6d2743ce5fe8e9f7c9301af02cd66" }
```
danyalprout, line 162: Are there going to be release tags for this repo?

crates/client/proofs/src/proofs.rs
```diff
@@ -0,0 +1,127 @@
...
+        let args = self.config;
+        let proofs_history_enabled = args.proofs_history;
+        let proofs_history_window = args.proofs_history_window;
+        let proofs_history_prune_interval = args.proofs_history_prune_interval;
+        let proofs_history_verification_interval = args.proofs_history_verification_interval;
+
+        if proofs_history_enabled {
+            let path = args
+                .proofs_history_storage_path
+                .expect("Path must be provided if not using in-memory storage");
```
refcell, line 47: Ah maybe the `fn apply` method should return a `Result<BaseBuilder>`? This seems not-optimal.

refcell: Alternatively, we can make the field inside `ProofsHistoryExtension` a smaller subset of `RollupArgs` and have just a proofs path without the enabled flag. The user then needs to validate that both enabled and the path is set. Does that make any sense?

Cargo.toml
meyer9: Yeah! I think they're planning to merge this into the OP monorepo very soon (next week or so).

We can update this to a tag once that happens, but this rev is stable.

crates/client/proofs/src/proofs.rs
meyer9: yeah, we could expose `.builder` on our `BaseBuilder`, since it requires the `OpBuilder` (which is just a type alias for `LaunchContext`).

crates/client/proofs/src/proofs.rs
meyer9: Yeah, I think that makes sense, although I would lean towards matching the OP interface since there are docs on it: https://docs.optimism.io/node-operators/tutorials/reth-historical-proofs#running-op-reth-with-historical-proofs

People might get confused if we have different CLI args. I'll ask about changing this upstream.

meyer9: Ah, ok, so if the user sets `--proofs-history.storage-path`, it automatically sets `--proofs-history` to enabled, so this expect does make sense. The only invalid case I guess is if they enable it without setting a storage path.

refcell (approved):
Nothing blocking. Need to follow up with configuring the wal warning threshold. Will take care of that.

PR #615, opened 2026-01-29
fix(client): Flashblocks WS RPC Subscriptions

### LINKED ISSUE #613: fix(flashblocks): eth_subscribe for flashblock types should closely mirror non FB variants
### New Flashblock Transactions
`{"jsonrpc":"2.0", "id": 1, "method": "eth_subscribe", "params": ["newFlashblockTransactions"]}`

* Stream of transaction data
* Can be enriched with transaction and log (lets dedup the fields here) info by passing full=true
* Should return one transaction per websocket message

### Pending Logs
`{"jsonrpc":"2.0", "id": 1, "method": "eth_subscribe", "params": ["pendingLogs", {"address": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913", "topics": ["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef"]}]}`

* Stream of logs in Flashblocks that match the filter
* Should return one matching log per websocket message (similar to logs)

### New Flashblocks
`{"jsonrpc":"2.0", "id": 1, "method": "eth_subscribe", "params": ["newFlashblocks"]}`

* Should return the current RPC block as it's modifying on each new message ✅

refcell:
## Summary

Closes #613

This PR fixes the flashblocks WebSocket subscription RPCs to emit one item per message instead of arrays, aligning with how standard Ethereum subscription variants work -- `newFlashblockTransactions` (both hash-only and full) and `pendingLogs` now stream individual transactions/logs per WebSocket message by flat-mapping the internal batch streams, with corresponding test updates to validate the new per-item format.

This change also needs to be backported to `0.3.0` I believe.

### DIFF
--- a/crates/client/flashblocks/src/rpc/pubsub.rs
+++ b/crates/client/flashblocks/src/rpc/pubsub.rs
@@ -91,9 +91,11 @@ impl<Eth, FB> EthPubSub<Eth, FB> {
         })
     }
 
-    /// Returns a stream that yields individual logs from pending flashblocks matching the filter.
+    /// Returns a stream that yields individual logs from only the latest flashblock matching the
+    /// filter.
     ///
     /// Each matching log is emitted as a separate stream item (one log per WebSocket message).
+    /// Only logs from the most recent flashblock are emitted to avoid duplicates.
     fn pending_logs_stream(flashblocks_state: Arc<FB>, filter: Filter) -> impl Stream<Item = Log>
     where
         FB: FlashblocksAPI + Send + Sync + 'static,
@@ -112,7 +114,7 @@ impl<Eth, FB> EthPubSub<Eth, FB> {
                             return None;
                         }
                     };
-                    let logs = pending_blocks.get_pending_logs(&filter);
+                    let logs = pending_blocks.get_latest_flashblock_logs(&filter);
                     if logs.is_empty() { None } else { Some(logs) }
                 },
             ),
@@ -120,11 +122,12 @@ impl<Eth, FB> EthPubSub<Eth, FB> {
         )
     }
 
-    /// Returns a stream that yields individual full transactions with logs from pending
-    /// flashblocks.
+    /// Returns a stream that yields individual full transactions with logs from only the latest
+    /// flashblock.
     ///
     /// Each transaction (with its associated logs) is emitted as a separate stream item
-    /// (one transaction per WebSocket message).
+    /// (one transaction per WebSocket message). Only transactions from the most recent
+    /// flashblock are emitted to avoid duplicates.
     fn new_flashblock_transactions_full_stream(
         flashblocks_state: Arc<FB>,
     ) -> impl Stream<Item = TransactionWithLogs>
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

haardikk21:
Is this overlapping with the work @danyalprout was doing? 

We also need to include logs in the `newFlashblockTransactions` method if `full = true` - which should probably be in the same PR

crates/client/flashblocks/src/rpc/pubsub.rs
```diff
@@ -91,76 +92,98 @@ impl<Eth, FB> EthPubSub<Eth, FB> {
...
+                        Ok(blocks) => blocks,
+                        Err(err) => {
+                            error!(
+                                message = "Error in flashblocks stream for pending logs",
+                                error = %err
+                            );
+                            return None;
+                        }
+                    };
+                    let logs = pending_blocks.get_pending_logs(&filter);
```
danyalprout: I think this pending method will cause duplicates to be sent as it gets called with the full pending block on each new Flashblock. E.g.

Fb1=(a)
Fb2=(b)

I believe this will cause us to send

a, a, b

I think we need to narrow it down to just the transactions in the latest FB 

Same with the get pending transaction hashes too

danyalprout (commented):
It would be great if we could also extend the full transaction data to have the log info too.

Happy to do that in a follow up

danyalprout (dismissed):

danyalprout (approved):

cb-heimdall:
Review Error for danyalprout @ 2026-01-30 10:20:31 UTC

danyalprout:
Cut a release branch for this [here](https://github.com/base/base/tree/releases/v0.3.1).

haardikk21 (approved):

PR #621, opened 2026-01-30
fix(flashblocks): adds bounds checking for collection access

### LINKED ISSUE #600: fix(flashblocks): Add bounds checking for collection access
## Summary

Replace `.unwrap()` calls on collection operations with proper error handling to prevent panics when collections are empty.

### Problem

The `pending_blocks.rs` module calls `.first().unwrap()` and `.last().unwrap()` on vectors without checking if they're empty. If these methods are called before any data is populated, the node will crash.

Similarly, `processor.rs` calls `.keys().min().unwrap()` on a `BTreeMap` without verifying it contains entries.

### Affected Files

- `crates/client/flashblocks/src/pending_blocks.rs` (lines 171, 176, 181, 186, 191)
- `crates/client/flashblocks/src/processor.rs` (line 290)

### Current Code

```rust
// pending_blocks.rs
fn latest_block_number(&self) -> u64 {
    self.headers.last().unwrap().number  // Panics if empty
}

fn earliest_block_number(&self) -> u64 {
    self.headers.first().unwrap().number  // Panics if empty
}

// processor.rs
let earliest_block_number = flashblocks_per_block.keys().min().unwrap();  // Panics if empty
```

### Proposed Solution

Return `Option` or `Result` instead of panicking:

```rust
// Option 1: Return Option
fn latest_block_number(&self) -> Option<u64> {
    self.headers.last().map(|h| h.number)
}

// Option 2: Return Result with custom error
fn latest_block_number(&self) -> Result<u64, FlashblocksError> {
    self.headers
        .last()
        .map(|h| h.number)
        .ok_or(FlashblocksError::NoHeaders)
}
```

mw2000:
Replaces `.unwrap()` calls on collection operations with proper error handling to prevent panics when collections are empty.

Fixes #600

### DIFF
--- a/crates/client/flashblocks-node/benches/pending_state.rs
+++ b/crates/client/flashblocks-node/benches/pending_state.rs
@@ -156,8 +156,8 @@ async fn wait_for_pending_state(
     let deadline = Instant::now() + Duration::from_secs(10);
     loop {
         if let Some(pending) = state.get_pending_blocks().as_ref()
-            && pending.latest_block_number().is_ok_and(|n| n == target_block)
-            && pending.latest_flashblock_index().is_ok_and(|i| i == expected_index)
+            && pending.latest_block_number() == target_block
+            && pending.latest_flashblock_index() == expected_index
         {
             return;
         }
--- a/crates/client/flashblocks-node/benches/pending_state.rs
+++ b/crates/client/flashblocks-node/benches/pending_state.rs
@@ -156,8 +156,8 @@ async fn wait_for_pending_state(
     let deadline = Instant::now() + Duration::from_secs(10);
     loop {
         if let Some(pending) = state.get_pending_blocks().as_ref()
-            && pending.latest_block_number() == target_block
-            && pending.latest_flashblock_index() == expected_index
+            && pending.latest_block_number().is_ok_and(|n| n == target_block)
+            && pending.latest_flashblock_index().is_ok_and(|i| i == expected_index)
         {
             return;
         }
--- a/crates/client/flashblocks/src/error.rs
+++ b/crates/client/flashblocks/src/error.rs
@@ -108,12 +108,6 @@ pub enum BuildError {
     NoFlashblocks,
 }
 
-impl From<BuildError> for jsonrpsee_types::ErrorObjectOwned {
-    fn from(err: BuildError) -> Self {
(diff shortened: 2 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

danyalprout:
If we have a `PendingBlocks` object I believe these should all have a value. I agree the unwrap calls are not ideal, but I don't like the idea of propagating Results throughout the code when I think this should never happen. (Specifically we have [PendingBlocks wrapped in an Option for the nil case](https://github.com/base/base/blob/main/crates/client/flashblocks/src/state.rs#L30)).

I wonder if there's alternative typing where we could pull the first/last of items out as fields in the PendingBlocks, to remove these cases (and we could extract that in the builder)

danyalprout (changes requested):
Based on the above comment, would love it if we could structure this code in a way that avoids propagating Result's for cases that should never happen

mw2000:
Yup now extracting earliest and latest values in `build` function and not propogating Results

refcell (approved):

danyalprout (approved):

PR #639, opened 2026-02-01
feat: add devnet crate with programmatic E2E testing

haardikk21:
## Summary

Adds a new `devnet` crate that provides programmatic E2E testing infrastructure using testcontainers. This enables writing system tests that span multiple components (L1, L2 builder, L2 client) without manual Docker Compose orchestration.

## Changes

- **New `devnet` crate** with testcontainers-based infrastructure for spinning up complete L1+L2 devnet stacks
- **In-process builder and client** - runs builder/client nodes in the test process for faster iteration and better debugging
- **Devnet CLI binary** (`devnet`) - standalone CLI for managing local devnets with commands:
  - `devnet start` - starts a full L1+L2 devnet with stable container names and ports
  - `devnet status` - shows block numbers and sync status for all nodes
  - `devnet smoke` - sends test transactions to L1 and L2
  - `devnet accounts` - displays funded test accounts
  - `devnet flashblocks` - streams flashblocks from builder via websocket
  - `devnet clean` - removes devnet data directory
- **Platform-aware host connectivity** - uses `host.docker.internal` on macOS (no SSHD containers) and testcontainers SSHD tunnel on Linux
- **Smoke tests** for:
  - L1/L2 block production
  - Builder-client block sync via P2P
  - Client pending state tracking via flashblocks
- **Separate CI job** (`ci.yml` now contains a devnet tests job) - devnet tests run independently from unit tests since they require Docker
- **Updated justfile** - `just test` excludes devnet, new `just devnet-tests` command added

## Testing

All 3 smoke tests pass (~44 seconds):
```
just devnet-tests
```

Run devnet locally:
```
cargo run -p devnet --bin devnet -- start
```


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

devnet/src/bin/devnet.rs
refcell, line 1: What do you think about just naming this `system` as opposed to `system_tests`?

It's not really just tests since it's a full orchestration service, and we may use it as a library elsewhere down the road for various reasons. And if someone/us want to use it, it'll be strange to do `use system_tests::`

Justfile
```diff
@@ -46,10 +46,23 @@ zepter-fix:
...
+    @command -v cargo-nextest >/dev/null 2>&1 || cargo install cargo-nextest
+    cargo nextest run -p system_tests
+
+# Pre-pulls Docker images needed for system tests
+system-tests-pull-images:
+    docker build -t devnet-setup:local -f docker/Dockerfile.devnet .
+    docker pull ghcr.io/paradigmxyz/reth:v1.10.2
+    docker pull sigp/lighthouse:v8.0.1
+    docker pull us-docker.pkg.dev/oplabs-tools-artifacts/images/op-node:v1.16.5
+    docker pull us-docker.pkg.dev/oplabs-tools-artifacts/images/op-batcher:v1.16.3
```
refcell, line 67: I think this belongs in the `docker/` directory. Then maybe we can also split out the versions into version files with optional overrides. So e.g. `reth.version` contains `v1.10.2` or something.

Justfile
```diff
@@ -46,10 +46,23 @@ zepter-fix:
...
+# Runs tests across workspace with all features enabled (excludes system_tests)
 test: build-contracts
     @command -v cargo-nextest >/dev/null 2>&1 || cargo install cargo-nextest
-    RUSTFLAGS="-D warnings" cargo nextest run --workspace --all-features
+    RUSTFLAGS="-D warnings" cargo nextest run --workspace --all-features --exclude system_tests
+
+# Runs system tests (requires Docker)
+system-tests: build-contracts
+    @command -v cargo-nextest >/dev/null 2>&1 || cargo install cargo-nextest
+    cargo nextest run -p system_tests
```
refcell: I wonder if
```
@command -v cargo-nextest >/dev/null 2>&1 || cargo install cargo-nextest
```

belongs in a `install-nextest` target that all these targets can inherit and dedup the logic.

Then the Justfile could look like:

```Justfile
install-nextest:
    @command -v cargo-nextest >/dev/null 2>&1 || cargo install cargo-nextest

system-tests: install-nextest build-contracts
    cargo nextest run -p system_tests
```

devnet/src/bin/devnet.rs
haardikk21: yeah i was also considering just naming it `devnet` but not sure. im good w either

Justfile
haardikk21: either that or do it as part of the devnet CLI so all versions can be managed in `images.rs`

system_tests/src/lib.rs
refcell, line 1: Preference for `lib.rs` to follow the format of other `lib.rs` files in the workspace, but not a blocker.

system_tests/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+//! ```
+
+#![warn(missing_docs)]
+#![cfg_attr(not(test), warn(unused_crate_dependencies))]
+
+use alloy_network as _;
+use alloy_rpc_types as _;
+use base_flashtypes as _;
+use futures_util as _;
+use tokio_tungstenite as _;
```
refcell: These are likely dev-dependencies if they need to be used?

system_tests/src/smoke.rs
```diff
@@ -0,0 +1,241 @@
...
+};
+
+const DEFAULT_L1_CHAIN_ID: u64 = 1337;
+const DEFAULT_L2_CHAIN_ID: u64 = 84538453;
+const DEFAULT_SLOT_DURATION: u64 = 2;
+
+/// A complete L1+L2 devnet stack.
+pub struct Devnet {
+    #[allow(dead_code)]
+    temp_dir: TempDir,
```
refcell: rm
```suggestion
```

devnet/src/smoke.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    /// Returns the public RPC URL of the L1 Reth node.
+    pub async fn l1_rpc_url(&self) -> Result<Url> {
+        self.l1_stack.rpc_url().await
+    }
+
+    /// Returns the public RPC URL of the L2 builder node.
+    pub fn l2_rpc_url(&self) -> Result<Url> {
+        self.l2_stack().rpc_url()
+    }
```
refcell, line 63: I think it'd be nice if instead of just returning the url, they returned an instance of [`Http`](https://docs.rs/alloy-transport-http/latest/alloy_transport_http/struct.Http.html#method.url). In the smoke tests, there's a lot of parsing the Urls and constructing clients that could just be nicely handled by the `Devnet` (and the individual stacks).

Since [`Http`](https://docs.rs/alloy-transport-http/latest/alloy_transport_http/struct.Http.html) provides both the client and url method, either the raw url or client can be easily accessed.

system_tests/src/lib.rs
haardikk21: theyre used by the binary and the tests but not by the lib so its not in dev dep but was causing the lib to complain about unused deps

devnet/src/smoke.rs
haardikk21: yes agreed, with proper typings too would be nice. forgot to change it to that

devnet/src/l2/stack.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// Rollup configuration JSON.
+    pub rollup_config: Vec<u8>,
+    /// L1 genesis JSON (for op-node chain spec).
+    pub l1_genesis: Vec<u8>,
+    /// JWT secret hex for Engine API authentication.
+    pub jwt_secret_hex: Vec<u8>,
+    /// P2P private key (hex-encoded, no 0x prefix) for op-node libp2p identity.
+    pub p2p_key: Vec<u8>,
+    /// Sequencer private key (hex-encoded with 0x prefix) for block signing.
+    pub sequencer_key: Vec<u8>,
```
refcell: We should probably use the sized alloy type here for the length of the Bytes.

Alloy's rpc engine types crate even provides a [`JwtSecret`](https://docs.rs/alloy-rpc-types-engine/latest/alloy_rpc_types_engine/struct.JwtSecret.html) type we can use out of the box.

devnet/src/l2/op_node.rs
```diff
@@ -0,0 +1,287 @@
...
+        "--p2p.ban.peers=false".to_string(),
+        "--rpc.addr=0.0.0.0".to_string(),
+        format!("--rpc.port={RPC_PORT}"),
+    ]
+}
+
+fn derive_libp2p_peer_id() -> Result<String> {
+    use crate::setup::BUILDER_LIBP2P_PEER_ID;
+    Ok(BUILDER_LIBP2P_PEER_ID.to_string())
+}
```
haardikk21: useless

system_tests/src/l2/batcher.rs
```diff
@@ -0,0 +1,105 @@
...
+        let container =
+            container_builder.start().await.wrap_err("Failed to start batcher container")?;
+
+        Ok(Self { container, name })
+    }
+
+    #[allow(dead_code)]
+    pub(crate) const fn container(&self) -> &ContainerAsync<GenericImage> {
+        &self.container
+    }
```
refcell: rm
```suggestion
```

devnet/src/l2/in_process_builder.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    /// Returns the Engine API URL.
+    pub fn engine_url(&self) -> Result<Url> {
+        Url::parse(&format!("http://{}", self.engine_addr)).wrap_err("Failed to parse Engine URL")
+    }
+
+    /// Returns the WebSocket URL.
+    pub fn ws_url(&self) -> Result<Url> {
+        Url::parse(&format!("ws://{}", self.ws_api_addr)).wrap_err("Failed to parse WebSocket URL")
+    }
```
refcell, line 216: There's a lot of fallibility that I've experienced doing this because if the addr already contains an `http://` prefix it'll fail, and it can be really annoying to trace because everything _looks_ right until you realize this duplication. Since we're constructing these addresses ourselves, shouldn't run into this issue. Calling it out though in case it shows up.

system_tests/src/config/jwt.rs
```diff
@@ -0,0 +1,16 @@
...
+pub fn random_jwt_secret() -> B256 {
+    let mut bytes = [0u8; 32];
+    rand::rng().fill_bytes(&mut bytes);
+    B256::from(bytes)
+}
+
+/// Generates a random JWT secret encoded as a hex string.
+pub fn random_jwt_secret_hex() -> String {
+    hex::encode(random_jwt_secret())
+}
```
refcell: We can just use [`JwtSecret::random()`](https://docs.rs/alloy-rpc-types-engine/latest/alloy_rpc_types_engine/struct.JwtSecret.html#method.random) instead of rolling out own randomness.

system_tests/src/bin/devnet.rs
```diff
@@ -0,0 +1,317 @@
...
+        std::env::current_dir().wrap_err("Failed to get current directory")?.join(".devnet");
+
+    let _ = fs::remove_dir_all(&devnet_dir);
+    fs::create_dir_all(devnet_dir.join("shared"))?;
+
+    println!("Starting devnet...");
+
+    let devnet = DevnetBuilder::new()
+        .with_l1_chain_id(1337)
+        .with_l2_chain_id(84538453)
```
refcell: defaults, don't need to be specified
```suggestion
```

system_tests/src/bin/devnet.rs
```diff
@@ -0,0 +1,317 @@
...
+    let l2_builder_op_rpc = devnet.l2_stack().op_node().rpc_url().await?;
+    let l2_client_op_rpc = "http://localhost:8549".to_string();
+
+    let urls = DevnetUrls {
+        l1_rpc: l1_rpc.to_string(),
+        l2_builder_rpc: l2_builder_rpc.to_string(),
+        l2_client_rpc: l2_client_rpc.to_string(),
+        l2_builder_op_rpc: l2_builder_op_rpc.to_string(),
+        l2_client_op_rpc,
+    };
```
refcell: This feels like something that would be easy to return from the `Devnet` e.g.

```rust
let urls = devnet.urls();
```

And since the `Devnet` has the urls, could just `devnet.write_urls()`.

system_tests/src/bin/devnet.rs
```diff
@@ -0,0 +1,317 @@
...
+    fs::write(devnet_dir.join("urls.json"), serde_json::to_string_pretty(&urls)?)?;
+
+    println!("\nDevnet started successfully!");
+    println!("\nRPC URLs:");
+    println!("  L1:          {l1_rpc}");
+    println!("  L2 Builder:  {l2_builder_rpc}");
+    println!("  L2 Client:   {l2_client_rpc}");
```
refcell: This can be a `Display` implementation on the `DevnetUrls` type.

```rust
impl std::fmt::Display for DevnetUrls {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        writeln!(f)?;
        writeln!(f, "RPC URLs:")?;
        writeln!(f, "  L1:          {}", self.l1_rpc)?;
        writeln!(f, "  L2 Builder:  {}", self.l2_builder_rpc)?;
        write!(f, "  L2 Client:   {}", self.l2_client_rpc)
    }
}
```

devnet/src/config/l1_beacon.rs
```diff
@@ -0,0 +1,106 @@
+//! L1 beacon chain (consensus layer) configuration.
+
+/// Generates the L1 beacon chain configuration as YAML.
+pub fn l1_beacon_config_yaml(chain_id: u64, genesis_time: u64, slot_duration: u64) -> String {
```
refcell, line 4: I wonder if we can template this out so it's not a long inline format string.

i.e. there's a `.yaml` file with `{genesis_time}` and that's read in as a string, with those template values filled.

Ideally we could actually have a valid beacon config yaml, parse it from file string into the actual type, and then just set/overwrite the values but I'm not sure if there's a beacon config type available to us.

system_tests/src/config/accounts.rs
```diff
@@ -0,0 +1,101 @@
...
+const ANVIL_PRIVATE_KEY_5: B256 =
+    b256!("0x8b3a350cf5c34c9194ca85829a2df0ec3153be0318b5e2d3348e872092edffba");
+const ANVIL_PRIVATE_KEY_6: B256 =
+    b256!("0x92db14e403b83dfe3df233f83dfa3a0d7096f21ca9b0d6d6b8d88b2b4ec1564e");
+const ANVIL_PRIVATE_KEY_7: B256 =
+    b256!("0x4bbbf85ce3377467afe5d46f804f221813b2bb87f24d81f60f1fcdbf7cbf4356");
+const ANVIL_PRIVATE_KEY_8: B256 =
+    b256!("0xdbda1821b80551c9d65939329250298aa3472ba22feea921c0cf5d620ea67b97");
+const ANVIL_PRIVATE_KEY_9: B256 =
+    b256!("0x2a871d0798f97d79848a013d4936a73bf4cc922c825d33c1cf7073dff6d409c6");
```
refcell: Since these keys are just derived from the dev mnemonic, we could just hardcode the test mnemonic and then derive each private key and account rather than hardcoding the addresses.

e.g.

```rust
use alloy_signer_local::{coins_bip39::English, MnemonicBuilder, PrivateKeySigner};

/// The test mnemonic used to derive dev accounts.
pub const TEST_MNEMONIC: &str = "test test test test test test test test test test test junk";

// Generate wallets from the default mnemonic
let builder = MnemonicBuilder::<English>::default().phrase(TEST_MNEMONIC);
let wallet: PrivateKeySigner = builder
    .derivation_path("m/44'/60'/0'/0/0")  // Index 0
    .unwrap()
    .build()
    .unwrap();
```

> [!INFO]
>
> Side note, it would be _really_ nice if someone, like alloy, published a lightweight crate that just contained these testing values instead of everybody rolling it themselves. (even reth rolls it themselves).

system_tests/src/bin/devnet.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    Ok(())
+}
+
+fn format_ether(wei: U256) -> String {
+    let wei_u128: u128 = wei.try_into().unwrap_or(u128::MAX);
+    let eth = wei_u128 as f64 / 1e18;
+    format!("{eth:.4}")
+}
```
refcell: `alloy-primitives` provides a [`format_ether`](https://docs.rs/alloy-primitives/latest/alloy_primitives/utils/fn.format_ether.html) function that we should re-use

system_tests/src/bin/devnet.rs
```diff
@@ -0,0 +1,317 @@
...
+    l2_builder_op_rpc: String,
+    l2_client_op_rpc: String,
+}
+
+fn read_urls() -> Result<DevnetUrls> {
+    let path = PathBuf::from(".devnet/urls.json");
+    let content = fs::read_to_string(&path)
+        .wrap_err("Failed to read .devnet/urls.json - is devnet running?")?;
+    serde_json::from_str(&content).wrap_err("Failed to parse urls.json")
+}
```
refcell: Maybe worth pulling this type into its own module and inlining methods.

For example

```rust
#[derive(serde::Deserialize, serde::Serialize)]
struct DevnetUrls {
    l1_rpc: String,
    l2_builder_rpc: String,
    l2_client_rpc: String,
    l2_builder_op_rpc: String,
    l2_client_op_rpc: String,
}

impl DevnetUrls {
    pub fn read_from_file() -> Result<Self> {
        ...
    }
}
```

system_tests/src/rpc.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Safe L2 block info.
+    pub safe_l2: Option<L2BlockRef>,
+}
+
+/// L2 block reference from sync status.
+#[derive(Debug, Clone, Deserialize)]
+pub struct L2BlockRef {
+    /// Block number.
+    pub number: u64,
+}
```
refcell: We'll need these in the future inside our crates, so I'd prefer to pull these types in soon and just re-use that crate. Can put up a PR to do so.

refcell: PR up https://github.com/base/base/pull/648

system_tests/src/rpc.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        #[derive(Deserialize)]
+        struct RpcResponse {
+            result: SyncStatus,
+        }
+
+        let rpc_resp: RpcResponse =
+            resp.json().await.wrap_err("Failed to parse syncStatus response")?;
+        Ok(rpc_resp.result)
+    }
```
refcell: Instead of manually constructing this call we can use `jsonrpsee` like so:
https://github.com/base/base/pull/649

system_tests/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+
+const ALPHANUMERIC: &[char] = &[
+    'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's',
+    't', 'u', 'v', 'w', 'x', 'y', 'z', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
+];
+
+/// Generates a unique container name with the given prefix.
+pub fn unique_name(prefix: &str) -> String {
+    format!("{}-{}", prefix, nanoid::nanoid!(8, ALPHANUMERIC))
+}
```
refcell: Can we place this in a module?

devnet/src/images.rs
```diff
@@ -0,0 +1,10 @@
+/// Docker image for Reth.
+pub const RETH_IMAGE: &str = "ghcr.io/paradigmxyz/reth:v1.10.2";
+/// Docker image for op-node.
+pub const OP_NODE_IMAGE: &str = "us-docker.pkg.dev/oplabs-tools-artifacts/images/op-node:v1.16.5";
+/// Docker image for op-batcher.
+pub const OP_BATCHER_IMAGE: &str =
+    "us-docker.pkg.dev/oplabs-tools-artifacts/images/op-batcher:v1.16.3";
+/// Docker image for op-deployer.
+pub const OP_DEPLOYER_IMAGE: &str =
+    "us-docker.pkg.dev/oplabs-tools-artifacts/images/op-deployer:v0.6.0-rc.3";
```
refcell, line 12: Elsewhere we also hardcode the versions of these images, ideally we pin them in version files in the `docker/` directory or something of the sorts.

refcell (dismissed):
First review pass, will leave it here and take another pass once these changes land

devnet/src/images.rs
haardikk21: do you have a pref for the approach here? https://github.com/base/base/pull/639#discussion_r2755785595

system_tests/src/smoke.rs
haardikk21: This is actually needed, as otherwise `TempDir` gets dropped when it goes out of scope. I'm gonna just prefix with an underscore instead of this macro.

devnet/src/images.rs
refcell: The least error-prone approach to me feels like pinning the versions in a file that's not programmatic. That said, I'm not super opinionated yet and I think we'll build good intuition about the right approach as we use these system tests over time.

For now, as long as we de-dup the versioning I'm happy -- if that means using `images.rs`, that SGTM!

haardikk21: i think its a matter of also figuring out if we have ergonomic opinions on using a just command to do `docker pull` for all the images up front or if we want it to be a CLI command managed via the devnet binary `cargo run -p devnet --bin devnet pull-images`

i dont mind either and no strong opinions either - but we just need to pick one to get this going

refcell: Yep. Probably worth getting other folks opinions including @danyalprout.

refcell (dismissed):
I think we could get this merged. There's a lot of cleanup that can be done, but would rather this not go stale and we do that work iterative

refcell (approved):

PR #640, opened 2026-02-01
chore: remove build scripts

danyalprout:
### Description
- Remove the build scripts and setup macros for initializing the metrics/reth version info.
- Update builder, client and consensus to use these

Reth implementation:
https://github.com/paradigmxyz/reth/blob/v1.10.2/crates/node/builder/src/builder/mod.rs#L682-L690
https://github.com/paradigmxyz/reth/blob/main/crates/node/core/src/args/engine.rs#L399

### Testing
Based on the version on main being `0.0.0`, via CLI:
```
~/c/e/base > ./target/debug/base-builder --version
base-builder 0.0.0
~/c/e/base > ./target/debug/base-reth-node --version
base-reth-node 0.0.0
~/c/e/base > ./target/debug/base-consensus --version
base-consensus 0.0.0
```

Via metrics:
```
reth_info{version="1.10.2/0.0.0",build_timestamp="2026-01-31T15:58:43.893335387Z",cargo_features="jemalloc,otlp",git_sha="8e3b5e6a",target_triple="aarch64-unknown-linux-gnu",build_profile="debug"} 1
# TYPE reth_base_info gauge
reth_base_info{version="0.0.0"} 1
```


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/shared/cli-utils/src/lib.rs
```diff
@@ -29,8 +29,11 @@ pub use logging::{
 mod tracing;
 pub use tracing::{LogfmtFormatter, init_test_tracing};
 
-mod version;
-pub use version::Version;
+pub mod version;
+
+mod cli;
```
refcell, line 35: Not a fan of the style change here.

```suggestion
mod version
pub use version::Version;

mod cli;
pub use cli::parse_cli;
```

crates/shared/cli-utils/src/lib.rs
```diff
@@ -29,8 +29,11 @@ pub use logging::{
...
 
 mod runtime;
+// Re-export clap for use in parse_cli! macro
+pub use clap;
```
refcell: We don't need this re-export. The macro already pulls in clap deps. If clap itself is needed, we should add a `use clap;` at the top of the macro.

```suggestion
```

crates/shared/cli-utils/src/version.rs
refcell, line 1: Can we place these free-floating global methods behind the `Version` struct that was removed.

This way its usage nicely looks like the following.
```rust
base_cli_utils::Version::init_reth();
base_cli_utils::Version::register_metrics();
```

bin/builder/src/main.rs
```diff
@@ -3,14 +3,30 @@
...
     base_cli_utils::SigsegvHandler::install();
 
-    if let Err(err) = op_rbuilder::launcher::launch() {
+    // Initialize Reth version metadata for P2P identification and logging.
+    base_cli_utils::init_reth_version!();
+
+    let logs_dir = dirs_next::cache_dir()
+        .map(|root| root.join(format!("{}/logs", env!("CARGO_PKG_NAME"))))
+        .unwrap()
+        .into_os_string();
```
refcell, line 21: Can be in a follow-on PR, but this is probably a good candidate for a macro. This way we keep the entrypoint a bit more slim.

crates/builder/op-rbuilder/src/launcher.rs
refcell, line 43: I think we should get to the point where this function doesn't exist.

Primarily, this function just wires telemetry into the cli app after it's configured.

I'm thinking the best way we can trait abstract this behavior is by creating a `WithTelemetry<Cli>` wrapper. The `WithTelemetry` type just wraps the `Cli` and has a `configure() -> CliApp` method that wires in the telemetry.

What this means is in the `main.rs` entrypoint, we can then just define by type what functionality wraps the cli, just like how reth's node builder does (e.g. `WithNodeContext<>` etc).

refcell (dismissed):
Some small changes requested, but love the direction of this PR

crates/builder/op-rbuilder/src/launcher.rs
danyalprout: Agreed, I'm hoping we can move base-builder to have a setup that's 90% identical to base-node-reth

refcell (approved):
Nice!

bin/builder/src/main.rs
refcell: Opened an issue to track [here](https://github.com/base/base/issues/644)

PR #650, opened 2026-02-02
fix(builder): Transactions not being priority fee ordered within Flashblocks

haardikk21:
## Summary

Fixes transaction priority fee ordering bug across flashblocks, where transactions were not being correctly re-ordered by priority fee between flashblock boundaries.

**Root cause**: When a low-priority transaction (`TX_A`) is consumed in flashblock 1, the pool still considers it pending. This means subsequent flashblocks see the nonce-chain as blocked, preventing the high-priority dependent transaction (`TX_B`) from surfacing at its correct position relative to other senders' transactions.

**Fix**: After `mark_committed` between flashblocks, call `pool.prune_transactions` to remove already-executed transactions from the pool. This unblocks dependent nonce-chains so the iterator produces the correct priority ordering in subsequent flashblocks.

### Upstream dependency

The upstream reth PR ([paradigmxyz/reth#21765](https://github.com/paradigmxyz/reth/pull/21765)) that added `prune_transactions` to the `TransactionPool` trait **has been merged** and is available in reth v1.11.0 (the current workspace dependency). This PR is no longer blocked.

### Changes

- **`payload.rs`**: Added `self.pool.prune_transactions(new_transactions)` after `mark_committed` between flashblocks
- **`best_txs.rs`**: Added regression test simulating the nonce-chain gating scenario (based on real Base Mainnet block 41628995). The test recreates the pool without the already-executed transaction to simulate `prune_transactions` behavior, and verifies correct priority ordering (`TX_B` 100 gwei before `TX_C` 10 gwei) in the second flashblock

### DIFF
--- a/crates/builder/core/src/flashblocks/best_txs.rs
+++ b/crates/builder/core/src/flashblocks/best_txs.rs
@@ -197,7 +197,12 @@ mod tests {
         assert_eq!(iterator.next(()), None);
 
         // Simulate: flashblock 1 is complete after TX_A was executed
-        iterator.mark_commited(vec![*tx_a.hash()]);
+        iterator.mark_committed(&[*tx_a.hash()]);
+        // We cannot simulate the transaction getting pruned from the pool
+        // so we instead recreate the pool without TX_A
+        let mut pool = PendingPool::new(CoinbaseTipOrdering::<MockTransaction>::default());
+        pool.add_transaction(Arc::new(f.validated(tx_b.clone())), 0);
+        pool.add_transaction(Arc::new(f.validated(tx_c.clone())), 0);
 
         // === FLASHBLOCK 2 ===
         // We refresh the iterator with the latest best transactions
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/builder/core/src/flashblocks/payload.rs
```diff
@@ -554,6 +554,7 @@ where
             .map(|tx| tx.tx_hash())
             .collect::<Vec<_>>();
         best_txs.mark_committed(&new_transactions);
+        self.pool.prune_transactions(new_transactions);
```
danyalprout, line 557: @haardikk21 what happens if there is a Flashblocks reorg? How do we recover the transactions?

haardikk21: They will be lost without a way to know about it again. Couple assumptions here: combined builder etc should aim to bring back FB reorg rate to internal SLAs of >99%, and mempool redesign should resend txns that didn't get included in canon blocks but still continue to be considered valid

crates/builder/core/src/flashblocks/best_txs.rs
```diff
@@ -135,4 +136,94 @@ mod tests {
...
+        let fb2_first = iterator.next(()).unwrap();
+        let fb2_second = iterator.next(()).unwrap();
+
+        assert_eq!(fb2_first.sender(), sender_a);
+        assert_eq!(fb2_second.sender(), sender_b);
+        assert!(
+            fb2_second.effective_tip_per_gas(MIN_PROTOCOL_BASE_FEE)
+                < fb2_first.effective_tip_per_gas(MIN_PROTOCOL_BASE_FEE)
+        );
+    }
```
danyalprout, line 228: @haardikk21 is it possible to update this test to use the actual builder code? I suspect it maybe very difficult due to the code not being overly testable.

danyalprout (commented):
Looks good to me, but can we take a quick look at the test. Would like to get this merged into 0.6.0

crates/builder/core/src/flashblocks/best_txs.rs
haardikk21: we will have very flaky tests if we do that because it needs to be "just right" with the timing and race conditions that exist here

danyalprout (approved):

PR #675, opened 2026-02-06
Backport tracing to 0.4.0

meyer9:


refcell (commented):
All we need to do is enable the `otlp` feature? There's no config needed?

meyer9:
yup, no config needed. It's all configured through env vars

refcell (approved):

PR #681, opened 2026-02-08
fix(engine): use default launch to respect all options

danyalprout:
### Description
Our current launch function doesn't respect all the properties. We should use the upstream version instead. See the implementation:

* [launch()](https://github.com/paradigmxyz/reth/blob/v1.10.2/crates/node/builder/src/builder/mod.rs#L682)
* [engine_api_launcher()](https://github.com/paradigmxyz/reth/blob/v1.10.2/crates/node/builder/src/builder/mod.rs#L717)
* [tree_config()](https://github.com/paradigmxyz/reth/blob/v1.10.2/crates/node/core/src/args/engine.rs#L380)

The tree config loads all the options, whereas we're currently only loading a subset.

### DIFF
--- a/crates/client/node/src/builder.rs
+++ b/crates/client/node/src/builder.rs
@@ -5,13 +5,14 @@ use std::fmt;
 use eyre::Result;
 use reth_exex::ExExContext;
 use reth_node_builder::{
-    NodeAdapter, NodeComponentsBuilder,
+    NodeAdapter, NodeComponentsBuilder, NodeHandleFor,
     node::FullNode,
     rpc::{RethRpcAddOns, RpcContext},
 };
 
 use crate::{
     OpBuilder,
+    node::BaseNode,
     types::{OpAddOns, OpComponentsBuilder, OpNodeTypes},
 };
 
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (approved):
~Looks like our `TreeConfig` construction is the same as what's in upstream reth, so I don't see why we were rolling this construction ourselves in the first place.~

Strike that, our tree config was missing a bunch of the configurable options in upstream reth. What was actually happening is since these weren't overridden by CLI, the default values were being used which is what reth does anyway. So this PR effectively just adds support to configure the tree config through cli options.

PR #691, opened 2026-02-10
fix: add tls feature for wss connections

danyalprout:
### Description
Re-add the TLS feature. This is required to connect to `wss://` websockets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -280,7 +280,7 @@ op-revm = { version = "15.0.0", default-features = false }
 tokio = "1.48.0"
 tokio-util = "0.7.4"
 tokio-stream = "0.1.17"
-tokio-tungstenite = "0.28.0"
+tokio-tungstenite = { version = "0.28.0", features = ["native-tls"] }
 
 # async
 futures = "0.3.31"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -280,7 +280,7 @@ op-revm = { version = "15.0.0", default-features = false }
 tokio = "1.48.0"
 tokio-util = "0.7.4"
 tokio-stream = "0.1.17"
-tokio-tungstenite = { version = "0.28.0", features = ["native-tls"] }
+tokio-tungstenite = "0.28.0"
 
 # async
 futures = "0.3.31"
--- a/crates/client/flashblocks/Cargo.toml
+++ b/crates/client/flashblocks/Cargo.toml
@@ -47,7 +47,7 @@ op-alloy-consensus.workspace = true
 
 # tokio
 tokio.workspace = true
-tokio-tungstenite.workspace = true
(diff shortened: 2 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Cargo.toml
```diff
@@ -280,7 +280,7 @@ op-revm = { version = "15.0.0", default-features = false }
 tokio = "1.48.0"
 tokio-util = "0.7.4"
 tokio-stream = "0.1.17"
-tokio-tungstenite = "0.28.0"
+tokio-tungstenite = { version = "0.28.0", features = ["native-tls"] }
```
meyer9: I think this should be added to the crates where this is included. There are no features in the root Cargo.toml.

meyer9 (changes requested):

BrianBland (dismissed):

Cargo.toml
danyalprout: Happy to do, out of curiosity why is that? Is it better practice?

meyer9: I think it is better practice. Reth has almost no features in the root Cargo.toml, but now that I look at it, only TLS features are specified there, so now I'm less sure.

This is the PR that I remember seeing: https://github.com/base/base/pull/582

danyalprout: TBH this feels like a sensible default to enable for the whole project. But I don't feel strongly. Updated it

meyer9: The idea I guess is that not all crates will require TLS (devnet probably doesn't need it for example). But if you specify it in the root Cargo.toml, you're forcing all crates to include it. Also makes it harder to separate out what features a crate actually requires.

meyer9 (approved):

PR #875, opened 2026-02-25
chore: hoist imports to top of file/mod

danyalprout:
### Description
Hoist imports to top of file/mod. Update claude.md


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/alloy/hardforks/src/chain.rs
```diff
@@ -132,15 +125,14 @@ impl Index<EthereumHardfork> for OpChainHardforks {
 
 #[cfg(test)]
 mod tests {
+    use OpHardfork::*;
```
haardikk21, line 128: This is a pre-existing issue so not a blocker for this PR but `use`ing enum variants as wildcards  is a pretty bad Rust practice

for e.g.

```rs
enum LongEnumName {
    A,
    B,
    C,
}

use LongEnumName::*;
match my_var {
    A => { ... },
    B => { ... },
    C => { ... },
}
```

If we have something like this, there's a few things that can go wrong now:

1. What happens if a variant is removed from the enum? Let's say `C` gets removed

Now, the third match arm - `C` is not recognized as an enum variant but rather a variable name that becomes a catch-all for all variants. The best the compiler can do here is give you an unreachable code warning, but no errors.

2. What happens if then a new variant `D` is added?

`C` being a catch-all now, the match statement will just enter the `C => { ... }` branch upon encountering the variant `D` because `C` is now a variable name, not an enum variant. You will not get any compiler errors about missing match arms.

---

For this particular case, it's a pre-existing issue in a narrow scoped test module and is an enum of hard forks that will likely never have something deleted from it - but still wanted to flag because this sort of practice can have unintended side effects in the future if we look at this and decide to reuse this code pattern elsewhere.

haardikk21 (approved):
One non-blocking comment, rest looks pretty mechanical :)

PR #942, opened 2026-03-01
refactor(proof): replace `BlockDeriver` with proper abstractions in `base-proof`

refcell:
## Summary

PR #925 introduced a `BlockDeriver` struct in `crates/proof/tee/core` to orchestrate oracle-backed L2 block derivation. While the intent was sound -- encapsulating the pipeline setup and execution needed by the TEE executor -- the implementation placed orchestration logic in the wrong layer. `BlockDeriver` was never called anywhere in the codebase, and its body was a near-verbatim recombination of primitives that already exist across `base-proof` and `base-proof-driver`. This PR removes it entirely and distributes the missing pieces to where they naturally belong.

The root issue was that `base-proof` lacked two things: a way to decode an `OutputRoot` from its preimage bytes (the inverse of the existing `OutputRoot::encode`), and a factory that could construct a fully-initialized `OraclePipeline` directly from a `BootInfo`. Without these, callers were forced to re-implement the setup dance at every call site: fetch preimage, decode output root, construct providers, initialize cursor, build pipeline. `BlockDeriver` was one such re-implementation.

`OutputRoot::decode` is now added to `base-protocol` as a symmetric counterpart to `encode`. It returns `None` if the byte slice is the wrong length or specifies an unsupported version, and a new `OracleProviderError::InvalidOutputRootPreimage` variant is added to `base-proof` to represent this failure in oracle context.

The safe-head fetch pattern -- looking up an output root preimage via the oracle and extracting the block hash -- is encapsulated in a `SafeHeadFetcher` unit struct in `base-proof::sync`. Per the project's convention of exporting types rather than loose functions, this lives alongside `new_oracle_pipeline_cursor` and is explicitly re-exported from the crate root.

The most significant addition is `OraclePipeline::from_boot_info`, a `pub async fn` on a specialized impl block pinned to the concrete oracle-backed provider types. This is the single entry point for constructing a ready-to-drive pipeline from a `BootInfo` and an oracle. It calls `SafeHeadFetcher::fetch`, constructs the L1 and L2 chain providers, the blob provider, the DA provider, and the pipeline cursor, then delegates to the existing `OraclePipeline::new`. Error propagation uses `?` throughout, relying on the existing `From<OracleProviderError> for PipelineErrorKind` implementation rather than manual `map_err` wrapping.

The result of this work is that TEE executor code calling into the proof layer no longer needs to know anything about providers, cursors, or pipeline construction. The entire setup is a single `await`:

  ```rust
  use base_proof::{OraclePipeline, boot::BootInfo};
  use base_proof_driver::Driver;

  pub async fn run(oracle: Arc<CachingOracle>, boot_info: &BootInfo) -> Result<()> {
      let (pipeline, cursor) = OraclePipeline::from_boot_info(Arc::clone(&oracle), boot_info).await?;
      let mut driver = Driver::new(pipeline, cursor, oracle);
      let attributes = driver.produce_payload(boot_info.claimed_l2_block_number).await?;
      // ... verify and seal
      Ok(())
  }
  ```


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

0x00101010 (approved):

PR #964, opened 2026-03-02
feat: add hook for sync validation backpressure

meyer9:
Adds a hook that's called when a block is successfully validated. The validation process will wait for the hook to complete before continuing.

For now, the hook is a no-op. In the future, the proofs ExEx will ensure the main sync process never gets too far ahead of the proofs sync process.

This will fix an issue where once the main sync process is 1024 blocks ahead of the proofs sync process, the proofs sync process slows down considerably.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/engine-tree/src/validator.rs
```diff
@@ -77,6 +79,36 @@ use crate::cached_execution::{
...
+/// This allows external components (e.g., the proofs history collector) to receive trie
+/// updates as blocks are validated, avoiding expensive re-execution later.
+pub trait OnValidatedBlockHook: Send + Sync + 'static {
+    /// Called after a block is validated with its execution output and trie updates.
+    fn on_validated_block(
+        &self,
+        block: &SealedBlock<OpBlock>,
+        output: &BlockExecutionOutput<OpReceipt>,
+        trie_updates: &TrieUpdates,
+    ) -> Pin<Box<dyn Future<Output = ()> + Send + '_>>;
```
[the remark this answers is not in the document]
meyer9: We probably don't want this to error. If an error occurs, we should just print an error log. We can do this from inside the hook fn.

crates/client/proofs/src/proofs.rs
```diff
@@ -42,6 +42,14 @@ impl BaseNodeExtension for ProofsHistoryExtension {
         let proofs_history_verification_interval = args.proofs_history_verification_interval;
 
         if proofs_history_enabled {
+            // TODO: replace with real hook that forwards trie updates to proofs collector
+            hooks = hooks.add_add_ons_hook(move |add_ons| {
```
0x00101010, line 64: should this be after `MdbxProofsStorage::new`?

Is this just for holding or will it write to db as well?

0x00101010 (commented):
one comment, other parts looks good

crates/execution/engine-tree/src/validator.rs
```diff
@@ -159,6 +194,7 @@ where
         cached_execution_provider: C,
         changeset_cache: ChangesetCache,
         runtime: reth_tasks::Runtime,
+        on_validated_block: Option<Arc<dyn OnValidatedBlockHook>>,
```
0x00101010, line 188: seems like we have 2 hooks now, invalid_block_hook and this new addition, wondering if we foresee any other potential additions and worth consolidating all the proofs as some kind of a registry to pass in here?

refcell (commented):
This LGTM

crates/execution/engine-tree/src/validator.rs
meyer9: `invalid_block_hook` is from Reth itself. I do agree the arguments to this are getting a bit unwieldy, but I think minimizing diff here makes more sense than trying to refactor (yet).

crates/client/proofs/src/proofs.rs
meyer9: yup, good callout! I'll make that change once I add back pressure logic

PR #1009, opened 2026-03-03
feat(txpool): add transaction forwarder with sliding window rate limiter

haardikk21:
## Summary

Transaction forwarder for the mempool→builder propagation pipeline (PR 2 of 4). Spawns one async task per builder URL, subscribing to the consumer's broadcast channel.

Uses a sliding window rate limiter (default 200 RPS) so transactions are forwarded immediately under normal load. Under spike pressure, incoming txs buffer and flush as a single batch when the window opens (capped at 500 txs per request).

Wire format: `ValidTransaction { sender, raw }` — sender included so the builder skips signer recovery.

## Open Questions

**Buffer bounding**: The forwarder's internal buffer is currently unbounded. Under a sustained spike that exceeds the rate limit for an extended period, the buffer will grow. In practice, the consumer's dedup window (4s) and broadcast channel capacity (10k) limit the inflow rate. Should we add a buffer cap with drop-oldest, or is unbounded acceptable given the natural upstream limits?

## Pipeline Status

| PR | Component | Status |
|----|-----------|--------|
| PR 1 | Consumer | ✅ Merged |
| **PR 2** | **Forwarder** | 🔄 |
| PR 3 | Builder RPC | Planned |
| PR 4 | Node integration | Planned |

### DIFF
--- a/crates/txpool/src/forwarder/mod.rs
+++ b/crates/txpool/src/forwarder/mod.rs
@@ -39,7 +39,10 @@ impl ForwarderHandle {
         let mut tasks = Vec::with_capacity(config.builder_urls.len());
 
         for url in &config.builder_urls {
-            let client = match HttpClientBuilder::default().build(url) {
+            let client = match HttpClientBuilder::default()
+                .request_timeout(Duration::from_secs(5))
+                .build(url)
+            {
                 Ok(client) => client,
                 Err(err) => {
                     error!(
--- a/crates/txpool/src/forwarder/mod.rs
+++ b/crates/txpool/src/forwarder/mod.rs
@@ -12,6 +12,9 @@ pub use config::ForwarderConfig;
 mod metrics;
 pub use metrics::ForwarderMetrics;
 
+mod rpc;
+pub use rpc::{BuilderApiClient, BuilderApiServer};
+
 mod task;
 pub use task::{Forwarder, ValidTransaction};
 
@@ -37,11 +40,12 @@ impl ForwarderHandle {
     {
         let cancel = CancellationToken::new();
         let mut tasks = Vec::with_capacity(config.builder_urls.len());
+        let config = Arc::new(config);
 
         for url in &config.builder_urls {
             let client = match HttpClientBuilder::default()
                 .request_timeout(Duration::from_secs(5))
-                .build(url)
+                .build(url.as_str())
             {
                 Ok(client) => client,
                 Err(err) => {
@@ -55,12 +59,12 @@ impl ForwarderHandle {
             };
 
             let receiver = sender.subscribe();
-            let metrics = ForwarderMetrics::new(url);
+            let metrics = ForwarderMetrics::new(url.as_str());
             let forwarder = Forwarder::new(
                 url.clone(),
                 client,
                 receiver,
-                config.clone(),
+                Arc::clone(&config),
                 metrics,
                 cancel.child_token(),
             );
@@ -84,17 +88,17 @@ impl ForwarderHandle {
 impl ForwarderHandle {
     /// Gracefully shuts down all forwarder tasks.
     ///
-    /// Signals cancellation and waits up to 5 seconds for in-flight RPC
-    /// requests to complete before abandoning them.
+    /// Signals cancellation and waits up to 30 seconds for each forwarder to
+    /// drain its buffer and complete in-flight RPC requests.
     pub async fn shutdown(mut self) {
         self.cancel.cancel();
 
         let tasks = std::mem::take(&mut self.tasks);
-        let results =
-            tokio::time::timeout(Duration::from_secs(30), futures::future::join_all(tasks)).await;
-
-        if let Err(err) = results {
-            warn!(error = %err, "forwarder tasks did not finish within shutdown timeout");
+        if tokio::time::timeout(Duration::from_secs(30), futures::future::join_all(tasks))
+            .await
+            .is_err()
+        {
+            warn!("forwarder tasks did not finish within shutdown timeout");
         }
     }
 }
--- a/crates/txpool/src/forwarder/mod.rs
+++ b/crates/txpool/src/forwarder/mod.rs
@@ -22,6 +22,7 @@ pub use task::{Forwarder, ValidTransaction};
 ///
 /// Cancels all forwarder tasks on drop. For a clean shutdown that waits for
 /// in-flight RPC requests to complete, use [`ForwarderHandle::shutdown`].
+#[must_use = "dropping the handle cancels all forwarder tasks without draining buffers — call shutdown() for graceful termination"]
 pub struct ForwarderHandle {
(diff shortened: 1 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/txpool/src/forwarder/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+        T: PoolTransaction + 'static,
+        <T as PoolTransaction>::Consensus: alloy_eips::Encodable2718,
+    {
+        let cancel = CancellationToken::new();
+        let mut tasks = Vec::with_capacity(config.builder_urls.len());
+        let config = Arc::new(config);
+
+        for url in &config.builder_urls {
+            let client = match HttpClientBuilder::default()
+                .request_timeout(Duration::from_secs(5))
```
refcell: Should the timeout here be a bit tighter than 5s?

crates/txpool/src/forwarder/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+
+            let receiver = sender.subscribe();
+            let metrics = ForwarderMetrics::new(url.as_str());
+            let forwarder = Forwarder::new(
+                url.clone(),
+                client,
+                receiver,
+                Arc::clone(&config),
+                metrics,
+                cancel.child_token(),
```
refcell, line 70: Very nice use of the cancellation token

crates/txpool/src/forwarder/mod.rs
haardikk21: yeah im gonna just make it configurable actually with a 1s default

crates/txpool/src/forwarder/task.rs
```diff
@@ -0,0 +1,322 @@
...
+                    self.handle_recv(result)
+                }
+            };
+            if closed {
+                break;
+            }
+            if !self.buffer.is_empty() && self.limiter.check_rate_limit().is_none() {
+                self.flush_buffer().await;
+            }
+        }
```
refcell, line 166: I _think_ this could be collapsed into a single rather than multiple tokio selects nested in logic.

```rust
loop {
    tokio::select! {
        // cancellation token
        // receiver
        // timer
    }
}
```

haardikk21: i dont think we can because line 139 (first `select!`) is selecting on cancellation, sleep, and recv. The timer sleep arm is the rate limit waiter, we need to wake up when the rate limit window gives us new capacity, but if a new tx arrives while we are waiting we can just buffer it

line 154 is the happy path which only selects on cancellation and receive, no timer there because we are not rate limited yet, just waiting for the next txn

if we collapsed these, we will always need to have a timer future and need to set it to something like `sleep(Duration::MAX)` or something to  treat it like a no-op match arm which is arguably uglier? 

the top level rate limiter match statement is used to just determine which of the two select sets we go into because it changes based on whether or not we are currently being rate limited

crates/txpool/src/forwarder/task.rs
```diff
@@ -0,0 +1,322 @@
...
+/// `max_batch_size`) once the window opens.
+pub struct Forwarder<T: PoolTransaction> {
+    builder_url: url::Url,
+    client: HttpClient,
+    receiver: broadcast::Receiver<Arc<ValidPoolTransaction<T>>>,
+    config: Arc<ForwarderConfig>,
+    metrics: ForwarderMetrics,
+    cancel: CancellationToken,
+    limiter: RateLimiter,
+    buffer: Vec<ValidTransaction>,
```
wlawt, line 96: i think with the consumer's channel cap and deduping window, along with the forwarder's waiting period, this should be ok. 

one scenario i can think of would be that transactions that are too old by the minimum timestamp definition reaches the builder and is no longer the best based on CoinbaseTip so we'd have to wait for the builder's txpool eviction to clear those out (i mean this is always a thing, but the goal is the reduce the builder's load as much as possible). 

perhaps we should have a metric to keep track of the buffer size? then we could see if we need a buffer cap

crates/txpool/src/forwarder/rpc.rs
```diff
@@ -0,0 +1,22 @@
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+
+use super::task::ValidTransaction;
+
+/// RPC interface for submitting pre-validated transactions to a block builder.
+#[rpc(server, client, namespace = "base")]
+pub trait BuilderApi {
+    /// Inserts a batch of transactions with pre-recovered senders.
+    #[method(name = "insertValidatedTransactions")]
+    async fn insert_validated_transactions(&self, txs: Vec<ValidTransaction>) -> RpcResult<()>;
```
wlawt, line 10: @haardikk21  should we just accept a `ValidTransaction` instead of a Vec here and let the json-rpc handle the batching?

haardikk21: i dont have a strong preference, is there a reason to prefer one over the other? i think this format may actually be a bit more concise vs. jsonrpc batch encoding

either way, let's address that in the builder PR if we do need the change

wlawt: kk sounds good, Julian pointed out handling errors might be better so i just wanted to bring it back up again

wlawt (approved):

crates/txpool/src/forwarder/rpc.rs
haardikk21: thats fair, if we do a batch like that we can get individual errors on a per-tx basis back as a response, and then maybe remove txs from the buffer that have been invalidated during the time delta of mempool vs builder ingestion

edit: tho i think the txpool eviction policies may take care of that for us eventually anyway, now that txs are no longer local

PR #1041, opened 2026-03-04
feat: implement `base_insertValidatedTransaction`

wlawt:
- Implements `base_insertValidatedTransaction` in the txpool crate so that the crate/builder can import crate/txpool
- Changes fn signature from `Vec<ValidatedTransaction>` to `ValidatedTransaction` because:
   - JSON-RPC can handle batches
   - We can return errors on a per-tx basis
   - This makes the implementation logic of the RPC much simpler and easier to read/reason about
- Moved `txpool/forwarder/rpc.rs` -> `txpool/builder/rpc.rs` . easier to reason about how this RPC is w.r.t the builder and not the forwarder
- Currently the RPC calls `pool.add_external_transaction`. This does validation again, however, from our internal benchmark this doesn't really seem like a bottleneck. Using `PoolInner` to bypass the validation seems like a bigger lift with setting up the `TransactionValidationOutcome` struct properly. We can do this in a follow-up if there's significant performance degradation

Next up:
- [ ] Create a dedicated thread in the crate/builder for handling this RPC.
- [ ] Create the TxForwardingExtension which will let us write a proper integration test using TestHarness, similar to how `client/metering/rpc.rs` does the testing

TODO:
- [x] Unit tests for tx decoding


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/txpool/src/builder/rpc.rs
```diff
@@ -0,0 +1,92 @@
...
+            ErrorObjectOwned::owned(
+                ErrorCode::InvalidParams.code(),
+                format!("failed to decode transaction: {e}"),
+                None::<()>,
+            )
+        })?;
+        let encoded_len = tx.raw.len();
+
+        // Create recovered transaction without needing signer recovery
+        let recovered = Recovered::new_unchecked(consensus_tx, sender);
```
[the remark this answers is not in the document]
wlawt: This is ok, only the mpool nodes will be able to call this fn

niran:
> Currently the RPC calls pool.add_external_transaction. This does validation again, however, from our internal benchmark this doesn't really seem like a bottleneck.

If this is the case, is there any remaining reason to not just use `sendRawTransaction`? Maybe I'm missing something.

wlawt:
atm, i'm still exploring a minimal diff to side stepping validation

however, there are plans for a fast follow to extend the wire format to include the fully decoded tx to send to this RPC that sendRawTransaction wouldn't be able to accept

edit: just a thought, it may be worthwhile to keep the validation in the builder here because a tx that takes longer to reach this RPC due to backoff or getting throttled by the RPS limit could end up being invalid by the time it reaches the builder. this is still holding the assumption that validation is relatively inexpensive in comparison to signer recovery etc.

crates/txpool/src/builder/rpc.rs
```diff
@@ -0,0 +1,93 @@
...
+        let pool_tx = BasePooledTransaction::new(recovered, encoded_len);
+
+        // Insert into the pool
+        let start = Instant::now();
+        let result = self.pool.add_external_transaction(pool_tx).await;
```
wlawt, line 77: some thoughts related to the design:

- re-validating at the builder might make sense. this is because the tx might not be valid by the time it arrives at the builder (i.e., nonce gap etc.)
- alternatively, we should update the wire format to send `TransactionValidationOutcome` where we make the mpool code more verbose. That is, instead of calling `pool.add_transaction`, we would call `pool.inner().validate_transaction()` to get the `TransactionValidationOutcome` and then call `pool.inner().add_transactions()`. We can then just call that exact same fn in the builder. This again, brings up the staleness issue where a tx might not be valid anymore when it arrives to the builder

i think for now, we should just measure the insertion duration w/ this validation on Mainnet/Sepolia and go from there. Bypassing the validation gives me the feeling that there'll be a lot of unknown edge cases/race conditions that come up

haardikk21:
> > Currently the RPC calls pool.add_external_transaction. This does validation again, however, from our internal benchmark this doesn't really seem like a bottleneck.
> 
> If this is the case, is there any remaining reason to not just use `sendRawTransaction`? Maybe I'm missing something.

There's gonna be a follow-up PR shortly to add minimal bundles support, for which `eth_sendRawTransaction` is not a good fit - so designing around this new RPC from the get go makes that code diff simpler

crates/txpool/src/builder/rpc.rs
haardikk21: revalidating is fine imo, the only thing i would have liked to skip is sender recovery but the other validations still make sense to do - especially because they're using different ordering mechanisms + we need to have custom validation code on top soon for bundles

crates/txpool/src/forwarder/task.rs
```diff
@@ -277,6 +304,17 @@ where
         }
     }
 
+    async fn send_batch(
+        &self,
+        batch: &[ValidTransaction],
+    ) -> Result<BatchResponse<'_, ()>, ClientError> {
+        let mut request = BatchRequestBuilder::new();
+        for tx in batch {
+            request.insert("base_insertValidatedTransaction", (tx,)).expect("valid method name");
```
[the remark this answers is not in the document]
haardikk21: this is a sad upstream limitation of not being able to use the generated client traits when doing batch requests. as long as this remains the only place where the method name is hardcoded, im fine with it without extracting it to a constant as we should have tests to test this workflow anyway which will start failing if the name ever changes.

crates/txpool/src/builder/metrics.rs
```diff
@@ -0,0 +1,20 @@
...
+pub struct BuilderApiMetrics {
+    /// Transactions successfully inserted into the pool.
+    #[metric(describe = "Transactions successfully inserted into the pool")]
+    pub txs_inserted: Counter,
+    /// Transactions that failed to decode.
+    #[metric(describe = "Transactions that failed to decode")]
+    pub decode_errors: Counter,
+    /// Transactions rejected by the pool.
+    #[metric(describe = "Transactions rejected by the pool")]
+    pub pool_rejections: Counter,
```
haardikk21: to keep things consistent with the above metric of `txs_inserted`, can we rename this to `txs_rejected`?

crates/txpool/src/builder/metrics.rs
```diff
@@ -0,0 +1,20 @@
...
+    /// Insert duration.
+    #[metric(describe = "Insert duration")]
+    pub insert_duration: Histogram,
+}
```
haardikk21: a slightly more verbose description here would be nice "Time to insert a transaction in the local txpool"

crates/txpool/src/builder/rpc.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// Inserts a single pre-validated transaction into the builder's pool.
+    ///
+    /// The transaction is EIP-2718 encoded with its sender address pre-recovered,
+    /// allowing the builder to skip signature verification.
+    ///
+    /// Returns an error if decoding or pool insertion fails.
+    /// Use JSON-RPC batch requests for efficient bulk submission.
+    #[method(name = "insertValidatedTransaction")]
+    async fn insert_validated_transaction(&self, tx: ValidTransaction) -> RpcResult<()>;
+}
```
haardikk21: rename `ValidTransaction` -> `ValidatedTransaction` imo (apologies if i had it as that before)

crates/txpool/src/forwarder/metrics.rs
```diff
@@ -9,6 +9,8 @@ pub struct ForwarderMetrics {
     pub txs_forwarded: Counter,
     /// Total RPC send errors (after all retries exhausted).
     pub rpc_errors: Counter,
+    /// Total transactions rejected by the builder's pool within successful batch calls.
+    pub tx_batch_rejections: Counter,
```
haardikk21: this name is unintuitive to me based on the given description. i get what its trying to communicate based on the description, but the name by itself sounds like "# of tx batches that got rejected" rather than "# of txs per-batch that got rejected"

perhaps slightly more verbose name of `tx_rejections_in_batch` or something could be better here

haardikk21 (commented):
minor nits, otherwise lgtm

crates/txpool/src/forwarder/metrics.rs
wlawt: how about `num_tx_rejected_in_batch` ?

haardikk21: yea thats fine

haardikk21 (approved):

PR #1047, opened 2026-03-04
feat: gate follow sync behind proofs progress with --proofs flag

meyer9:
When --proofs is passed, the `DelegateL2DerivationActor` queries `debug_proofsSyncStatus` on the local L2 RPC and caps sync at proofs_latest + 512 blocks to prevent proofs from falling behind.


bin/consensus/src/cli.rs
```diff
@@ -97,6 +97,10 @@ pub struct Follow {
     #[command(flatten)]
     pub logging: LogArgs,
 
+    /// Gate sync behind proofs progress via `debug_proofsSyncStatus`.
+    #[arg(long = "proofs", env = "BASE_NODE_PROOFS")]
+    pub proofs: bool,
+
```
refcell, line 102: Thought: if the user runs `base-consensus follow` without the `--proofs` flag maybe we should have it return early with a warning message that this subcommand is only intended to be used with proofs? And they can run it without by passing a cli flag like `--dangerously-without-proofs` or something

refcell (commented):
Looks good, just one ux idea

crates/consensus/service/src/actors/derivation/delegate_l2/actor.rs
```diff
@@ -167,6 +184,43 @@ where
             .await
             .map_err(|e| DerivationError::Sender(Box::new(e)))?;
 
+        let sync_limit = if self.proofs_enabled {
+            match self
+                .local_l2_provider
+                .raw_request::<_, ProofsSyncStatus>("debug_proofsSyncStatus".into(), ())
+                .await
```
refcell, line 232: We probably want to fire this right away on startup and make sure it doesn't error with x amount of retries in case the user accidentally configured the wrong node. Otherwise it's just spam warn logs that may be less obvious. If we error right away on startup, we can output an error message to stdout that says "debug_proofsSyncStatus endpoint doesn't appear to be supported by endpoint: <...>"

crates/consensus/service/src/actors/derivation/delegate_l2/actor.rs
```diff
@@ -167,6 +184,43 @@ where
...
+            {
+                Ok(status) => {
+                    let cap = status.latest + PROOFS_MAX_BLOCKS_AHEAD;
```
refcell: Maybe we should make `PROOFS_MAX_BLOCKS_AHEAD` configurable and use this as a sensible default? May make it easier for us to use.

bin/consensus/src/cli.rs
meyer9: I added a warning, but I don't think we need `--dangerously-without-proofs` since it isn't really dangerous. Using it without proofs could be useful to run a node without a CL.

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (approved):
Makes sense ✅

meyer9:
ah sorry for ping, I thought it would require re-approving heimdall

PR #1063, opened 2026-03-05
feat(challenger): Challenger output root validation

jackchuma:
### What changed? Why?
Adds `OutputValidator` to `base-challenger` that verifies final and intermediate output roots for candidate dispute games. It fetches L2 headers and storage proofs, recomputes expected roots via `output_root_v0`, and compares them against onchain claims — this is the core logic that determines whether a game should be challenged.

### How has it been tested?
All 43 tests pass.

### Notes to reviewers
Includes defense-in-depth checks beyond original requirements: RPC header hash verification against consensus hash, u64 overflow protection for adversarial onchain values, and checkpoint count validation to prevent silently skipping unverified roots.

### Change management
type=routine
risk=low
impact=sev5

### Backwards Compatibility
backwards_compatible=true

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/challenge/Cargo.toml
```diff
@@ -13,7 +13,9 @@ workspace = true
 
 [dependencies]
 # Alloy
+alloy-consensus.workspace = true
```
0x00101010, line 16: style: waterfall for CLAUDE.md

jackchuma: Isn't this already correct? The lines are sorted in ascending order of line length

0x00101010:
TLDR: feel like there's oppourtunity to consolidate common logic and have validate_final_root / validate_intermediate_roots as thin wrappers of core logic

## Suggestion: consolidate `validate_final_root` and `validate_intermediate_roots`

Both methods repeat the same ~20 lines of boilerplate:
- `Instant::now()` → `elapsed` → `VALIDATION_LATENCY_SECONDS` histogram
- `compute_output_root` error handling → `VALIDATION_ERRORS_TOTAL` counter + `warn!`
- Validity check → `GAMES_INVALID_TOTAL` counter + `warn!`
- Constructing `ValidationResult`

This appears 3 times total (once in `validate_final_root`, twice in `validate_intermediate_roots` — once for the stream error path, once for the success comparison).

**Recommendation:** keep both public methods, but extract a private helper that owns the compute → compare → metrics pipeline:

```rust
struct Checkpoint {
    index: usize,
    block: u64,
    claimed_root: B256,
}

/// Computes roots for all checkpoints, returns first mismatch (if any).
async fn validate_output_roots(
    &self,
    game_address: Address,
    checkpoints: &[Checkpoint],
) -> Result<Option<(usize, B256, B256)>, ValidatorError> { ... }
```

Then both public methods become thin wrappers:
- `validate_final_root` → logging + calls helper with a single-element slice + builds `ValidationResult`
- `validate_intermediate_roots` → interval/range/count validation + checkpoint computation + logging + calls helper with the checkpoint vec + builds `ValidationResult`

Metrics recording (latency, errors, invalids) lives in one place — either inside the helper or in small `record_latency` / `record_compute_failure` associated functions.

**Why not a single public method:** the two methods have legitimately different input validation (interval ≠ 0, block range, checkpoint count, overflow checks). An enum input saves zero lines and hurts readability at call sites. The public API is clean as-is.

**One semantic question:** does `validate_intermediate_roots` also need to validate the final claimed root? Currently it only checks intermediate checkpoints. If the caller always calls `validate_final_root` first, the helper should stay checkpoint-only. If they should be combined into one pass, the helper should compute the final root too. Worth clarifying.

crates/proof/challenge/Cargo.toml
jackchuma: The other groupings are not sorted though. Fixing now

crates/proof/challenge/src/validator.rs
```diff
@@ -0,0 +1,918 @@
...
+            checkpoint = checkpoint
+                .checked_add(intermediate_block_interval)
+                .ok_or(ValidatorError::ArithmeticOverflow { block_number: checkpoint })?;
+        }
+
+        let checkpoints: Vec<Checkpoint> = checkpoint_blocks
+            .iter()
+            .zip(intermediate_roots.iter())
+            .map(|(&block, &root)| Checkpoint { block, claimed_root: root })
+            .collect();
```
0x00101010: nit: This block can be simplified — the while-loop has a redundant double-guard (`len < count` in the while condition and `len >= count` break inside the body), and builds an intermediate `Vec<u64>` just to zip it into `Vec<Checkpoint>` one line later.

Since `intermediate_roots.len() == expected_count` is already validated above, the checkpoint block numbers are a simple arithmetic sequence: `starting + (i+1) * interval`. This can be a single iterator chain that builds `Vec<Checkpoint>` directly — no mutable state, no intermediate allocation:

```rust
let checkpoints = intermediate_roots
    .iter()
    .enumerate()
    .map(|(i, &root)| {
        let multiplier = u64::try_from(i + 1)
            .map_err(|_| ValidatorError::ArithmeticOverflow { block_number: starting_block_number })?;
        let offset = intermediate_block_interval
            .checked_mul(multiplier)
            .ok_or(ValidatorError::ArithmeticOverflow { block_number: starting_block_number })?;
        let block = starting_block_number
            .checked_add(offset)
            .ok_or(ValidatorError::ArithmeticOverflow { block_number: starting_block_number })?;
        Ok(Checkpoint { block, claimed_root: root })
    })
    .collect::<Result<Vec<_>, ValidatorError>>()?;
```

Overflow protection is preserved via `checked_mul` + `checked_add`.

0x00101010 (commented):
## Suggestion: make `validate_output_roots` more idiomatic

Three changes to reduce duplication and simplify control flow:

### 1. Drop guard for latency recording

The latency metric is recorded in 3 separate places. A scope guard records it exactly once regardless of exit path:

```rust
/// Records validation latency on drop, ensuring exactly one measurement
/// regardless of exit path (success, error, or early return).
struct RecordOnDrop(Instant);

impl Drop for RecordOnDrop {
    fn drop(&mut self) {
        metrics::histogram!(ChallengerMetrics::VALIDATION_LATENCY_SECONDS)
            .record(self.0.elapsed().as_secs_f64());
    }
}
```

### 2. Remove `mut first_invalid` + `break` pattern, drop dead empty-checkpoints branch

`first_invalid` is set once then immediately `break`ed. Return directly instead. The empty-checkpoints early return is also unnecessary since the `while let` loop naturally does nothing for an empty stream.

### 3. Use `inspect_err` + `?` instead of `match` + explicit `return Err`

Signals "log on error, then propagate" without transforming the error.

### Recommended code

```rust
async fn validate_output_roots(
    &self,
    game_address: Address,
    checkpoints: &[Checkpoint],
) -> Result<Option<(usize, B256)>, ValidatorError> {
    let _latency = RecordOnDrop(Instant::now());

    let mut stream = stream::iter(checkpoints.iter().enumerate())
        .map(|(idx, cp)| async move { (idx, cp, self.compute_output_root(cp.block).await) })
        .buffered(Self::VALIDATION_CONCURRENCY);

    while let Some((idx, cp, result)) = stream.next().await {
        let expected_root = result.inspect_err(|e| {
            metrics::counter!(ChallengerMetrics::VALIDATION_ERRORS_TOTAL).increment(1);
            warn!(
                game = %game_address,
                block = cp.block,
                index = idx,
                error = %e,
                "output root computation failed"
            );
        })?;

        if expected_root != cp.claimed_root {
            warn!(
                game = %game_address,
                block = cp.block,
                index = idx,
                expected = %expected_root,
                claimed = %cp.claimed_root,
                "invalid output root detected"
            );
            metrics::counter!(ChallengerMetrics::GAMES_INVALID_TOTAL).increment(1);
            return Ok(Some((idx, expected_root)));
        }
    }

    Ok(None)
}
```

24 lines vs 38 — no `mut` bindings, no duplicated metrics calls, no dead branch, and standard `inspect_err` + `?` idiom.

0x00101010 (approved):

PR #1174, opened 2026-03-08
refactor(proposer): Consolidate signal handling into RuntimeManager

jackchuma:
### What changed? Why?
Consolidates the proposer's duplicated signal handling (`SignalHandler`) into `RuntimeManager::install_signal_handler()` in `base-cli-utils`, making SIGINT+SIGTERM support with cooperative `CancellationToken`-based shutdown available to all workspace crates. The proposer's `signal.rs` is deleted and `service.rs` now calls the shared implementation. Runtime behavior is unchanged.

### How has it been tested?
All 129 tests pass. CI checks (cargo check, cargo clippy, linting) pass without warnings. No existing tests were modified; the moved logic is functionally identical to the deleted code.

### Notes to reviewers
Removing `SignalHandler` from `base-proposer`'s public API is technically a breaking change, but it was only used internally. This will be used by the challenger as well in a follow-up.

### Change management
type=routine
risk=low
impact=sev5

### Backwards Compatibility
backwards_compatible=true

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proposer/src/service.rs
```diff
@@ -51,7 +52,7 @@ pub async fn run(config: ProposerConfig) -> Result<()> {
 
     // ── 1. Global cancellation token and signal handler ──────────────────
     let cancel = CancellationToken::new();
-    crate::SignalHandler::install(cancel.clone());
+    let _signal_handle = RuntimeManager::install_signal_handler(cancel.clone());
```
leopoldjoy: Since the `_signal_handle` is unused (the returned `JoinHandle<()>` is bound to `_signal_handle` but never awaited or aborted), it may be worth considering whether:

1. the API should just return `()` like the old `SignalHandler::install` did if no caller ever needs the handle, or
2. continue returning `JoinHandle` since its arguably the better API for future flexibility, but then consider adding a documentation comment noting "the handle can be safely dropped".

crates/utilities/cli/src/runtime.rs
```diff
@@ -15,6 +21,40 @@ impl RuntimeManager {
...
+    /// [`CancellationToken`] is cancelled, allowing all holders of child tokens
+    /// to begin cooperative shutdown.
+    pub fn install_signal_handler(cancel: CancellationToken) -> JoinHandle<()> {
+        tokio::spawn(async move {
+            #[cfg(unix)]
+            {
+                use tokio::signal::unix::{SignalKind, signal};
+                let mut sigterm =
+                    signal(SignalKind::terminate()).expect("failed to register SIGTERM handler");
+                tokio::select! {
```
leopoldjoy, line 37: Nit: it seems like there may now be [inconsistency with `run_until_ctrl_c`](https://github.com/base/base/pull/1174/changes#diff-8ef9a50cbbbfc9597f67914e1deabd720c58992091c2a8b5fc147d123877e8e4R68): The pre-existing `run_until_ctrl_c` method on the same struct only handles `SIGINT (Ctrl-C)`, not `SIGTERM`. Now that this `install_signal_handler` function handles both, the two methods have divergent shutdown semantics.

This doesn't need to be addressed on this PR, but it may be worth addressing in a follow-up PR to align them.

crates/utilities/cli/src/runtime.rs
```diff
@@ -15,6 +21,40 @@ impl RuntimeManager {
...
+    /// only SIGINT (Ctrl-C) is handled. When a signal is received the
+    /// [`CancellationToken`] is cancelled, allowing all holders of child tokens
+    /// to begin cooperative shutdown.
+    pub fn install_signal_handler(cancel: CancellationToken) -> JoinHandle<()> {
+        tokio::spawn(async move {
+            #[cfg(unix)]
+            {
+                use tokio::signal::unix::{SignalKind, signal};
+                let mut sigterm =
+                    signal(SignalKind::terminate()).expect("failed to register SIGTERM handler");
```
leopoldjoy, line 36: The `.expect()` calls (on this line and the ones below) will panic on the spawned task if signal registration fails. Tokio swallows panics on spawned tasks by default (they don't propagate to the caller), so a failure here would silently leave the process without any signal handler.

I realise this is pre-existing behaviour, but now that this is a shared utility, it may be worth considering returning a `Result` or logging an error before panicking.

crates/utilities/cli/src/runtime.rs
jackchuma: yeah I think the inconsistency is fine - they are two separate functions that do different things

crates/utilities/cli/src/runtime.rs
jackchuma: That's what the `JoinHandle` return type is for

leopoldjoy: Fair point in principle. However, as noted [here](https://github.com/base/base/pull/1174/changes/BASE..3f84eb8ca7634efae0bff269fab7448d33648e29#r2905659104), the current callsite binds it to `_signal_handle` and never awaits the `JoinError`, so in practice the panic is silently swallowed today.

crates/proof/proposer/src/service.rs
jackchuma: I think returning the value still makes sense - just added proper handling to the propser

0x00101010 (approved):

leopoldjoy (approved):

leopoldjoy (approved):

PR #1198, opened 2026-03-09
feat(actions): Sequencer Actions Tests

refcell:
## Summary

> [!WARNING]
> Stacked on #1200. Do not merge until that PR lands.

Adds `L2Sequencer` to the action test harness, ported to use the production `OpBlock` types introduced by #1200.

The sequencer builds `OpBlock`s anchored to the L1 chain: each block contains a correct L1-info deposit transaction (built via `L1BlockInfoTx::try_new_with_deposit_tx`) as its first transaction, mirroring real sequencer output. No EVM execution is performed and no user transactions are included — state roots are `B256::ZERO` — which is sufficient for testing epoch selection and end-to-end derivation without the overhead of full block execution.

Because the sequencer implements `L2BlockProvider<Item = OpBlock>`, it feeds directly into `Batcher`, which decodes the epoch from the deposit tx per the logic introduced in #1200.

### DIFF
--- a/actions/harness/src/harness.rs
+++ b/actions/harness/src/harness.rs
@@ -6,8 +6,8 @@ use base_protocol::{BlockInfo, L2BlockInfo};
 
 use crate::{
     ActionDataSource, ActionL1ChainProvider, ActionL2ChainProvider, Batcher, BatcherConfig,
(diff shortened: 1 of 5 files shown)


refcell:
Closing — the sequencer is substantially redundant with `L2BlockBuilder` introduced in #1200. Both implement the same epoch selection logic and L1-info deposit construction. The derivation and batch submission tests in #1200 already cover the same scenarios. Will delete the branch.

PR #1242, opened 2026-03-11
chore(proof/tee): remove unused code from base-enclave

0x00101010:
## Summary

- Remove ~3,200 lines of dead code from `base-enclave` and `base-enclave-client`
- Delete `executor/`, `providers/`, and `config/` modules that formed an unused internal execution pipeline (the nitro server uses `base-proof-client` instead)
- Remove 12 unused Cargo.toml dependencies (revm, base-revm, alloy-evm, base-proof-executor, etc.)
- Remove dead `EnclaveProvider` trait from `base-enclave-client`

### Relocated before deletion
- `ExecutionWitness` → `types/witness.rs` (used by proposer)
- `l2_block_to_block_info()` → `types/l2_block_ref.rs` with new standalone `L2BlockRefError` (used by proposer)

### Verification
- `cargo check -p base-enclave -p base-enclave-client -p base-proposer -p base-challenger -p base-proof-rpc` — 0 warnings
- `cargo test -p base-enclave` — 34/34 pass
- All downstream imports unchanged (same public API paths)

Net: **+132 / -3,399 lines** across 27 files


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/tee/core/src/error.rs
```diff
@@ -28,163 +14,10 @@ pub enum CryptoError {
...
-
-    /// Provider error.
-    #[error(transparent)]
-    Provider(#[from] ProviderError),
-}
-
-/// Errors that can occur during provider operations (L1/L2 data fetching).
+/// Errors that can occur during account proof verification.
 #[derive(Debug, Clone, Eq, PartialEq, Error)]
 pub enum ProviderError {
```
leopoldjoy, line 19: After the removal, `ProviderError` has a single variant `AccountProofFailed`. The name "ProviderError" is a leftover from when it covered many provider-related errors. Consider renaming it to something like `AccountProofError` since it's only about account proof verification now. This is cosmetic though.

crates/proof/tee/core/src/lib.rs
```diff
@@ -1,44 +1,18 @@
...
-pub use base_alloy_consensus::OpReceiptEnvelope;
-pub use base_consensus_genesis::{
-    ChainConfig, ChainGenesis, HardForkConfig, L1ChainConfig, SystemConfig,
-};
-
-mod config;
-pub use config::{
-    default_l1_config, default_rollup_config, l1_config_for_l2_chain_id, sepolia_l1_config,
-};
+pub use base_consensus_genesis::ChainConfig;
```
leopoldjoy, line 5: Nit: the crate re-exports `Header`, `Address`, `B256`, `Bytes`, `U256`, and `ChainConfig` directly from `alloy_*` / `base_consensus_genesis`. These are convenience re-exports that aren't types defined in this crate. If the intent was to slim down, these could also be removed (downstream consumers can import from the original crates). But I understand if you prefer to keep them for API convenience.

crates/proof/tee/core/src/lib.rs
```diff
@@ -1,44 +1,18 @@
...
+pub use proof::{PROOF_TYPE_TEE, ProofEncoder};
 
 mod types;
 pub use types::{
-    AccountResult, AggregateRequest, BlockId, ExecuteStatelessRequest, Genesis,
-    GenesisSystemConfig, MARSHAL_BINARY_SIZE, PerChainConfig, RollupConfig, StorageProof,
-    output_root_v0, output_root_v0_with_hash,
+    AccountResult, AggregateRequest, BlockId, ExecuteStatelessRequest, ExecutionWitness, Genesis,
+    GenesisSystemConfig, L2BlockRefError, PerChainConfig, RollupConfig, StorageProof,
+    l2_block_to_block_info, output_root_v0, output_root_v0_with_hash,
```
leopoldjoy, line 17: `l2_block_to_block_info` is exported as a bare function rather than as a method on a type. The `CLAUDE.md` convention says to "prefer placing functions as methods on a type (even a unit struct) rather than as bare functions." This pre-dates this PR though, so not something to fix here (probably better in a follow-up).

jackchuma (approved):

crates/proof/tee/core/src/error.rs
0x00101010: This will be deleted in later PR, this won't be used any more

crates/proof/tee/core/src/lib.rs
0x00101010: same as above, will be deleted in later PR, we need to change proposer to use the new enclave client

crates/proof/tee/core/src/lib.rs
0x00101010: same as above

PR #1249, opened 2026-03-12
feat(tx-manager): NonceManager with RAII NonceGuard

jackchuma:
### What changed? Why?
Implements a production-ready `NonceManager` for concurrent nonce reservation in `base-tx-manager`. Key design:

- **Lazy initialization** — nonce is fetched from chain via `get_transaction_count()` only on first use; subsequent calls increment locally from cache.
- **Lock-free RPC fetch** — on cache miss the mutex is dropped before the RPC call so concurrent callers aren't blocked by the network round-trip. A generation counter detects `reset()` races mid-flight and retries (up to 5 attempts).
- **RAII `NonceGuard`** — returned by `next_nonce()`, holds an `OwnedMutexGuard` for the caller's entire signing duration, serializing nonce assignment. Supports `rollback()` to restore the nonce on failure; implicit drop finalizes consumption.
- **`Send`-safe** — `OwnedMutexGuard` (via `Arc<Mutex>`) makes the guard `'static` so it can cross `tokio::spawn` boundaries, required for the `send_async()` pattern.
- **RPC timeout** — configurable timeout on chain fetches to prevent indefinite hangs.
- **Error variants** — adds `NonceOverflow` (checked at `u64::MAX`) and `NonceAcquisitionFailed` to `TxManagerError`.

Also adds `RpcErrorClassifier` with `classify_rpc_error()` and `err_string_contains_any()` for mapping raw geth RPC error strings into structured `TxManagerError` variants with retry semantics (`is_retryable()`, `is_already_known()`).

### How has it been tested?
14 integration tests using Anvil cover:
- Initial chain fetch and local increment
- Rollback and drop semantics
- Reset with fresh chain fetch and reset-then-rollback interaction
- Concurrent unique sequential nonces (10 tasks)
- Concurrent next_nonce + reset stress test (5 rounds x 20 tasks, asserting contiguous nonces)
- Reset blocking while guard held (lock contention)
- Provider failure and RPC timeout error paths
- Nonce overflow at `u64::MAX`
- Compile-time `Send` bound assertion

### Notes to reviewers
- The `OwnedMutexGuard` (via `Arc<Mutex>`) design is intentional — it makes `NonceGuard` `'static` so it can cross `tokio::spawn` boundaries, which is required for the `send_async()` pattern where nonce reservation happens before task spawn.
- `NonceState` fields are private with read-only accessors; a test-only constructor (`new_with_nonce`) pre-seeds cache for edge-case testing without chain interaction.
- RPC error classification relies on case-insensitive substring matching against known geth error strings. Other clients (Erigon, Besu) may have different wording and could fall through to the `Rpc(String)` fallback (treated as retryable).

### Change management
type=routine
risk=low
impact=sev4

### Backwards Compatibility
backwards_compatible=true

automerge=false

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -276,6 +276,7 @@ alloy-sol-types = { version = "1.5.0", default-features = false }
 alloy-sol-macro = "1.5.0"
 alloy-rpc-client = "1.5.2"
 alloy-signer-local = "1.5.2"
+alloy-node-bindings = "1.5.2"
 alloy-rpc-types-eth = { version = "1.5.2", default-features = false }
 alloy-transport-http = "1.5.2"
 alloy-rpc-types-debug = { version = "1.6.3", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -276,8 +276,7 @@ alloy-sol-types = { version = "1.5.0", default-features = false }
 alloy-sol-macro = "1.5.0"
 alloy-rpc-client = "1.5.2"
 alloy-signer-local = "1.5.2"
-# pinned: alloy-node-bindings 1.5.3+ pulls in incompatible alloy-hardforks
-alloy-node-bindings = "=1.5.2"
+alloy-node-bindings = "1.5.2"
 alloy-rpc-types-eth = { version = "1.5.2", default-features = false }
 alloy-transport-http = "1.5.2"
 alloy-rpc-types-debug = { version = "1.6.3", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -276,6 +276,7 @@ alloy-sol-types = { version = "1.5.0", default-features = false }
 alloy-sol-macro = "1.5.0"
 alloy-rpc-client = "1.5.2"
 alloy-signer-local = "1.5.2"
+# pinned: alloy-node-bindings 1.5.3+ pulls in incompatible alloy-hardforks
 alloy-node-bindings = "=1.5.2"
 alloy-rpc-types-eth = { version = "1.5.2", default-features = false }
 alloy-transport-http = "1.5.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -276,7 +276,7 @@ alloy-sol-types = { version = "1.5.0", default-features = false }
 alloy-sol-macro = "1.5.0"
 alloy-rpc-client = "1.5.2"
 alloy-signer-local = "1.5.2"
-alloy-node-bindings = "1.5.2"
+alloy-node-bindings = "=1.5.2"
 alloy-rpc-types-eth = { version = "1.5.2", default-features = false }
 alloy-transport-http = "1.5.2"
 alloy-rpc-types-debug = { version = "1.6.3", default-features = false }
--- a/crates/utilities/tx-manager/src/lib.rs
+++ b/crates/utilities/tx-manager/src/lib.rs
@@ -28,7 +28,7 @@ mod traits;
 pub use traits::{SendHandle, SendResponse, TxManager};
 
 mod nonce;
-pub use nonce::{NonceGuard, NonceManager};
+pub use nonce::{NonceGuard, NonceManager, NonceState};
 
 mod manager;
 pub use manager::SimpleTxManager;
--- a/crates/utilities/tx-manager/src/lib.rs
+++ b/crates/utilities/tx-manager/src/lib.rs
@@ -28,7 +28,7 @@ mod traits;
 pub use traits::{SendHandle, SendResponse, TxManager};
 
 mod nonce;
-pub use nonce::NonceManager;
+pub use nonce::{NonceGuard, NonceManager};
 
 mod manager;
 pub use manager::SimpleTxManager;
--- a/crates/utilities/tx-manager/src/nonce.rs
+++ b/crates/utilities/tx-manager/src/nonce.rs
@@ -146,15 +146,12 @@ impl NonceManager {
     pub async fn next_nonce(&self) -> Result<NonceGuard, TxManagerError> {
         for attempt in 0..Self::MAX_RETRY_ATTEMPTS {
             // Acquire the owned lock once upfront.
-            let mut guard = Arc::clone(&self.inner).lock_owned().await;
+            let guard = Arc::clone(&self.inner).lock_owned().await;
 
             // Fast path: cache is populated — read-and-increment in a
             // single lock round-trip.
             if let Some(n) = guard.nonce {
-                let next = n.checked_add(1).ok_or(TxManagerError::NonceOverflow)?;
-                guard.nonce = Some(next);
-                debug!(nonce = n, "nonce reserved");
-                return Ok(NonceGuard { guard: Some(guard), nonce: n });
+                return Self::reserve_nonce(guard, n);
             }
 
             // Cache miss: snapshot the generation, then drop the lock so
@@ -199,14 +196,27 @@ impl NonceManager {
                 debug!(nonce = fetched, "nonce fetched from chain");
                 fetched
             });
-            let next = nonce.checked_add(1).ok_or(TxManagerError::NonceOverflow)?;
-            guard.nonce = Some(next);
-            debug!(nonce, "nonce reserved");
-            return Ok(NonceGuard { guard: Some(guard), nonce });
+            return Self::reserve_nonce(guard, nonce);
         }
 
         warn!(attempts = Self::MAX_RETRY_ATTEMPTS, "nonce acquisition failed after max retries",);
         Err(TxManagerError::NonceAcquisitionFailed)
+    }
+
+    /// Advances the cached nonce by one and returns a [`NonceGuard`]
+    /// holding the lock and the reserved value.
+    ///
+    /// Both the fast path (cache hit) and Phase 3 (after RPC fetch) use
+    /// this single call-site for the `checked_add` overflow check,
+    /// ensuring consistent behavior.
+    fn reserve_nonce(
+        mut guard: OwnedMutexGuard<NonceState>,
+        nonce: u64,
+    ) -> Result<NonceGuard, TxManagerError> {
+        let next = nonce.checked_add(1).ok_or(TxManagerError::NonceOverflow)?;
+        guard.nonce = Some(next);
+        debug!(nonce, "nonce reserved");
+        Ok(NonceGuard { guard: Some(guard), nonce })
     }
 
     /// Clears the cached nonce, forcing a fresh chain fetch on the next
@@ -304,6 +314,20 @@ mod tests {
                 test_barrier: Some(barrier),
             }
         }
+    }
+
+    #[test]
+    fn nonce_state_new_is_none_and_generation_zero() {
+        let state = NonceState::new();
+        assert_eq!(state.nonce(), None);
+        assert_eq!(state.generation(), 0);
+    }
+
+    #[test]
+    fn nonce_state_default_matches_new() {
+        let state = NonceState::default();
+        assert_eq!(state.nonce(), None);
+        assert_eq!(state.generation(), 0);
     }
 
     #[tokio::test]
--- a/crates/utilities/tx-manager/src/nonce.rs
+++ b/crates/utilities/tx-manager/src/nonce.rs
@@ -9,10 +9,38 @@ use tracing::{debug, info, warn};
 
 use crate::TxManagerError;
 
+/// Internal state tracked by [`NonceManager`].
+///
+/// Pairs the optional cached nonce with a generation counter that
+/// distinguishes "never initialized" (`generation == 0`) from
+/// "cleared by [`NonceManager::reset`]" (`generation > 0`).
+#[derive(Debug)]
+pub struct NonceState {
+    /// The cached nonce value, or `None` if uninitialized / reset.
+    pub nonce: Option<u64>,
+    /// Monotonically increasing counter bumped on every
+    /// [`NonceManager::reset`].
+    pub generation: u64,
+}
+
+impl NonceState {
+    /// Creates a new [`NonceState`] with no cached nonce and generation
+    /// zero.
+    pub const fn new() -> Self {
+        Self { nonce: None, generation: 0 }
+    }
+}
+
+impl Default for NonceState {
+    fn default() -> Self {
+        Self::new()
+    }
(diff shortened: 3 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/utilities/tx-manager/src/nonce.rs
```diff
@@ -1,5 +1,113 @@
...
+///
+/// The mutex is held for the duration of transaction signing via the
+/// returned [`NonceGuard`], ensuring sequential nonce assignment even
+/// under concurrent access.
+#[derive(Debug, Clone)]
+pub struct NonceManager {
+    inner: Arc<Mutex<Option<u64>>>,
+    provider: RootProvider,
+    address: Address,
+}
```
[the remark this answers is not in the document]
jackchuma: Worth noting that `RootProvider` in alloy is already the type-erased provider — it uses `BoxTransport` internally, so it works with HTTP, WS, IPC, and any other transport out of the box. The constructors differ (`new_http`, `new_ws`, etc.) but the resulting type is the same concrete `RootProvider`. So the transport concern isn't actually a limitation here.

The testability angle is fair — a generic `P: Provider` would let tests inject a mock without Anvil. That said, adding a type parameter here cascades through every consumer (`SimpleTxManager`, etc.), and since the manager integration is still stubbed out, I'd prefer to defer the generic refactor until we wire `NonceManager` into `SimpleTxManager` and can evaluate the ergonomics holistically. Happy to revisit when we build out the full send path.

crates/utilities/tx-manager/src/nonce.rs
```diff
@@ -1,5 +1,144 @@
...
+                *guard = Some(n + 1);
+                debug!(nonce = n, "nonce reserved");
+                return Ok(NonceGuard { guard: Some(guard), nonce: n });
+            }
+
+            // Rare: reset() cleared the cache between our peek and lock
+            // acquisition. Drop the lock and retry.
+            drop(guard);
+            debug!("nonce cache cleared during acquisition, retrying");
+        }
```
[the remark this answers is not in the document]
jackchuma: Good observation about the redundant RPCs during init. I considered this trade-off and opted for the current approach because: (1) `tokio::sync::OnceCell` doesn't support clearing/resetting, so it wouldn't work with `reset()` without wrapping it in another synchronization layer. (2) A `Notify`-based barrier adds meaningful complexity for a cold-path that only fires on first use or after an explicit `reset()` — both low-frequency events. (3) The current approach is correct under concurrency (first writer wins, others safely use the winning value), and the redundant RPCs are harmless — just slightly wasteful during a narrow startup window. If this becomes a problem in practice (e.g., high-concurrency startup against rate-limited RPCs), we can revisit.

Cargo.toml
```diff
@@ -276,6 +276,8 @@ alloy-sol-types = { version = "1.5.0", default-features = false }
 alloy-sol-macro = "1.5.0"
 alloy-rpc-client = "1.5.2"
 alloy-signer-local = "1.5.2"
+# pinned: alloy-node-bindings 1.5.3+ pulls in incompatible alloy-hardforks
+alloy-node-bindings = "=1.5.2"
```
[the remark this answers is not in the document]
jackchuma: the cargo deny CI check is succeeding. this is not a valid concern

crates/utilities/tx-manager/src/nonce.rs
```diff
@@ -1,5 +1,363 @@
...
+/// distinguishes "never initialized" (`generation == 0`) from
+/// "cleared by [`NonceManager::reset`]" (`generation > 0`).
+#[derive(Debug)]
+pub struct NonceState {
+    /// The cached nonce value, or `None` if uninitialized / reset.
+    pub nonce: Option<u64>,
+    /// Monotonically increasing counter bumped on every
+    /// [`NonceManager::reset`].
+    pub generation: u64,
+}
```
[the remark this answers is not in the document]
jackchuma: This seems like a valid concern but goes against project conventions. Going to ignore

crates/utilities/tx-manager/src/nonce.rs
```diff
@@ -1,5 +1,363 @@
...
+    /// # Caution
+    ///
+    /// The fresh fetch uses the `latest` block tag, so pending-but-unconfirmed
+    /// transactions are not counted. Callers should avoid calling `reset()`
+    /// while transactions are still in-flight, as the freshly fetched nonce
+    /// may conflict with pending transactions in the mempool.
+    pub async fn reset(&self) {
+        let mut guard = self.inner.lock().await;
+        guard.nonce = None;
+        guard.generation = guard.generation.wrapping_add(1);
```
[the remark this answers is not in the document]
jackchuma: This is not a practical concern. ignoring

crates/utilities/tx-manager/src/nonce.rs
```diff
@@ -1,5 +1,363 @@
...
+                // bumped — discard the stale fetch and retry.
+                let mut guard = Arc::clone(&self.inner).lock_owned().await;
+
+                if guard.generation != generation {
+                    drop(guard);
+                    debug!(attempt, "nonce cache reset during RPC fetch, retrying");
+                    continue;
+                }
+
+                let nonce = *guard.nonce.get_or_insert_with(|| {
```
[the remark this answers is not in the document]
jackchuma: nothing worth addressing here

crates/utilities/tx-manager/src/lib.rs
```diff
@@ -28,7 +28,7 @@ mod traits;
 pub use traits::{SendHandle, SendResponse, TxManager};
 
 mod nonce;
-pub use nonce::NonceManager;
+pub use nonce::{NonceGuard, NonceManager, NonceState};
```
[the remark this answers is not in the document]
jackchuma: The workspace conventions (CLAUDE.md) require that all structs, types, enums, and functions within modules be `pub` and properly re-exported from `lib.rs` — no private or `pub(crate)` types. `NonceState` follows this rule. While there's no `NonceManager::state()` accessor today, keeping it exported is consistent with the rest of the workspace and avoids a breaking API change if we add diagnostics later. The public accessors (`nonce()`, `generation()`) are already useful for testing and debugging downstream.

crates/utilities/tx-manager/tests/nonce.rs
```diff
@@ -0,0 +1,286 @@
...
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn concurrent_next_nonce_and_reset_stress() {
+    let (manager, _anvil) = setup();
+
+    // Run several rounds of concurrent next_nonce() calls separated by
+    // resets. Within each round all assigned nonces must be unique —
+    // concurrent callers must never receive the same slot.
+    for round in 0u32..5 {
```
refcell, line 285: I think we should have a test that uses [rayon's parallel iterator](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html) which should help test nonce conflicts

crates/utilities/tx-manager/src/nonce.rs
```diff
@@ -1,5 +1,268 @@
...
+
+/// Internal state tracked by [`NonceManager`].
+///
+/// Pairs the optional cached nonce with a generation counter that
+/// distinguishes "never initialized" (`generation == 0`) from
+/// "cleared by [`NonceManager::reset`]" (`generation > 0`).
+#[derive(Debug, Default)]
+pub struct NonceState {
+    /// The cached nonce value, or `None` if uninitialized / reset.
+    pub nonce: Option<u64>,
```
leopoldjoy, line 20: `NonceState` fields are `pub`, but this is mutex-guarded internal state. External code can construct a `NonceState` and bypass `NonceManager`'s invariants. Other types in this crate (e.g., `SendState`) keep their inner fields private. Could we consider making `nonce` and `generation` private with read-only accessors?

crates/utilities/tx-manager/tests/nonce.rs
```diff
@@ -0,0 +1,312 @@
...
+    );
+
+    // After the reset, next nonce should re-fetch from chain (0).
+    let g = manager.next_nonce().await.unwrap();
+    assert_eq!(g.nonce(), 0);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn rayon_parallel_nonce_acquisition_produces_unique_nonces() {
+    use rayon::prelude::*;
```
leopoldjoy, line 254: `use rayon::prelude::*;` is inside a function body, violating [the codebase rule](https://github.com/base/base/blob/main/CLAUDE.md): "All use imports must be at the top of the file or the top of a mod block. Never place use statements inside function bodies." Move it to the top of the file.

refcell (approved):

crates/utilities/tx-manager/src/nonce.rs
```diff
@@ -1,5 +1,268 @@
...
+/// [`Provider::get_transaction_count`] on first use. Subsequent calls
+/// increment locally without making RPC calls.
+///
+/// The mutex is held for the duration of transaction signing via the
+/// returned [`NonceGuard`], ensuring sequential nonce assignment even
+/// under concurrent access.
+#[derive(Debug, Clone)]
+pub struct NonceManager {
+    inner: Arc<Mutex<NonceState>>,
+    provider: RootProvider,
```
leopoldjoy, line 44: Nit: `NonceManager` is hardcoded to `RootProvider` rather than being generic over `P: Provider`. This forces callers to extract the root provider from their provider stack and prevents use with middleware / fill providers. Consider making it generic (or at least accepting `impl Provider + Clone`).

crates/utilities/tx-manager/src/nonce.rs
```diff
@@ -1,5 +1,268 @@
...
+    /// [`NonceGuard::rollback`] on failure to restore the nonce.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`TxManagerError::Rpc`] if the provider call fails,
+    /// [`TxManagerError::NonceAcquisitionFailed`] if the nonce cache is
+    /// repeatedly cleared by concurrent [`reset`](Self::reset) calls during
+    /// acquisition, or [`TxManagerError::NonceOverflow`] if the nonce
+    /// exceeds `u64::MAX`.
+    pub async fn next_nonce(&self) -> Result<NonceGuard, TxManagerError> {
```
leopoldjoy, line 90: `next_nonce()` returns a `NonceGuard` that holds the mutex for the entire signing duration, serialising all transaction submission through this manager. If signing or sending takes non-trivial time, this becomes a throughput bottleneck — only one transaction can be in-flight at a time. Is serial signing intentional for the target use cases, or should nonce reservation be decoupled from the signing lock?

crates/utilities/tx-manager/src/nonce.rs
```diff
@@ -1,5 +1,268 @@
...
+            // the RPC fetch doesn't block concurrent callers.
+            let generation = guard.generation;
+            drop(guard);
+
+            // Phase 2: fetch from chain *without* holding the lock so
+            // concurrent callers are not blocked by the RPC round-trip.
+            // Multiple concurrent callers may fetch redundantly; only
+            // the first writer's value is used.
+            let fetched = tokio::time::timeout(
+                self.rpc_timeout,
```
leopoldjoy, line 111: `get_transaction_count(self.address)` defaults to the latest block tag, so pending-but-unconfirmed transactions are not counted. After a `reset()` with in-flight transactions, the fetched nonce will collide with pending mempool entries. The `reset()` doc warns about this, but consider accepting a `BlockNumberOrTag` parameter (defaulting to pending) to let callers choose.

crates/utilities/tx-manager/tests/nonce.rs
```diff
@@ -0,0 +1,312 @@
...
+    let handle = tokio::spawn(async move {
+        notify.notify_one();
+        mgr.reset().await;
+        flag.store(true, Ordering::SeqCst);
+    });
+
+    // Wait until the spawned task is about to call reset(), then yield
+    // to let it enter the lock wait.
+    about_to_reset.notified().await;
+    tokio::task::yield_now().await;
```
leopoldjoy, line 233: The `yield_now()` after `notified().await` is a best-effort synchronization — there's no guarantee the spawned task has entered the lock wait by the time the assertion runs. The assertion passes trivially if `reset()` hasn't been called yet. Consider a second `Notify` or `Barrier` so the spawned task signals after it has actually called reset() (and is thus blocked), making the test deterministic.

crates/utilities/tx-manager/tests/nonce.rs
```diff
@@ -0,0 +1,312 @@
...
+        .collect();
+
+    let mut sorted = nonces;
+    sorted.sort();
+    let expected: Vec<u64> = (0..100).collect();
+    assert_eq!(sorted, expected, "all nonces must be unique and contiguous");
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn concurrent_next_nonce_and_reset_stress() {
```
leopoldjoy, line 279: Test name `concurrent_next_nonce_and_reset_stress` is misleading — `reset()` is called sequentially between rounds, never concurrently with `next_nonce()`. The test validates sequential batches separated by resets, not concurrent interleaving of the two operations.

PR #1251, opened 2026-03-12
V1 of load tester

kmchicoine:
Architecture Overview: This introduces a load testing framework with a clean separation of concerns:
 - LoadRunner orchestrates the test: manages accounts, rate-limits tx submission, and coordinates with a background Confirmer
 - WorkloadGenerator + Payload trait handle tx payload creation (transfers, calldata, ERC20, precompiles, Uniswap)
 - Confirmer runs in a separate tokio task polling receipts and tracking per-sender in-flight counts to prevent mempool flooding
 - MetricsCollector aggregates latency/throughput/gas stats

 The flow: LoadRunner::run() → rate-limited loop → WorkloadGenerator creates payloads → NonceTracker allocates nonces → batch submit → Confirmer polls receipts → MetricsCollector summarizes.

Other Features:
- YAML-configurable: RPC, accounts, TPS, duration, transaction types
- Weighted transaction mix: transfers, calldata, ERC20, precompiles
- Per-sender in-flight limiting with adaptive backoff
- Mnemonic or seed-based account derivation
- Metrics: latency percentiles, TPS, gas usage, success rate

Future work not yet supported:
- Flashblocks support
- Multi-endpoint submission


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/infra/load-tests/base-load-tests-core/src/runner/load_runner.rs
kmchicoine, line 1: Core orchestration lives here. Key design decisions:
 - Per-sender in-flight limiting (max_in_flight_per_sender) prevents any single account from flooding the mempool—the runner skips accounts at their limit and round-robins to the next
 - Batched submission with configurable batch_size and batch_timeout balances throughput vs latency
 - Adaptive backoff handles txpool is full errors with exponential backoff rather than failing immediately
 - Providers are pre-cached per address at init time to avoid repeated wallet/provider construction per tx

crates/infra/load-tests/base-load-tests-core/src/runner/confirmer.rs
kmchicoine, line 1: Confirmation tracking is decoupled from submission via channels. The Confirmer:
 - Runs in a background task polling receipts every 500ms
 - Maintains atomic per-sender in-flight counters that LoadRunner reads to enforce rate limits
 - Expires pending txs after 60s to prevent unbounded memory growth
 - Returns TransactionMetrics through a channel for the collector to aggregate

crates/infra/load-tests/base-load-tests-core/src/runner/nonce.rs
kmchicoine, line 1: Nonces are managed explicitly rather than relying on eth_getTransactionCount per tx. The NonceTracker:
- Pre-initializes from on-chain state, then tracks allocations locally
- Detects "gaps" (failed nonces between confirmed and expected) for potential healing
- Handles nonce too low errors by marking as confirmed rather than failing

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/mod.rs
kmchicoine, line 1: The Payload trait defines a pluggable tx generation interface. Each payload type (Transfer, Calldata, ERC20, Precompile, Uniswap) implements generate(rng, from, to) -> TransactionRequest. The WorkloadGenerator selects payloads by weighted random distribution configured via TxConfig.

crates/infra/load-tests/base-load-tests-core/src/workload/accounts.rs
kmchicoine, line 1: AccountPool supports two derivation modes:
 - Seeded RNG: deterministic key generation from a seed (for reproducible tests)
 - Mnemonic: BIP-39 derivation with configurable offset (for pre-funded accounts)

 The sender_offset parameter enables multiple load test instances to use non-overlapping account ranges from the same mnemonic.

crates/infra/load-tests/base-load-tests-core/src/config/test_config.rs
kmchicoine, line 1: Configuration is layered: TestConfig (YAML-loadable) wraps user-facing options, while LoadConfig is the validated runtime config consumed by LoadRunner. This separation allows the example/CLI to handle ergonomics (humantime durations, hex keys) while the core library works with typed values.

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/calldata.rs
```diff
@@ -0,0 +1,59 @@
+use alloy_primitives::{Address, Bytes, U256};
+
+use super::Payload;
+use crate::{rpc::TransactionRequest, workload::SeededRng};
+
+const GAS_PER_NONZERO_BYTE: u64 = 16;
+
+/// Generates ETH transfer transactions with random calldata.
+#[derive(Debug, Clone)]
+pub struct CalldataPayload {
```
BrianBland, line 10: It would be nice to configure compressibility of this calldata, potentially by repeating the same sequence N times (default to no repeating)

crates/infra/load-tests/base-load-tests-core/examples/spam.rs
```diff
@@ -0,0 +1,75 @@
+#![allow(missing_docs)]
```
refcell: Could we remove the missing docs allow, should be pretty quick to get claude to write up some concise docs

crates/infra/load-tests/base-load-tests-core/src/config/test_config.rs
```diff
@@ -0,0 +1,312 @@
...
+    Some("30s".to_string())
+}
+
+const fn default_target_tps() -> Option<u32> {
+    Some(100)
+}
+
+fn default_precompile() -> String {
+    "sha256".to_string()
+}
```
refcell, line 160: I think we should be able to inline these in the `#[serde()]` attribute e.g. `#[serde(default = "10")]` rather than defining methods for each.

crates/infra/load-tests/base-load-tests-core/src/runner/nonce.rs
refcell, line 1: I think we can almost 1:1 re-use the nonce manager that @jackchuma built in https://github.com/base/base/pull/1249 and should be available via the `base-tx-manager` crate in the workspace.

crates/infra/load-tests/base-load-tests-core/src/utils/logging.rs
refcell, line 1: We should be able to re-use the logging setup from our [`base-cli-utils`](https://github.com/base/base/tree/main/crates/utilities/cli) crate rather than duplicating impl here.

crates/infra/load-tests/base-load-tests-core/examples/config.yaml
```diff
@@ -0,0 +1,69 @@
...
+sender_offset: 0
+
+# Max in-flight transactions per sender (default: 16)
+in_flight_per_sender: 16
+
+# Test duration using humantime format: "30s", "5m", "1h 30m" (default: "30s")
+duration: "30s"
+
+# Target transactions per second (default: 100)
+target_tps: 100
```
meyer9: How does this work with differing gas? Maybe we should make this gas per second instead?

crates/infra/load-tests/base-load-tests-core/examples/config.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+# Seed for deterministic account generation when mnemonic not provided (default: 42)
+seed: 42
```
meyer9: maybe the default could be random seed? This will help with sending to random new accounts

crates/infra/load-tests/base-load-tests-core/examples/spam.rs
```diff
@@ -0,0 +1,75 @@
...
+use eyre::{Result, bail};
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<()> {
+    init_tracing();
+
+    let config_path = std::env::args()
+        .nth(1)
+        .map(PathBuf::from)
+        .unwrap_or_else(|| PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("examples/config.yaml"));
```
meyer9: should we require this argument to be set instead of defaulting? `CARGO_MANIFEST_DIR` is only set when using `cargo run` I believe

meyer9: Another option is only erroring if `CARGO_MANIFEST_DIR` is missing.

crates/infra/load-tests/base-load-tests-core/src/config/test_config.rs
```diff
@@ -0,0 +1,312 @@
...
+    Erc20 {
+        /// ERC20 contract address.
+        contract: String,
+    },
+
+    /// Precompile call.
+    Precompile {
+        /// Target precompile (sha256, identity, ecrecover, etc.).
+        #[serde(default = "default_precompile")]
+        target: String,
```
meyer9, line 122: Might make sense to make this an enum with different test scenarios. We may want to test blake2b with a different number of rounds. I think this can be generalized to just choosing different test scenarios that are defined somewhere.

crates/infra/load-tests/base-load-tests-core/src/runner/config.rs
```diff
@@ -0,0 +1,199 @@
...
+
+use crate::{
+    utils::{BaselineError, Result},
+    workload::PrecompileTarget,
+};
+
+/// Configuration for a single transaction type with its weight.
+#[derive(Debug, Clone)]
+pub struct TxConfig {
+    /// Weight for random selection (higher = more frequent).
```
meyer9: might be worth specifying if this is number of transactions or gas that's being weighted

crates/infra/load-tests/base-load-tests-core/src/runner/confirmer.rs
```diff
@@ -0,0 +1,289 @@
...
+            pending_tx,
+            in_flight_per_sender: Arc::new(self.in_flight_per_sender.clone()),
+            total_in_flight: Arc::clone(&self.total_in_flight),
+        };
+        (handle, pending_rx)
+    }
+
+    /// Runs the confirmation loop until stopped.
+    pub async fn run(mut self, mut pending_rx: mpsc::Receiver<PendingTx>) {
+        let client = RpcClient::new(self.rpc_url.clone());
```
meyer9: nit: should this be responsible for initializing the RpcClient? It might make this piece harder to test since it has to accept Url. If this accepted a trait instead that was generic, we could pass in a mock RpcClient.

crates/infra/load-tests/base-load-tests-core/src/runner/confirmer.rs
```diff
@@ -0,0 +1,289 @@
...
+        let current_block = match client.get_block_number().await {
+            Ok(block) => block,
+            Err(e) => {
+                warn!(error = %e, "failed to get current block number");
+                return;
+            }
+        };
+
+        let start_block = if self.last_checked_block == 0 {
+            current_block.saturating_sub(5)
```
meyer9: maybe add comment here explaining reasoning for default 5 block window for checking confirmations (also can move to const to make it more clear)

crates/infra/load-tests/base-load-tests-core/src/runner/confirmer.rs
```diff
@@ -0,0 +1,289 @@
...
+        self.last_checked_block = current_block;
+
+        self.lookup_stragglers(client, &found_in_blocks, &mut confirmed).await;
+
+        for (tx_hash, from) in confirmed {
+            self.pending.remove(&tx_hash);
+            self.decrement_in_flight(&from);
+        }
+
+        for tx_hash in expired {
```
meyer9, line 247: same comment here as the AI about expired + confirmed not necessarily being disjoint. Could cause improper tracking of inflight txs.

crates/infra/load-tests/base-load-tests-core/src/runner/confirmer.rs
```diff
@@ -0,0 +1,289 @@
...
+            if let Some(pending) = self.pending.remove(&tx_hash) {
+                warn!(tx_hash = %tx_hash, from = %pending.from, "transaction expired without confirmation");
+                self.decrement_in_flight(&pending.from);
+            }
+        }
+    }
+
+    async fn lookup_stragglers(
```
meyer9, line 264: can you add docs on what this function is supposed to do?

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/precompile.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    fn generate(&self, rng: &mut SeededRng, _from: Address, _to: Address) -> TransactionRequest {
+        let data = match self.target {
+            PrecompileTarget::Identity => Self::encode_identity_data(rng),
+            PrecompileTarget::Sha256 | PrecompileTarget::Ripemd160 => Self::encode_sha256_data(rng),
+            _ => Bytes::from(rng.gen_bytes::<32>().to_vec()),
+        };
+
+        TransactionRequest::contract_call(self.target.address(), data).with_gas_limit(100_000)
```
meyer9: doesn't have to be this PR, but would be nice to be able to call these in a loop instead of only one per transaction. Also, blake2f with high number of rounds would exceed the gas limit set here, but probably can just be a follow up

meyer9 (commented):
nice, going to be awesome to have this for testing in the future!

crates/infra/load-tests/base-load-tests-core/src/workload/accounts.rs
refcell, line 1: We already have something similar to this in the [`base-test-utils`](https://github.com/base/base/blob/main/crates/utilities/test-utils/src/accounts.rs) crate. Maybe we ought to merge this into `base-test-utils`. By default, it seems like this tool is just using the default accounts anyways which base-test-utils provides, so we could probably just use those by default? Not sure the best way to merge this but worth trying to de-duplicate I think

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/precompile.rs
refcell, line 1: We depend on `revm` in our workspace, and revm already [exports precompiles](https://github.com/bluealloy/revm/blob/main/crates/precompile/src/id.rs). we should just be able to re-use that type rather than maintain our own precompile enum.

refcell (dismissed):
Few comments mainly on code re-use

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/precompile.rs
```diff
@@ -0,0 +1,99 @@
...
+    EcPairing,
+    /// BLAKE2F (0x09) - BLAKE2 compression.
+    Blake2f,
+}
+
+impl PrecompileTarget {
+    const fn address(&self) -> Address {
+        match self {
+            Self::EcRecover => {
+                Address::new([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1])
```
BrianBland: Generally recommend using the `address!` macro, but in this case you can use `Address::with_last_byte(1)`

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/uniswap.rs
```diff
@@ -0,0 +1,150 @@
...
+        router: Address,
+        weth: Address,
+        token: Address,
+        min_amount: U256,
+        max_amount: U256,
+    ) -> Self {
+        Self { router, weth, token, min_amount, max_amount }
+    }
+
+    fn encode_swap_exact_eth_for_tokens(
```
BrianBland: I'd recommend using the alloy `sol!` [macro](https://alloy.rs/guides/static-dynamic-abi-in-alloy/#alloy-abi-101) and `abi_encode()` to cleanly encode these function calls

crates/infra/load-tests/base-load-tests-core/src/config/test_config.rs
kmchicoine: I like this idea- it's a bit bigger change than expected, so I'll implement it in an immediate followup PR

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/precompile.rs
kmchicoine: Nice, I'll note this and make a quick followup PR

crates/infra/load-tests/base-load-tests-core/src/workload/accounts.rs
kmchicoine: Main issue I see is that the `base-test-utils` crate targets local devnet testing- Anvil test accounts, 4 fixed accounts, hardcoded chain id. To hit sepolia/sepolia-alpha, we'll need to use funded accounts on those networks (not Anvil, unless I'm misunderstanding), flexible account numbers (more than 4, maybe on the order of 100+ to hit really high throughput), and more robust parallelization support. 

Merging the load-tests account.rs into base-test-utils could work, but it'd add a lot of bloat and potentially unneeded features that load-tests uses for dynamic account pools. Happy to chat more, but maybe the best/easiest path forward for now is to keep them separate and add deduping as a future effort once load-tests is stable and in use

crates/infra/load-tests/base-load-tests-core/src/runner/load_runner.rs
```diff
@@ -0,0 +1,590 @@
...
+                                backoff_ms = delay.as_millis(),
+                                from = %prepared.from,
+                                nonce,
+                                "txpool full, retrying with adaptive backoff"
+                            );
+                            tokio::time::sleep(delay).await;
+                            continue;
+                        }
+
+                        if error_str.contains("nonce too low") {
```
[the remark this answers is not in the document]
kmchicoine: The code is correct - NonceGuard::drop() auto-commits (does not rollback), which is what we want for nonce too low. No code change needed, but the behavior is verified as correct.

crates/infra/load-tests/base-load-tests-core/src/runner/load_runner.rs
```diff
@@ -0,0 +1,590 @@
...
+                .with_max_fee_per_gas(max_fee)
+                .with_max_priority_fee_per_gas((self.gas_price / 10).max(1))
+                .with_gas_limit(prepared.gas_limit);
+
+            let mut attempts = 0;
+            let max_attempts = 3;
+            let mut should_rollback = false;
+
+            loop {
+                match provider.send_transaction(tx.clone()).await {
```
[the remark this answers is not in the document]
kmchicoine: This is fine for now

crates/infra/load-tests/base-load-tests-core/src/metrics/aggregator.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        let mean = Duration::from_nanos((sum.as_nanos() / len as u128) as u64);
+        LatencyMetrics {
+            min: latencies[0],
+            max: latencies[len - 1],
+            mean,
+            p50: Self::percentile(&latencies, 50),
+            p95: Self::percentile(&latencies, 95),
+            p99: Self::percentile(&latencies, 99),
+        }
```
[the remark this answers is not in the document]
kmchicoine: This is not a concern

crates/infra/load-tests/base-load-tests-core/src/config/test_config.rs
meyer9: sounds good!

meyer9 (dismissed):

crates/infra/load-tests/base-load-tests-core/src/rpc/types.rs
```diff
@@ -0,0 +1,46 @@
+use alloy_primitives::{Address, Bytes, U256};
+
+/// Represents a transaction request to be submitted to the blockchain.
+#[derive(Debug, Clone)]
+pub struct TransactionRequest {
```
refcell, line 5: I think this could be pulled in from alloy crates rather than defining locally here. Can be follow-on

Cargo.toml
```diff
@@ -41,7 +41,13 @@ members = [
...
   "crates/batcher/*",
   "crates/builder/*",
-  "crates/infra/*",
+  "crates/infra/audit",
+  "crates/infra/basectl",
+  "crates/infra/based",
+  "crates/infra/ingress-rpc",
+  "crates/infra/mempool-rebroadcaster",
+  "crates/infra/websocket-proxy",
+  "crates/infra/load-tests/*",
```
refcell, line 50: Can we revert this? We should leave `crates/infra/*` glob match and then just add a new `crates/infra/load-tests/*` glob for the new crate.

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/uniswap.rs
refcell, line 1: Wonder if uniswap has a rust lib, probably not but would be cool

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/precompile.rs
```diff
@@ -0,0 +1,150 @@
...
+        "identity" => Ok(PrecompileId::Identity),
+        "modexp" => Ok(PrecompileId::ModExp),
+        "bn254_add" | "bn128_add" | "ecadd" => Ok(PrecompileId::Bn254Add),
+        "bn254_mul" | "bn128_mul" | "ecmul" => Ok(PrecompileId::Bn254Mul),
+        "bn254_pairing" | "bn128_pairing" | "ecpairing" => Ok(PrecompileId::Bn254Pairing),
+        "blake2f" | "blake2" => Ok(PrecompileId::Blake2F),
+        "kzg_point_evaluation" | "kzg" => Ok(PrecompileId::KzgPointEvaluation),
+        other => Err(format!("unknown precompile: {other}")),
+    }
+}
```
refcell, line 22: Does revm's `PrecompileId` not have a `from_str` impl?

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/uniswap.rs
kmchicoine: There are a few 3rd party SDKs but nothing official

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/precompile.rs
kmchicoine: It does not! Which surprised me

crates/infra/load-tests/base-load-tests-core/src/workload/payloads/precompile.rs
```diff
@@ -0,0 +1,150 @@
...
+            PrecompileId::Bn254Add => 0x06,
+            PrecompileId::Bn254Mul => 0x07,
+            PrecompileId::Bn254Pairing => 0x08,
+            PrecompileId::Blake2F => 0x09,
+            PrecompileId::KzgPointEvaluation => 0x0a,
+            _ => 0x01,
+        };
+        Address::with_last_byte(byte)
+    })
+}
```
refcell, line 40: We should be able to use revm's [`.precompile()` method](https://github.com/bluealloy/revm/blob/e03804e304e081eec297409a9a012b3237286a23/crates/precompile/src/id.rs#L89) to grab the actual `Precompile` and then we can just `.address` on the `Precompile` to get the address.

crates/infra/load-tests/base-load-tests-core/src/utils/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! Utility types, error handling, and tracing setup.
+
+mod errors;
+pub use base_cli_utils::init_test_tracing as init_tracing;
```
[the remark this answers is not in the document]
refcell: Should rm this and use the proper crate as a dependency at call sites

refcell (approved):
Nothing really blocking, happy to land and improve as we go

PR #1269, opened 2026-03-12
feat(tx-manager): SimpleTxManager Transaction Construction

jackchuma:
### What changed? Why?
Implements the core transaction construction pipeline in `SimpleTxManager`, replacing `todo!()` stubs with a working `prepare()` / `craft_tx()` flow.

**`craft_tx()` pipeline (6 steps):**
1. Query provider for gas price caps (tip cap, base fee) via `suggest_gas_price_caps()`
2. Enforce fee limit ceiling via `FeeCalculator::check_limits()` — detects when enforced minimums (`min_tip_cap`, `min_basefee`) inflate fees beyond `multiplier × suggested`
3. Build `TransactionRequest` manually (no alloy fillers) with all EIP-1559 fields
4. Estimate gas and use `max(estimate, candidate.gas_limit)` as a floor to enforce intrinsic gas
5. Acquire nonce from `NonceManager` via RAII `NonceGuard`
6. Sign, encode to RLP — on success the guard is dropped (consuming the nonce); on failure `guard.rollback()` restores it for reuse

**`prepare()` wraps `craft_tx()` with retry logic:**
- `backon` crate with fixed 2-second delay, up to 30 attempts
- Exits immediately on non-retryable errors (`Unsupported`, `FeeLimitExceeded`, `Sign`, `InvalidConfig`, etc.)
- Only retries transient failures (`Underpriced`, `FeeTooLow`, `Rpc`, etc.)

**Other additions:**
- Constructor validates `chain_id` against the provider at startup to fail fast on misconfiguration
- All RPC calls wrapped with `tokio::time::timeout(config.network_timeout)`
- RPC error classification via string matching on geth error messages with correct ordering (e.g. `"replacement transaction underpriced"` before `"transaction underpriced"`)
- Cancellation safety documented — `prepare()` and `craft_tx()` are **not** cancellation-safe (nonce gap risk if dropped mid-flight)
- `#[non_exhaustive]` on `GasPriceCaps`

### How has it been tested?
All 234 tests pass — 19 integration tests (Anvil), ~100 fee unit tests (rstest + proptest), ~50 error classification tests, 14 nonce manager tests, and 6 doc tests.

**Integration tests cover:** valid EIP-1559 signing, explicit gas limit floors, blob rejection, contract creation, gas price cap suggestion, prepare retry wrapper, close/shutdown behavior, accessor methods, sequential nonce increments, chain ID mismatch rejection, calldata preservation, min tip/basefee enforcement, fee limit ceiling enforcement, non-retryable immediate exit, nonce rollback on sign failure, and `Send`+`Sync` bounds.

Clippy and fmt pass cleanly.

### Notes to reviewers
- Blob transaction support is explicitly rejected (`TxManagerError::Unsupported`) — will be implemented in a follow-up ticket
- `send()` / `send_async()` trait methods remain `todo!()` pending a separate receipt-polling ticket
- Fee bumping infrastructure exists (`FeeCalculator::update_fees`, `calc_threshold_value`) but is not wired into `SimpleTxManager` yet
- Signer sidecar support will be added in a separate PR

### Change management
type=routine
risk=low
impact=sev5

### Backwards Compatibility
backwards_compatible=true

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/utilities/tx-manager/tests/craft_tx.rs
```diff
@@ -0,0 +1,495 @@
...
+        "expected TxManagerError::FeeLimitExceeded, got {err:?}",
+    );
+}
+
+/// Verifies that `prepare()` exits immediately on a non-retryable error
+/// rather than looping through all 30 retry attempts. A 5-second timeout
+/// guard catches regressions where `Unsupported` is accidentally marked
+/// retryable.
+#[tokio::test]
+async fn prepare_exits_immediately_on_non_retryable_error() {
```
leopoldjoy, line 379: Nit: the `prepare()` docs state explicitly: "Callers must not wrap this future in `tokio::select!`, `tokio::time::timeout`, or similar combinators that may drop it mid-flight." Yet this test does exactly that:

```
let result = tokio::time::timeout(Duration::from_secs(5), manager.prepare(&candidate)).await;
```

The test is incidentally safe here because `Unsupported` is returned in the blob check before any nonce is ever acquired. But the test establishes a pattern that looks like approved usage. A future developer copying this for a candidate that does reach nonce acquisition will silently introduce the nonce-gap hazard the docs warn about.

The timeout guard is unnecessary anyway: `prepare()` already exits immediately on non-retryable errors via the `.when(|e| e.is_retryable())` predicate — no timeout is needed to prevent a 60-second hang. The test can assert the behaviour directly:

```
let err = manager.prepare(&candidate).await.expect_err("should reject blob tx");
assert!(matches!(err, TxManagerError::Unsupported(_)));
```

If a regression timer is still desired, it belongs at the test-harness level, not wrapping `prepare()`.

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -1,23 +1,386 @@
...
+        // Step 3: Build TransactionRequest.
+        let from = self.sender_address();
+        let mut tx_request = TransactionRequest::default()
+            .with_input(candidate.tx_data.clone())
+            .with_max_fee_per_gas(caps.gas_fee_cap)
+            .with_max_priority_fee_per_gas(caps.gas_tip_cap)
+            .with_value(candidate.value)
+            .with_chain_id(self.chain_id);
+
+        tx_request.from = Some(from);
```
leopoldjoy, line 319: Nit: Every other field on `tx_request` is set via the consuming builder chain (`.with_input()`, `.with_max_fee_per_gas()`, etc.) or via the `TransactionBuilder` mutation API (`.set_to()`, `.into_create()`). But here this is set by reaching into the struct directly.

`TransactionBuilder` provides `set_from(&mut self, from: Address)` and a `with_from(self, from: Address) -> Self` default. The from assignment should use the trait API for consistency — either folded into the chain as `.with_from(from)` or kept as a post-chain call with `tx_request.set_from(from)`.

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -1,23 +1,386 @@
...
+
+    /// Returns a reference to the nonce manager.
+    pub const fn nonce_manager(&self) -> &NonceManager {
+        &self.nonce_manager
+    }
+
+    /// Returns the chain ID.
+    pub const fn chain_id(&self) -> u64 {
+        self.chain_id
+    }
```
0x00101010, line 141: do all the internal fields need to be exposed as public?

leopoldjoy (approved):
Two nits here, they can be addressed later:

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -1,23 +1,386 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns immediately with [`TxManagerError::ChannelClosed`] if the
+    /// manager is closed. Otherwise returns the first non-retryable error,
+    /// or the last retryable error after exhausting all retry attempts.
+    pub async fn prepare(&self, candidate: &TxCandidate) -> TxManagerResult<Bytes> {
+        if self.is_closed() {
+            return Err(TxManagerError::ChannelClosed);
+        }
```
0x00101010, line 182: nit, prefer removing this for less code duplication

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -1,23 +1,386 @@
...
+            None => tx_request = tx_request.into_create(),
+        }
+
+        // Step 4: Gas estimation.
+        //
+        // Always call estimate_gas to enforce intrinsic gas checks (e.g.
+        // the 21,000 minimum for plain value transfers). When the caller
+        // supplies an explicit gas_limit, it is used as a floor via max()
+        // so the transaction never under-provisions gas.
+        let estimated = tokio::time::timeout(
```
0x00101010, line 332:
```
let client = reqwest::Client::builder()
    .timeout(config.network_timeout)
    .build()?;
let http = Http::with_client(client, endpoint);
let provider = RootProvider::new(RpcClient::new(http, false));
```

We can get rid of tokio::time::timeout all together

PR #1335, opened 2026-03-15
fix(batcher): Batcher Coverage And DA

refcell:
## Summary

This PR addresses several gaps in batcher test coverage and makes a number of correctness improvements to the encoder and derivation pipeline.

On the encoder side, the `BatchType::Span` path was wired up but never actually produced submissions. The fix adds a size-based flush trigger that estimates the compressed size of the accumulated span batch after each block and closes the channel when it reaches the configured size budget, mirroring the approach taken in op-batcher's `SpanChannelOut`. A timeout-based flush path was also added for spans that grow slowly. Two correctness bugs in `close_current_channel` were fixed: previously, a failed `append_singular_batch` would still write a zero-block span batch to the channel, and the blocks were permanently lost because the accumulator was consumed via `std::mem::take` before the loop. Now the accumulator is only cleared after both batch construction and channel write succeed, so blocks are always preserved on any failure path. The size estimation was also made O(1) per step by maintaining a running byte counter instead of re-scanning the entire accumulator on every call.

A bug in the `ChannelBank` derivation stage was fixed where `signal()` was unconditionally clearing the channel map on every signal, including `ProvideBlock`. This caused multi-block channels to be discarded at L1 block boundaries, breaking channel timeout tests. Now only `Reset`, `Activation`, and `FlushChannel` signals clear state.

On the test infrastructure side, the `TestRollupConfigBuilder` was extended with new builder methods and all action tests that were manually constructing or mutating `RollupConfig` values have been updated to use it, removing several duplicated local helper functions. New action tests were added covering span batch end-to-end encoding, blob DA submission, multi-blob packing, calldata DA, DA type switching, DA throttling, large-transaction stress, and sequencer drift scenarios.

### DIFF
--- a/actions/harness/tests/batcher_channels.rs
+++ b/actions/harness/tests/batcher_channels.rs
@@ -52,10 +52,6 @@ use base_action_harness::{
 /// issues, `ChannelDriver` needs a `with_channel_id(id)` builder or random
 /// ID generation.
 #[tokio::test]
-#[ignore = "channel timeout semantics in the IndexedTraversal mode need verification; \
-            the indexed traversal clears the ChannelBank per L1 block, which may mean \
-            multi-block channel timeout is not exercisable in the current harness — \
-            the test skeleton documents the expected flow for when the harness supports it"]
 async fn channel_timeout_triggers_channel_invalidation() {
     use base_batcher_encoder::EncoderConfig;
 
@@ -167,9 +163,10 @@ async fn channel_timeout_triggers_channel_invalidation() {
 /// No new methods needed — uses existing `Batcher::advance()` with a fresh
 /// `ChannelDriver` instance for the recovery submission.
 #[tokio::test]
-#[ignore = "blocked on channel timeout test infrastructure; once the channel bank \
-            correctly prunes timed-out channels in the IndexedTraversal path, \
-            this recovery test can be enabled"]
+#[ignore = "test design issue: batcher.advance() submits all frames in a single L1 \
+            block, so the channel completes immediately and the safe head advances \
+            before the timeout window elapses — needs restructuring to use selective \
+            per-frame submission to make the channel actually expire before recovery"]
 async fn channel_timeout_recovery_resubmits_successfully() {
     let batcher_cfg = BatcherConfig::default();
     let rollup_cfg =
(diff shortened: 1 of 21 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/batcher/core/src/driver.rs
```diff
@@ -167,41 +168,61 @@ where
...
+                            self.pipeline.requeue(id);
+                            drop(permit);
+                            continue;
+                        }
+                    },
+                    DaType::Calldata => {
+                        // Calldata mode: one frame per submission (enforced by
+                        // EncoderConfig::validate). Encode as
+                        // [DERIVATION_VERSION_0] ++ frame.encode().
+                        let frame = &sub.frames[0];
```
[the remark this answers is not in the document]
refcell: user error and everywhere calls validate()

actions/harness/tests/batcher_channels.rs
```diff
@@ -167,9 +163,10 @@ async fn channel_timeout_triggers_channel_invalidation() {
 /// No new methods needed — uses existing `Batcher::advance()` with a fresh
 /// `ChannelDriver` instance for the recovery submission.
 #[tokio::test]
-#[ignore = "blocked on channel timeout test infrastructure; once the channel bank \
-            correctly prunes timed-out channels in the IndexedTraversal path, \
-            this recovery test can be enabled"]
+#[ignore = "test design issue: batcher.advance() submits all frames in a single L1 \
+            block, so the channel completes immediately and the safe head advances \
+            before the timeout window elapses — needs restructuring to use selective \
+            per-frame submission to make the channel actually expire before recovery"]
```
refcell, line 169: Will address this in a follow-on pr

danyalprout (approved):

actions/harness/tests/batcher_throttling.rs
```diff
@@ -120,3 +120,53 @@ fn test_throttle_step_strategy_full_intensity() {
...
+
+/// Verify that `da_backlog_bytes()` returns 0 after encoding (all blocks
+/// consumed by the pipeline) and that [`ThrottleController`] correctly
+/// signals throttling for a simulated high backlog then clears when the
+/// backlog reaches zero.
+#[test]
+fn test_throttle_batcher_integration() {
+    let l1_cfg = L1MinerConfig { block_time: 2 };
+    let batcher_cfg = BatcherConfig::default();
+    let rollup_cfg = {
```
jackchuma, line 132: Nit: could clean this up with `let rollup_cfg = TestRollupConfigBuilder::base_mainnet(&batcher_cfg).build();`

refcell: Good catch, will follow up with this ty

PR #1342, opened 2026-03-15
chore(actions): auto-register block hashes in L2Sequencer

### LINKED ISSUE #1313: chore(actions): Auto-Register Block Hashes
## Summary

Every async derivation test in the harness must manually call `verifier.register_block_hash(number, hash)` for each L2 block after building it with `L2Sequencer`. This is necessary because `L2Verifier` maintains an internal `HashMap<u64, B256>` for parent-hash chain consistency, and tests that omit a `register_block_hash` call will fail with a confusing error rather than a clear assertion failure. The boilerplate is present in `tests/derivation.rs`, `tests/finalization.rs`, `tests/unsafe_gossip.rs`, and both skeleton test files.

This pattern is a contributor footgun. A new test author writing a custom derivation test will not know to call `register_block_hash` and will get a cryptic failure. The registration should be automatic.

The fix is to have `L2Sequencer::build_next_block` and `L2Sequencer::build_empty_block` at [`actions/harness/src/l2.rs`](https://github.com/base/base/blob/main/actions/harness/src/l2.rs) push each built block hash into a shared registry that `L2Verifier` can read directly, rather than requiring the test to pass hashes manually. Alternatively, `ActionTestHarness` could return a `SharedBlockHashRegistry` from `create_l2_sequencer` and accept it in `create_verifier`, wiring the two together at construction time. Either approach eliminates the per-block `register_block_hash` calls from all test files.

All calls to `verifier.register_block_hash` across the test suite should be removed as part of this change.

vikions:
Closes #1313

## Summary

Introduces `SharedBlockHashRegistry` and wires it between `L2Sequencer` and
`L2Verifier` so that block hashes are registered automatically as blocks are
built, eliminating the per-block `verifier.register_block_hash(number, hash)`
boilerplate that was required in every derivation test.

## Changes

- **`actions/harness/src/verifier.rs`** — adds `SharedBlockHashRegistry`
  (`Arc<Mutex<HashMap<u64, B256>>>`) and a `with_block_hash_registry()` builder
  method on `L2Verifier`; `register_block_hash` is kept as a manual override API
- **`actions/harness/src/l2.rs`** — `L2Sequencer` stores a `SharedBlockHashRegistry`
  and auto-inserts `(number, hash)` in `build_next_block` and `build_empty_block`;
  adds `with_block_hash_registry()` builder method
- **`actions/harness/src/harness.rs`** — `ActionTestHarness` creates one shared
  registry at construction time and passes it to both `create_l2_sequencer` and
  `create_verifier`
- **`actions/harness/src/lib.rs`** — re-exports `SharedBlockHashRegistry`
- **All test files** — removed all manual `verifier.register_block_hash(...)` calls

### DIFF
--- a/actions/harness/src/harness.rs
+++ b/actions/harness/src/harness.rs
@@ -7,7 +7,7 @@ use base_protocol::{BlockInfo, L2BlockInfo};
 use crate::{
     ActionBlobDataSource, ActionDataSource, ActionL1ChainProvider, ActionL2ChainProvider, Batcher,
     BatcherConfig, BlobVerifierPipeline, L1Miner, L1MinerConfig, L2BlockProvider, L2Sequencer,
-    L2Verifier, SharedL1Chain, VerifierPipeline, block_info_from,
+    L2Verifier, SharedBlockHashRegistry, SharedL1Chain, VerifierPipeline, block_info_from,
 };
 
 /// Top-level test harness that owns all actors for a single action test.
@@ -37,12 +37,18 @@ pub struct ActionTestHarness {
     pub l1: L1Miner,
     /// The rollup configuration shared by all actors.
     pub rollup_config: RollupConfig,
+    /// Shared L2 block hashes written by sequencers and read by verifiers.
+    block_hashes: SharedBlockHashRegistry,
 }
 
 impl ActionTestHarness {
     /// Create a harness with the given configurations.
     pub fn new(l1_config: L1MinerConfig, rollup_config: RollupConfig) -> Self {
-        Self { l1: L1Miner::new(l1_config), rollup_config }
+        Self {
+            l1: L1Miner::new(l1_config),
+            rollup_config,
+            block_hashes: SharedBlockHashRegistry::default(),
+        }
     }
 
     /// Mine `n` L1 blocks and return the latest block number after mining.
@@ -122,6 +128,7 @@ impl ActionTestHarness {
         let system_config = self.rollup_config.genesis.system_config.unwrap_or_default();
 
         L2Sequencer::new(genesis_head, l1_chain, self.rollup_config.clone(), system_config)
+            .with_block_hash_registry(self.block_hashes.clone())
     }
 
     /// Create an [`L2Verifier`] wired to the harness's L1 chain.
@@ -176,7 +183,8 @@ impl ActionTestHarness {
             l2_provider,
             safe_head,
             genesis_l1,
-        );
+        )
+        .with_block_hash_registry(self.block_hashes.clone());
 
         (verifier, chain)
     }
@@ -218,7 +226,8 @@ impl ActionTestHarness {
             ActionL2ChainProvider::from_genesis(&self.rollup_config),
             safe_head,
             genesis_l1,
-        );
+        )
+        .with_block_hash_registry(self.block_hashes.clone());
 
         (verifier, chain)
     }
--- a/actions/harness/src/harness.rs
+++ b/actions/harness/src/harness.rs
@@ -7,7 +7,7 @@ use base_protocol::{BlockInfo, L2BlockInfo};
 use crate::{
     ActionBlobDataSource, ActionDataSource, ActionL1ChainProvider, ActionL2ChainProvider, Batcher,
     BatcherConfig, BlobVerifierPipeline, L1Miner, L1MinerConfig, L2BlockProvider, L2Sequencer,
-    L2Verifier, SharedBlockHashRegistry, SharedL1Chain, VerifierPipeline, block_info_from,
+    L2Verifier, SharedL1Chain, VerifierPipeline, block_info_from,
 };
 
 /// Top-level test harness that owns all actors for a single action test.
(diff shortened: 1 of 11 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (commented):
Thanks for tackling this. The goal is right and the test cleanup is clean, but there are a few design issues worth discussing.

**Centralized registry on the harness is the wrong unit of sharing**

The core concern is that `ActionTestHarness` owns one `SharedBlockHashRegistry` and passes it to every `create_l2_sequencer` and `create_verifier` call. This means all sequencers and all verifiers created from the same harness instance share one map. Any future test that creates two sequencers — for example to simulate a reorg with conflicting forks, or to test a backup unsafe head scenario — will have both sequencers writing into the same registry. Block 1 from sequencer A will silently overwrite block 1 from sequencer B, and the verifier will use whichever hash was written last. This is the kind of hidden global state that produces hard-to-diagnose test failures later.

The registry belongs on the sequencer, not the harness. Each sequencer creates its own and the caller explicitly wires a specific sequencer to a specific verifier. That way the coupling between actors is visible at the test callsite. Something like `h.create_verifier_from_sequencer(&sequencer)` makes the relationship explicit and composes cleanly when there are multiple sequencers.

**Type lives in the wrong module**

`SharedBlockHashRegistry` is defined in `verifier.rs`, but it is primarily a sequencer-side concept — the sequencer writes to it, the verifier just reads. Defining it in the verifier module and having `L2Sequencer` import it from `crate::` is inverted. It should live in `l2.rs` alongside `L2Sequencer`.

**Builder step on the sequencer is awkward**

`L2Sequencer::with_block_hash_registry()` allocates a default `HashMap` in `new()` and then immediately replaces it. More importantly, any sequencer created directly in a test (not via the harness) silently gets a dead registry that no verifier will ever read. If the registry is the sequencer's own state from the start, this problem goes away.

refcell (changes requested):

vikions:
> Thanks for tackling this. The goal is right and the test cleanup is clean, but there are a few design issues worth discussing.
> 
> **Centralized registry on the harness is the wrong unit of sharing**
> 
> The core concern is that `ActionTestHarness` owns one `SharedBlockHashRegistry` and passes it to every `create_l2_sequencer` and `create_verifier` call. This means all sequencers and all verifiers created from the same harness instance share one map. Any future test that creates two sequencers — for example to simulate a reorg with conflicting forks, or to test a backup unsafe head scenario — will have both sequencers writing into the same registry. Block 1 from sequencer A will silently overwrite block 1 from sequencer B, and the verifier will use whichever hash was written last. This is the kind of hidden global state that produces hard-to-diagnose test failures later.
> 
> The registry belongs on the sequencer, not the harness. Each sequencer creates its own and the caller explicitly wires a specific sequencer to a specific verifier. That way the coupling between actors is visible at the test callsite. Something like `h.create_verifier_from_sequencer(&sequencer)` makes the relationship explicit and composes cleanly when there are multiple sequencers.
> 
> **Type lives in the wrong module**
> 
> `SharedBlockHashRegistry` is defined in `verifier.rs`, but it is primarily a sequencer-side concept — the sequencer writes to it, the verifier just reads. Defining it in the verifier module and having `L2Sequencer` import it from `crate::` is inverted. It should live in `l2.rs` alongside `L2Sequencer`.
> 
> **Builder step on the sequencer is awkward**
> 
> `L2Sequencer::with_block_hash_registry()` allocates a default `HashMap` in `new()` and then immediately replaces it. More importantly, any sequencer created directly in a test (not via the harness) silently gets a dead registry that no verifier will ever read. If the registry is the sequencer's own state from the start, this problem goes away.

Thanks for the detailed feedback! Addressed everything:

- `SharedBlockHashRegistry` moved to `l2.rs`
- Removed the centralized registry from `ActionTestHarness`, added `create_verifier_from_sequencer()` so the wiring is explicit at each call site
- `L2Sequencer` now owns its registry from construction, `with_block_hash_registry()` removed
- Restored `register_block_hash` in the two ignored timeout tests

refcell (approved):
This is awesome, thanks for working on this

PR #1353, opened 2026-03-15
fix(engine): capture apply_update return value in test state setters

kinderjoypresents:
## Problem

`EngineSyncState` derives `Copy`, and `EngineSyncState::apply_update` takes
`self` by value (`fn apply_update(self, ...) -> Self`). When called as
`self.sync_state.apply_update(...)` without assigning the result back, Rust
copies the field, runs the method on the copy, and discards the returned
`EngineSyncState`. The original `self.sync_state` is never mutated.

All five test-only state setters on `EngineState` had this defect:

- `set_unsafe_head`
- `set_cross_unsafe_head`
- `set_local_safe_head`
- `set_safe_head`
- `set_finalized_head`

The existing `test_chain_label_metrics` test did **not** catch this because
it only asserts on Prometheus gauge values. Those gauges are updated as a
side-effect *inside* `apply_update` before the new state is returned, so the
metrics fire correctly even though the `sync_state` field is never written.
Any test that reads sync state after calling one of these setters would
receive the zero default value.

## Solution

Assign the return value of `apply_update` back to `self.sync_state` in all
five setters, matching the pattern used everywhere else in the codebase
(`consolidate/task.rs`, `synchronize/task.rs`, `test_utils/engine_state.rs`):

```rust
self.sync_state = self.sync_state.apply_update(EngineSyncStateUpdate { ... });
```

Add `test_set_state_helpers_actually_update_state` which directly asserts
that each setter persists the updated block number into `sync_state`. This
test requires no feature flags and would have caught the regression without
`--features metrics`.

Also lift `use base_protocol::BlockInfo` to the top of the test module
(removing its `#[cfg(feature = "metrics")]` guard) since it is now used by
both the metrics test and the new unconditional test.

## Impact

Low blast-radius: the affected methods are `#[cfg(test)]`-only helpers, so
no production logic is changed. However, the fix ensures these setters
actually work as documented, preventing silent test failures in any future
test that sets engine state via these helpers and then asserts on `sync_state`
fields.
`

### DIFF
--- a/crates/consensus/engine/src/state/core.rs
+++ b/crates/consensus/engine/src/state/core.rs
@@ -78,45 +78,40 @@ impl EngineSyncState {
         }
     }
 
-    /// Applies the update to the provided sync state, using the current state values if the update
-    /// is not specified. Returns the new sync state.
-    pub fn apply_update(self, sync_state_update: EngineSyncStateUpdate) -> Self {
+    /// Applies the update to the sync state in place, keeping current field values for any
+    /// fields not specified in the update.
+    pub fn apply_update(&mut self, sync_state_update: EngineSyncStateUpdate) {
         if let Some(unsafe_head) = sync_state_update.unsafe_head {
             Self::update_block_label_metric(
                 Metrics::UNSAFE_BLOCK_LABEL,
                 unsafe_head.block_info.number,
             );
+            self.unsafe_head = unsafe_head;
         }
         if let Some(cross_unsafe_head) = sync_state_update.cross_unsafe_head {
             Self::update_block_label_metric(
                 Metrics::CROSS_UNSAFE_BLOCK_LABEL,
                 cross_unsafe_head.block_info.number,
             );
+            self.cross_unsafe_head = cross_unsafe_head;
         }
         if let Some(local_safe_head) = sync_state_update.local_safe_head {
             Self::update_block_label_metric(
                 Metrics::LOCAL_SAFE_BLOCK_LABEL,
                 local_safe_head.block_info.number,
             );
+            self.local_safe_head = local_safe_head;
         }
         if let Some(safe_head) = sync_state_update.safe_head {
             Self::update_block_label_metric(Metrics::SAFE_BLOCK_LABEL, safe_head.block_info.number);
+            self.safe_head = safe_head;
         }
         if let Some(finalized_head) = sync_state_update.finalized_head {
             Self::update_block_label_metric(
                 Metrics::FINALIZED_BLOCK_LABEL,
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/consensus/engine/src/state/core.rs
```diff
@@ -194,39 +193,39 @@ mod test {
     impl EngineState {
         /// Set the unsafe head.
         pub fn set_unsafe_head(&mut self, unsafe_head: L2BlockInfo) {
-            self.sync_state.apply_update(EngineSyncStateUpdate {
+            self.sync_state = self.sync_state.apply_update(EngineSyncStateUpdate {
```
refcell: I think the right fix here is to actually fix the `apply_update` method to mutate state rather than change all these call sites.

refcell (changes requested):

kinderjoypresents:
Refactored `apply_update` to take `&mut self` instead of returning a new `Self`, eliminating the Copy-type footgun at the source rather than patching individual call sites.

Updated all call sites accordingly:
- `consolidate/task.rs`, `test_utils/engine_state.rs` — dropped the redundant lhs assignment
- `synchronize/task.rs` — saves `old_sync_state` before mutation and restores it on any error path, preserving the invariant that state is committed only after a successful FCU
- `build/task.rs` — copies `sync_state` to a local temp (Copy type) and applies the update there, since the build task must not mutate engine state

PR #1356, opened 2026-03-15
fix(protocol): correct estimated_size accounting when pruning future 

jrootcache:
## Summary

Fixes a logical bug in `Channel::add_frame` where `estimated_size` was incorrectly decremented for **retained** frames during the out-of-order closing-frame prune path.

## Problem

When a closing frame (`is_last = true`) arrives with a frame number **lower** than the highest previously ingested frame number, `add_frame` prunes the out-of-range frames via `HashMap::retain`. The closure unconditionally subtracted `frame.size()` from `estimated_size` for **every** iterated entry — including frames that were being **kept** — and only then returned the boolean that decided whether the entry was removed:

```rust
// BEFORE (buggy)
self.inputs.retain(|id, frame| {
    self.estimated_size -= frame.size(); // runs for kept frames too
    *id < self.last_frame_number
});
```

This leaves `estimated_size` lower than the actual bytes held in `self.inputs`.

## Impact

`estimated_size` is the value returned by `Channel::size()`, which `ChannelBank::prune` accumulates across all channels to decide whether the bank exceeds `MAX_CHANNEL_BANK_SIZE` / `FJORD_MAX_CHANNEL_BANK_SIZE`. An underestimated size means the channel bank can **silently exceed its memory limit** when channels with out-of-order frames are present. Under adversarial or unusual L1 data ordering this could lead to unbounded memory growth in the derivation pipeline.

## Fix

Gate the `estimated_size` decrement on the retain predicate — only frames that are actually **removed** (`id >= last_frame_number`) should reduce the counter:

```rust
// AFTER (fixed)
self.inputs.retain(|id, frame| {
    if *id < self.last_frame_number {
        true
    } else {
        self.estimated_size -= frame.size();
        false
    }
});
```

## Test

Added a new `FrameValidityTestCase` — `"prune future frames on out-of-order close"` — that exercises this exact code path:

1. Ingest frame 2 (non-last) → `estimated_size = 203`
2. Ingest frame 0 (non-last) → `estimated_size = 407`
3. Ingest frame 1 (`is_last = true`) → triggers prune of frame 2, retains frame 0 → `estimated_size = 407`

With the bug the final size would be `203` (frame 0's contribution is erroneously zeroed out). The test asserts `407` and verifies `frame_data` returns `b"zeroone"` (frames 0 + 1 concatenated).

## Files changed

- `crates/consensus/protocol/src/channel.rs`
`

### DIFF
--- a/crates/consensus/protocol/src/channel.rs
+++ b/crates/consensus/protocol/src/channel.rs
@@ -121,8 +121,12 @@ impl Channel {
             // closing frame.
             if self.last_frame_number < self.highest_frame_number {
                 self.inputs.retain(|id, frame| {
-                    self.estimated_size -= frame.size();
-                    *id < self.last_frame_number
+                    if *id < self.last_frame_number {
+                        true
+                    } else {
+                        self.estimated_size -= frame.size();
+                        false
+                    }
                 });
                 self.highest_frame_number = self.last_frame_number;
             }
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (approved):
Very good catch.

PR #1265, opened 2026-03-12
feat(txpool): scaffold minimal eth_sendBundle support

haardikk21:
## Summary

Adds minimal bundle transaction support to the mempool propagation pipeline in `crates/txpool/`.

- **Extended `BasePooledTransaction`** with optional `target_block_number`, `min_timestamp`, and `max_timestamp` fields (all timestamps in milliseconds)
- **New `BundleTransaction` trait** with `is_bundle_expired()` and `is_bundle_not_yet_valid()` for checking bundle constraints against block number/timestamp
- **New `eth_sendBundle` RPC endpoint** (`bundle/rpc.rs`) with full parameter validation:
  - Single transaction only
  - Block number ≤ 30 blocks ahead, not in the past
  - Timestamps ≤ 60s ahead, not in the past, min ≤ max
  - Reverting tx hashes, replacement UUID, builders must be empty
  - Runtime-gated via `enabled` flag
- **New `maintain_bundle_transactions` eviction task** (`bundle/maintain.rs`) that listens to canonical state notifications and removes expired bundle txs from the pool on each new block — reusable on both client and builder nodes
- **Extended `ValidatedTransaction` wire format** with optional bundle metadata fields (backward compatible via `serde(skip_serializing_if, default)`)
- **Updated forwarder** to populate bundle metadata when forwarding to builders
- **Updated builder RPC** to read bundle metadata from wire format and apply to pool transactions

### Constants
- `BLOCK_TIME_SECS = 2`
- `MAX_BUNDLE_ADVANCE_SECS = 60` (→ 30 blocks, 60_000ms)

### Validation layers
| Layer | Checks |
|-------|--------|
| RPC submission (client) | Structural + time window (≤ 60s ahead, not past) |
| Pool maintenance (client + builder) | Evict expired `max_timestamp` / `target_block_number` |
| Block inclusion (builder) | Out of scope — trait methods provided for future use |

### Out of scope
- Builder block-inclusion checks (will use `is_bundle_not_yet_valid`)
- Node extension wiring (spawning the RPC + maintenance task)

PR #1423, opened 2026-03-17
feat(bundle): wire bundle validation, RPC, and maintenance into node

haardikk21:
## Summary

Stacked on #1265. Wires the bundle transaction infrastructure into the builder and client nodes:

- **Builder validation**: Skip bundle txs during block building when target_block_number doesn't match, bundle is expired (max_timestamp), or not yet valid (min_timestamp). Not-yet-valid bundles are deferred without permanent invalidation.
- **Bundle extension**: New `base-bundle-extension` crate that registers `eth_sendBundle` RPC (always enabled, no config flags) and spawns the `maintain_bundle_transactions` background task.
- **Block number sync**: The maintenance task now updates the shared `current_block_number` AtomicU64 on each canonical block, keeping RPC validation accurate.

## Test Plan

**Builder integration tests** (5 tests in `crates/builder/core/tests/bundles.rs`):
- Bundle tx targeting current block is included
- Bundle tx targeting wrong block is excluded  
- Expired bundle tx is excluded
- Not-yet-valid bundle tx is deferred (stays in pool)
- Normal tx is unaffected by bundle checks

**Devnet e2e tests** (4 tests in `devnet/tests/bundles.rs`):
- Valid bundle accepted and included in target block
- Invalid bundles rejected (empty txs, block too far ahead)
- Bundle targets specific block number only
- Expired bundles cleaned up by maintenance task

`just check` passes (fmt, udeps, clippy, 3150 tests).

### DIFF
--- a/crates/builder/core/src/flashblocks/context.rs
+++ b/crates/builder/core/src/flashblocks/context.rs
@@ -484,7 +484,6 @@ impl OpPayloadBuilderCtx {
         let block_timestamp = self.attributes().timestamp();
 
         while let Some(tx) = best_txs.next(()) {
-
             if let Some(target) = tx.target_block_number()
                 && target != block_number
             {
--- a/crates/builder/core/src/flashblocks/context.rs
+++ b/crates/builder/core/src/flashblocks/context.rs
@@ -516,8 +516,9 @@ impl OpPayloadBuilderCtx {
                     tx_hash = ?tx.hash(),
                     block = block_number,
                     timestamp = block_timestamp,
-                    "deferring bundle tx: not yet valid"
+                    "skipping bundle tx: not yet valid"
                 );
+                best_txs.mark_invalid(tx.sender(), tx.nonce());
                 continue;
             }
 
--- a/crates/builder/core/src/flashblocks/context.rs
+++ b/crates/builder/core/src/flashblocks/context.rs
@@ -16,7 +16,7 @@ use base_execution_payload_builder::error::OpPayloadBuilderError;
 use base_execution_primitives::{OpReceipt, OpTransactionSigned};
 use base_node_core::OpPayloadBuilderAttributes;
 use base_revm::{L1BlockInfo, OpSpecId};
-use base_txpool::estimated_da_size::DataAvailabilitySized;
+use base_txpool::{BundleTransaction, estimated_da_size::DataAvailabilitySized};
 use reth_basic_payload_builder::PayloadConfig;
 use reth_chainspec::{EthChainSpec, EthereumHardforks};
 use reth_evm::{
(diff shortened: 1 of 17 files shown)


crates/builder/core/src/flashblocks/context.rs
```diff
@@ -480,7 +480,48 @@ impl OpPayloadBuilderCtx {
...
+
+            if tx.is_bundle_not_yet_valid(block_timestamp) {
+                trace!(
+                    target: "payload_builder",
+                    tx_hash = ?tx.hash(),
+                    block = block_number,
+                    timestamp = block_timestamp,
+                    "skipping bundle tx: not yet valid"
+                );
+                best_txs.mark_invalid(tx.sender(), tx.nonce());
```
meyer9, line 521: I think this will cause all txs from the sender to not be yielded. Is there a case where a bundle is submitted with a nonce gap and we still want to consider lower nonce transactions? As it stands, it seems like anyone with a bundle that isn't valid yet will not be able to submit other transactions.

I may be misunderstanding something with how bundles work though.

meyer9 (commented):
QQ about whether we should allow lower none tx before bundle is included

crates/builder/core/tests/bundles.rs
wlawt, line 1: can we add tests for:

- the same sender sending multiple bundles (i.e., 2) targeting the same block, where the bundles are valid
- the same sender sends a valid/invalid bundle for the same block

crates/builder/core/src/flashblocks/context.rs
haardikk21: `mark_invalid` will mark this particular txn (due to the nonce) and any descendants (higher nonce) to be invalid, which is fair. lower nonce txns will not be affected, and anyway should have been yielded earlier by the best txs iterator rather than later

crates/builder/core/tests/bundles.rs
haardikk21: added

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

meyer9 (approved):

PR #1431, opened 2026-03-17
fix(proposer): improve proposer retry handling

jackchuma:
### What changed? Why?

Fixes the proposer retry handling so that RPC errors are correctly classified for retry decisions.

Previously, raw `TransportError`s were immediately wrapped in domain-specific `RpcError` variants (e.g. `WitnessNotFound`, `InvalidResponse`) *inside* the retry closure. This meant `is_retryable()` was checked against the already-wrapped variant, which could prevent retries from firing on transient transport errors.

Now:
- Inside retry closures, errors convert via `RpcError::from(TransportError)`, preserving the original error category (`Transport`, `InvalidResponse`, etc.) so `is_retryable()` works correctly.
- Domain-specific wrapping (`WitnessNotFound`, `InvalidResponse` with context) is applied *after* retries are exhausted.
- Error message truncation (to avoid multi-MB witness JSON in logs) is centralized in a `truncate_error_msg` helper called from the `From<TransportError>` impl, rather than being duplicated at individual call sites.

### Notes to reviewers

The `L2Client`, `RethL2Client` `execution_witness` and `db_get` methods all follow the same updated pattern.

### How has it been tested?

Existing tests + manual verification of retry behavior with transient RPC failures.

### Change management ([definitions](http://go/change-management))

type=routine
risk=low
impact=sev5

### Backwards Compatibility ([learn more](http://go/backwards-compatibility))

backwards_compatible=true

automerge=false

### DIFF
--- a/crates/proof/rpc/src/error.rs
+++ b/crates/proof/rpc/src/error.rs
@@ -58,13 +58,33 @@ impl RpcError {
     }
 }
 
+/// Maximum length for error messages stored in [`RpcError`] variants.
+///
+/// RPC error payloads (e.g. from `debug_executionWitness`) can contain multi-MB
+/// JSON responses. Truncating at the conversion boundary prevents large strings
+/// from propagating through retry loops, log lines, and error chains.
+const MAX_ERROR_MSG_LEN: usize = 500;
+
+/// Truncates a string to [`MAX_ERROR_MSG_LEN`], appending a `(truncated)` suffix
+/// if it exceeds the limit.
+fn truncate_error_msg(msg: String) -> String {
+    if msg.len() > MAX_ERROR_MSG_LEN {
+        let end = msg.floor_char_boundary(MAX_ERROR_MSG_LEN);
+        format!("{}... (truncated)", &msg[..end])
+    } else {
+        msg
+    }
+}
+
 impl From<TransportError> for RpcError {
     fn from(err: TransportError) -> Self {
         match err {
             AlloyRpcError::SerError(e) => Self::Serialization(e.to_string()),
             AlloyRpcError::DeserError { err, .. } => Self::InvalidResponse(err.to_string()),
             AlloyRpcError::NullResp => Self::InvalidResponse("null response".into()),
-            AlloyRpcError::ErrorResp(payload) => Self::InvalidResponse(payload.to_string()),
+            AlloyRpcError::ErrorResp(payload) => {
+                Self::InvalidResponse(truncate_error_msg(payload.to_string()))
+            }
             err @ (AlloyRpcError::Transport(_)
             | AlloyRpcError::UnsupportedFeature(_)
             | AlloyRpcError::LocalUsageError(_)) => Self::Transport(err.to_string()),
--- a/crates/proof/rpc/src/error.rs
+++ b/crates/proof/rpc/src/error.rs
@@ -48,6 +48,24 @@ pub enum RpcError {
 }
 
 impl RpcError {
+    /// Maximum length for error messages stored in [`RpcError`] variants.
+    ///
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/rpc/src/error.rs
```diff
@@ -58,13 +58,33 @@ impl RpcError {
...
+/// Truncates a string to [`MAX_ERROR_MSG_LEN`], appending a `(truncated)` suffix
+/// if it exceeds the limit.
+fn truncate_error_msg(msg: String) -> String {
+    if msg.len() > MAX_ERROR_MSG_LEN {
+        let end = msg.floor_char_boundary(MAX_ERROR_MSG_LEN);
+        format!("{}... (truncated)", &msg[..end])
+    } else {
+        msg
+    }
+}
```
refcell: These could be moved inside the `impl RpcError`

e.g.
```rust
impl RpcError {
  /// Maximum length for error messages stored in [`RpcError`] variants.
  ///
  /// RPC error payloads (e.g. from `debug_executionWitness`) can contain multi-MB
  /// JSON responses. Truncating at the conversion boundary prevents large strings
  /// from propagating through retry loops, log lines, and error chains.
  pub const MAX_ERROR_MSG_LEN: usize = 500;
  
  /// Truncates a string to [`Self::MAX_ERROR_MSG_LEN`], appending a `(truncated)` suffix
  /// if it exceeds the limit.
  pub fn truncate_msg(msg: String) -> String {
      if msg.len() > Self::MAX_ERROR_MSG_LEN {
          let end = msg.floor_char_boundary(Self::MAX_ERROR_MSG_LEN);
          format!("{}... (truncated)", &msg[..end])
      } else {
          msg
      }
  }
}
```

So later on they can be used like:
```rust
impl From<TransportError> for RpcError {
    fn from(err: TransportError) -> Self {
        match err {
            AlloyRpcError::SerError(e) => Self::Serialization(e.to_string()),
            AlloyRpcError::DeserError { err, .. } => Self::InvalidResponse(err.to_string()),
            AlloyRpcError::NullResp => Self::InvalidResponse("null response".into()),
            AlloyRpcError::ErrorResp(payload) => {
                Self::InvalidResponse(Self::truncate_msg(payload.to_string()))
            }
```

refcell: This just makes methods and constants scoped a bit better while allowing them to be accessed publicly without having to export them from `lib.rs`.

refcell (changes requested):
Small nit

refcell (approved):

PR #1453, opened 2026-03-18
feat(CHAIN-3455): Registration driver with Boundless / Direct proving and full binary wiring

leopoldjoy:
## Summary                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                    
- Adds `RegistrationDriver` to the registrar library crate — core orchestration loop that discovers prover instances, generates ZK proofs for unregistered signers, and submits `registerSigner` transactions to L1 via `SimpleTxManager`.
- Adds `--proving-mode boundless|direct` CLI flag with conditional args (`--image-id` for Boundless, `--elf-path` for Direct), `--l1-chain-id`, `--prover-timeout-secs`
- Replaces the `TODO(CHAIN-3455)` in the binary with full component construction: AWS SDK clients, proof provider, tx manager, driver loop with ctrl-c shutdown
- Extends `RegistrarConfig` with `ProvingConfig` enum, `l1_chain_id`, `prover_timeout`
- Transaction signing uses `base-tx-manager`'s `SimpleTxManager` (CHAIN-3454 cancelled)
- Follows the proposer driver's `tokio::select!` + `CancellationToken` pattern

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo clippy --tests -- -D warnings` passes
- [x] Registrar lib: 29 tests pass (all pre-existing, unchanged)
- [x] Binary CLI: 21 tests pass (rstest parametrized, shared constants, parse_image_id coverage)
- [ ] Integration test with Anvil L1 fork (deferred — requires mock prover instances)
- [ ] Deregistration logic (CHAIN-3456)
- [ ] Health checks and metrics (CHAIN-3457)

### DIFF
--- a/bin/prover-registrar/Cargo.toml
+++ b/bin/prover-registrar/Cargo.toml
@@ -36,6 +36,7 @@ hex.workspace = true
 url.workspace = true
 eyre.workspace = true
 tracing.workspace = true
+tracing-subscriber.workspace = true
 tokio-util.workspace = true
 tokio = { workspace = true, features = ["full"] }
 clap = { workspace = true, features = ["derive", "env"] }
--- a/bin/prover-registrar/Cargo.toml
+++ b/bin/prover-registrar/Cargo.toml
@@ -20,7 +20,7 @@ workspace = true
 base-cli-utils.workspace = true
 base-tx-manager.workspace = true
 base-proof-tee-registrar.workspace = true
-base-proof-tee-nitro-attestation-prover.workspace = true
+base-proof-tee-nitro-attestation-prover = { workspace = true, features = ["prove"] }
 
 # Alloy
 alloy-provider.workspace = true
--- a/bin/prover-registrar/Cargo.toml
+++ b/bin/prover-registrar/Cargo.toml
@@ -16,10 +16,29 @@ path = "src/main.rs"
 workspace = true
 
 [dependencies]
-url.workspace = true
-eyre.workspace = true
+# Internal
+base-tx-manager.workspace = true
+base-proof-tee-registrar.workspace = true
+base-proof-tee-nitro-attestation-prover.workspace = true
+
+# Alloy
 alloy-primitives.workspace = true
 alloy-signer-local.workspace = true
-base-proof-tee-registrar.workspace = true
+alloy-provider.workspace = true
+
+# AWS
+aws-config = { workspace = true, features = ["default-https-client", "rt-tokio"] }
+aws-sdk-ec2.workspace = true
+aws-sdk-elasticloadbalancingv2.workspace = true
+
+# CLI / runtime
+hex.workspace = true
+url.workspace = true
+eyre.workspace = true
+tracing.workspace = true
+tokio-util.workspace = true
 tokio = { workspace = true, features = ["full"] }
 clap = { workspace = true, features = ["derive", "env"] }
+
+[dev-dependencies]
+rstest.workspace = true
--- a/bin/prover-registrar/Cargo.toml
+++ b/bin/prover-registrar/Cargo.toml
@@ -17,28 +17,29 @@ workspace = true
 
 [dependencies]
 # Internal
+base-cli-utils.workspace = true
 base-tx-manager.workspace = true
 base-proof-tee-registrar.workspace = true
 base-proof-tee-nitro-attestation-prover.workspace = true
 
 # Alloy
+alloy-provider.workspace = true
 alloy-primitives.workspace = true
 alloy-signer-local.workspace = true
-alloy-provider.workspace = true
 
 # AWS
-aws-config = { workspace = true, features = ["default-https-client", "rt-tokio"] }
 aws-sdk-ec2.workspace = true
 aws-sdk-elasticloadbalancingv2.workspace = true
+aws-config = { workspace = true, features = ["default-https-client", "rt-tokio"] }
 
 # CLI / runtime
 hex.workspace = true
 url.workspace = true
 eyre.workspace = true
-humantime.workspace = true
 tracing.workspace = true
-tracing-subscriber.workspace = true
+humantime.workspace = true
 tokio-util.workspace = true
+tracing-subscriber.workspace = true
 tokio = { workspace = true, features = ["full"] }
 clap = { workspace = true, features = ["derive", "env"] }
 
--- a/bin/prover-registrar/Cargo.toml
+++ b/bin/prover-registrar/Cargo.toml
@@ -35,6 +35,7 @@ aws-sdk-elasticloadbalancingv2.workspace = true
 hex.workspace = true
 url.workspace = true
 eyre.workspace = true
+humantime.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
 tokio-util.workspace = true
--- a/bin/prover-registrar/src/cli.rs
+++ b/bin/prover-registrar/src/cli.rs
@@ -139,10 +139,6 @@ struct BoundlessArgs {
     )]
     boundless_verifier_program_url: Option<Url>,
 
-    /// Minimum price in wei per cycle for Boundless proof requests.
-    #[arg(long, env = "REGISTRAR_BOUNDLESS_MIN_PRICE", default_value_t = 100_000)]
-    boundless_min_price: u64,
-
     /// Maximum price in wei per cycle for Boundless proof requests.
     #[arg(long, env = "REGISTRAR_BOUNDLESS_MAX_PRICE", default_value_t = 1_000_000)]
     boundless_max_price: u64,
@@ -216,25 +212,26 @@ impl Cli {
         // Build proving config based on mode.
         let proving = match self.proving_mode {
             ProvingMode::Boundless => {
-                if self.boundless.boundless_min_price > self.boundless.boundless_max_price {
-                    return Err(RegistrarError::Config(
-                        "--boundless-min-price must not exceed --boundless-max-price".into(),
-                    ));
-                }
                 if self.boundless.boundless_timeout_secs == 0 {
                     return Err(RegistrarError::Config(
                         "--boundless-timeout-secs must be greater than 0".into(),
                     ));
                 }
 
+                let boundless_key =
+                    self.boundless.boundless_private_key.as_deref().ok_or_else(|| {
+                        RegistrarError::Config("--boundless-private-key is required".into())
+                    })?;
+                let image_id_hex = self
+                    .image_id
+                    .as_deref()
+                    .ok_or_else(|| RegistrarError::Config("--image-id is required".into()))?;
+
                 ProvingConfig::Boundless(Box::new(BoundlessConfig {
                     rpc_url: self.boundless.boundless_rpc_url.ok_or_else(|| {
                         RegistrarError::Config("--boundless-rpc-url is required".into())
                     })?,
-                    signer: parse_private_key(
-                        "--boundless-private-key",
-                        self.boundless.boundless_private_key.as_deref().unwrap_or(""),
-                    )?,
+                    signer: parse_private_key("--boundless-private-key", boundless_key)?,
                     verifier_program_url: self
                         .boundless
                         .boundless_verifier_program_url
@@ -243,8 +240,7 @@ impl Cli {
                                 "--boundless-verifier-program-url is required".into(),
                             )
                         })?,
-                    image_id: parse_image_id(self.image_id.as_deref().unwrap_or(""))?,
-                    min_price: self.boundless.boundless_min_price,
+                    image_id: parse_image_id(image_id_hex)?,
                     max_price: self.boundless.boundless_max_price,
                     poll_interval: Duration::from_secs(self.boundless.boundless_poll_interval_secs),
                     timeout: Duration::from_secs(self.boundless.boundless_timeout_secs),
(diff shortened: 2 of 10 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -0,0 +1,197 @@
...
+            cancel,
+        }
+    }
+
+    /// Runs the registration loop until cancelled.
+    ///
+    /// Runs `step()` immediately on startup, then waits `poll_interval` between
+    /// subsequent ticks. Individual instance failures are logged and skipped —
+    /// the loop continues with the next instance.
+    pub async fn run(&self) -> Result<()> {
```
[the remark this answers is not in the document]
leopoldjoy: Matches the proposer driver pattern (`crates/proof/proposer/src/driver/mod.rs:247`) which also returns `Result<()>` but always `Ok(())`; the signature is forward-compatible for fatal error propagation (e.g., `N`
  consecutive failures) in CHAIN-3456.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -0,0 +1,197 @@
...
+impl<D, P, R, T> RegistrationDriver<D, P, R, T>
+where
+    D: InstanceDiscovery,
+    P: AttestationProofProvider,
+    R: RegistryClient,
+    T: TxManager,
+{
+    /// Creates a new registration driver.
+    #[allow(clippy::too_many_arguments)]
+    pub const fn new(
```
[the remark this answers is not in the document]
leopoldjoy: Removing `const` would trigger the workspace `clippy lint missing_const_for_fn` (enabled via `[workspace.lints.clippy]` in the root `Cargo.toml`); if the constructor ever needs allocation, removing `const` is a one-line change.

bin/prover-registrar/src/cli.rs
```diff
@@ -143,53 +211,171 @@ impl Cli {
...
+        let discovery = AwsTargetGroupDiscovery::new(
+            elb_client,
+            ec2_client,
+            config.discovery.target_group_arn.clone(),
+            config.discovery.port,
+        );
+
+        // Build registry client.
+        let registry = RegistryContractClient::new(
+            config.tee_prover_registry_address,
```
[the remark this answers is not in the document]
leopoldjoy: `RegistryContractClient::new(address, url)` was designed as self-contained in CHAIN-3450; refactoring it to accept a shared `RootProvider` changes the public API across multiple call sites and is better as its own follow-up — the overhead of two HTTP clients at one RPC call per instance per 30-second cycle is negligible.

bin/prover-registrar/src/cli.rs
```diff
@@ -143,53 +211,171 @@ impl Cli {
...
-        // `["default-https-client", "rt-tokio"]` to construct real AWS SDK clients
-        // for AwsTargetGroupDiscovery.
+        let config = self.into_config()?;
+
+        info!(config = ?config, "starting prover registrar");
+
+        // Build L1 provider.
+        let provider = RootProvider::new_http(config.l1_rpc_url.clone());
+
+        // Build signing config for SimpleTxManager.
```
jackchuma: This is unnecessary - see https://github.com/base/base/blob/main/crates/proof/challenge/src/cli.rs#L98-L100

bin/prover-registrar/src/cli.rs
```diff
@@ -143,53 +211,171 @@ impl Cli {
...
+            SigningConfig::Local(pk) => SignerConfig::local(B256::new(pk.to_bytes().0)),
+            SigningConfig::Remote(cfg) => {
+                SignerConfig::Remote { endpoint: cfg.endpoint.clone(), address: cfg.address }
+            }
+        };
+
+        let tx_manager = SimpleTxManager::new(
+            provider,
+            signer_config,
+            TxManagerConfig::default(),
```
jackchuma: This sets defaults for the tx manager without allowing env var configurability. See the challenge crate for how to wire this in properly

bin/prover-registrar/src/cli.rs
```diff
@@ -143,53 +211,171 @@ impl Cli {
...
+            config.l1_chain_id,
+            Arc::new(NoopTxMetrics),
```
jackchuma, line 282: Noticing now that this is also wrong in the challenger but we shouldn't use noop metrics here

leopoldjoy: Good catch, I've added this as a part of CHAIN-3457, so I'll take care of it there 👍

bin/prover-registrar/src/cli.rs
leopoldjoy: Done here: https://github.com/base/base/pull/1453/changes/c806962518ff772a04710d95ca2a19f3055f62c1

bin/prover-registrar/src/cli.rs
leopoldjoy: Addressed: https://github.com/base/base/pull/1453/changes/c806962518ff772a04710d95ca2a19f3055f62c1

crates/proof/tee/registrar/src/driver.rs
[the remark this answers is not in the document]
leopoldjoy: Matches the proposer driver pattern (`crates/proof/proposer/src/driver/mod.rs:247`) which also returns `Result<()>` but always `Ok(())`; the signature is forward-compatible for propagating fatal errors (e.g., `N` consecutive failures) when deregistration logic is added in CHAIN-3456.

bin/prover-registrar/src/cli.rs
```diff
@@ -1,195 +1,355 @@
...
+            ec2_client,
+            config.discovery.target_group_arn.clone(),
+            config.discovery.port,
+        );
+
+        // Build registry client.
+        let registry = RegistryContractClient::new(
+            config.tee_prover_registry_address,
+            config.l1_rpc_url.clone(),
+        );
```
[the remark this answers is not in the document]
leopoldjoy: `RegistryContractClient::new(address, url)` is a pre-existing self-contained API from CHAIN-3450; refactoring it to accept a shared `RootProvider` changes its public interface across multiple consumers and is better as a follow-up — the overhead of two HTTP clients at one call per instance per 30-second cycle is negligible.

bin/prover-registrar/Cargo.toml
jackchuma, line 1: Nit: waterfall order

bin/prover-registrar/src/cli.rs
```diff
@@ -1,195 +1,353 @@
...
-            .required(true)
-            .args(["private_key", "signer_endpoint"])
-    )
-)]
+#[command(name = "prover-registrar", version, about)]
 pub(crate) struct Cli {
     // ── L1 ────────────────────────────────────────────────────────────────────
     /// L1 Ethereum RPC endpoint.
-    #[arg(long, env = "REGISTRAR_L1_RPC_URL")]
+    #[arg(long, env = "BASE_REGISTRAR_L1_RPC_URL")]
```
jackchuma: Nit: can use the `cli_env!` macro to reduce prefix duplication

bin/prover-registrar/Cargo.toml
```diff
@@ -16,10 +16,31 @@ path = "src/main.rs"
...
+# AWS
+aws-config = { workspace = true, features = ["default-https-client", "rt-tokio"] }
+aws-sdk-ec2.workspace = true
+aws-sdk-elasticloadbalancingv2.workspace = true
+
+# CLI / runtime
+hex.workspace = true
+url.workspace = true
+eyre.workspace = true
+humantime.workspace = true
```
jackchuma: I think the `humantime` dep can be removed

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// Cancellation token for graceful shutdown.
+    pub cancel: CancellationToken,
+}
+
+/// Core registration loop tying together discovery, attestation polling,
+/// ZK proof generation, and on-chain submission.
+///
+/// Generic over the discovery, proof generation, registry, and transaction
+/// manager backends so each can be mocked independently in tests.
+pub struct RegistrationDriver<D, P, R, T> {
```
jackchuma, line 40: Can we add unit testing for this?

crates/proof/tee/registrar/src/error.rs
```diff
@@ -55,5 +56,11 @@ impl From<ProverError> for RegistrarError {
     }
 }
 
+impl From<TxManagerError> for RegistrarError {
```
jackchuma, line 63: `From<TxManagerError>` maps to the wrong error variant

  https://github.com/base/base/blob/leopoldjoy/chain-3455-registration-driver/crates/proof/tee/registrar/src/error.rs#L59-L62

  ```rust
  impl From<TxManagerError> for RegistrarError {
      fn from(e: TxManagerError) -> Self {
          Self::Signing(Box::new(e))
      }
  }
```

  TxManagerError covers much more than signing — RPC failures, fee estimation errors, nonce conflicts, send timeouts, receipt polling failures, etc. Mapping all of these to RegistrarError::Signing means an operator seeing "signing error" in logs
   could be chasing the wrong problem when the real cause is e.g. an RPC timeout or nonce exhaustion.

  Suggest adding a dedicated variant:

```rust
  /// Transaction submission or confirmation failed.
  #[error("transaction error")]
  Transaction(#[source] Box<dyn std::error::Error + Send + Sync>),
```

  and mapping TxManagerError → Transaction instead.

bin/prover-registrar/Cargo.toml
leopoldjoy: Fixed: https://github.com/base/base/pull/1453/commits/18e3776ecfc5cb8b99c743b40c799d7f458a5fb1

bin/prover-registrar/src/cli.rs
leopoldjoy: Fixed: https://github.com/base/base/pull/1453/changes/18e3776ecfc5cb8b99c743b40c799d7f458a5fb1

crates/proof/tee/registrar/src/error.rs
leopoldjoy: Good call! Done: https://github.com/base/base/pull/1453/changes/18e3776ecfc5cb8b99c743b40c799d7f458a5fb1

bin/prover-registrar/Cargo.toml
leopoldjoy: Required by the `define_tx_manager_cli!` macro expansion which references `::humantime::parse_duration` via absolute path — removing it breaks compilation.

crates/proof/tee/registrar/src/driver.rs
leopoldjoy: `process_instance()` creates `ProverClient` internally as a concrete HTTP client, so unit testing the driver requires abstracting it behind a trait — that's a design change tracked as a follow-up, not a test addition.

bin/prover-registrar/Cargo.toml
```diff
@@ -16,10 +16,32 @@ path = "src/main.rs"
...
+aws-sdk-ec2.workspace = true
+aws-sdk-elasticloadbalancingv2.workspace = true
+aws-config = { workspace = true, features = ["default-https-client", "rt-tokio"] }
+
+# CLI / runtime
+hex.workspace = true
+url.workspace = true
+eyre.workspace = true
+tracing.workspace = true
+humantime.workspace = true
```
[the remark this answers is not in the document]
leopoldjoy: `humantime` is required by the `define_tx_manager_cli!` macro expansion which references `::humantime::parse_duration` via absolute path — it won't appear in use statements but removing it breaks compilation.

bin/prover-registrar/Cargo.toml
```diff
@@ -16,10 +16,32 @@ path = "src/main.rs"
...
```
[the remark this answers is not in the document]
leopoldjoy: `humantime` is consumed by the `define_tx_manager_cli!` macro expansion which generates `value_parser = ::humantime::parse_duration` in the clap arg definitions — removing it breaks compilation (the challenger binary has the same dependency for the same reason).

bin/prover-registrar/src/cli.rs
```diff
@@ -1,195 +1,354 @@
...
+
     /// Proof generation timeout in seconds.
-    #[arg(long, env = "REGISTRAR_BOUNDLESS_TIMEOUT_SECS", default_value_t = 600)]
+    #[arg(long, env = cli_env!("BOUNDLESS_TIMEOUT_SECS"), default_value_t = 600)]
     boundless_timeout_secs: u64,
 
     /// `NitroEnclaveVerifier` contract address for certificate caching (optional).
-    #[arg(long, env = "REGISTRAR_NITRO_VERIFIER_ADDRESS")]
+    #[arg(long, env = cli_env!("NITRO_VERIFIER_ADDRESS"))]
     nitro_verifier_address: Option<Address>,
```
jackchuma, line 150: Seems like this might be unused

jackchuma (approved):

PR #1515, opened 2026-03-20
feat(CHAIN-3456): add orphan signer deregistration

leopoldjoy:
## Summary
- Add orphan detection to `RegistrationDriver`: each cycle compares on-chain signers against active instances and deregisters any signer not backed by a healthy instance (CHAIN-3456)
- Change `process_instance` return type from `Result<()>` to `Result<Address>` so `step()` can collect the active signer set
- Add `deregister_orphans()` method that fetches all on-chain signers via `get_registered_signers()`, diffs against the active set, and submits `deregisterSigner` transactions for orphans
- Guard against mass deregistration from transient failures: skip orphan cleanup when all instance processing fails but registerable instances exist
- Per-signer error handling in deregistration loop (warn and continue, matching registration pattern)

## Test plan
- [x] `cargo +nightly fmt -p base-proof-tee-registrar --check`
- [x] `RISC0_SKIP_BUILD_KERNELS=1 cargo clippy -p base-proof-tee-registrar --tests -- -D warnings`
- [x] `RISC0_SKIP_BUILD_KERNELS=1 cargo test -p base-proof-tee-registrar --lib` (32/32 passing)
- [x] New rstest `deregister_calldata_encodes_correctly` covers zero address, Hardhat account #0, and all-ones address

### DIFF
--- a/crates/proof/tee/registrar/src/driver.rs
+++ b/crates/proof/tee/registrar/src/driver.rs
@@ -117,7 +117,11 @@ where
 
         // Resolve signer addresses for ALL reachable instances (regardless of
         // health status) to build a complete active set. This protects draining
-        // and transiently-unhealthy instances from premature deregistration.
+        // instances (still running, usually reachable) from premature
+        // deregistration. Truly unreachable instances will fail the RPC and be
+        // excluded — the majority guard below is the safeguard for that case.
+        // A signer-address cache across cycles would strengthen this but adds
+        // state management complexity; deferred for now.
         // Registration is only attempted for instances that pass should_register().
         let mut active_signers = HashSet::new();
 
@@ -149,12 +153,13 @@ where
             return Ok(());
         }
 
-        // Guard against mass deregistration from transient failures: if fewer
-        // than half of the discovered instances could be reached, our view of
-        // the active set is unreliable. Only proceed when we have a clear
-        // majority, or when discovery itself returns zero instances (truly no
-        // instances exist).
-        if !instances.is_empty() && active_signers.len() * 2 < instances.len() {
+        // Guard against mass deregistration from transient failures: require a
+        // strict majority (>50%) of discovered instances to be reachable before
+        // proceeding with orphan cleanup. When discovery returns zero instances
+        // (e.g. after ASG scale-down removes them from the target group),
+        // deregistration proceeds normally — scaled-down instances leave the
+        // target group entirely, so they don't inflate `instances.len()`.
+        if !instances.is_empty() && active_signers.len() * 2 <= instances.len() {
             warn!(
                 active = active_signers.len(),
                 total = instances.len(),
--- a/crates/proof/tee/registrar/src/driver.rs
+++ b/crates/proof/tee/registrar/src/driver.rs
@@ -2,9 +2,10 @@
 //!
 //! Discovers prover instances, checks on-chain registration status, generates
 //! ZK proofs for unregistered signers, and submits registration transactions
-//! to L1 via the [`TxManager`].
+//! to L1 via the [`TxManager`]. Also detects orphaned on-chain signers (those
+//! no longer backed by a healthy instance) and deregisters them.
 
-use std::{fmt, time::Duration};
+use std::{collections::HashSet, fmt, time::Duration};
 
 use alloy_primitives::{Address, Bytes, hex};
 use alloy_sol_types::SolCall;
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

leopoldjoy:
> One observation (non-blocking): the test coverage only covers ABI encoding for the `deregisterSigner` calldata. The guard conditions in `step()` (cancellation, empty active set, partial failures) and the `deregister_orphans` orchestration logic don't have dedicated unit tests. Given these guards are the primary defense against mass deregistration, consider adding mock-based tests that exercise each guard path.

Acknowledged, deferred to CHAIN-3599 👍

jackchuma (commented):
## Must-fix items

**1. Signer address lost on partial `process_instance` failure**

Once `signer_address` is derived (L178), a transient error from `is_registered()` (L180) propagates via `?` and the address is never returned to the caller. That signer won't be in `active_signers`, so `deregister_orphans` will remove it from the chain even though the instance is alive. Suggest separating address discovery from registration so the active set is built from all *reachable* instances, not just fully-processed ones.

**2. Guard only covers total failure**

The `active_signers.is_empty() && !instances.is_empty()` guard (L155) only fires when every instance fails. If 1-of-5 succeeds and 4 fail transiently, you proceed to deregister with an active set of size 1. Consider a proportional threshold or failure-rate check.

**3. No tests for new logic**

The only test verifies `alloy_sol_types` ABI encoding — a library guarantee. None of the actual new paths are covered: `step()` orchestration, the guard condition, cancellation mid-loop vs inside `process_instance`, `deregister_orphans` with mocks, partial failure, or the empty-instances-deregisters-everything edge case. The trait boundaries are already mock-able, so this should be straightforward.

**4. Draining instances deregistered with no grace period**

`should_register()` excludes `Draining`, so a draining instance's signer is immediately orphaned. The doc comment acknowledges this depends on `poll_interval > ALB drain timeout`, but nothing enforces that. If the interval is shorter, you deregister a signer while the instance is still serving. Either query draining instances just for their signer address (they're still reachable) and include them in the active set, or add a configurable grace period.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -100,50 +101,127 @@ where
...
+            debug!("shutdown requested, skipping orphan deregistration");
+            return Ok(());
+        }
+
+        // Guard against mass deregistration from transient failures: if fewer
+        // than half of the discovered instances could be reached, our view of
+        // the active set is unreliable. Only proceed when we have a clear
+        // majority, or when discovery itself returns zero instances (truly no
+        // instances exist).
+        if !instances.is_empty() && active_signers.len() * 2 < instances.len() {
```
0x00101010: what if we want to scale down the TEE instances? Wouldn't this bork orphan deregistration

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -198,4 +276,303 @@ where
...
+
+        assert_eq!(calldata.len(), DEREGISTER_CALLDATA_LEN);
+        assert_eq!(&calldata[..4], &ITEEProverRegistry::deregisterSignerCall::SELECTOR);
+        // The 12 bytes between the selector and the address must be zero-padding.
+        assert_eq!(&calldata[4..ABI_ADDRESS_OFFSET], &[0u8; ABI_ADDRESS_PAD]);
+        // The last 20 bytes must be the raw signer address.
+        assert_eq!(&calldata[ABI_ADDRESS_OFFSET..], signer.as_slice());
+    }
+
+    // ── deregister_orphans tests ────────────────────────────────────────
```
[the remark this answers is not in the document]
leopoldjoy: Already tracked in CHAIN-3599 with the specific paths listed. Blocked on `ProverClient` being behind a trait.

crates/proof/tee/registrar/src/driver.rs
leopoldjoy: This is not an issue as scaled-down instances are removed from the target group by the ASG, so `discover_instances()` returns fewer instances and the guard doesn't fire. The comment on the guard now documents this.

0x00101010 (approved):

PR #1574, opened 2026-03-21
feat(actions): In-Process Engine Client With EVM

### LINKED ISSUE #1559: feat(actions): Create an In-Process EngineClient
## Summary

The actions test framework has no implementation of \`EngineClient\` that performs real computation. \`MockEngineClient\` at https://github.com/base/base/blob/main/crates/consensus/engine/src/test_utils/engine_client.rs returns pre-configured static responses and is intended for unit-testing individual components that call into the engine API. It does not execute transactions, does not maintain EVM state across blocks, and cannot compute state roots.

\`L2Sequencer\` at https://github.com/base/base/blob/main/actions/harness/src/l2.rs performs real EVM execution using \`InMemoryDB\` (revm's in-memory state database) and computes a proper Merkle Patricia Trie state root via \`compute_state_root\` for each block it builds. The verifier currently receives \`OpAttributesWithParent\` from the derivation pipeline and advances the safe head counter without re-executing anything or validating the output.

\`ActionEngineClient\` closes this gap by providing a full \`EngineClient\` implementation that runs the same revm execution path, starting from an identical genesis state, independently of \`L2Sequencer\`. When \`new_payload_vX\` is called with a derived payload, the client executes every transaction through the EVM, commits the resulting state, and computes the state root. This result can then be compared against the state root stored by \`L2Sequencer\` in the shared block registry to confirm that derivation produced the correct payload.

## Why

Without real EVM execution in the engine client, a test node that runs a full derivation pipeline still cannot validate that the derived payload content is correct. A mismatch between what the sequencer built and what the derivation pipeline reconstructed would go undetected. \`ActionEngineClient\` provides the same assurance as a node backed by a real execution engine: every derived block is independently executed and its state root verified.

## Changeset

New file: \`actions/harness/src/engine.rs\`.

\`ActionEngineClient\` holds an \`InMemoryDB\` that must be initialised from the same genesis state as the \`L2Sequencer\` it is paired with. Both use the \`compute_state_root\` free function defined at https://github.com/base/base/blob/main/actions/harness/src/l2.rs#L527 and the \`OpEvmConfig\` + \`OpChainSpecBuilder\` already imported by the harness.

```rust
pub struct ActionEngineClient {
    db: InMemoryDB,
    rollup_config: Arc<RollupConfig>,
    chain_spec: Arc<OpChainSpec>,
    canonical_head: L2BlockInfo,
    block_registry: SharedBlockRegistry,
}
```

\`SharedBlockRegistry\` is an extension of the existing \`SharedBlockHashRegistry\` that stores \`Sealed<Header>\` values (carrying \`state_root\`, \`receipts_root\`, and all other header fields) rather than just block hashes. \`L2Sequencer\` writes a \`Sealed<Header>\` for every block it builds; \`ActionEngineClient\` reads from it after executing a derived payload to compare state roots.

The \`new_payload_vX\` methods decode the \`ExecutionPayload\` into a list of \`OpTxEnvelope\` values, build the EVM block environment from the payload fields (timestamp, block number, gas limit, base fee, parent hash), execute each transaction through revm using \`OpEvmConfig::evm_with_env\`, commit state after each transaction, compute the MPT state root, and return \`PayloadStatus::Valid { latest_valid_hash: computed_block_hash }\`. If the computed state root differs from the value in \`SharedBlockRegistry\` for that block number the method panics with a diagnostic showing both roots and the block number.

The \`fork_choice_updated_vX\` methods update \`canonical_head\` to the supplied \`ForkchoiceState::head_block_hash\` and return \`ForkchoiceUpdated\` with \`PayloadStatus::Valid\`.

\`get_l2_block\` and \`l2_block_by_label\` serve from the block registry. \`get_l1_block\` delegates to the \`ActionL1ChainProvider\` passed at construction time. \`get_proof\` returns an empty proof (not required for derivation testing).

Both \`EngineClient\` and \`OpEngineApi<Base, Http<HyperAuthClient>>\` are implemented. The \`Http<HyperAuthClient>\` phantom parameter in the supertrait does not constrain the implementation; \`MockEngineClient\` already demonstrates this pattern.

refcell:
## Summary

Implements `ActionEngineClient`, a full `EngineClient` backed by an in-process revm EVM that starts from the same genesis state as `L2Sequencer`. When `new_payload_vX` is called with a derived payload, every transaction is executed through the EVM, the MPT state root is computed, and the result is compared against the value stored in the shared `SharedBlockHashRegistry` — panicking with a diagnostic if they diverge. This closes the gap where a verifier running a complete derivation pipeline had no way to confirm that the reconstructed payload content was correct.

Closes #1559

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2502,21 +2502,30 @@ dependencies = [
  "alloy-consensus",
  "alloy-eips",
  "alloy-hardforks 0.4.7",
+ "alloy-network",
  "alloy-primitives",
+ "alloy-provider",
  "alloy-rlp",
+ "alloy-rpc-types-engine",
  "alloy-rpc-types-eth",
  "alloy-signer",
  "alloy-signer-local",
+ "alloy-transport",
+ "alloy-transport-http",
  "alloy-trie",
  "async-trait",
  "base-alloy-chains",
  "base-alloy-consensus",
+ "base-alloy-network",
+ "base-alloy-provider",
+ "base-alloy-rpc-types",
  "base-alloy-rpc-types-engine",
  "base-batcher-core",
  "base-batcher-encoder",
  "base-batcher-source",
  "base-blobs",
  "base-consensus-derive",
+ "base-consensus-engine",
  "base-consensus-genesis",
  "base-consensus-gossip",
  "base-consensus-node",
@@ -9863,7 +9872,7 @@ dependencies = [
  "cesu8",
  "cfg-if",
  "combine",
- "jni-sys",
+ "jni-sys 0.3.1",
  "log",
  "thiserror 1.0.69",
  "walkdir",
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

danyalprout (approved):

PR #1575, opened 2026-03-21
Fix proof journal to use proposer address instead of enclave signer address

leopoldjoy:
The enclave was using its own signer key address as the `proposer` field in the `ProofJournal`. On-chain, [`AggregateVerifier`](https://github.com/base/contracts/blob/main/src/multiproof/AggregateVerifier.sol) recomputes the journal with `gameCreator()` (the actual transaction sender / KMS key), causing a hash mismatch that made `ecrecover` return a garbage address and every dispute game creation revert with `SignerNotRegistered`.

Fix: thread the proposer address from the proposer service through the existing preimage pipeline to the enclave, so the journal is signed with the correct address. No changes to the `ProverBackend` trait, transport layer, or vsock protocol — the address flows through `BootInfo` preimages.

Changes:
- Add `proposer: Address` to `ProofRequest` and `BootInfo`
- Serve it via `BootKeyValueStore` as preimage key 8
- Read it in enclave `Server::prove` from `boot_info.proposer`
- Set it from `tx_manager.sender_address()` in the proposer driver

### DIFF
--- a/crates/proof/challenge/src/driver.rs
+++ b/crates/proof/challenge/src/driver.rs
@@ -358,6 +358,7 @@ impl<L2: L2Provider, P: ZkProofProvider, T: TxManager> Driver<L2, P, T> {
             agreed_l2_output_root,
             claimed_l2_output_root,
             claimed_l2_block_number,
+            proposer: self.submitter.sender_address(),
         };
 
         let result = tee.provider.prove(request).await.map_err(|e| eyre::eyre!(e))?;
--- a/crates/proof/primitives/src/proof.rs
+++ b/crates/proof/primitives/src/proof.rs
@@ -1,6 +1,6 @@
 use alloc::vec::Vec;
 
-use alloy_primitives::B256;
+use alloy_primitives::{Address, B256};
 use base_proof_preimage::PreimageKey;
 
 use crate::Proposal;
@@ -51,6 +51,12 @@ pub struct ProofRequest {
     pub claimed_l2_output_root: B256,
     /// L2 block number that the claimed output root commits to.
     pub claimed_l2_block_number: u64,
+    /// Address of the proposer that will submit the proof transaction on-chain.
+    ///
+    /// Included in the proof journal so on-chain verification can match it against
+    /// the actual `msg.sender` (gameCreator).
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub proposer: Address,
 }
 
 /// A proof request bundled with the witness data needed to fulfill it.
--- a/crates/proof/challenge/src/submitter.rs
+++ b/crates/proof/challenge/src/submitter.rs
@@ -21,6 +21,11 @@ impl<T: TxManager> ChallengeSubmitter<T> {
         Self { tx_manager }
     }
 
+    /// Returns the address that will submit transactions on-chain.
+    pub fn sender_address(&self) -> Address {
+        self.tx_manager.sender_address()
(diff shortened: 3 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/primitives/src/proof.rs
```diff
@@ -51,6 +51,12 @@ pub struct ProofRequest {
     pub claimed_l2_output_root: B256,
     /// L2 block number that the claimed output root commits to.
     pub claimed_l2_block_number: u64,
+    /// Address of the proposer that will submit the proof transaction on-chain.
+    ///
+    /// Included in the proof journal so on-chain verification can match it against
+    /// the actual `msg.sender` (gameCreator).
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub proposer: Address,
```
jackchuma, line 59: Nit: this is unnecessary - we have a sender_address fn exposed from the tx manager: https://github.com/base/base/blob/main/crates/utilities/tx-manager/src/traits.rs#L65

jackchuma: oh nevermind I misread this file

leopoldjoy: The `sender_address` on the tx manager lives on the proposer side, but the `ProofRequest` crosses a JSON-RPC boundary to the prover (a separate process without access to the tx manager), and passing it per-request means the proposer address can change dynamically (e.g., key rotation) without needing to restart / reconfigure the prover.

crates/proof/challenge/src/driver.rs
```diff
@@ -358,6 +358,7 @@ impl<L2: L2Provider, P: ZkProofProvider, T: TxManager> Driver<L2, P, T> {
             agreed_l2_output_root,
             claimed_l2_output_root,
             claimed_l2_block_number,
+            proposer: self.submitter.sender_address(),
```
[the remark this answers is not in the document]
leopoldjoy: Confirmed — `nullify()` uses `msg.sender` (line 568 of `AggregateVerifier.sol`), not `gameCreator()`, so the challenger's own `sender_address()` is the correct value here.

refcell (approved):
lgtm

PR #1592, opened 2026-03-23
chore(el): derive reth chainspecs from base-alloy-chains vs. duplicating configs

danyalprout:
### Description
Part one of deduplicating chain configurations and settings. Creates `base-alloy-chains` which has the chain configurations defined in Rust.

Upstream consumers will map this generic type into CL/EL specific types via a `TryFrom/From` impl.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>-1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/alloy/chains/src/config.rs
```diff
@@ -350,5 +370,26 @@ const DEVNET: BaseChainConfig = BaseChainConfig {
 
     bootnodes: &[],
 
+    prune_delete_limit: 0,
```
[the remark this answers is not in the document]
danyalprout: This is a correct change

crates/alloy/chains/src/config.rs
```diff
@@ -2,12 +2,18 @@
 
 use alloy_primitives::{Address, B256, U256, address, b256, uint};
 
+/// All supported chain names for the CLI.
+pub const SUPPORTED_CHAINS: &[&str] =
+    &["base", "base_sepolia", "base-sepolia", "base-devnet-0-sepolia-dev-0", "dev"];
```
refcell, line 7: I think these could be placed on `BaseChainConfig` i.e. `BaseChainConfig::ALL`

crates/execution/chainspec/src/spec.rs
```diff
@@ -320,6 +307,60 @@ impl From<Genesis> for OpChainSpec {
...
+        Self {
+            inner: ChainSpec {
+                chain: Chain::from_id(cfg.chain_id),
+                genesis_header,
+                genesis,
+                paris_block_and_final_difficulty: Some((0, U256::ZERO)),
+                hardforks,
+                base_fee_params,
+                prune_delete_limit: cfg.prune_delete_limit,
+                ..Default::default()
```
refcell, line 358: Are we sure about this `..Default::default()`? I think we ought to list everything out here so it's very explicit about what's made default.

crates/execution/chainspec/src/lib.rs
```diff
@@ -10,23 +10,11 @@
...
 
-mod base;
-pub use base::BASE_MAINNET;
-
-mod base_devnet_0_sepolia_dev_0;
-pub use base_devnet_0_sepolia_dev_0::BASE_DEVNET_0_SEPOLIA_DEV_0;
-
-mod base_sepolia;
-pub use base_sepolia::BASE_SEPOLIA;
-
```
refcell, line 21: I think we should at least lazy init (e.g. how the hardfork crate does it) the chain specs here for base so we're not doing constructions and conversions at runtime.

crates/execution/chainspec/src/lib.rs
```diff
@@ -10,23 +10,11 @@
...
 
 mod builder;
 pub use builder::OpChainSpecBuilder;
 
-mod dev;
-pub use dev::OP_DEV;
-
 mod spec;
-pub use spec::{OpChainSpec, OpGenesisInfo, SUPPORTED_CHAINS};
+pub use spec::{OpChainSpec, OpGenesisInfo};
```
refcell, line 20: Would be nice to rename this to `BaseChainSpec` and `BaseGenesisInfo` while we're in here. If there are too many call sites, fine to leave it for later

crates/execution/hardforks/src/lib.rs
```diff
@@ -11,7 +11,6 @@
 extern crate alloc;
 
 mod chain;
-pub use chain::{
-    BASE_DEVNET_0_SEPOLIA_DEV_0_HARDFORKS, BASE_MAINNET_HARDFORKS, BASE_SEPOLIA_HARDFORKS,
-    BaseChainUpgradesExt, DEV_HARDFORKS,
-};
+// Re-export base-alloy-upgrades types.
+pub use base_alloy_chains::BaseChainUpgrades;
+pub use chain::{BASE_MAINNET_HARDFORKS, BaseChainUpgradesExt};
```
refcell, line 16: rm re-exports

crates/execution/node/src/node.rs
```diff
@@ -274,12 +274,15 @@ impl OpNode {
...
+    /// use std::sync::Arc;
+    /// use base_alloy_chains::BaseChainConfig;
+    /// use base_execution_chainspec::OpChainSpec;
     /// use base_node_core::OpNode;
     ///
     /// fn demo(runtime: reth_tasks::Runtime) {
+    ///     let chain_spec = Arc::new(OpChainSpec::from(BaseChainConfig::mainnet()));
     ///     let factory = OpNode::provider_factory_builder()
-    ///         .open_read_only(BASE_MAINNET.clone(), "datadir", runtime)
+    ///         .open_read_only(chain_spec, "datadir", runtime)
```
refcell, line 285: I think this api is noticeably worse, especially since we have effectively a fixed number of chains we build.

The easiest fix IMO would be to just lazy init these constants again in `base_execution_chainspec`

refcell (changes requested):
Couple comments

crates/execution/chainspec/src/lib.rs
danyalprout: I maybe missing something (please feel free to correct), but I believe they're only setup once when the ChainSpec is created. 

I'm not sure this lazy init was really buying anything

crates/execution/node/src/node.rs
danyalprout: I don't really want to have to create new constants throughout the codebase whenever we create a new testnet (e.g. vibenet).


Like ideally this is 
```
let flags = Cli::parse(...);
let config = BaseChainConfig::by_name(flags.chain_name)
let chain_spec: OpChainSpec = config.into();
```

refcell: Gotcha ok, that's good enough justification for me!

crates/execution/chainspec/src/lib.rs
refcell: Right, but then it doesn't have to initialize again thereafter. Not a big deal though. Personal style preference that I don't really care all too much about. can leave out of this pr

PR #1620, opened 2026-03-23
Kaley/load tests updates

kmchicoine:
In this PR:

- Flatten crate structure: Hoist base-load-tests-core up one level to crates/infra/load-tests/, removing the unnecessary nested subdirectory and renaming the crate to base-load-tests
- Switch RPC provider to RootProvider<Base>: Use the Base-specific network type for proper OP-stack receipt handling (OpTransactionReceipt)
- Fix precompile and calldata payloads: Add EIP-7623 floor cost calculation for calldata gas limits, fix Blake2F to use round-based gas estimation, and use proper precompile addresses
- Improve funding logic: Fund accounts by deficit rather than full amount, use pending nonce for funder, and add progress reporting during load tests
- Replace example runner and configs: Replace spam.rs with load_test.rs (adds FUNDER_KEY env var support) and add network-specific configs (devnet.yaml, sepolia.yaml, sepolia-alpha.yaml)
- Add Justfile for load tests: Add just load-test recipes for devnet, sepolia-alpha, and sepolia; also add _skip_kernels for macOS risc0 builds in the root Justfile
- Clean up TestConfig: Add Default impl with post-parse validation, consolidate serde defaults into #[serde(default)] on the struct, remove const fn helpers


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/infra/load-tests/src/config/test_config.rs
```diff
@@ -175,8 +158,41 @@ impl TestConfig {
...
-            .map_err(|e| BaselineError::Config(format!("failed to parse YAML: {e}")))
+        let config: Self = serde_yaml::from_str(yaml)
+            .map_err(|e| BaselineError::Config(format!("failed to parse YAML: {e}")))?;
+        config.validate()?;
+        Ok(config)
+    }
+
+    /// Validates that all required fields are set and values are sensible.
+    pub fn validate(&self) -> Result<()> {
+        if self.rpc.is_empty() {
```
meyer9: nit: could check that this is a valid URL, or use `Url` type for Serde to do it for you

crates/infra/load-tests/src/config/test_config.rs
```diff
@@ -175,8 +158,41 @@ impl TestConfig {
...
+
+    /// Resolves the funder key, checking the `FUNDER_KEY` environment variable
+    /// first, then falling back to the config value. Returns an error if the
+    /// key is a placeholder or env var reference that cannot be resolved.
+    pub fn resolve_funder_key(&self) -> Result<String> {
+        if let Ok(env_key) = std::env::var("FUNDER_KEY") {
+            return Ok(env_key);
+        }
+
+        if self.funder_key.starts_with("${") || self.funder_key == "<your key here>" {
```
meyer9: maybe this field can be optional instead of using a placeholder value here? Also, if it's always hex, you can check that the format is a hex string to validate.

Since users can still access the pub `.funder_key`, it might not be clear they have to call `resolve_funder_key()`. Maybe you can move this to the parsing/validation phase instead? Or make the interface `config.with_env().with_yaml(file)` so it's clear that it loads env and YAML during parsing?

meyer9 (commented):
These updates are looking great! Left a few suggestions (feel free to take or leave).

crates/infra/load-tests/src/config/test_config.rs
kmchicoine: Yeah, this is confusing. I think I'll just remove funder_key from the config and make it env_var only- that way we don't risk someone editing the yaml and accidentally committing their key, and it cleans up this interface as well

meyer9 (approved):

refcell (approved):

PR #1629, opened 2026-03-23
fix(consensus): Small Sealing Bugs

refcell:
## Summary

This PR fixes two sequencing bugs that caused the unsafe head to fall behind the safe head under catch-up conditions. The root cause of the first bug was a race between `insert_unsafe_payload` and the next `build()` call: because `insert_unsafe_payload` is fire-and-forgot to the engine, the unsafe-head watch channel still reflects the old (pre-insert) block at the point `build()` reads it, causing the next block to be built on the grandparent rather than the just-sealed parent. The fix introduces a `next_build_parent: Option<L2BlockInfo>` field on `SequencerActor` and a new `PayloadBuilder::build_on(parent)` method. After a successful seal, the sealed envelope's `L2BlockInfo` is derived immediately and stashed as `next_build_parent`; the subsequent build call in the `Ok(true)` sealer arm then uses `build_on` with that explicit parent instead of reading the watch channel, guaranteeing the engine's `InsertTask` is enqueued before `BuildTask`.

The second bug was that the `build_ticker` and sealer arms were not mutually exclusive: in catch-up mode `reset_immediately()` could make the ticker `Poll::Ready` while the sealer's `step().await` was `Poll::Pending`, allowing the ticker arm to fire mid-seal and start a second build before the first three-step pipeline (commit → gossip → insert) completed. The ticker arm is now gated on `self.sealer.is_none()`, making the two arms structurally exclusive. Two additional fixes land alongside these: pipeline reset errors in `build_attributes` (e.g. `BlockMismatch`, `BrokenTimeInvariant`) no longer trigger an engine reset that would rewind the unsafe head to the safe head — they now return `Ok(None)` and retry on the next tick; and `schedule_initial_reset` now services admin API queries throughout both the reset-attempt loop and the backoff sleep so the sequencer remains controllable during EL sync.

### DIFF
--- a/crates/consensus/service/src/actors/sequencer/actor.rs
+++ b/crates/consensus/service/src/actors/sequencer/actor.rs
@@ -10,6 +10,7 @@ use async_trait::async_trait;
 use base_consensus_derive::AttributesBuilder;
 use base_consensus_genesis::RollupConfig;
 use base_consensus_rpc::SequencerAdminAPIError;
+use base_protocol::L2BlockInfo;
 use tokio::{
     select,
     sync::{mpsc, oneshot},
@@ -26,8 +27,7 @@ use crate::{
             conductor::Conductor,
             error::SequencerActorError,
             metrics::{
-                inc_seal_error, inc_seal_pipeline_overlap, update_seal_duration_metrics,
-                update_total_transactions_sequenced,
+                inc_seal_error, update_seal_duration_metrics, update_total_transactions_sequenced,
             },
             origin_selector::OriginSelector,
             recovery::RecoveryModeGuard,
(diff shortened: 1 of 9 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

danyalprout (approved):

PR #1743, opened 2026-03-28
Add trie node counts to meterBundle

niran:
## Summary
- add `stateRootAccountNodeCount` and `stateRootStorageNodeCount` to `meterBundle`
- expose deterministic trie-work counts alongside wall-clock state root timing
- count only the bundle’s own post-state leaves plus the incremental trie updates returned by reth

## Why
The mainnet metering review showed that state root work is the binding resource dimension, but it also showed that simulated state root time can vary in ways that are hard to interpret from timing alone.

The clearest example is [block `43925035`](https://base-mainnet-tips.cbhq.net/block/0x78359f654d473d9b0ae0155675365ee0cea936e63af80789d4c54efb0fcd78d2). In the review, that block had:
- `82` transactions with simulated state root time above `200ms`
- `38,324ms` of combined simulated state root time
- `921ms` max simulated state root time
- only `150ms` of actual finalization time

That is exactly the kind of mismatch we need to explain. It could reflect pending-state context in the simulation, transient node load during metering, cache effects, or something else. What matters is that wall-clock simulation time by itself is not enough to tell us whether a transaction actually introduced a large amount of incremental trie work.

This PR adds a structural metric that helps isolate those effects.

## What This Measures
The new fields are:

- `stateRootAccountNodeCount`
- `stateRootStorageNodeCount`

They are computed from two sources:

- surviving/inserted leaves in the bundle’s hashed post-state
- trie update nodes returned by reth during state root calculation

In practice:

- account count = surviving/inserted account leaves + account trie branch updates/removals
- storage count = surviving/inserted storage slot leaves + storage trie branch updates/removals/deletes

Deletion-style changes are not double counted as leaves. Deleted accounts, zero-valued storage removals, and pure storage wipes are represented through trie updates rather than emitted leaf nodes.

## Pending State Behavior
When `meterBundle` runs on top of pending state, the metric is still intended to reflect the bundle’s incremental work rather than the full pending trie.

We start from the bundle’s hashed post-state and then prepend the cached pending trie only to compute the final root. That lets us measure the trie nodes attributable to the bundle while still using the pending-state root path for correctness.

## Why This Is Useful
These counts give us a side-effect-free structural metric to compare against simulated state root time.

That should let us answer questions like:

- did this transaction really force reth to touch a large amount of trie structure?
- or did the simulation only look slow because of pending-state context, cache state, or transient node conditions?

This does not replace timing data. It gives us an independent metric that helps explain timing anomalies.

## Testing
- `cargo test -p base-bundles --lib`
- `cargo test -p base-builder-metering --lib`
- `cargo check -p base-metering --lib`

### DIFF
--- a/crates/client/metering/src/meter.rs
+++ b/crates/client/metering/src/meter.rs
@@ -450,9 +450,10 @@ where
 mod tests {
     use alloy_eips::Encodable2718;
     use alloy_primitives::{Address, Bytes, keccak256, utils::Unit};
+    use alloy_sol_types::SolCall;
     use base_bundles::{Bundle, ParsedBundle};
     use base_execution_primitives::OpTransactionSigned;
-    use base_node_runner::test_utils::{Account, TestHarness};
+    use base_node_runner::test_utils::{Account, SimpleStorage, TestHarness};
     use eyre::Context;
     use reth_provider::StateProviderFactory;
     use reth_revm::{bytecode::Bytecode, state::AccountInfo};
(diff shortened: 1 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/client/metering/src/meter.rs
```diff
@@ -336,6 +440,8 @@ where
         bundle_hash,
         total_time_us,
         state_root_time_us,
+        state_root_account_node_count: trie_node_counts.account_trie_nodes,
+        state_root_storage_node_count: trie_node_counts.storage_trie_nodes,
```
wlawt, line 445: should we also emit these as metrics to datadog?

crates/client/metering/src/meter.rs
```diff
@@ -449,6 +557,8 @@ mod tests {
         let result = &output.results[0];
         assert!(output.total_time_us > 0);
         assert!(output.state_root_time_us > 0);
+        assert!(output.state_root_account_node_count > 0);
+        assert_eq!(output.state_root_storage_node_count, 0);
```
wlawt, line 563: wonder if it makes sense to add a unit test where both counts are non-zero ? (just for sanity testing)

niran: good catch, done

crates/client/metering/src/meter.rs
niran: Yeah, emitting these in the builder during payload building would probably be valuable

wlawt (approved):
LGTM - might need to rerun the CI Test due to flakey-ness

PR #1771, opened 2026-03-29
fix(proposer): propagate RPC errors in recovery scan to prevent cache poisoning

0x00101010:
## Summary

- **Bug**: After 1 successful on-chain proposal, the proposer gets permanently stuck and never submits another proposal.
- **Root cause**: `scan_range_for_recovery()` silently swallowed `game_at_index` and `game_info` RPC errors, treating them as "no match." When `recover_latest_state()` received `Ok(None)` from the incremental scan, it advanced the cache watermark (`cached.game_count = count`) while keeping stale state. Future ticks matched `game_count == cached.game_count`, returned the stale cached state, and the pipeline never advanced.
- **Fix**: Propagate RPC errors as `Err` instead of silently skipping, so `recover_latest_state()` never reaches the watermark-advance code path. The next tick retries with the cache intact.

## Stuck sequence (before fix)

1. Submit proposal for block A → success → cache set to `{gc:1, state:block_A}`
2. Submit proposal for block B → success → `game_count` increases to 2
3. `recover_latest_state` → gc=2 > cached gc=1 → incremental scan of 1 new game
4. `game_info` RPC call fails transiently → scan returns `Ok(None)` 
5. Watermark advances to gc=2, cached state remains block_A
6. Every subsequent tick: gc=2 == cached gc=2 → cache hit → stale block_A forever
7. Re-dispatches block B → `GameAlreadyExists` → never advances

### `pipeline.rs`
- `game_at_index` errors: was `filter_map` returning `None` (silent skip), now `warn!` + `return Err(e.into())`
- `game_info` errors: was `continue` (silent skip), now `warn!` + `return Err(e.into())`
- 3 new tests:
  - `test_scan_range_returns_err_on_game_at_index_failure`
  - `test_scan_range_returns_err_on_game_info_failure`
  - `test_recovery_incremental_rpc_error_preserves_cache`

### `test_utils.rs`
- Added `failing_addresses: HashSet<Address>` to `MockAggregateVerifier` for error injection

## Notes

- `output_at_block` RPC errors were already correctly handled (`return Err`) — no change needed
- Root mismatch still uses `continue` (intentional — skip invalid proposals, not an RPC issue)
- Pre-existing build failure in `cli.rs` (`clap` `Arg::env` method) is unrelated to this change

### DIFF
--- a/crates/proof/proposer/src/driver/pipeline.rs
+++ b/crates/proof/proposer/src/driver/pipeline.rs
@@ -638,36 +638,39 @@ where
         // Fetch game_at_index concurrently with order-preserving buffering.
         // Because indices are emitted highest-first, the first type-match is
         // the best — stop consuming the stream immediately.
-        let mut stream = std::pin::pin!(stream::iter(0..len)
-            .map(|i| {
-                let factory = &self.factory_client;
-                async move {
-                    let game_index = count - 1 - i;
-                    let result = factory.game_at_index(game_index).await;
-                    (game_index, result)
-                }
-            })
-            .buffered(RECOVERY_SCAN_CONCURRENCY)
-            .filter_map(|(idx, result)| async move {
-                match result {
-                    Ok(game) if game.game_type == game_type => Some((idx, game)),
-                    Ok(_) => None,
-                    Err(e) => {
-                        warn!(error = %e, game_index = idx, "Failed to read game at index during recovery");
-                        None
+        //
+        // RPC failures are propagated as errors rather than silently skipped,
+        // so the caller does not advance the cache watermark past games that
+        // were never successfully read.
+        let mut stream = std::pin::pin!(
+            stream::iter(0..len)
+                .map(|i| {
+                    let factory = &self.factory_client;
+                    async move {
+                        let game_index = count - 1 - i;
+                        let result = factory.game_at_index(game_index).await;
+                        (game_index, result)
                     }
-                }
-            }));
+                })
+                .buffered(RECOVERY_SCAN_CONCURRENCY)
+        );
 
-        // Try each matching game in order (most-recent first). If
-        // `game_info` fails for one match, fall through to the next rather
-        // than giving up entirely.
-        while let Some((game_index, game)) = stream.next().await {
+        // Try each game in order (most-recent first).
+        while let Some((game_index, result)) = stream.next().await {
+            let game = match result {
+                Ok(game) if game.game_type == game_type => game,
+                Ok(_) => continue,
+                Err(e) => {
+                    warn!(error = %e, game_index, "Failed to read game at index during recovery");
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proposer/src/driver/pipeline.rs
```diff
@@ -638,36 +638,39 @@ where
...
+
+        // Try each game in order (most-recent first).
+        while let Some((game_index, result)) = stream.next().await {
+            let game = match result {
+                Ok(game) if game.game_type == game_type => game,
+                Ok(_) => continue,
+                Err(e) => {
+                    warn!(error = %e, game_index, "Failed to read game at index during recovery");
+                    return Err(e.into());
                 }
```
leopoldjoy, line 666: A single transient `game_at_index` RPC failure for any index in the buffered window now aborts the entire scan; since up to `RECOVERY_SCAN_CONCURRENCY` calls are in-flight, one flaky index (e.g. a game of a different type whose index happens to hit a bad RPC node) will block recovery even though the proposer's own game type might be at a different index that succeeded — consider whether non-matching-type errors should also be fatal, or only errors for games that match `game_type`.

0x00101010: This is intended. Two things:

1. The scan window should be very small, we expect to find latest usable game normally within 1-2 scans.
2. This is correct behavior, if scan at the latest index failed, we shouldn't proceed and go to the next available one, this would cause proposer to start proving on a potentially created proposal and wasting even more compute cycles

leopoldjoy (approved):

PR #1823, opened 2026-03-31
test(builder): add build_block and FlashblocksExtraCtx unit tests

mw2000:
Adds four stateless unit tests using the `NoopProvider` + in-memory `State` infra from #1810:

- `build_block_empty_with_state_root` — `calculate_state_root = true` path
- `build_block_rejects_block_number_mismatch` — stale parent header guard
- `build_block_rejects_missing_beacon_block_root` — missing beacon root error
- `extra_ctx_next_advances_index_and_updates_targets` — index increment, target updates, field preservation through `..self`

PR #1830, opened 2026-03-31
test(builder): add flashblock context progression test

mw2000:
Stacked on #1823.

## Summary

- Add `extra_ctx_next_advances_index_and_updates_targets` — verifies that `FlashblocksExtraCtx::next()` increments the flashblock index, updates all per-batch target fields, and preserves the per-batch limit fields and target count

This catches the case where a new field is added to `FlashblocksExtraCtx` but not carried through the `..self` struct update in `next()`.

Standalone in-memory test with no external dependencies.

## Test plan

```
cargo test -p base-builder-core --lib
# 48 passed, 0 failed
cargo clippy -p base-builder-core --tests
# 0 warnings
```


mw2000:
Folded into #1823 — the single test didn't justify its own PR.

PR #1911, opened 2026-04-03
bench(builder): add state root criterion benchmarks

mw2000:
## Summary
- Add criterion benchmarks comparing two state root computation modes with 10 flashblocks at varying accounts-per-flashblock sizes (10/100/1000):
  - `finalize_only`: accumulate all changes, compute state root once at the end (current production behavior)
  - `per_flashblock`: compute state root after every flashblock (currently recomputes from scratch each time)
- `per_flashblock` is ~8-10x slower than `finalize_only` today — this establishes the baseline we'll optimize against

## Run
```
cargo bench --bench state_root -p base-builder-core
```


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/builder/core/src/flashblocks/payload.rs
```diff
@@ -915,6 +915,22 @@ where
...
+/// trie nodes from the previous flashblock via
+/// [`StateRootProvider::state_root_from_nodes_with_updates`].
+pub fn compute_state_root(
+    provider: &(impl StateRootProvider + ?Sized),
+    hashed_state: HashedPostState,
+    _prev_trie_updates: Option<&TrieUpdates>,
+) -> Result<(B256, TrieUpdates), ProviderError> {
+    provider.state_root_with_updates(hashed_state)
+}
+
```
niran: From Claude: `_prev_trie_updates` is accepted but silently ignored — the implementation always takes the cold `state_root_with_updates` path regardless. The doc comment above describes an incremental `state_root_from_nodes_with_updates` path that doesn't exist yet, which overpromises to callers.

Consider either deferring the parameter until the incremental path lands, or adding an explicit `// TODO` and adjusting the doc comment to say the incremental path is not yet wired up.

crates/builder/core/benches/state_root.rs
```diff
@@ -0,0 +1,143 @@
...
+    let mut g = c.benchmark_group("state_root/per_flashblock");
+    g.sample_size(10);
+
+    for &accounts_per_fb in &[10, 100, 1_000] {
+        let (deltas, _full_state) = setup_flashblock_data(accounts_per_fb);
+
+        g.bench_function(BenchmarkId::new("accounts_per_fb", accounts_per_fb), |b| {
+            b.iter(|| {
+                let storage = OpProofsStorage::from(InMemoryProofsStorage::new());
+                let provider =
```
niran: From Claude: The `OpProofsStorage` and provider are created inside `b.iter()` here, but in `finalize_only_benches` they're created outside it. That means this benchmark includes provider allocation overhead that the other one doesn't, so the 5.4x comparison isn't apples-to-apples.

If the intent is to measure pure state-root computation cost, consider moving the provider setup outside the measured loop (or using `iter_batched`). If the allocation is intentionally included, a comment explaining why would help.

niran:
I don't have a solution for this yet, but I think these benchmarks will be misleading without I/O costs incorporated. We want to minimize the time between flashblock 10 being built and the state root being calculated, and that's almost entirely an I/O problem.

niran:
I just came across flashbots/op-rbuilder#445, which I assume is the inspiration for this. I don't fully have my head wrapped around it, but they're handling the state differently. From Claude:

The key difference is the database layer. This benchmark uses `NoopProvider` + empty `InMemoryProofsStorage`, so the trie is built from scratch with only the generated accounts — no underlying state to traverse. op-rbuilder's benchmark pre-populates a real MDBX-backed database with 50k accounts and 10 storage slots each:

```rust
let provider_factory = create_test_provider_factory_with_chain_spec(MAINNET.clone());
let provider_rw = provider_factory.provider_rw().unwrap();
provider_rw.insert_account_for_hashing(accounts_iter).unwrap();
provider_rw.insert_storage_for_hashing(storage_entries).unwrap();
```

Then queries go through `LatestStateProvider` backed by that real DB. This matters because:
- In production, state root computation is dominated by reading existing trie nodes from storage, not just hashing new accounts
- Trie depth with 50k+ accounts is meaningfully deeper than a fresh trie
- The overlay cursor merge path (interleaving new changes with existing state) doesn't get exercised against empty storage

We have the infrastructure for this already — `create_test_provider_factory_with_chain_spec`, `insert_account_for_hashing`, and `LatestStateProvider` are all used in `crates/execution/trie/tests/live.rs`. Wiring them in here would make the benchmark numbers more representative of production behavior.

mw2000:
Oh nice I actually hadn't come across [flashbots/op-rbuilder#445](https://github.com/flashbots/op-rbuilder/pull/445)

Agreed on the current benchmark not taking into account I/O, will rework this 🫡

crates/builder/core/src/flashblocks/payload.rs
mw2000: Removed this, will be adding it back in a future PR when we introduce intermediate state root caching

crates/builder/core/src/flashblocks/payload.rs
```diff
@@ -915,6 +915,20 @@ where
...
+/// Computes the state root over `hashed_state` using `provider`.
+///
+/// Performs a full state root computation via
+/// [`StateRootProvider::state_root_with_updates`], returning both the root
+/// hash and the [`TrieUpdates`] produced during the walk.
+pub fn compute_state_root(
+    provider: &(impl StateRootProvider + ?Sized),
+    hashed_state: HashedPostState,
+) -> Result<(B256, TrieUpdates), ProviderError> {
+    provider.state_root_with_updates(hashed_state)
```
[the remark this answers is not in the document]
mw2000: I think this is fine for now

mw2000:
1. Should be fine since we will extend this function when we add incremental state root

2. Is fine since we haven't added incremental state root logic yet

niran (approved):

PR #1971, opened 2026-04-06
[bcp]: BCP-0001 BFT Sequencer Consensus

refcell:
## Summary

This PR introduces BCP-0001, which proposes replacing op-conductor with a purpose-built BFT consensus engine embedded directly in the sequencer node using commonware's threshold_simplex protocol. The new design provides Byzantine-fault-tolerant leader election, VRF-based view rotation, and BLS12-381 threshold signature attestation for unsafe blocks. It also adds the BCP-0001 entry to the BCPs list component in the specs site.

### DIFF
--- a/docs/specs/pages/bcps/bcp-0001.md
+++ b/docs/specs/pages/bcps/bcp-0001.md
@@ -1,0 +1,1009 @@
+# BCP-0001: Byzantine-Fault-Tolerant Sequencer Consensus via Commonware Simplex
+
+## Abstract
+
+op-conductor provides sequencer high availability through Raft consensus: a Go
+process co-deployed beside each `base-consensus` instance, operating a
+distributed state machine that tracks which node is the active sequencer and
+records the latest unsafe block hash. Raft is crash-fault-tolerant (CFT) — it tolerates node
+failures, not Byzantine behavior. It runs as a separate process, making it a
+single point of failure for the very failover it is meant to provide. And it
+injects a `CommitUnsafePayload` call directly into the block production
+critical path, where a degraded Raft cluster stalls block production for up to
+30 seconds per block.
+
+This BCP replaces op-conductor with a purpose-built consensus engine embedded
+in the sequencer node, using commonware's `threshold_simplex` protocol. The new
+engine provides Byzantine-fault-tolerant (BFT) leader election, VRF-based view
+rotation that is not predictable or manipulable by any participant, certified
+unsafe block attestation backed by BLS12-381 threshold signatures, and
+in-process operation that removes inter-process coordination overhead from the
+block production critical path entirely.
+
+---
+
+## 1. Status Quo
+
+### 1.1 Architecture
+
+Each sequencer deployment currently runs three processes per physical node:
+`base-reth-node` (execution layer), `base-consensus` (consensus and
+derivation), and `op-conductor` (sequencer lifecycle management). There is an
+ongoing effort to collapse `base-reth-node` and `base-consensus` into a single
+process. This BCP addresses the third process: `op-conductor`.
+
+op-conductor performs a narrow but critical function: it answers one distributed
+question — which of the N sequencer nodes is the currently active sequencer —
+and enforces that answer by starting and stopping sequencing on the local
+`base-consensus` node accordingly. It does this through Raft consensus, with
+one `op-conductor` instance per node forming a Raft cluster of typically three
+members.
+
+The integration between `op-conductor` and `base-consensus` is bidirectional
+and tightly coupled:
+
+- After sealing each block, `base-consensus` calls `conductor.commit_unsafe_payload()`
+  on its local conductor client with the sealed execution payload envelope. The
+  conductor propagates this to the Raft log with a 30-second timeout. If the
+  Raft cluster is degraded, this call blocks for the full 30 seconds before
+  failing — stalling block production for the duration.
+- On bootstrap, `base-consensus` calls `conductor.leader()` to determine whether
+  to start in `ActiveSequencer` or `ConductorFollower` mode. A non-leader result
+  causes the node to enter follower mode, running derivation but not building
+  blocks.
+- When a conductor detects that its node has become unhealthy (via its
+  `SequencerHealthMonitor`), it stops the local sequencer and calls
+  `TransferLeader()` on its Raft instance, triggering a new election. The newly
+  elected conductor starts the sequencer on its node.
+
+### 1.2 The Raft Consensus Core
+
+The conductor's Raft implementation uses `hashicorp/raft` backed by BoltDB for
+log and stable storage. The Raft finite state machine (`unsafeHeadTracker`)
+maintains exactly one piece of state: the latest committed
+`ExecutionPayloadEnvelope`. When a new leader starts sequencing, it calls
+`LatestUnsafePayload()`, which issues a Raft barrier (a linear consistency
+guarantee) and returns the most recently Raft-committed block. The new leader
+posts this block to its `base-consensus` node if needed to align the unsafe
+head before sequencing begins.
+
+The Raft FSM stores only the latest unsafe head; prior entries are compacted
+away through snapshots. Snapshot serialization is SSZ-encoded
+`ExecutionPayloadEnvelope`, with V4 and V3 fallback handling baked in to
+accommodate rolling upgrades. This creates a direct coupling between the Raft
+log format and the execution payload wire format — a coupling that has caused
+production incidents (issue #9064, Dencun upgrade) and requires
+format-version awareness in persistence code.
+
+### 1.3 Health Monitoring and the 8-State FSM
+
+The conductor's control loop evaluates three boolean dimensions —
+`leader × healthy × active` — producing eight possible states. In the
+healthy-leader-active state (normal operation), no action is taken. Any
+deviation triggers an action: stop sequencer, transfer leadership, start
+sequencer, or a combination.
+
+Health is determined by `SequencerHealthMonitor`, which checks:
+
+- Unsafe head recency (last unsafe block timestamp within configured threshold)
+- P2P peer count (above minimum)
+- Optional safe head progression
+- Optional EL P2P peer count
+- Optional rollup-boost health (with partial-health tolerance counters)
+
+Health check results are delivered to the control loop via a channel. The
+control loop dequeues actions with jittered backoff (0–2 seconds random sleep).
+This jitter prevents thundering-herd leadership elections but introduces
+non-determinism in failover timing.
+
+### 1.4 Cluster Membership
+
+Membership is managed manually through Raft's configuration API. Adding a node
+requires the operator to call `conductor_addServerAsVoter` from the current
+leader — there is no automatic discovery or join mechanism. Removing a node
+requires `conductor_removeServer`. All membership operations require passing the
+current Raft configuration version for optimistic concurrency control, and all
+require that the caller is the current Raft leader.
+
+Upgrades require pausing each node's conductor, waiting for unsafe head sync,
+then resuming. Leadership must be manually transferred away from any node being
+upgraded. The `pause` state is not persisted across restarts, which has caused
+production incidents where a newly restarted node incorrectly marked itself
+healthy before syncing its unsafe head (issue #12912).
+
+---
+
+## 2. Failure Analysis — Why op-conductor Falls Short
+
+### 2.1 Crash Fault Tolerance Is Insufficient for Decentralized Sequencing
+
+Raft is a CFT protocol. It tolerates the failure of nodes that crash but
+assumes all responding nodes are honest. In a permissioned three-node cluster
+operated by a single entity, this assumption holds. As the sequencer set moves
+toward decentralization — multiple independent operators — it breaks down. A
+single Byzantine node in a three-node Raft cluster can:
+
+- Delay elections indefinitely by responding but returning incorrect state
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

niran:
I talked through this proposal with Claude. Here are my notes.

**The value of consensus-based HA is worth questioning.** This isn't something the proposal needs to address, but something we should think about as a team. My sense is that op-conductor made individual recoveries faster while making them far more frequent — health checks (particularly with rollup-boost) triggered failovers that wouldn't have been necessary otherwise, and the workloads that degrade one node tend to degrade all of them, making failover to another node relatively useless. If the dominant failure mode is correlated degradation rather than isolated node failure, the complexity cost of any consensus-based HA system should be weighed against that.

**Calibrating the decentralization value.** Op-conductor is out-of-protocol — invisible to the derivation pipeline, not part of the OP Stack specs. This proposal replaces it with a BFT consensus engine, but it's still out-of-protocol. The threshold certificates aren't posted to L1 or consumed by derivation. The protocol still sees a single sequencer address. This is closer to Ethereum's DVT effort: better fault tolerance for a single logical sequencer role, not a step toward multiple independent sequencers in the protocol. I'd discount the decentralization framing in sections 2.1 and 8 when evaluating this.

**Do the cited failures motivate BFT?** The issues in the failure analysis (#9064, #12448, #12912, #17835) appear to be architectural and implementation problems — serialization versioning, EL/CL interaction bugs, resource contention — not Byzantine fault scenarios. Every documented conductor incident I'm aware of is a liveness failure (chain halts), not a safety failure (invalid blocks accepted by peers). Are there incidents I'm missing where Byzantine tolerance would have helped?

**Invalid block propagation is already prevented without BFT.** Gossip happens after Raft commit, every node validates blocks via `new_payload` before adopting them as the unsafe head, and L1 derivation is the ultimate safety backstop. A Byzantine leader's invalid block gets rejected by every receiving node's EL. Under the current system, health checks transfer leadership in ~1-2 seconds and the new leader backfills. Under BFT with tenure-based rotation, the same scenario produces nullified views for the rest of the tenure with no mid-tenure eviction — wouldn't that be strictly worse on liveness with identical safety?

**No-skip-slots semantics.** The derivation pipeline expects continuous blocks — every L2 timestamp slot gets a block. Today the new leader backfills any gap. Simplex's nullify mechanism skips the view entirely. Adapting threshold_simplex to produce backfilled blocks deviates from the library's expected usage — how much customization is needed here, and does that erode the value of using an off-the-shelf protocol?

**Do the architectural improvements require this change?** Embedding consensus in-process, removing CommitUnsafePayload from the critical path, and decoupling the log format are all changes that could be made to the existing Raft-based system. But it's also worth scrutinizing whether they're needed at all:

- *Sidecar SPOF (section 2.2):* The proposal claims base-consensus "loses conductor governance and continues sequencing independently" when the conductor crashes. But both the Rust and Go paths gate block progression on the conductor commit — doesn't a dead conductor stall sequencing rather than cause split-brain? Has split-brain from a conductor crash actually occurred?
- *CommitUnsafePayload latency (section 2.3):* The effective timeout appears to be `ConductorRpcTimeout` (1 second), not the 30-second parent context, and Raft Apply during elections fails immediately with `ErrNotLeader`. Could the Rust seal pipeline be enhanced to pipeline the commit with the next block's build without replacing the consensus system?
- *Payload format coupling (section 2.7):* The proposal frames storing payloads in the Raft log as a liability, but the full payload is there for a reason — during leadership transfer, the new leader's EL may not have the latest block, and the Raft log provides it directly. The proposed design stores only 32-byte hashes and relies on the Resolver to backfill from peers, adding a network dependency at a latency-sensitive moment.

---

**Edit:** Added the point below, which I think is the most practically important concern.

**Proposer health and fast failover.** The proposal's tenure-based rotation nullifies views when the leader fails to propose, retrying the same leader until the tenure boundary. If the leader goes unhealthy mid-tenure with K=100 and 80 views remaining, that's up to 40 seconds of missed blocks. Off-the-shelf BFT protocols are designed to tolerate missed turns, but we want to actively route to a healthy node — during crashes, deploys, or degraded performance. Today, op-conductor's health monitor transfers leadership in ~1-2 seconds and the new leader backfills. How would this design handle a routine rolling deploy without extended block production gaps?

Sertug17:
Per BCP-0000's own format requirement, each BCP must include a metadata table after the H1 heading. BCP-0001 is missing this table. Happy to add it if helpful.

docs/specs/pages/bcps/bcp-0001.md
```diff
@@ -0,0 +1,1009 @@
...
+Raft's leader election is not random: the first node to time out starts an
+election, and the outcome is determined by network timing and term numbers. In
+a three-node cluster with static configuration, the leader is predictable to
+any observer who knows the cluster's current term and heartbeat timeout.
+`TransferLeader()` uses Raft's built-in transfer, which selects "the most
+up-to-date follower" — a fully deterministic criterion. A strategic actor who
+knows which node will win the next election can position orders accordingly.
+Even if this is not currently being exploited, it becomes more significant as
+the sequencer moves toward decentralization and higher-value transaction
+ordering.
```
BrianBland, line 183: Predictable leader election not only enables MEV extraction, but also reduces the effectiveness of failover during cluster degradation. In a 5 node cluster with 2 partially-degraded sequencers, these two may take turns sequencing if they happen to have the highest priority in the raft ordering.
This particular issue could be addressed via an alternative selection mechanism, such as round robin.

docs/specs/pages/bcps/bcp-0001.md
```diff
@@ -0,0 +1,1009 @@
...
+
+---
+
+## 4. Proposal — Embedded Commonware Threshold Simplex Consensus
+
+### 4.1 Core Design Principles
+
+1. **In-process, not a sidecar.** The consensus engine is a crate embedded in
+   the sequencer node binary. There is no separate process to manage, no
+   inter-process RPC for the critical-path `Leader()` and
```
BrianBland, line 287: "no inter-process RPC" actually makes a material performance difference when handling multi-MB payload serialization/deserialization

docs/specs/pages/bcps/bcp-0001.md
```diff
@@ -0,0 +1,1009 @@
...
+
+3. **Off the critical path.** Block production does not wait for consensus.
+   The consensus engine certifies blocks asynchronously; the sequencer builds
+   the next block while vote collection for the current block proceeds
+   concurrently.
+
+4. **VRF-based leader rotation.** The active sequencer for view `v+1` is
+   determined by the threshold VRF embedded in view `v`'s consensus certificate
+   — knowable only after view `v` concludes, unpredictable before it, and not
+   manipulable by any coalition of fewer than `f+1` participants.
```
BrianBland, line 302: Recommend presenting this as a future extension point, as this is not strictly beneficial in a centrally sequenced cluster.

docs/specs/pages/bcps/bcp-0001.md
```diff
@@ -0,0 +1,1009 @@
...
+
+5. **Threshold signatures for certified blocks.** An unsafe block is certified
+   when `2f+1` validators have co-signed a notarization certificate over its
+   hash. The certificate is compact (~240 bytes, a single BLS12-381 signature),
+   externally verifiable by anyone holding the threshold public key, and can
+   serve as a cross-chain inclusion receipt for bridges and light clients.
```
BrianBland, line 308: Depending on committee size, it might actually be better to use N independent ed25519 signatures vs one aggregated BLS12-381 signature. I believe that this is the case for N<100, and the signature size is comparable for N<5

docs/specs/pages/bcps/bcp-0001.md
```diff
@@ -0,0 +1,1009 @@
...
+│     ├── threshold_simplex::Engine                                        │
+│     │     ├── Voter          — Simplex protocol FSM                      │
+│     │     ├── Batcher        — BLS12-381 lazy signature aggregation      │
+│     │     └── Resolver       — on-demand cert/payload backfill           │
+│     ├── SequencerAutomaton (implements CertifiableAutomaton)             │
+│     │     ├── propose()  → sends ProposeBlock to SequencerActor,         │
+│     │     │               returns sealed block hash                      │
+│     │     ├── verify()   → delivers peer ExecutionPayload to EL          │
+│     │     │               via new_payload, returns EL result             │
+│     │     └── certify()  → approves notarization after 2f+1 votes       │
```
BrianBland, line 414: Nit: alignment

docs/specs/pages/bcps/bcp-0001.md
```diff
@@ -0,0 +1,1009 @@
...
+4. `CertifiedBlock` events arrive asynchronously, consumed by the Batcher and
+   Proposer.
+5. Block production for view `v+1` begins immediately after `propose(v)`
+   returns — it does not wait for notarization.
+
+A view timeout (`leader_timeout`, suggested 500ms) fires only if `2f+1` votes
+are not collected in time. In that case, the engine emits a `Nullify` and the
+elector picks a new leader for the next view. The previous leader yields;
+the new leader begins block production. The view timeout is the *only* path
+by which slow consensus affects block production — and it requires a majority
```
BrianBland, line 748: Nit: this requires a superminority to be unresponsive

docs/specs/pages/bcps/bcp-0001.md
```diff
@@ -0,0 +1,1009 @@
...
+2. `propose()` triggers block production and returns the block hash when
+   sealing completes.
+3. Vote collection (`2f+1` notarize votes) happens concurrently while the next
+   block is already being built.
+4. `CertifiedBlock` events arrive asynchronously, consumed by the Batcher and
+   Proposer.
+5. Block production for view `v+1` begins immediately after `propose(v)`
+   returns — it does not wait for notarization.
+
+A view timeout (`leader_timeout`, suggested 500ms) fires only if `2f+1` votes
```
BrianBland, line 744: Should this be defined as a proportion of the block time, e.g. `block_time / 4`?

docs/specs/pages/bcps/bcp-0001.md
```diff
@@ -0,0 +1,1009 @@
...
+
+**Phase 3 — Full replacement.** op-conductor is retired. The
+`sequencer-consensus` engine has sole authority over sequencer lifecycle. Health
+monitoring is absorbed into the node's `HealthBus`. The conductor's proxy
+RPC endpoints (`eth_`, `optimism_`, `admin_` namespaces) are replaced by direct
+routing in the node's RPC server. The `overrideLeader` disaster recovery
+mechanism is replaced by `basectl conductor override`, which sets an in-process
+flag with the same semantics as op-conductor's `OverrideLeader(true)` — the
+escape hatch for operating a single-node sequencer when the cluster is fully
+unavailable.
```
BrianBland, line 864: And just to clarify, this would be a special block production mode without BFT safety.
The single sequencer would be unable to certify blocks as it would lack sufficient key material, which means that other nodes may treat these as unsafe/preconfirmations only.

We would also be unable to batch these transactions to L1 unless we explicitly signal a fallback mechanism on the L1 contracts.

docs/specs/pages/bcps/bcp-0001.md
```diff
@@ -0,0 +1,1009 @@
...
+  removed entirely.
+
+- **BFT membership changes.** Adding or removing validators requires `2f+1`
+  agreement. No single operator can unilaterally alter cluster composition.
+
+- **Resharing without key rotation.** Validator set changes preserve the
+  threshold public key, enabling light clients and bridges to hold a stable
+  long-term verification key across all sequencer set changes.
+
+**Sacrificed:**
```
BrianBland, line 915: We also sacrifice some latency for unsafe head commitments. While before we could generally trust that the unsafe head could not be reorged, this certification will be arrive after a delay.

BrianBland (commented):
My overall recommendation would be to focus this BCP on enhancing the sequencer configuration of the chain, while flagging the potential for future expansion into decentralized block production.

We may find that BLS12-381 and VRF are both overkill from the sequencer lens, but these may more valuable with larger / more diverse sets of block producers.

One area that we still need to further investigate is disaster recovery. In the case where we lose quorum, how can we continue to operate the chain with a reduced level of preconfirmation safety while retaining finality via L1 settlement?

PR #1973, opened 2026-04-06
feat(challenger): add continuous bond discovery scanning

jackchuma:
## Summary

- Adds `discover_claimable_games` to `BondManager` with two-tier scanning: **incremental** scans on every driver tick to catch newly created games, and **periodic full rescans** (every 5 minutes) of the lookback window to detect state transitions (games challenged or resolved by other actors).
- Refactors `startup_scan` and the new discovery method to share game evaluation logic via `evaluate_game_for_bonds` / `evaluate_bond_range`, eliminating code duplication.
- Stores an `Arc<dyn DisputeGameFactoryClient>` and scan watermark state inside `BondManager` so it can independently query the factory without external arguments on each tick.

## Changes

- **`bond.rs`**: New `discover_claimable_games` method, extracted `evaluate_game_for_bonds` and `evaluate_bond_range` helpers, simplified `startup_scan` signature, added `factory_client`/`bond_scan_head`/`last_full_scan`/`lookback` fields, custom `Debug` impl.
- **`driver.rs`**: Calls `discover_claimable_bonds()` in `step()` before `poll_bond_claims()`.
- **`metrics.rs`**: New `bond_discovery_scans_total` (labeled by scan type) and `bond_discovery_games_found_total` counters.
- **`service.rs`**: Passes `factory_client` and `lookback` to `BondManager::new`, simplifies `startup_scan` call.
- **`test_utils.rs`**: Adds `empty_factory()` helper.
- **`tests/driver.rs`**: Updates all `BondManager::new` call sites with the new signature.
- **`Cargo.toml`**: Enables `derive` and `env` features for `clap` at the crate level.

### DIFF
--- a/crates/proof/challenge/src/bond.rs
+++ b/crates/proof/challenge/src/bond.rs
@@ -328,7 +328,7 @@ impl BondManager {
     }
 
     /// Evaluates all games in `range` concurrently for bond tracking
-    /// eligibility, returning one entry per index.
+    /// eligibility, returning one entry per evaluated game.
     async fn evaluate_bond_range(
         range: std::ops::Range<u64>,
         factory_client: &Arc<dyn DisputeGameFactoryClient>,
--- a/crates/proof/challenge/src/bond.rs
+++ b/crates/proof/challenge/src/bond.rs
@@ -28,7 +28,7 @@
 use std::{
     collections::{HashMap, HashSet},
     sync::Arc,
-    time::{Duration, Instant, SystemTime},
+    time::{Duration, SystemTime},
 };
 
 use alloy_primitives::Address;
@@ -104,6 +104,7 @@ pub struct TrackedGame {
 /// discovers claimable games via [`discover_claimable_games`](Self::discover_claimable_games),
 /// scanning both newly created games and periodically rescanning the
 /// lookback window to catch games challenged or resolved by other actors.
+#[derive(derive_more::Debug)]
 pub struct BondManager {
     /// Games being tracked, keyed by proxy address.
     tracked: HashMap<Address, TrackedGame>,
@@ -118,29 +119,18 @@ pub struct BondManager {
     /// Injectable clock for obtaining the current Unix timestamp. Defaults
     /// to [`unix_now`] in production; tests can substitute a controlled
     /// clock.
+    #[debug(skip)]
     clock: ClockFn,
     /// Factory client for querying game indices during bond discovery.
+    #[debug(skip)]
     factory_client: Arc<dyn DisputeGameFactoryClient>,
     /// Highest game index scanned for bond discovery. Incremental scans
     /// start from this index; periodic full rescans reset it backward.
     bond_scan_head: u64,
-    /// When the last full rescan of the lookback window completed.
-    last_full_scan: Instant,
+    /// Unix timestamp (seconds) of the last full rescan completion.
+    last_full_scan: u64,
     /// Number of games to look back during periodic full rescans.
     lookback: u64,
-}
-
-impl std::fmt::Debug for BondManager {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("BondManager")
-            .field("tracked", &self.tracked)
-            .field("claim_addresses", &self.claim_addresses)
-            .field("weth_delay", &self.weth_delay)
-            .field("l1_rpc_url", &self.l1_rpc_url)
-            .field("bond_scan_head", &self.bond_scan_head)
-            .field("lookback", &self.lookback)
-            .finish_non_exhaustive()
-    }
 }
 
 impl BondManager {
@@ -170,7 +160,7 @@ impl BondManager {
             clock: unix_now,
             factory_client,
             bond_scan_head: 0,
-            last_full_scan: Instant::now(),
+            last_full_scan: unix_now(),
             lookback,
         }
     }
@@ -391,15 +381,18 @@ impl BondManager {
         // Process results sequentially: insert tracked games and resolve the
         // WETH delay from the first relevant game.
         for (game_address, bond_recipient, phase) in results.into_iter().flatten() {
-            let Some(phase) = phase else {
-                continue;
-            };
-
+            // Resolve the WETH delay from the first available game,
+            // including already-claimed ones, so that the delay is
+            // bootstrapped as early as possible.
             if self.weth_delay.is_none()
                 && let Err(e) = self.resolve_weth_delay(verifier_client, game_address).await
             {
                 warn!(error = %e, "failed to read DelayedWETH delay, will retry later");
             }
+
+            let Some(phase) = phase else {
+                continue; // already claimed, skip
+            };
 
             info!(
                 game = %game_address,
@@ -413,7 +406,7 @@ impl BondManager {
         // Set the discovery watermark so continuous scanning starts from
         // where startup left off.
         self.bond_scan_head = game_count;
-        self.last_full_scan = Instant::now();
+        self.last_full_scan = (self.clock)();
 
         ChallengerMetrics::bonds_tracked().set(self.tracked.len() as f64);
         info!(count = self.tracked.len(), "bond startup scan complete");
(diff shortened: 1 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/challenge/src/bond.rs
```diff
@@ -132,6 +168,10 @@ impl BondManager {
             weth_delay: None,
             l1_rpc_url,
             clock: unix_now,
+            factory_client,
+            bond_scan_head: 0,
+            last_full_scan: Instant::now(),
```
leopoldjoy: `Instant::now()` bypasses the existing injectable `ClockFn` abstraction. `BondManager` already has a `ClockFn` field for testable time, but `last_full_scan` uses `Instant::now()` directly, making the rescan interval non-controllable through the same mechanism; tests work around this via direct field mutation (`mgr.last_full_scan = Instant::now() - ...`) which is fragile and inconsistent with the `set_clock` pattern used elsewhere in this struct.

crates/proof/challenge/src/service.rs
```diff
@@ -143,11 +144,19 @@ impl ChallengerService {
...
+                l1_rpc_url,
+                Arc::clone(&factory_client) as Arc<dyn DisputeGameFactoryClient>,
+                config.lookback_games,
+            );
             info!("starting bond recovery scan");
-            if let Err(e) =
-                bm.startup_scan(&*factory_client, &*verifier_client, config.lookback_games).await
-            {
+            if let Err(e) = bm.startup_scan(&*verifier_client).await {
+                // On failure `bond_scan_head` stays at 0, so the first
```
leopoldjoy, line 155: Unbounded full-factory scan after `startup_scan` failure is a latent operational risk. If `startup_scan` fails, `bond_scan_head` stays at 0, so the first `discover_claimable_games` tick fires `evaluate_bond_range(0..game_count)` as a single unbounded batch with `SCAN_CONCURRENCY` parallelism — for factories with tens of thousands of games this could saturate RPC rate limits or cause a timeout. Consider capping the incremental scan span or batching.

crates/proof/challenge/src/bond.rs
```diff
@@ -114,6 +119,28 @@ pub struct BondManager {
...
+impl std::fmt::Debug for BondManager {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("BondManager")
+            .field("tracked", &self.tracked)
+            .field("claim_addresses", &self.claim_addresses)
+            .field("weth_delay", &self.weth_delay)
+            .field("l1_rpc_url", &self.l1_rpc_url)
+            .field("bond_scan_head", &self.bond_scan_head)
+            .field("lookback", &self.lookback)
+            .finish_non_exhaustive()
```
leopoldjoy: `last_full_scan` is omitted from this `Debug` output. The impl includes `bond_scan_head` and `lookback` but omits `last_full_scan`, which would be the most useful field for diagnosing why a full rescan did or didn't fire in production logs.

crates/proof/challenge/src/bond.rs
```diff
@@ -122,8 +149,17 @@ impl BondManager {
     /// succeed earlier; if longer, the attempt reverts and is retried.
     const DEFAULT_WETH_DELAY: Duration = Duration::from_secs(7 * 24 * 60 * 60);
 
+    /// How often a full rescan of the lookback window is performed to catch
+    /// state transitions (games challenged or resolved by other actors).
+    const BOND_DISCOVERY_INTERVAL: Duration = Duration::from_secs(300);
```
leopoldjoy, line 144: The 300-second rescan interval is a hardcoded constant with no CLI override. Unlike `lookback` which is configurable via `config.lookback_games`, the full-rescan interval is baked in; if operators need to tune the tradeoff between RPC cost and discovery latency (e.g., during incident response), they currently cannot without a code change.

crates/proof/challenge/src/bond.rs
```diff
@@ -211,128 +358,49 @@ impl BondManager {
...
-                    weth_delay_resolved = true;
-                }
-            }
-
             let Some(phase) = phase else {
-                continue; // already claimed, skip
+                continue;
             };
 
+            if self.weth_delay.is_none()
```
leopoldjoy, line 387: Subtle behavioral change from the original `startup_scan`: WETH delay is no longer resolved from already-claimed games. The old code attempted `resolve_weth_delay` before the `let Some(phase) = phase else { continue }` guard, so even a fully-claimed game could bootstrap the WETH delay; the new code skips claimed games first, deferring resolution if the first N results are all claimed. Not a bug since `poll()` resolves it lazily, but worth noting as an intentional change.

crates/proof/challenge/src/bond.rs
refcell: We should be able to use `derive_more::Debug` and just skip fields we can't or don't want to output in the debug

refcell (dismissed):
One small change

crates/proof/challenge/src/service.rs
jackchuma: Will fix this in a follow-up

crates/proof/challenge/src/bond.rs
jackchuma: Will fix this in a follow-up

crates/proof/challenge/src/bond.rs
```diff
@@ -200,6 +230,113 @@ impl BondManager {
...
+        verifier_client: &dyn AggregateVerifierClient,
+        claim_addresses: &HashSet<Address>,
+    ) -> Option<(Address, Address, Option<BondPhase>)> {
+        let game_at = match factory_client.game_at_index(index).await {
+            Ok(g) => g,
+            Err(e) => {
+                warn!(index, error = %e, "failed to fetch game at index");
+                return None;
+            }
+        };
```
refcell, line 252: Can be in a follow-on but would be nice to record metrics here for failure so we can graph the rate of failure, that could be useful to alert on if there's a spike in failure rates

crates/proof/challenge/src/bond.rs
```diff
@@ -200,6 +230,113 @@ impl BondManager {
...
+            verifier_client.zk_prover(game_address),
+        ) {
+            Ok(pair) => pair,
+            Err(e) => {
+                debug!(
+                    game = %game_address,
+                    error = %e,
+                    "failed to read bondRecipient/zkProver"
+                );
+                return None;
```
refcell, line 267: Same here for metrics

crates/proof/challenge/src/bond.rs
```diff
@@ -342,11 +403,127 @@ impl BondManager {
...
+    /// **Periodic full rescan** (every [`BOND_DISCOVERY_INTERVAL`](Self::BOND_DISCOVERY_INTERVAL)):
+    /// resets the watermark backward by `lookback` to re-evaluate games
+    /// whose state may have changed (e.g. challenged or resolved by
+    /// another actor since the last scan).
+    pub async fn discover_claimable_games(
+        &mut self,
+        verifier_client: &dyn AggregateVerifierClient,
+    ) -> eyre::Result<()> {
+        if !self.is_enabled() {
+            return Ok(());
```
refcell, line 432: Should probably warn here right?

crates/proof/challenge/src/bond.rs
```diff
@@ -342,11 +403,127 @@ impl BondManager {
...
+        }
+
+        let factory_client = Arc::clone(&self.factory_client);
+        let game_count = factory_client.game_count().await?;
+        if game_count == 0 {
+            return Ok(());
```
refcell, line 438: Probably worth a debug log if 0 game count

refcell (approved):
Nothing blocking. Logs and Metrics can be improved in future PRs

PR #1999, opened 2026-04-06
refactor(challenger): replace ClockFn with Clock trait from base-runtime

jackchuma:
## Summary

Replaces the ad-hoc `ClockFn = fn() -> u64` function pointer in `BondManager` with the `Clock` trait from `base-runtime`, consistent with the batcher's established pattern for time abstraction.

- Make `BondManager` generic over `C: Clock`, with `TokioRuntime` as the default on `DriverComponents` and `Driver` to minimize churn
- Change time fields (`last_full_scan`, `AwaitingDelay::unlocked_at`) from `u64` (Unix seconds) to monotonic `Duration`
- Add `unix_to_monotonic()` helper for converting onchain timestamps at recovery boundaries
- Remove `unix_now()`, `ClockFn` type alias, and `set_clock()` method
- Thread `clock: &C` through static methods (`evaluate_game_for_bonds`, `evaluate_bond_range`, `determine_phase`)
- Replace test helper `fixed_clock()` with a `FixedClock` struct implementing `Clock`

Resolves CHAIN-3898

### DIFF
--- a/crates/proof/challenge/src/bond.rs
+++ b/crates/proof/challenge/src/bond.rs
@@ -36,24 +36,11 @@ use base_proof_contracts::{
     AggregateVerifierClient, DelayedWETHClient, DelayedWETHContractClient,
     DisputeGameFactoryClient, encode_claim_credit_calldata, encode_resolve_calldata,
 };
+use base_runtime::Clock;
 use futures::stream::{self, StreamExt};
 use tracing::{debug, info, warn};
 
 use crate::{ChallengerMetrics, GameScanner};
-
-/// Returns the current Unix timestamp in seconds.
-fn unix_now() -> u64 {
-    SystemTime::now()
-        .duration_since(SystemTime::UNIX_EPOCH)
-        .expect("system clock before UNIX epoch")
-        .as_secs()
-}
-
-/// A function that returns the current Unix timestamp in seconds.
-///
-/// Used as an injectable clock in [`BondManager`] so that tests can
-/// control time without relying on the real wall clock.
-pub type ClockFn = fn() -> u64;
 
 /// Reason a game was removed from tracking after [`BondManager::advance_game`].
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -75,8 +62,8 @@ pub enum BondPhase {
     /// The unlock has been submitted; waiting for the `DelayedWETH` delay
     /// to elapse before the second `claimCredit()` call.
     AwaitingDelay {
-        /// Unix timestamp (seconds) at which the unlock occurred.
-        unlocked_at: u64,
+        /// Monotonic timestamp at which the unlock occurred.
+        unlocked_at: Duration,
     },
     /// The delay has elapsed; needs the second `claimCredit()` call to
     /// complete the withdrawal.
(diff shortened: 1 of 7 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/challenge/src/bond.rs
```diff
@@ -844,6 +832,35 @@ impl BondManager {
...
+    ///
+    /// `unix_now` is accepted as a parameter (rather than calling
+    /// `SystemTime::now()` internally) so that the function is fully
+    /// deterministic and testable.
+    ///
+    /// If `unix_secs` is ahead of `unix_now` (e.g. L1 clock skew),
+    /// the age is treated as zero and `unlocked_at` equals the current
+    /// monotonic time — re-imposing the full delay. This is the safe
+    /// conservative fallback.
+    fn unix_to_monotonic(clock: &C, unix_secs: u64, unix_now: u64) -> Duration {
```
leopoldjoy, line 851: Nit: when process uptime is shorter than the event's age (the common case right after a restart), `clock.now().saturating_sub(age)` clamps `unlocked_at` to `Duration::ZERO`, which means the full `DelayedWETH` delay is effectively re-imposed from process start — a behavioral regression from the previous Unix-timestamp approach, which correctly preserved elapsed delay progress across restarts. The doc comment documents the future-timestamp/clock-skew clamping, but this more impactful uptime-based clamping (tested in `unix_to_monotonic_age_exceeds_monotonic`) deserves an explicit note about its production consequence (up to `delay` seconds of extra waiting after every restart).

crates/proof/challenge/src/bond.rs
```diff
@@ -731,20 +718,22 @@ impl BondManager {
             // better than resetting to "now" (which would re-impose the full
             // delay after every restart).
             let resolved_at = verifier_client.resolved_at(game_address).await?;
+            let unlocked_at =
+                Self::unix_to_monotonic(&self.clock, resolved_at, Self::wall_clock_unix_secs());
```
leopoldjoy, line 722: Nit: the comment above (lines 715-719) says this approach is *"strictly better than resetting to 'now' (which would re-impose the full delay after every restart)"*, but after the switch to monotonic time, when the event age exceeds process uptime (the normal restart scenario), this code **does** effectively reset to process-start time — which is the very thing the comment says it avoids. The comment should be updated to reflect the new monotonic-time semantics.

leopoldjoy (approved):

PR #2026, opened 2026-04-08
perf(load-tests): parallelize funding/draining and add early balance validation

haardikk21:
## Summary

- **Early balance validation**: Before sending any funding TXs, compute the total cost (deficits + gas for all accounts) and compare against the funder balance. Returns a clear error with balance/needed/shortfall in ETH instead of silently failing mid-way and leaving stranded funds that can't be drained.
- **Progress bars**: Added `indicatif` progress bars for every phase — balance checks, TX sending, TX confirmation, draining, and post-funding refresh — so users can see real-time progress instead of staring at a frozen terminal.
- **Parallel funding**: Pre-assign nonces and send all funding TXs concurrently via `buffer_unordered(32)` instead of sequential send-and-await. Also parallelized the initial balance/nonce queries and post-funding state refresh.
- **Parallel draining**: Each account has its own signer so drain TXs are fully independent — now sent concurrently via `buffer_unordered(32)`.
- **Parallel confirmation polling**: Receipt lookups within each poll cycle now use `join_all` instead of sequential iteration.

## Motivation

With a large number of senders, the load tester funding process was extremely slow and appeared stuck with zero progress feedback. Three specific issues:
1. If the funder didn't have enough balance, it would fund as many accounts as possible then fail mid-way, leaving funded accounts stranded (drain logic never runs on error)
2. No progress reporting during funding — impossible to tell if stuck or working
3. Sequential TX submission and drain — O(n) round trips to the RPC for n accounts

### DIFF
--- a/crates/infra/load-tests/src/runner/load_runner.rs
+++ b/crates/infra/load-tests/src/runner/load_runner.rs
@@ -26,6 +26,11 @@ use tracing::{debug, error, info, instrument, warn};
 /// Maximum number of concurrent RPC requests during funding/draining operations.
 const FUNDING_CONCURRENCY: usize = 32;
 
+/// Maximum number of funding TXs to send before waiting for confirmation.
+/// Kept below typical per-sender txpool limits (e.g. reth default is 16) to
+/// avoid "txpool is full" rejections when all TXs originate from one funder.
+const FUNDING_BATCH_SIZE: usize = 16;
+
 use super::{
     AdaptiveBackoff, Confirmer, ConfirmerHandle, DisplaySnapshot, LoadConfig, LoadTestDisplay,
     RateLimiter, TxType,
@@ -337,7 +342,8 @@ impl LoadRunner {
         let replacement_max_fee = max_fee.saturating_mul(3);
         let replacement_priority_fee = max_priority_fee.saturating_mul(3);
 
-        // Phase 3: Build all TXs upfront with pre-assigned nonces, send concurrently.
+        // Phase 3+4: Send funding TXs in batches and confirm each batch before
+        // sending the next. This avoids overwhelming the txpool's per-sender limit.
         let txs: Vec<(TransactionRequest, Address, U256, u64)> = accounts_to_fund
             .iter()
             .enumerate()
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (dismissed):

kmchicoine (approved):
Love this!

cb-heimdall:
Review Error for kmchicoine @ 2026-04-08 15:44:09 UTC

crates/infra/load-tests/src/runner/load_runner.rs
```diff
@@ -264,116 +338,159 @@ impl LoadRunner {
...
+        let total_txs = txs.len() as u64;
+        let pb_fund = Self::progress_bar(total_txs, "Funding accounts");
+        let mut txs_remaining = txs.into_iter().peekable();
+        while txs_remaining.peek().is_some() {
+            let batch: Vec<_> = txs_remaining.by_ref().take(FUNDING_BATCH_SIZE).collect();
+            let mut batch_pending: Vec<(TxHash, Address)> = Vec::with_capacity(batch.len());
+            let mut retries: Vec<(Address, U256, u64)> = Vec::new();
+            let mut fatal_errors: Vec<String> = Vec::new();
+
+            let send_futs = batch.into_iter().map(|(tx, address, deficit, nonce)| {
```
kmchicoine, line 379: Could the parallelization here lead to nonce issues if one of the transactions fails? May not be a huge concern since funding, etc. were checked earlier

kmchicoine (approved):
Just had 1 non blocking question

crates/infra/load-tests/src/runner/load_runner.rs
haardikk21: yeah if it fails then it will block future transactions, but it shouldnt really happen and is likely a separate bug somewhere that needs to be fixed? if we find that balance estimates are wrong, we should fix that. if we find there are edge cases not thought about, worst case we can add a retry here on the same nonce

PR #2094, opened 2026-04-10
fix(txpool-tracing): record fb_inclusion_duration only once per transaction

haardikk21:
## Summary

- `track_flashblock_transactions` calls `get_pending_transaction_hashes()` which returns **all** tx hashes across all flashblocks in `PendingBlocks`, not just the latest delta. Without a guard, `transaction_fb_included` re-records the `fb_inclusion_duration` metric on every subsequent flashblock notification with an increasingly stale `received_at`, inflating the histogram significantly (e.g. a real 200ms inclusion gets recorded as 200, 400, 600, 800ms across successive flashblocks).
- Adds an `fb_included` flag to `EventLog` that gets set on first measurement and causes subsequent calls to early-return, ensuring each transaction's flashblock inclusion time is recorded exactly once.
- Adds `test_fb_inclusion_recorded_only_once` to verify the dedup behavior.

## Details

On Base Sepolia the reported `fb_inclusion_duration` was ~700ms — far higher than expected given low traffic and fast mempool→builder propagation. With ~10 flashblocks per block at ~200ms intervals, and assuming a roughly uniform distribution of transactions landing at the mempool nodes, the average reported for transactions would be ~1s. However because it stops tracking transactions upon canonical block inclusion, the real average is lower around 700ms, because it doesn't typically get all the way up to the 10th flashblock for the same tx hash.

After this fix, each tx is measured once at the time the first flashblock containing it is received, which should bring the metric down to the actual inclusion latency (~200ms or lower).

### DIFF
--- a/crates/client/txpool-tracing/src/events.rs
+++ b/crates/client/txpool-tracing/src/events.rs
@@ -48,6 +48,7 @@ pub enum Pool {
 pub struct EventLog {
     pub(crate) mempool_time: Instant,
     pub(crate) pending_time: Option<Instant>,
+    pub(crate) fb_included: bool,
     pub(crate) events: Vec<(DateTime<Local>, TxEvent)>,
     pub(crate) limit: usize,
 }
@@ -57,7 +58,13 @@ impl EventLog {
     pub fn new(t: DateTime<Local>, event: TxEvent) -> Self {
         let now = Instant::now();
         let pending_time = if event == TxEvent::Pending { Some(now) } else { None };
-        Self { mempool_time: now, pending_time, events: vec![(t, event)], limit: 10 }
+        Self {
+            mempool_time: now,
+            pending_time,
+            fb_included: false,
+            events: vec![(t, event)],
+            limit: 10,
+        }
     }
 
     /// Append a new `(timestamp, event)` tuple to the log.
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

wlawt (approved):

PR #2236, opened 2026-04-16
feat(consensus): Safe Head Tracking For Follow Nodes

refcell:
## Summary

Follow nodes previously hardcoded DisabledSafeDB, causing the optimism_safeHeadAtL1Block RPC endpoint to unconditionally return an error. This is a bug because follow nodes are legitimate L2 nodes that expose the rollup RPC surface, and callers reasonably expect safe head queries to work. The bug went unnoticed because follow nodes are typically used in contexts where the RPC endpoint is not queried directly, and DisabledSafeDB silently returns a well-typed error rather than crashing. This PR wires up real SafeDB support via a new with_safedb_path builder on FollowNode and a with_safe_head_listener builder on DelegateL2DerivationActor, which decodes the L1 info deposit transaction on each sync cycle to extract l1_origin and record the L1→L2 safe head mapping. The DB write is skipped when the local EL's block hash disagrees with the source to avoid recording data during fork divergence.

### DIFF
--- a/crates/consensus/service/src/actors/derivation/delegate_l2/actor.rs
+++ b/crates/consensus/service/src/actors/derivation/delegate_l2/actor.rs
@@ -224,14 +224,16 @@ where
                     sync_task = Some(tokio::spawn(async move {
                         SyncFromSourceTask::new(
                             engine_client,
-                            engine_actor_request_tx,
-                            cancellation_token,
-                            local_l2_provider,
+                            SyncFromSourceTaskConfig {
+                                engine_actor_request_tx,
+                                cancellation_token,
+                                local_l2_provider,
+                                safe_head_listener,
+                                genesis,
+                            },
                             sent_head,
                             target_block,
                             l2_source,
-                            safe_head_listener,
-                            genesis,
                         )
                         .sync_from_source()
                         .await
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/consensus/service/src/actors/derivation/delegate_l2/actor.rs
```diff
@@ -396,17 +424,60 @@ where
...
-                target: "derivation",
-                block_number = clamped_safe,
-                local_hash = %local_block.header.hash,
-                source_hash = %source_hash,
-                "Delegated safe block hash mismatch between source and local EL"
-            );
+        // When the chains disagree we skip the safe head DB write: we are in a fork-diverged
+        // state and recording either version would be misleading. The next sync cycle will
+        // retry once the local engine has caught up.
+        let chains_agree =
```
meyer9: I think we send the safe head to the EL while syncing. We send `min(current_unsafe, source_safe)`, which isn't the actual safe at this unsafe block number. Won't this still be true even in this case? https://github.com/base/base/pull/2236/changes#diff-f594debf1c72fff29acff2e0fe4fb41482bf851018524b17883704d26762a99eR419-R422

PR #2360, opened 2026-04-23
feat: perf improvements to `MeteringStore` 

wlawt:
- Add TTL
- Change from LFU to LRU, ensures that `insert()` always adds a new entry and evicts an old one. LFU is based on the probability a candidate may be used again which means that an insertion might not actually add the new tx hash when we want it to

### DIFF
--- a/bin/builder/src/cli.rs
+++ b/bin/builder/src/cli.rs
@@ -348,7 +348,11 @@ mod tests {
         use base_builder_core::MeteringProvider;
         use base_bundles::MeterBundleResponse;
 
-        let args = Args { metering_store_ttl_secs: 60, ..Default::default() };
+        let args = Args {
+            metering_store_ttl_secs: 60,
+            enable_resource_metering: true,
+            ..Default::default()
+        };
         let store = args.build_metering_store();
         let tx_hash = TxHash::random();
         store.insert(
--- a/bin/builder/src/cli.rs
+++ b/bin/builder/src/cli.rs
@@ -121,10 +121,9 @@ pub struct Args {
     pub tx_data_store_buffer_size: usize,
 
     /// TTL in seconds for entries in the metering store cache.
-    /// Stale entries are evicted after this duration, preventing `TinyLFU` admission
-    /// rejections for new metering data.
+    /// Stale entries are evicted after this duration.
     #[arg(long = "builder.metering-store-ttl-secs", default_value = "30")]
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

bin/builder/src/cli.rs
```diff
@@ -120,6 +120,11 @@ pub struct Args {
     #[arg(long = "builder.tx-data-store-buffer-size", default_value = "10000")]
     pub tx_data_store_buffer_size: usize,
 
+    /// TTL in seconds for entries in the metering store cache.
+    /// Stale entries are evicted after this duration.
+    #[arg(long = "builder.metering-store-ttl-secs", default_value = "30")]
```
wlawt, line 125: should default value be lower?

niran: I think the only reason to have a TTL would be to ensure that there's space for bursts of activity. That means the TTL would need to be lower than the current time it takes for the whole queue to roll over. So since a 10,000 entry queue can handle 333 entries per second for 30 seconds, that TTL should give us a good amount of headroom over the current eviction rate of ~150 entries per second.

But lower is fine too. 10 seconds? Since the `needed_at` cache has the same TTL, we don't want to go too low.

niran (approved):

PR #2434, opened 2026-04-28
feat: flag to use new `basev0` discovery network

meyer9:
This will allow nodes to find Base peers faster.

### DIFF
--- a/crates/execution/cli/src/commands/p2p/bootnode.rs
+++ b/crates/execution/cli/src/commands/p2p/bootnode.rs
@@ -2,10 +2,11 @@
 
 use std::{net::SocketAddr, path::PathBuf};
 
+use base_node_core::BASE_V0_PROTOCOL_VERSION;
 use clap::Parser;
 use reth_cli_util::{get_secret_key, load_secret_key::rng_secret_key};
 use reth_discv4::{DiscoveryUpdate, Discv4, Discv4Config};
-use reth_discv5::{Config, Discv5, NetworkStackId, discv5::Event};
+use reth_discv5::{Config, Discv5, discv5::Event};
 use reth_net_nat::{NatResolver, external_addr_with};
 use reth_network_peers::NodeRecord;
 use secp256k1::SecretKey;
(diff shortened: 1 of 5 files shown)


crates/execution/node/src/node.rs
```diff
@@ -1044,24 +1045,30 @@ where
...
+    /// Enable the Base discv5 protocol identity
+    pub base_protocol: bool,
 }
 
-impl Clone for BaseNetworkBuilder {
-    fn clone(&self) -> Self {
-        Self::new(self.disable_txpool_gossip, self.disable_discovery_v4)
+impl Default for BaseNetworkBuilder {
+    fn default() -> Self {
+        Self { disable_txpool_gossip: false, disable_discovery_v4: false, base_protocol: true }
```
danyalprout: We should default `disable_txpool_gossip` to `true`

crates/execution/cli/src/commands/p2p/bootnode.rs
```diff
@@ -37,6 +37,10 @@ pub struct Command {
     /// Run a discv5 topic discovery bootnode in addition to discv4.
     #[arg(long)]
     pub v5: bool,
+
+    /// Enable the Base discv5 protocol identity.
+    #[arg(long = "v5.base-protocol", default_value_t = true, action = clap::ArgAction::Set)]
+    pub base_protocol: bool,
```
danyalprout: Should this default to true? Would this break peering by default?

meyer9: Going forward, it should be true, but if we want to default to false until we have bootnodes setup, I think that's reasonable. I'll just open another PR to default to true once we test it fully.

crates/execution/node/src/node.rs
meyer9: What's the reasoning for changing this? Before, I believe it defaulted to false (default bool)

Mjofialda: nice

refcell (approved):

PR #2441, opened 2026-04-29
consensus: bound snappy gossip decompression

niran:
## Summary
- Adds a bounded Snappy `DataTransform` for consensus gossipsub messages.
- Moves Snappy compression/decompression out of `NetworkPayloadEnvelope` and into the libp2p gossipsub transform boundary.
- Changes `NetworkPayloadEnvelope::encode_v[1-4]` and `decode_v[1-4]` to operate on canonical uncompressed SSZ bytes, matching how the handler consumes messages after gossipsub transforms them.
- Removes the production `snap` dependency from `base-common-rpc-types-engine`; Snappy is now a consensus gossip transport concern.

## Compression Boundary
Previously, block gossip bytes were decompressed in more than one layer: gossipsub message-id computation and envelope decoding both knew about Snappy-compressed payloads. This made the envelope API look like a wire-compression API and left multiple places where an attacker-controlled compressed buffer could trigger allocation-heavy decompression.

This PR follows the Lighthouse-style layering instead:
- Inbound wire bytes enter gossipsub as Snappy-compressed data.
- `SnappyTransform::inbound_transform` first checks the compressed length, reads the declared decompressed length with `snap::raw::decompress_len`, rejects anything above `MAX_GOSSIP_SIZE`, and only then allocates and decompresses.
- After the transform, message-id computation, topic handlers, and `NetworkPayloadEnvelope::decode_v[1-4]` all see uncompressed SSZ bytes.
- Outbound block publishing encodes the envelope to uncompressed SSZ, then `SnappyTransform::outbound_transform` validates the uncompressed size, compresses for the wire, and validates the compressed size.

Keeping compression at the gossipsub boundary makes the invariant explicit: consensus gossip messages are bounded before Snappy allocation, while envelope encode/decode remains pure payload serialization. It also avoids carrying compressed-byte helper APIs through non-gossip call sites where the transport boundary is not visible.

## Test plan
- `cargo test -p base-common-rpc-types-engine -p base-consensus-gossip`
- `cargo check -p base-common-rpc-types-engine -p base-consensus-gossip --no-default-features`
- `cargo clippy --locked -p base-common-rpc-types-engine -p base-consensus-gossip --all-targets --profile ci -- -D warnings`
- `git diff --check`

`cargo fmt --all` was also run; stable rustfmt prints existing warnings for unstable options in the workspace rustfmt config.

PR #2444, opened 2026-04-29
fix(consensus): cap snappy-decoded gossip payload size

0x00101010:
## Summary

`compute_message_id` (`crates/consensus/gossip/src/config.rs`) and `NetworkPayloadEnvelope::decode_v{1..4}` (`crates/common/rpc-types-engine/src/envelope.rs`) snappy-decompress untrusted gossip data without bounding the *decoded* size. The existing `max_transmit_size(MAX_GOSSIP_SIZE)` only caps **compressed** wire bytes, so a wire-valid ~9 MiB snappy frame can declare ~200 MiB decoded and force the gossipsub task to allocate that much heap before any signature, length, or SSZ check runs. Both call sites fire on every received message — `compute_message_id` runs on every subscribed topic (not just block topics), and `decode_v*` runs again for block-topic dispatch. Sustained delivery from a small number of unauthenticated peers can starve the single-threaded gossipsub task and halt block propagation.

This patch pre-checks the declared decoded length via `snap::raw::decompress_len` (parses only the varu32 preamble; O(1) and never allocates) and rejects when the declared size exceeds the cap. Mirrors the canonical guard used by `paradigmxyz/reth` (`crates/net/eth-wire/src/p2pstream.rs`) and required by the Ethereum consensus-specs p2p-interface. Equivalent guards ship in Lighthouse, Prysm, Teku, and Nimbus.

## Changes

| File | Change |
|---|---|
| `crates/consensus/gossip/src/config.rs` | `compute_message_id` pre-checks `snap::raw::decompress_len` against `MAX_GOSSIP_SIZE`. Frames whose declared decoded size exceeds the cap hash under the existing invalid-snappy domain tag — no allocation, no behavioral surprise for downstream dedup. |
| `crates/common/rpc-types-engine/src/envelope.rs` | New `pub const MAX_DECOMPRESSED_ENVELOPE_BYTES = 10 * (1 << 20)` (mirrors `MAX_GOSSIP_SIZE`). New error variant `PayloadEnvelopeError::DecodedTooLarge { given, max }`. New private helper `NetworkPayloadEnvelope::decompress_bounded` that performs the bounded preamble-check + decompress; all four `decode_v1..v4` functions are swapped to use it. |
| `crates/common/rpc-types-engine/src/lib.rs` | Re-export `MAX_DECOMPRESSED_ENVELOPE_BYTES`. |

## Tests

- `crates/common/rpc-types-engine/src/envelope.rs::tests::test_decode_rejects_oversized_decompressed_payload` — builds a snappy frame whose declared decoded length is `MAX_DECOMPRESSED_ENVELOPE_BYTES + 1` (compressed wire size well under the cap) and asserts all four `decode_v{1..4}` functions return `DecodedTooLarge { given, max }`.
- `crates/consensus/gossip/src/config.rs::tests::test_compute_message_id_rejects_oversized_snappy_bomb` — same shape; asserts the message id is hashed under the invalid-snappy domain (`[0,0,0,0]`) rather than the valid-snappy domain (`[1,0,0,0]`), proving the early-reject path was taken before any allocation.

Both new tests pass; all existing tests continue to pass.

## Verification

\`\`\`
cargo test  -p base-common-rpc-types-engine -p base-consensus-gossip --lib
# 32 passed (envelope) / 45 passed (gossip), 0 failed

cargo clippy -p base-common-rpc-types-engine -p base-consensus-gossip --all-targets -- -D warnings
# clean
\`\`\`

## No protocol or wire-compatibility change

`MAX_GOSSIP_SIZE = 10 MiB` is the documented invariant at `config.rs:13-15`:
> Limits the total size of gossip RPC containers as well as decompressed individual messages.

This patch enforces what the doc comment already claims. Honest Base block payloads compress to far less than 10 MiB decoded, so honest peers are unaffected. The 10 round-trip-encoding tests (`test_roundtrip_encode_envelope_v1..v4` and friends) against recorded mainnet envelopes continue to pass without modification.

### DIFF
--- a/crates/common/rpc-types-engine/src/envelope.rs
+++ b/crates/common/rpc-types-engine/src/envelope.rs
@@ -14,6 +14,15 @@ use alloy_rpc_types_engine::{
 
 use crate::{BaseExecutionPayload, BaseExecutionPayloadSidecar, BaseExecutionPayloadV4};
 
+/// Maximum allowed decoded size for a snappy-compressed [`NetworkPayloadEnvelope`].
+///
+/// Mirrors `MAX_GOSSIP_SIZE` in `base-consensus-gossip` and bounds the heap
+/// allocation performed by [`NetworkPayloadEnvelope::decode_v1`] and friends.
+/// Without this cap, a wire-valid 9 `MiB` snappy frame can declare a 200 `MiB`
+/// decoded length and force the decoder to allocate that buffer before any
+/// downstream length, SSZ, or signature check runs.
+pub const MAX_DECOMPRESSED_ENVELOPE_BYTES: usize = 10 * (1 << 20);
+
 /// A thin wrapper around [`BaseExecutionPayload`] that includes the parent beacon block root.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
@@ -249,13 +258,31 @@ pub struct NetworkPayloadEnvelope {
 }
 
 impl NetworkPayloadEnvelope {
+    /// Snappy-decompresses `data` after checking that the declared decoded size
+    /// does not exceed [`MAX_DECOMPRESSED_ENVELOPE_BYTES`].
+    ///
+    /// `snap::raw::decompress_len` parses only the varu32 preamble and never
+    /// allocates, so this guard runs in O(1) and prevents an unauthenticated
+    /// peer from forcing arbitrarily large heap allocations on the consensus
+    /// node before any signature or SSZ check has run.
+    #[cfg(feature = "std")]
+    fn decompress_bounded(data: &[u8]) -> Result<Vec<u8>, PayloadEnvelopeError> {
+        let declared = snap::raw::decompress_len(data)?;
+        if declared > MAX_DECOMPRESSED_ENVELOPE_BYTES {
+            return Err(PayloadEnvelopeError::DecodedTooLarge {
+                given: declared,
+                max: MAX_DECOMPRESSED_ENVELOPE_BYTES,
+            });
+        }
+        Ok(snap::raw::Decoder::new().decompress_vec(data)?)
+    }
+
     /// Decode a payload envelope from a snappy-compressed byte array.
     /// The payload version decoded is `ExecutionPayloadV1` from SSZ bytes.
     #[cfg(feature = "std")]
     pub fn decode_v1(data: &[u8]) -> Result<Self, PayloadEnvelopeError> {
         use ssz::Decode;
-        let mut decoder = snap::raw::Decoder::new();
-        let decompressed = decoder.decompress_vec(data)?;
+        let decompressed = Self::decompress_bounded(data)?;
 
         if decompressed.len() < 66 {
             return Err(PayloadEnvelopeError::InvalidLength);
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/rpc-types-engine/src/envelope.rs
```diff
@@ -249,13 +258,31 @@ pub struct NetworkPayloadEnvelope {
...
 impl NetworkPayloadEnvelope {
+    /// Snappy-decompresses `data` after checking that the declared decoded size
+    /// does not exceed [`MAX_DECOMPRESSED_ENVELOPE_BYTES`].
+    ///
+    /// `snap::raw::decompress_len` parses only the varu32 preamble and never
+    /// allocates, so this guard runs in O(1) and prevents an unauthenticated
+    /// peer from forcing arbitrarily large heap allocations on the consensus
+    /// node before any signature or SSZ check has run.
+    #[cfg(feature = "std")]
+    fn decompress_bounded(data: &[u8]) -> Result<Vec<u8>, PayloadEnvelopeError> {
```
BrianBland, line 269: Naming: should we just call this `decompress` to reinforce that this is the single blessed approach?

crates/consensus/gossip/src/config.rs
```diff
@@ -166,26 +166,48 @@ pub fn default_config() -> Config {
...
+            decoder.decompress_vec(&msg.data).map_or_else(
+                |_| {
+                    warn!(target: "cfg", "Failed to decompress message, using invalid snappy");
+                    invalid_snappy_id(&msg.data)
+                },
+                |data| {
+                    let domain_valid_snappy: Vec<u8> = vec![0x1, 0x0, 0x0, 0x0];
+                    sha256([domain_valid_snappy.as_slice(), data.as_slice()].concat().as_slice())
+                        [..20]
+                        .to_vec()
```
BrianBland, line 192: This should probably also use a helper function like `invalid_snappy_id`. Both implementations would be nearly identical but with different domains

niran:
#2441 is a version of this inspired by how Lighthouse uses libp2p. But no strong feelings either way.

BrianBland:
@niran my slight preference is for this as a more concise change

BrianBland (approved):

danyalprout (approved):

PR #2450, opened 2026-04-29
fix(gossip): bind message ids to topics

niran:
## Summary
- bind gossipsub message IDs to the topic using the OP Stack preimage shape: domain || topic_len_le || topic || payload
- preserve the existing Snappy decoded-size guard and valid/invalid Snappy domains
- add a regression test that byte-identical payloads on V3 and V4 block topics produce different message IDs

## Context
The relevant op-node implementation is `op-node/p2p/gossip.go::BuildMsgIdFn`, which includes the topic length and topic before hashing the payload. Without topic binding, libp2p-gossipsub duplicate-cache entries can collide across Base block-version topics.

## Test Plan
- `cargo test --package base-consensus-gossip --lib`
- `cargo clippy --package base-consensus-gossip --all-targets -- -D warnings`

### DIFF
--- a/crates/consensus/gossip/src/config.rs
+++ b/crates/consensus/gossip/src/config.rs
@@ -6,7 +6,7 @@ use libp2p::{
     connection_limits::ConnectionLimits,
     gossipsub::{Config, ConfigBuilder, Message, MessageId},
 };
-use openssl::sha::sha256;
+use openssl::sha::Sha256;
 use snap::raw::Decoder;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////
@@ -53,6 +53,12 @@ pub const DEFAULT_MAX_ESTABLISHED_CONNECTIONS: u32 = 30;
 
 /// The default maximum number of established libp2p connections per peer.
 pub const DEFAULT_MAX_ESTABLISHED_CONNECTIONS_PER_PEER: u32 = 1;
+
+/// Domain tag for malformed snappy messages in gossip message IDs.
+const DOMAIN_INVALID_SNAPPY: [u8; 4] = [0x0, 0x0, 0x0, 0x0];
+
+/// Domain tag for valid snappy messages in gossip message IDs.
+const DOMAIN_VALID_SNAPPY: [u8; 4] = [0x1, 0x0, 0x0, 0x0];
 
 ////////////////////////////////////////////////////////////////////////////////////////////////
 // Duration Constants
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (commented):
If we roll this out across some nodes, will it break p2p communication with other nodes? Aren't we hardforking the gossip protocol here?

niran:
From GPT:

> Good question. I checked this against both OP Stack `op-node` and libp2p-gossipsub.
>
> This should not be a gossip wire-protocol hardfork. The computed message ID is not part of the full published gossipsub message; peers send the raw message fields, including topic and data, and each receiver computes its own local message ID. Message IDs are used for local duplicate/mcache accounting and in IHAVE/IWANT control flow. In the IHAVE/IWANT case, the receiver echoes back the advertiser's ID, and the advertiser uses its own cache keyed by that ID, so mixed message-id functions can still retrieve the raw message.
>
> The important point is that OP Stack `op-node` already uses `domain || topic_len_le || topic || data` in `BuildMsgIdFn`. Base's Rust `base-consensus` implementation omitted the topic and diverged from that behavior. This PR restores the Rust implementation to the existing OP behavior rather than introducing a new Base-specific gossip protocol.
>
> During a mixed rollout, the realistic downside is some redundant IHAVE/IWANT/full-message traffic between old and new nodes because duplicate caches use different IDs. Old nodes also remain vulnerable to the cross-topic duplicate-cache issue until upgraded. I do not see a p2p communication break or network partition from this change.

refcell (approved):

PR #2455, opened 2026-04-30
fix(consensus): reject stale unsafe gossip payloads

niran:
## Summary
- Reject replayed or otherwise inapplicable unsafe payloads before `engine_newPayload` and forkchoice update.
- Drop unsafe payloads that are already the current unsafe head, older than the current unsafe head, or the next height with the wrong parent.
- Keep derived safe payloads and valid direct-child unsafe payloads on the existing insertion path.

## op-node context
The equivalent op-node flow does not insert gossip directly into the engine sync state. Unsafe gossip is decoded and forwarded through `BlockReceiver.OnUnsafeL2Payload` and `SyncDeriver.OnUnsafeL2Payload`, then `EngineController.AddUnsafePayload` queues it. Before insertion, op-node runs `PayloadsQueue.DropInapplicableUnsafePayloads`, which drops payloads that are older than or equal to the current unsafe head and drops next-height payloads whose parent is not the current unsafe head. `EngineController.onForkchoiceUpdate` then only processes an unsafe payload when it is the direct child of the current unsafe head.

Base consensus previously accepted a signed unsafe gossip payload through `ProcessUnsafeL2Block` into `InsertTask`, imported it, and then let `SynchronizeTask` apply it as the new `unsafe_head` without the same current-unsafe applicability check. This change adds the op-node-style guard at the Base insertion boundary so stale replays after gossip dedup loss are ignored before they can canonicalize or regress local unsafe state.

## Tests
- `cargo fmt --check` (passes; stable rustfmt prints existing warnings for unstable config options)
- `cargo test -p base-consensus-engine insert::task --lib`
- `cargo test -p base-consensus-engine --lib`

### DIFF
--- a/crates/consensus/engine/src/task_queue/tasks/insert/task.rs
+++ b/crates/consensus/engine/src/task_queue/tasks/insert/task.rs
@@ -89,6 +89,56 @@ impl<EngineClient_: EngineClient> InsertTask<EngineClient_> {
     const fn check_new_payload_status(&self, status: &PayloadStatusEnum) -> bool {
         matches!(status, PayloadStatusEnum::Valid | PayloadStatusEnum::Syncing)
     }
+
+    fn is_unsafe_payload_applicable(
+        &self,
+        state: &EngineState,
+        new_unsafe_ref: &L2BlockInfo,
+    ) -> bool {
+        if self.payload_safety.advances_safe_head() {
+            return true;
+        }
+
+        let unsafe_head = state.sync_state.unsafe_head();
+        if new_unsafe_ref.block_info.hash == unsafe_head.block_info.hash {
+            debug!(
+                target: "engine",
+                hash = %new_unsafe_ref.block_info.hash,
+                number = new_unsafe_ref.block_info.number,
+                "Skipping already processed unsafe payload"
+            );
+            return false;
+        }
+
+        if new_unsafe_ref.block_info.number <= unsafe_head.block_info.number {
+            info!(
+                target: "engine",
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (approved):
This LGTM

PR #2551, opened 2026-05-06
fix: error severity for invalid FC state -> reset

meyer9:
Change the error severity for an invalid forkchoice state error to Reset so the engine resets if we get this error. This matches `SynchronizationTask` and `op-node` behavior.

Disables flaky test `test_online_discv5_driver_bootstrap_mainnet`.

### DIFF
--- a/crates/consensus/disc/src/driver.rs
+++ b/crates/consensus/disc/src/driver.rs
@@ -475,6 +475,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore = "flaky: depends on live mainnet bootstrap nodes responding"]
     async fn test_online_discv5_driver_bootstrap_mainnet() {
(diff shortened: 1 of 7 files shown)


refcell (commented):
Took an initial pass and this LGTM

crates/consensus/disc/src/driver.rs
```diff
@@ -475,6 +475,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore = "flaky: depends on live mainnet bootstrap nodes responding"]
```
meyer9, line 478: This is failing rn, will open an issue to fix this flake in a forward PR.

meyer9: https://github.com/base/base/issues/2552

refcell (approved):

PR #2573, opened 2026-05-08
fix(engine): validate Isthmus withdrawals with overlay state

mw2000:
## Summary
- Validate Isthmus withdrawals roots against the in-memory-aware parent state provider used by engine-tree execution.
- Remove the silent success path when the legacy payload validator cannot load parent state for Isthmus validation.

## Test plan
- `RISC0_SKIP_BUILD_KERNELS=1 BASE_SUCCINCT_ELF_STUB=1 cargo check -p base-node-core -p base-engine-tree`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3897,6 +3897,7 @@ dependencies = [
  "alloy-primitives",
  "alloy-rpc-types-engine",
  "alloy-rpc-types-eth",
+ "base-common-chains",
  "base-common-consensus",
  "base-common-evm",
  "base-common-flashblocks",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3898,7 +3898,6 @@ dependencies = [
  "alloy-primitives",
  "alloy-rpc-types-engine",
  "alloy-rpc-types-eth",
- "base-common-chains",
  "base-common-consensus",
  "base-common-evm",
  "base-common-flashblocks",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3890,6 +3890,7 @@ dependencies = [
 name = "base-engine-tree"
 version = "0.0.0"
 dependencies = [
+ "alloy-chains",
  "alloy-consensus",
  "alloy-eip7928",
  "alloy-eips",
@@ -3913,6 +3914,7 @@ dependencies = [
  "reth-chain-state",
  "reth-chainspec",
  "reth-consensus",
+ "reth-db-common",
  "reth-engine-primitives",
  "reth-engine-tree",
  "reth-errors",
--- a/crates/execution/engine-tree/Cargo.toml
+++ b/crates/execution/engine-tree/Cargo.toml
@@ -31,6 +31,7 @@ reth-primitives-traits.workspace = true
 reth-engine-primitives.workspace = true
 reth-payload-primitives.workspace = true
 base-execution-chainspec.workspace = true
+base-common-chains.workspace = true
 base-common-consensus = { workspace = true, features = ["reth"] }
 
 revm.workspace = true
--- a/crates/execution/engine-tree/Cargo.toml
+++ b/crates/execution/engine-tree/Cargo.toml
@@ -31,7 +31,6 @@ reth-primitives-traits.workspace = true
 reth-engine-primitives.workspace = true
 reth-payload-primitives.workspace = true
 base-execution-chainspec.workspace = true
-base-common-chains.workspace = true
 base-common-consensus = { workspace = true, features = ["reth"] }
 
 revm.workspace = true
--- a/crates/execution/engine-tree/Cargo.toml
+++ b/crates/execution/engine-tree/Cargo.toml
@@ -53,9 +53,12 @@ derive_more.workspace = true
 crossbeam-channel.workspace = true
 
 [dev-dependencies]
+alloy-chains.workspace = true
 reth-chainspec.workspace = true
+reth-db-common.workspace = true
 reth-storage-errors.workspace = true
 alloy-rpc-types-eth.workspace = true
+base-common-rpc-types.workspace = true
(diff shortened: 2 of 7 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

0x00101010:
## Review feedback

The fix is directionally correct — caller-supplies-state is the only sound shape, since reth's `Node::Provider::state_by_block_hash` cannot resolve non-canonical pending blocks by design. The legacy path being made fail-closed is also right.

One suggestion: **the new `PayloadValidatorWithState` trait may be unnecessary.** It's a single-use abstraction that adds API surface, generic-bound surgery through the builder, and a third indirection layer over what is essentially one function call.

### Alternative: free function + direct call from engine-tree

Move the verifier into a free function in `node/src/engine.rs`:

```rust
pub fn verify_isthmus_withdrawals_root<DB, H, ChainSpec>(
    chain_spec: &ChainSpec,
    hashed_msg_passer_addr: B256,
    state_updates: &HashedPostState,
    state: DB,
    header: H,
) -> Result<(), ConsensusError>
where DB: StateProvider, H: BlockHeader, ChainSpec: Upgrades { ... }
```

Then:
- `PayloadValidator::validate_block_post_execution_with_hashed_state` calls it with `self.provider.state_by_block_hash(parent_hash)?` (fail-closed, same as this PR).
- `engine-tree/src/validator.rs` calls it directly with `parent_state_provider_builder.build()?`. Engine-tree already has `chain_spec` via `self.provider.chain_spec()`; store the hashed predeploy address as a field at construction (computed once via the same `KeyHasher` already used in `BaseEngineValidator::new`).

### What this changes vs. the current PR

| | Current PR | Trait-free |
|---|---|---|
| New public trait | `PayloadValidatorWithState` | none |
| Engine-tree generic bound | `V: PayloadValidatorWithState<T>` | unchanged from pre-PR (`V: PayloadValidator<T, Block = BaseBlock>`) |
| Builder bound changes | yes, propagates through `BaseEngineValidatorBuilder` | none |
| Source of truth for verifier | inherent method on validator (called from two trait methods) | one free function |
| Indirection layers | trait method → inherent method → `isthmus::verify_…` | free function → `isthmus::verify_…` |
| Net lines added | ~95 | ~40 |

### Why I think it's a better factoring

1. **Single source of truth, no drift risk.** Today the inherent `verify_isthmus_withdrawals_root` and the `validate_block_post_execution_with_hashed_state` impl both contain the timestamp gate + storage-update extraction. With one free function they share one body.
2. **No generic-bound surgery.** The engine-tree builder's bound reverts to the original `PayloadValidator<...>`. Smaller blast radius, easier to reason about backward compatibility.
3. **Structurally impossible to call without state.** A free function with a `DB: StateProvider` parameter cannot be invoked without supplying state — the silent-skip class of bug is eliminated at the type level, not just at this call site.
4. **YAGNI on the trait.** It pays its cost only if multiple validator impls or multiple state-aware checks appear. Today there's one of each.

The cost is one constant reference (`Predeploys::L2_TO_L1_MESSAGE_PASSER`) and a `KeyHasher` type param leaking from `node` into `engine-tree`'s `BaseEngineValidator::new`. The node-side `BaseEngineValidator::new` already takes `KH: KeyHasher`, so this is symmetric.

### Independent of the trait question — please add tests

`cargo check` is not a sufficient test plan for an Immunefi-grade consensus-validation fix. Suggested at minimum:

1. **Engine-tree regression test:** non-canonical pending parent in `tree_state.blocks_by_hash`, child block with bad `withdrawals_root` → assert `Err(ConsensusError)`. This is the test that would have caught the original bypass.
2. **Legacy-path test:** call `validate_block_post_execution_with_hashed_state` with a parent that `state_by_block_hash` cannot resolve → assert it errors instead of returning `Ok(())`.
3. **Happy path:** valid block with in-memory parent → validation passes (proves the overlay returns the right state).

Without (1), the next refactor can re-open the bypass without anything failing.

### Other small things

- The Isthmus timestamp check exists in both the engine-tree caller (gating the `parent_state_provider_builder.build()` call) and inside the inherent verifier — defensible (verifier is publicly callable and should be safe in isolation), but worth a one-line comment so the duplication doesn't read as accidental.
- A short comment justifying the fresh `parent_state_provider_builder.build()` (vs. reusing the execution-side provider that may be wrapped with caches) would help future readers.

Happy to ship a follow-up PR with the free-function refactor + tests if useful, or just the tests on top of this PR if the team prefers to keep the trait.

0x00101010 (approved):

cb-heimdall:
Review Error for 0x00101010 @ 2026-05-11 19:37:35 UTC

PR #2580, opened 2026-05-08
fix(challenge): back off on reverted withdraw before retrying

jackchuma:
## Summary
- When `claimCredit()` withdraw transactions revert, transition back to `AwaitingDelay` with a short backoff (60s, capped to the WETH delay) rather than retrying on every poll tick.
- Avoids hammering the chain with reverting transactions when the onchain WETH delay is longer than the locally cached value.
- Adds `revert_retry_delay` parameter to `submit_claim_credit` so the unlock path keeps its existing immediate-retry behavior while withdraws back off.

## Test plan
- New `reverted_withdraw_backs_off_before_retrying` unit test in `crates/proof/challenge/src/bond.rs` covers the back-off transition and verifies the next poll waits in `AwaitingDelay` instead of resubmitting.

### DIFF
--- a/crates/proof/challenge/src/bond.rs
+++ b/crates/proof/challenge/src/bond.rs
@@ -175,6 +175,9 @@ impl<C: Clock> BondManager<C> {
     /// waiting for its best-effort anchor update to finish.
     const DEFAULT_ANCHOR_UPDATE_RETENTION: Duration = Duration::from_secs(24 * 60 * 60);
 
+    /// How long to wait before retrying a reverted withdraw attempt.
+    const WITHDRAW_REVERT_RETRY_DELAY: Duration = Duration::from_secs(60);
+
     /// `DEFENDER_WINS` game status constant.
     const STATUS_DEFENDER_WINS: u8 = 2;
 
@@ -879,6 +882,7 @@ impl<C: Clock> BondManager<C> {
             submitter,
             "unlock",
             BondPhase::AwaitingDelay { unlocked_at: self.clock.now() },
+            None,
         )
         .await
     }
@@ -892,7 +896,7 @@ impl<C: Clock> BondManager<C> {
         // Fall back to 7 days if the onchain delay has not been read yet.
         // If the real delay is shorter, the withdraw attempt will simply
         // succeed earlier than expected. If longer, the attempt will revert
-        // and be retried on the next poll tick.
+        // and be retried after a short backoff.
         let delay = self.weth_delay.unwrap_or_else(|| {
             debug!(game = %game_address, "WETH delay not yet known, using default 7 days");
             Self::DEFAULT_WETH_DELAY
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/challenge/src/bond.rs
```diff
@@ -970,6 +983,21 @@ impl<C: Clock> BondManager<C> {
                 );
                 ChallengerMetrics::claim_credit_tx_outcome_total(ChallengerMetrics::STATUS_ERROR)
                     .increment(1);
+                if let Some(retry_delay) = revert_retry_delay
+                    && matches!(&e, crate::ChallengeSubmitError::TxReverted { .. })
+                {
+                    let delay = self.weth_delay.unwrap_or(Self::DEFAULT_WETH_DELAY);
```
leopoldjoy, line 989: Nit: this `self.weth_delay.unwrap_or(Self::DEFAULT_WETH_DELAY)` must stay in sync with the identical fallback in `check_delay` (line 896); if a future change alters one without the other, the retry delay arithmetic silently breaks — consider extracting a shared `fn effective_weth_delay(&self) -> Duration` to eliminate the coupling.

crates/proof/challenge/src/bond.rs
```diff
@@ -970,6 +983,21 @@ impl<C: Clock> BondManager<C> {
                 );
                 ChallengerMetrics::claim_credit_tx_outcome_total(ChallengerMetrics::STATUS_ERROR)
                     .increment(1);
+                if let Some(retry_delay) = revert_retry_delay
```
leopoldjoy, line 986: Nit: the warn log just above this block unconditionally says `"will retry"` even when this revert backoff immediately follows and delays the retry by 60 seconds; consider making the warn message conditional (e.g. `"will retry after backoff"` vs `"will retry"`) so operators aren't misled when triaging logs.

crates/proof/challenge/src/bond.rs
```diff
@@ -970,6 +983,21 @@ impl<C: Clock> BondManager<C> {
...
+                    && matches!(&e, crate::ChallengeSubmitError::TxReverted { .. })
```
leopoldjoy, line 987: Nit: the `matches!` guard correctly restricts backoff to `TxReverted` only, but no test exercises the negative case (e.g. a `TxManager(NonceTooLow)` error during withdraw) to prove the backoff does *not* trigger for non-revert failures — adding one would prevent regressions if someone broadens the match arm later.

refcell (approved):

refcell:
Removing from mq since @leopoldjoy had some comments

leopoldjoy:
@refcell mine are all nits, they can be addressed later 👍

PR #2593, opened 2026-05-08
fix(challenger): bound ZK proof sessions with explicit status handling and wall-clock timeout

xenoliss:
## Problem

`PendingProofs::poll` dispatched any non-Succeeded/non-Failed ZK prover status to a `_ => ProofUpdate::Pending` catch-all. The driver short-circuits on `Pending` without incrementing `retry_count`, changing the phase, or applying any timeout. Combined with the `process_candidate` skip-if-pending guard, any game whose proof session got stuck returned `Pending` indefinitely, making that game non-challengeable for the lifetime of the process.

Two concrete triggers:
- The ZK prover loses its session (crash, restart) and responds `STATUS_UNSPECIFIED` or a gRPC error.
- A future proto status code (e.g. `STATUS_CANCELLED = 6`) added before the Rust client is rebuilt: `try_from` returns `Err`, silently remapped to `Unspecified`, then caught by `_`.

## Fix

**Exhaustive status dispatch**

Replace `_ => ProofUpdate::Pending` with an explicit match on `ProofJobStatus::try_from(response.status)`:
- `Created | Pending | Running` -> `ProofUpdate::Pending` (legitimate intermediate states)
- `Unspecified | Err(_)` -> `NeedsRetry` (server-side error, treated as retryable failure)

Any future unknown status code now surfaces as `Err(_)` and flows into the bounded retry path instead of stalling forever.

**Wall-clock timeout on `AwaitingProof`**

`ProofPhase::AwaitingProof` now carries a `started_at: Instant`. On each poll, if `elapsed() > max_proof_duration` the entry transitions to `NeedsRetry`. The timeout check runs **before** the `get_proof` RPC so persistent transport errors (prover crashed/unreachable) can no longer keep the entry pinned in `AwaitingProof` indefinitely; the same bound applies whether the RPC fails or returns a stuck intermediate status.

The timeout is configurable via `--max-proof-duration` (env `CHALLENGER_MAX_PROOF_DURATION`, default `4h`). Zero values are rejected at startup.

## Files changed

- `src/pending.rs`: exhaustive match, `started_at` field, timeout check before RPC, `max_proof_duration` parameter on `poll`
- `src/driver.rs`: `max_proof_duration` field on `Driver` and `DriverConfig`, updated `poll` call site
- `src/cli.rs`: `--max-proof-duration` flag (default `4h`)
- `src/config.rs`: `max_proof_duration` in `ChallengerConfig`, zero-value rejection
- `src/service.rs`: wires `max_proof_duration` into `DriverConfig`

## Tests

New tests in `tests/driver.rs`:
- `test_poll_unspecified_status_triggers_retry`: `Unspecified` response increments `retry_count` and transitions to `NeedsRetry`
- `test_poll_running_within_timeout_stays_pending`: `Running` within timeout stays `Pending`
- `test_poll_running_timeout_triggers_retry`: timeout fires before the RPC, transitioning to `NeedsRetry`
- `test_max_proof_duration_zero_rejected`: zero `--max-proof-duration` is rejected at config validation

### DIFF
--- a/crates/proof/challenge/src/driver.rs
+++ b/crates/proof/challenge/src/driver.rs
@@ -44,6 +44,8 @@ use crate::{
 pub struct DriverConfig {
     /// How often the driver polls for new games.
     pub poll_interval: Duration,
+    /// Maximum wall-clock time to wait for a ZK proof session before treating it as failed.
+    pub max_proof_duration: Duration,
     /// Cancellation token for graceful shutdown.
     pub cancel: CancellationToken,
 }
@@ -119,6 +121,8 @@ where
     pub bond_manager: Option<BondManager<C>>,
     /// Interval between polling cycles.
     pub poll_interval: Duration,
+    /// Maximum wall-clock time to wait for a ZK proof session before treating it as failed.
+    pub max_proof_duration: Duration,
     /// Token used to signal graceful shutdown.
     pub cancel: CancellationToken,
 }
@@ -137,8 +141,6 @@ impl<L2: L2Provider, P: ZkProofProvider, T: TxManager, C: Clock> std::fmt::Debug
 impl<L2: L2Provider, P: ZkProofProvider, T: TxManager, C: Clock> Driver<L2, P, T, C> {
     /// Maximum number of times a failed proof job will be retried before being dropped.
     pub const MAX_PROOF_RETRIES: u32 = 3;
-    /// Maximum wall-clock time to wait for a proof session before treating it as failed.
-    pub const MAX_PROOF_DURATION: Duration = Duration::from_secs(30 * 60);
 
     /// Creates a new driver with the given components.
     pub fn new(config: DriverConfig, components: DriverComponents<L2, P, T, C>) -> Self {
@@ -152,6 +154,7 @@ impl<L2: L2Provider, P: ZkProofProvider, T: TxManager, C: Clock> Driver<L2, P, T
             pending_proofs: PendingProofs::new(),
             bond_manager: components.bond_manager,
             poll_interval: config.poll_interval,
+            max_proof_duration: config.max_proof_duration,
             cancel: config.cancel,
         }
     }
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/challenge/src/driver.rs
```diff
@@ -144,6 +144,8 @@ impl<L2: L2Provider, P: ZkProofProvider, T: TxManager, C: Clock> std::fmt::Debug
 impl<L2: L2Provider, P: ZkProofProvider, T: TxManager, C: Clock> Driver<L2, P, T, C> {
     /// Maximum number of times a failed proof job will be retried before being dropped.
     pub const MAX_PROOF_RETRIES: u32 = 3;
+    /// Maximum wall-clock time to wait for a proof session before treating it as failed.
+    pub const MAX_PROOF_DURATION: Duration = Duration::from_secs(30 * 60);
```
jackchuma: Should this be cli configurable? Very likely that 30 mins isn't long enough for the zk proof

crates/proof/challenge/tests/driver.rs
```diff
@@ -1444,3 +1444,108 @@ async fn test_driver_tracks_bond_after_successful_challenge() {
         "game should be tracked by bond manager after successful challenge"
     );
 }
+
+// ── Fix A: unexpected status codes treated as retryable failures ──────────
```
jackchuma: Nit: the `Fix A` and `Fix B` mentions are not necessary

leopoldjoy (commented):
Two questions on the timeout/retry interaction:

1. **Total wall-clock exposure per game**: Each retry resets `started_at` via `Instant::now()` in the re-initiation path. With `MAX_PROOF_RETRIES = 3` and a 4h default, a game stuck on a broken prover could stay non-challengeable for up to ~16h before being dropped. Is that acceptable for the dispute window, or should there be a `first_attempt_at` that doesn't reset?

2. **Transport errors don't consume retries**: If `get_proof()` hits a transport error (prover unreachable), the `?` propagates without incrementing `retry_count`, so the only defense is the full wall-clock timeout each time. Is that intentional? It means a dead prover burns through the entire `max_proof_duration` per attempt before transitioning to `NeedsRetry`, which compounds with point 1 above.

leopoldjoy (approved):

PR #2595, opened 2026-05-08
fix(payload): treat invalid sequencer tx as fatal under no_tx_pool

BrianBland:
## Summary

When ` no_tx_pool=true` the attribute-supplied transaction list is the consensus input for the payload (derived by `base-consensus` from batcher-authenticated L1 data), not a list of optional pre-includes. The EL payload builder was silently skipping `InvalidTx` from sequencer transactions and continuing, while the proof executor strictly executes the full list and rejects the block on any invalid tx.

This divergence breaks Holocene's deposit-only fallback: the EL freezes a skip-and-continue safe block while the proof path computes a deposit-only replacement root, so:

- An honest proof client cannot validate the canonical EL root (`Epilogue::validate` returns `InvalidClaim`).
- Even if the deposit-only proof root were surfaced, the proposer's JIT equality check rejects it as `RootMismatch`.
- A trigger sequence with a later valid tx that calls `L2ToL1MessagePasser` produces an EL-only withdrawal message that the deposit-only proof lineage does not contain.

This PR makes the EL strict in `no_tx_pool=true` mode: `InvalidTx` (and the flashblocks-side `is_invalid_tx_err()` equivalent) is propagated as a fatal `EvmExecutionError`, so the EL and proof executor reject the same payloads and Holocene's deposit-only fallback applies consistently across both consumers of the same L1 input. The pre-existing skip-and-continue path under `no_tx_pool=false` (mempool mode with skippable pre-includes) is unchanged.

Affected:
- `crates/execution/payload/src/builder.rs` — `BasePayloadBuilderCtx::execute_sequencer_transactions`
- `crates/builder/core/src/flashblocks/context.rs` — flashblocks variant of the same loop

### DIFF
--- a/crates/builder/core/src/flashblocks/context.rs
+++ b/crates/builder/core/src/flashblocks/context.rs
@@ -535,11 +535,22 @@ impl BasePayloadBuilderCtx {
     }
 
     /// Executes all sequencer transactions that are included in the payload attributes.
+    ///
+    /// When `no_tx_pool` is set the attribute-supplied transaction list is the consensus input
+    /// for the payload (derived from L1 batches by `base-consensus`), not a list of optional
+    /// pre-include candidates. In that mode any invalid-tx error must be propagated as fatal so
+    /// the EL rejects the payload, matching the strictness of the proof executor and allowing
+    /// Holocene's deposit-only fallback to apply consistently across both consumers.
+    ///
+    /// When `no_tx_pool` is `false` the builder is composing a new block from mempool plus
+    /// attribute pre-includes; pre-includes there may legitimately be skipped on invalid-tx
+    /// errors, so the historical skip-and-continue behavior is preserved.
     pub(super) fn execute_sequencer_transactions(
         &self,
         db: &mut State<impl Database>,
     ) -> Result<ExecutionInfo, PayloadBuilderError> {
         let mut info = ExecutionInfo::with_capacity(self.attributes().transactions.len());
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (approved):
This is a good catch

PR #2602, opened 2026-05-09
fix(proposer): refresh recovery after invalid parent failures

mw2000:
## Summary

In `handle_submit_result`, drop `state.cached_recovery` whenever the submission completes with `SubmitOutcome::Failed`. The next poll-tick re-walks from the anchor and rebuilds the parent chain from current on-chain state instead of replaying the stale cached parent.

## Fix

`handle_submit_result` now sets `state.cached_recovery = None` on `SubmitOutcome::Failed`. The next tick's `recover_latest_state` re-walks from the anchor. The on-chain contract (`_isValidGame`) remains the source of truth for parent validity; this change only ensures the proposer does not compound its own stale internal state across retry boundaries.

## Why this scope

- Mirroring `_isValidGame` in Rust would duplicate contract logic, add per-walk-step RPCs on the happy path, and still could not make progress past a permanently invalid occupied UUID slot — that is a contract-level constraint requiring operator intervention regardless.

### DIFF
--- a/crates/proof/contracts/src/aggregate_verifier.rs
+++ b/crates/proof/contracts/src/aggregate_verifier.rs
@@ -26,6 +26,10 @@ sol! {
         /// Error returned when the proof's L1 origin is older than the EIP-2935 history window.
         error L1OriginTooOld(uint256 l1OriginNumber, uint256 currentBlock);
 
+        /// Error returned by `initialize` when the parent game is unregistered,
+        /// unrespected, blacklisted, retired, or resolved as `CHALLENGER_WINS`.
+        error InvalidParentGame();
+
         /// Returns the root claim (output root) of this game.
         function rootClaim() external pure returns (bytes32);
 
@@ -295,6 +299,11 @@ pub trait AggregateVerifierClient: Send + Sync {
 /// The 4-byte selector for `L1OriginTooOld()`.
 pub const fn l1_origin_too_old_selector() -> [u8; 4] {
     IAggregateVerifier::L1OriginTooOld::SELECTOR
+}
+
+/// The 4-byte selector for `InvalidParentGame()`.
+pub const fn invalid_parent_game_selector() -> [u8; 4] {
+    IAggregateVerifier::InvalidParentGame::SELECTOR
 }
 
 /// Concrete implementation backed by Alloy's sol-generated contract bindings.
@@ -836,6 +845,14 @@ mod tests {
         let selector = l1_origin_too_old_selector();
         assert_eq!(selector.len(), 4);
         assert_ne!(selector, [0u8; 4]);
+    }
+
+    #[test]
+    fn test_invalid_parent_game_selector() {
+        let selector = invalid_parent_game_selector();
+        assert_eq!(selector.len(), 4);
+        assert_ne!(selector, [0u8; 4]);
+        assert_ne!(selector, l1_origin_too_old_selector());
     }
 
     #[test]
--- a/crates/proof/proposer/src/pipeline.rs
+++ b/crates/proof/proposer/src/pipeline.rs
@@ -584,6 +584,10 @@ where
                 );
                 state.proved.insert(target_block, proof);
                 state.submitting = None;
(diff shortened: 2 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proposer/src/pipeline.rs
```diff
@@ -899,6 +912,18 @@ where
...
+                warn!(
+                    candidate_parent = %lookup,
+                    expected_block,
+                    previous_parent = %parent_address,
+                    "Recovered UUID-linked game is not a valid parent candidate"
+                );
+                return Err(ProposerError::InvalidParentCandidate(format!(
+                    "recovered game {lookup} at block {expected_block} is not valid"
+                )));
+            }
```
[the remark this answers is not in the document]
mw2000: Agreed that the break-based approach would livelock, but we can't advance past an invalid game as if it were a valid parent

crates/proof/proposer/src/pipeline.rs
```diff
@@ -577,11 +577,20 @@ where
...
+                if error.is_invalid_parent_game() {
+                    warn!(
+                        error = %error,
+                        target_block,
+                        "Submission rejected: parent game invalid, dropping recovery cache"
+                    );
+                    state.cached_recovery = None;
+                } else {
+                    warn!(error = %error, target_block, "Submission failed, will retry");
+                }
```
leopoldjoy, line 593: Consider adding an inline comment making the limits of this fix explicit at the call site: dropping the cache only helps when on-chain state has *concurrently* advanced — `forward_walk` (pipeline.rs:823) matches games purely by UUID and never consults `_isValidGame`, so on the next tick recovery will deterministically re-discover the same blacklisted/retired parent and the same submission will fail again until an operator intervenes. The reasoning is in the PR description but isn't visible to a future reader scrolling this branch.

crates/proof/contracts/src/aggregate_verifier.rs
```diff
@@ -26,6 +26,10 @@ sol! {
         /// Error returned when the proof's L1 origin is older than the EIP-2935 history window.
         error L1OriginTooOld(uint256 l1OriginNumber, uint256 currentBlock);
 
+        /// Error returned by `initialize` when the parent game is unregistered,
+        /// unrespected, blacklisted, retired, or resolved as `CHALLENGER_WINS`.
+        error InvalidParentGame();
```
leopoldjoy, line 31: Consider adding a doc-comment reference to the Solidity source/commit defining `error InvalidParentGame();`. The Rust mirror here is matched by a 4-byte selector — if the Solidity contract ever evolves to add an argument (e.g. the offending parent address), the selector silently changes and `classify_tx_manager_error` will stop classifying these reverts as `InvalidParentGame` with no test failure, regressing this fix.

leopoldjoy (approved):
Approving, those nits can be addressed later.

PR #2632, opened 2026-05-11
fix(txpool): account for cached encoded_2718 bytes in pool size

0x00101010:
## Summary

`BasePooledTransaction` lazily caches a full EIP-2718 encoded copy of each pooled transaction in `encoded_2718: OnceLock<Bytes>`. The cache is populated on hot paths — the validator hits it during L1 data-fee checks for every admitted tx, and the payload builder hits it via `estimated_compressed_size` for DA accounting.

`InMemorySize::size()` did not count those cached bytes. Subpool budgets and eviction thresholds rely on `size()`, so a calldata-heavy mempool can retain materially more memory than the configured limits suggest. For a 64 KiB-calldata tx the uncounted cache is ~99% of the reported size, so process memory for the txpool storage can grow to roughly 2× the budgeted footprint while the pool reports itself as within budget.

This is not a leak — the cached bytes are released along with the transaction on inclusion / eviction / replacement / invalidation. The risk is sustained under-accounting while transactions are pooled.

## Fix

- Add the cache's heap length to `size()` via non-blocking `OnceLock::get`, so `size()` itself never triggers initialization as a side effect.
- Replace the manual sum of additional field sizes with `size_of::<Self>() - size_of_val(&inner)`, so future fields added to `BasePooledTransaction` are accounted for automatically and we do not silently regress this kind of bug again.

## Tests

- `size_includes_cached_encoded_2718_bytes` — populates the cache and asserts `size()` grows by at least the cached length. Verified to **fail** against the pre-fix implementation with the same numbers as the reported PoC (`before=65880, after=65880, cached_len=65582`).
- `size_unchanged_when_cache_uninitialized` — guards against a future refactor that turns `size()` into a side-effecting operation by switching to `get_or_init`.

```
running 3 tests
test transaction::tests::size_unchanged_when_cache_uninitialized ... ok
test transaction::tests::size_includes_cached_encoded_2718_bytes ... ok
test transaction::tests::validate_base_transaction ... ok
```

Full `cargo test -p base-execution-txpool` (41 tests) passes; `cargo clippy -p base-execution-txpool --all-targets` clean.

## Out of scope (potential follow-up)

The `OnceLock<Bytes>` cache itself is debatable now that `estimated_tx_compressed_size: OnceLock<u64>` already memoizes the only repeatedly-queried derivative. A follow-up could drop the bytes cache, encode once locally in the validator (feeding both L1-fee and DA-size computations), and re-encode on demand in the rare consumers that still need raw bytes. That would eliminate the variable-size term from `size()` entirely, at the cost of one extra encode per inclusion / RPC fetch. Worth benchmarking before pulling the trigger; intentionally not part of this PR.

### DIFF
--- a/crates/execution/txpool/src/transaction.rs
+++ b/crates/execution/txpool/src/transaction.rs
@@ -208,7 +208,16 @@ impl<Cons: Typed2718, Pooled> Typed2718 for BasePooledTransaction<Cons, Pooled> 
 
 impl<Cons: InMemorySize, Pooled> InMemorySize for BasePooledTransaction<Cons, Pooled> {
     fn size(&self) -> usize {
-        self.inner.size() + core::mem::size_of::<u128>() + core::mem::size_of::<Option<u64>>() * 3
+        // The pool uses this value for subpool budgeting; any retained
+        // allocation must be reflected here. The `size_of::<Self>() -
+        // size_of_val(&inner)` term auto-accounts for every inline field
+        // in `Self` except `inner` (whose heap-aware size is added via
+        // `inner.size()`). `encoded_2718` holds a heap allocation when
+        // populated; use non-blocking `get` so `size()` never triggers
+        // cache initialization as a side effect.
+        self.inner.size()
+            + (core::mem::size_of::<Self>() - core::mem::size_of_val(&self.inner))
+            + self.encoded_2718.get().map_or(0, |b| b.len())
     }
 }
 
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

0x00101010:
Close as this actually isn't a problem. We use the size to limit tx pool size, so the extra cache shouldn't be counted

PR #2637, opened 2026-05-11
fix(engine): drop poisoned build attrs on InvalidPayload

0x00101010:
## Summary

When the EL rejects payload attributes from a Build task (`engine_forkchoiceUpdated` returns `Invalid`), the engine used to retry the same task forever. That blocks every later task in the queue.

This PR drops the bad attributes instead: classify `InvalidPayload` as `Flush`, and have `Engine::drain` pop the head task on `Flush` before returning the error. The engine processor already turns `Flush` into `Signal::FlushChannel`, which tells derivation to discard the bad batch and produce new attributes (eventually deposits-only, which always passes).

Matches the spec: *"On payload processing errors the attributes must be dropped, and the forkchoice state must be left unchanged."* (`docs/specs/pages/protocol/consensus/derivation.md:843-844`)

## Changes

- `build/error.rs` — `InvalidPayload` is now `Flush` instead of `Temporary`.
- `task_queue/core.rs` — `Engine::drain` pops the head task on `Flush` (with metrics/watch kept in sync).
- Tests for the severity mapping, the drain-and-resume behaviour, and an end-to-end check that one bad Build dispatches exactly one `FlushChannel` signal and the queue drains back to zero.

### DIFF
--- a/crates/consensus/engine/src/task_queue/core.rs
+++ b/crates/consensus/engine/src/task_queue/core.rs
@@ -189,11 +189,15 @@ impl<EngineClient_: EngineClient> Engine<EngineClient_> {
     /// Exception: tasks that fail with [`EngineTaskErrorSeverity::Flush`] are popped from the
     /// queue before the error is returned. The poisoned task must not be retried in-place — the
     /// engine processor will signal the derivation pipeline to flush its buffered batch/channel
-    /// state and the next call to [`Engine::drain`] will proceed with the remaining tasks.
+    /// state and the next call to [`Engine::drain`] will proceed with the remaining tasks. Any
+    /// [`EngineState`] mutations the task made before failing are published to subscribers prior
+    /// to popping, mirroring the success path so watch consumers (e.g. the RPC state view) stay
+    /// in sync.
     pub async fn drain(&mut self) -> Result<(), EngineTaskErrors> {
         while let Some((task, _)) = self.tasks.peek() {
             if let Err(err) = task.execute(&mut self.state).await {
                 if matches!(err.severity(), EngineTaskErrorSeverity::Flush) {
+                    self.state_sender.send_replace(self.state);
                     self.pop_head();
                 }
                 return Err(err);
--- a/crates/consensus/engine/src/task_queue/core.rs
+++ b/crates/consensus/engine/src/task_queue/core.rs
@@ -197,10 +197,9 @@ impl<EngineClient_: EngineClient> Engine<EngineClient_> {
         // Drain tasks in order of priority, halting on errors for a retry to be attempted.
         while let Some((task, _)) = self.tasks.peek() {
             // Execute the task.
-            let flush_err = match task.execute(&mut self.state).await {
-                Ok(()) => None,
-                Err(err) if matches!(err.severity(), EngineTaskErrorSeverity::Flush) => Some(err),
-                Err(err) => return Err(err),
+            let outcome = match task.execute(&mut self.state).await {
+                Err(err) if err.severity() != EngineTaskErrorSeverity::Flush => return Err(err),
+                other => other,
             };
 
             // Update the state and notify the engine actor.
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/consensus/engine/src/task_queue/core.rs
```diff
@@ -183,24 +185,33 @@ impl<EngineClient_: EngineClient> Engine<EngineClient_> {
...
-        // Drain tasks in order of priority, halting on errors for a retry to be attempted.
         while let Some((task, _)) = self.tasks.peek() {
-            // Execute the task
-            task.execute(&mut self.state).await?;
+            if let Err(err) = task.execute(&mut self.state).await {
+                if matches!(err.severity(), EngineTaskErrorSeverity::Flush) {
+                    self.pop_head();
+                }
+                return Err(err);
+            }
```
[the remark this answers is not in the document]
0x00101010: done

crates/consensus/engine/src/task_queue/tasks/build/error.rs
```diff
@@ -77,3 +88,25 @@ impl EngineTaskError for BuildTaskError {
...
+        let err = BuildTaskError::EngineBuildError(EngineBuildError::InvalidPayload(
+            "malformed transaction at index 3".to_string(),
+        ));
+
+        assert_eq!(err.severity(), EngineTaskErrorSeverity::Flush);
+        assert!(
+            format!("{err:?}").contains("malformed transaction at index 3"),
+            "Debug must surface the EL validation error, got: {err:?}",
+        );
+    }
```
refcell, line 111: Thank you for adding tests

crates/consensus/engine/src/task_queue/core.rs
```diff
@@ -183,24 +185,37 @@ impl<EngineClient_: EngineClient> Engine<EngineClient_> {
...
         while let Some((task, _)) = self.tasks.peek() {
-            // Execute the task
-            task.execute(&mut self.state).await?;
+            if let Err(err) = task.execute(&mut self.state).await {
+                if matches!(err.severity(), EngineTaskErrorSeverity::Flush) {
+                    self.state_sender.send_replace(self.state);
+                    self.pop_head();
+                }
+                return Err(err);
+            }
```
refcell: So, the `EngineTaskErrorSeverity::Flush` code path here is the same now as if the task executed successfully. Both `Ok(())` and `Err(EngineTaskErrorSeverity::Flush)` perform these two operations:
```rust
self.state_sender.send_replace(self.state);
self.pop_head();
```

I think we can revert the `pop_head` refactor here, leave the previous code comments as well (they're useful to see what's going on) and then only return an error if `!matches!(err.severity(), EngineTaskErrorSeverity::Flush)`

refcell (dismissed):
One change requested

refcell:
FWIW this whole task queue paradigm will be removed in the near future, i've got PRs open to do just that.

refcell (approved):
This lgtm

PR #2722, opened 2026-05-15
refactor(action-harness): Use Production Sequencer Actor

refcell:
## Summary

This replaces the action-harness custom L2 sequencing glue with a harness wrapper that runs the production SequencerActor. The shared account, L2 source, and block-hash registry types move into common leaf modules so lib.rs and mod.rs stay as exports only. The action engine now keeps built payloads pending until explicit insertion, which lets the production actor drive build, seal, conductor commit, and unsafe insertion without double-committing state.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

actions/harness/src/sequencer/driver.rs
```diff
@@ -0,0 +1,413 @@
...
+        &mut self,
+        transactions: Vec<BaseTxEnvelope>,
+    ) -> BaseBlock {
+        self.try_build_next_block_with_transactions(transactions)
+            .await
+            .unwrap_or_else(|e| panic!("L2Sequencer::build_next_block failed: {e}"))
+    }
+
+    /// Build the next L2 block, returning an error instead of panicking.
+    pub async fn try_build_next_block_with_transactions(
```
0x00101010, line 197: # Orphaned Block Race in `inserted_rx`

## Problem
The driver waits for the actor to commit a block via an `mpsc` channel, with a 10s timeout. The actor doesn't know about the timeout — it keeps working independently.

## Race
1. Driver times out at t=10s, returns `Err(Timeout)`.
2. Actor commits the block at t=10.001s and pushes `(block, head)` into `inserted_rx`.
3. Driver's `stop_sequencer` cleanup runs but does **not** drain the channel.
4. The orphaned `(block, head)` sits in the channel.

## Why It Breaks
The next call to `try_build_next_block_with_transactions`:
1. Calls `wait_for_inserted_block`.
2. Immediately pops the **stale** tuple from the previous timed-out call.
3. Returns it as if it were the new block.
The driver is now permanently one block behind the actor. `self.head` diverges from the engine's real canonical head. Eventually surfaces as a cryptic `parent_hash mismatch` far from the original timeout.

## Root Cause
The channel has no notion of which build cycle a message belongs to. Async actor lifecycle + per-call timeout = state divergence when work completes after the caller gives up.

## Fix
Drain stale notifications at the start of every call:
```rust
if let Some(rx) = self.inserted_rx.as_mut() {
    while rx.try_recv().is_ok() {}
}

Safe because successful calls always consume their own message — a non-empty channel at start means leftover from a prior error.
Why It Matters
- Silent data corruption: returned block doesn't match requested txs.
- CI flakiness amplifies under load (slower runs → more timeouts → more orphans).
- The synchronous predecessor couldn't hit this; the refactor introduces it.

0x00101010 (approved):

PR #2769, opened 2026-05-19
feat(policy): implement full IPolicyRegistry precompile

eric-ships:
## Motivation

The `IPolicyRegistry` interface (base-std) defines the singleton that B-20 tokens call for every transfer, mint, and redeem authorization check. This PR delivers the full precompile, replacing the `helloWorld` stub.

**Storage namespacing:** We explored both ERC-7201 sequential namespacing and per-field named slots (e.g. `keccak256("base.policy_registry.policies")`) to allow field reordering across hardforks. Both approaches work with the `#[contract]` macro. Deferring to a follow-up PR for @refcell to decide the right convention — sequential slots (append-only) are safe for now since the precompile owns its address exclusively.

## What changed

- Full `IPolicyRegistry` implementation in `crates/common/precompiles/src/policy/`: all 13 ABI functions, `SolInterface` dispatch, 4-field storage layout (slots 0-3, append-only), and 22 unit tests
- `PolicyType` enum: `ALWAYS_ALLOW=0`, `ALWAYS_BLOCK=1`, `ALLOWLIST=2`, `BLOCKLIST=3`; built-in IDs are 0 and 1
- Policy ID encoding: top byte = `uint8(PolicyType)`, low 56 bits = global counter (checked add, 56-bit cap)
- Packed policy word: `[167:8] = admin`, `[7:0] = PolicyType`; packed == 0 reliably means never created

### DIFF
--- a/crates/common/precompiles/src/policy/storage.rs
+++ b/crates/common/precompiles/src/policy/storage.rs
@@ -39,6 +39,10 @@ impl PolicyRegistryStorage<'_> {
     /// Packs `admin` and `policy_type` into a single U256 storage word.
     ///
     /// Layout: `[255:168]` reserved (zero) | `[167:8]` admin (160 bits) | `[7:0]` `PolicyType`.
+    ///
+    /// Invariant: the result is always non-zero because ALLOWLIST = 2 and BLOCKLIST = 3 are
+    /// both non-zero. This means `policies[id] == 0` reliably signals "never created", even
+    /// after `renounce_admin` sets admin to `Address::ZERO` (the type byte is preserved).
     fn encode_packed(admin: Address, policy_type: u8) -> U256 {
         let mut word = [0u8; 32];
         word[12..32].copy_from_slice(admin.as_slice());
@@ -89,13 +93,13 @@ impl PolicyRegistryStorage<'_> {
             return Err(BasePrecompileError::revert(IPolicyRegistry::ZeroAddress {}));
         }
 
-        let policy_id = self.next_policy_id(policy_type)?;
-        let counter = policy_id & ((1u64 << 56) - 1);
+        let counter = self.next_counter.read()?;
         let next = counter
             .checked_add(1)
             .filter(|&n| n < (1u64 << 56))
             .ok_or_else(|| BasePrecompileError::revert(IPolicyRegistry::CounterExhausted {}))?;
         self.next_counter.write(next)?;
+        let policy_id = (policy_type_u8 as u64) << 56 | counter;
         let packed = Self::encode_packed(admin, policy_type_u8);
         self.policies.at_mut(&policy_id).write(packed)?;
 
@@ -125,7 +129,7 @@ impl PolicyRegistryStorage<'_> {
             return Err(BasePrecompileError::revert(IPolicyRegistry::BatchTooLarge {}));
         }
         let policy_id = self.create_policy(admin, policy_type)?;
-        let policy_type_u8 = policy_type as u8;
+        let policy_type_u8 = policy_type.as_discriminant()?;
         if !accounts.is_empty() {
             let caller = self.storage.caller();
             for account in &accounts {
--- a/crates/common/precompiles/src/policy/storage.rs
+++ b/crates/common/precompiles/src/policy/storage.rs
@@ -1,21 +1,1130 @@
-use alloy_primitives::{Address, address};
+use alloc::vec::Vec;
+
+use alloy_primitives::{Address, U256, address};
 use base_precompile_macros::contract;
-use base_precompile_storage::{Handler, Mapping, Result};
+use base_precompile_storage::{BasePrecompileError, Handler, Mapping, Result};
+
+use super::{IPolicyRegistry, IPolicyRegistry::PolicyType};
 
 /// Storage layout for the `PolicyRegistry` precompile.
 ///
 /// Slots are append-only — never reorder across hardforks.
 #[contract(addr = Self::ADDRESS)]
 pub struct PolicyRegistryStorage {
-    pub members: Mapping<u64, Mapping<Address, bool>>, // slot 0
+    pub policies: Mapping<u64, U256>,                  // slot 0
+    pub members: Mapping<u64, Mapping<Address, bool>>, // slot 1
+    pub pending_admins: Mapping<u64, Address>,         // slot 2
+    /// Global monotonic counter for the low 56 bits of all custom policy IDs.
+    /// Intentionally shared across ALLOWLIST and BLOCKLIST types — the type
+    /// discriminator is encoded in the top byte, so both types draw from the
+    /// same 56-bit space without collision.
+    pub next_counter: u64, // slot 3
 }
 
 impl PolicyRegistryStorage<'_> {
     /// Singleton precompile address for the `PolicyRegistry`.
     pub const ADDRESS: Address = address!("b030000000000000000000000000000000000000");
 
-    /// Returns `true` if `account` is authorized to send tokens under `policy_id`.
-    pub(super) fn is_authorized(&self, policy_id: u64, account: Address) -> Result<bool> {
-        self.members.at(&policy_id).at(&account).read()
+    /// Built-in policy ID that always authorizes every account.
+    pub const ALWAYS_ALLOW_ID: u64 = 0;
+    /// Built-in policy ID that always rejects every account.
+    pub const ALWAYS_BLOCK_ID: u64 = 1;
+
+    const ALLOWLIST_TYPE: u8 = PolicyType::ALLOWLIST as u8;
+    const BLOCKLIST_TYPE: u8 = PolicyType::BLOCKLIST as u8;
+
+    /// Maximum number of accounts accepted in any single membership call.
+    pub const MAX_ACCOUNTS: usize = 64;
+
+    /// Packs `admin` and `policy_type` into a single U256 storage word.
+    ///
+    /// Layout: `[255:168]` reserved (zero) | `[167:8]` admin (160 bits) | `[7:0]` `PolicyType`.
+    fn encode_packed(admin: Address, policy_type: u8) -> U256 {
+        let mut word = [0u8; 32];
+        word[12..32].copy_from_slice(admin.as_slice());
+        (U256::from_be_slice(&word) << 8) | U256::from(policy_type)
+    }
+
+    fn decode_admin(packed: U256) -> Address {
+        let bytes = (packed >> 8usize).to_be_bytes::<32>();
+        Address::from_slice(&bytes[12..])
+    }
+
+    const fn decode_type(packed: U256) -> u8 {
+        packed.to_be_bytes::<32>()[31]
+    }
+
+    fn require_write(&self) -> Result<()> {
+        if self.storage.is_static() {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::StaticCallNotAllowed {}));
+        }
+        Ok(())
+    }
+
+    fn require_custom(&self, policy_id: u64) -> Result<U256> {
+        let packed = self.policies.at(&policy_id).read()?;
+        if packed == U256::ZERO {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::PolicyNotFound {}));
+        }
+        Ok(packed)
+    }
+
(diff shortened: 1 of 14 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,638 @@
...
+    fn decode_admin(packed: U256) -> Address {
+        let bytes = (packed >> 8usize).to_be_bytes::<32>();
+        Address::from_slice(&bytes[12..])
+    }
+
+    const fn decode_type(packed: U256) -> u8 {
+        packed.to_be_bytes::<32>()[31]
+    }
+
+    fn require_custom(&self, policy_id: u64) -> Result<U256> {
```
rayyan224, line 68: These methods, that we added should be added to IPolicyRegistry interface then PolicyRegistryStorage should implement IPolicyRegistry. That way the token factory can use this

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,638 @@
...
 impl PolicyRegistryStorage<'_> {
     /// Singleton precompile address for the `PolicyRegistry`.
     pub const ADDRESS: Address = address!("b030000000000000000000000000000000000000");
 
-    /// Returns `true` if `account` is authorized to send tokens under `policy_id`.
-    pub(super) fn is_authorized(&self, policy_id: u64, account: Address) -> Result<bool> {
-        self.members.at(&policy_id).at(&account).read()
+    /// Built-in policy ID that always authorizes every account.
+    pub const ALWAYS_ALLOW_ID: u64 = 0;
+    /// Built-in policy ID that always rejects every account.
```
rayyan224, line 30: Instead of this, let's mapp this to enums

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,638 @@
...
+
+        Ok(policy_id)
+    }
+
+    /// Creates a new policy and populates its initial member list.
+    pub fn create_policy_with_accounts(
+        &mut self,
+        admin: Address,
+        policy_type: u8,
+        accounts: Vec<Address>,
```
rayyan224, line 126: Is there a max size of accounts that we have to verify ?

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,638 @@
...
+    ) -> Result<u64> {
+        let policy_id = self.create_policy(admin, policy_type)?;
+        if !accounts.is_empty() {
+            let caller = self.storage.caller();
+            for account in &accounts {
+                self.members.at_mut(&policy_id).at_mut(account).write(true)?;
```
rayyan224, line 136: Let's add a todo: comment to see if we can optimize this writing to storage. I think arrays, themselves are stored contiguous in memory

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,638 @@
...
+        }
+        if policy_type != Self::ALLOWLIST_TYPE && policy_type != Self::BLOCKLIST_TYPE {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::InvalidPolicyType {}));
+        }
+        if admin == Address::ZERO {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::ZeroAddress {}));
+        }
+
+        let counter = self.next_counter.read()?;
+        let next = counter
```
rayyan224, line 97: I think there's a function at the bottom    ?

` pub fn next_policy_id(&self, policy_type: u8) -> Result<u64> {
`

eric-ships: good catch - fb2f2ed6 `create_policy` now calls `self.next_policy_id(policy_type)`, removing duplicate counter logic

crates/common/precompiles/src/policy/storage.rs
eric-ships: is this what you were thinking? 1b7076e7
- added PolicyRegistryTrait mirroring the full IPolicyRegistry interface
- PolicyRegistryStorage implements it
- The trait is exported from the crate root for token factory use

crates/common/precompiles/src/policy/storage.rs
eric-ships: 203ab910 good catch, went with 256 - ~each batch write is an SSTORE~ each store needs its own sstore, will check with @ilikesymmetry

crates/common/precompiles/src/policy/storage.rs
eric-ships: done ba6a65a6

crates/common/precompiles/src/policy/storage.rs
ilikesymmetry: Good callout, can we lower to 64 for now to be conservative. @stevieraykatz actually has best intuition on rate-limiting SSTOREs. Seems like something we can easily change up until release cut and even adjust in future hardforks. Would be inclined to start smaller like 64 to start just to be safe (technically more backwards to increase later than decrease and this is a P2 decision).

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,1175 @@
...
+        add: bool,
+        accounts: &[Address],
+    ) -> Result<Address> {
+        if accounts.len() > Self::MAX_BATCH_ACCOUNTS {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::BatchTooLarge {}));
+        }
+        if self.storage.is_static() {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::StaticCallNotAllowed {}));
+        }
+        let packed = self.require_custom(policy_id)?;
```
rayyan224: Question I think we need to check that, membership old size + batch size is not > then max size here right @eric-ships ?

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,1175 @@
...
+
+/// Rust interface mirroring [`IPolicyRegistry`] for use by the token factory and other consumers.
+pub trait PolicyRegistryTrait {
+    /// See [`PolicyRegistryStorage::create_policy`].
+    fn create_policy(&mut self, admin: Address, policy_type: PolicyType) -> Result<u64>;
+    /// See [`PolicyRegistryStorage::create_policy_with_accounts`].
+    fn create_policy_with_accounts(
+        &mut self,
+        admin: Address,
+        policy_type: PolicyType,
```
rayyan224: Move this to base/crates/common/precompiles/src/common/policy.rs then implement. This is so the token interface can see the ops available in policy!

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,1146 @@
...
+        Ok(policy_id)
+    }
+
+    /// Stages `new_admin` as the pending admin for `policy_id`.
+    ///
+    /// Passing `address(0)` clears a previously-staged transfer per the interface spec.
+    pub fn stage_update_admin(&mut self, policy_id: u64, new_admin: Address) -> Result<()> {
+        if self.storage.is_static() {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::StaticCallNotAllowed {}));
+        }
```
refcell: Would like to refactor but not blocking -- cleanup for a follow-on pr

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,1146 @@
...
+    ) -> Result<u64> {
+        if accounts.len() > Self::MAX_ACCOUNTS {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::BatchTooLarge {}));
+        }
+        let policy_id = self.create_policy(admin, policy_type)?;
+        let policy_type_u8 = policy_type as u8;
+        if !accounts.is_empty() {
+            let caller = self.storage.caller();
+            // TODO: investigate whether accounts can be written as a contiguous
+            // array to storage rather than one SSTORE per address.
```
refcell: if creating policies is not a frequent op, this optimization probably doesn't matter

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,1146 @@
...
+
+    /// Creates a new ALLOWLIST or BLOCKLIST policy, returning its encoded ID.
+    pub fn create_policy(&mut self, admin: Address, policy_type: PolicyType) -> Result<u64> {
+        if self.storage.is_static() {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::StaticCallNotAllowed {}));
+        }
+        let policy_type_u8 = match policy_type {
+            PolicyType::ALLOWLIST | PolicyType::BLOCKLIST => policy_type as u8,
+            _ => return Err(BasePrecompileError::revert(IPolicyRegistry::InvalidPolicyType {})),
+        };
```
refcell: should be a method on `PolicyType`, i.e.

```rust
impl PolicyType {
    pub fn as_u8(&self) -> Result<u8> {
        match self {
            Self::ALLOWLIST | Self::BLOCKLIST => Ok(policy_type as u8),
            _ => return Err(BasePrecompileError::revert(IPolicyRegistry::InvalidPolicyType {})),
        }
    }
}
```

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -1,21 +1,1146 @@
...
+        }
+        let packed = self.require_custom(policy_id)?;
+        let pending = self.pending_admins.at(&policy_id).read()?;
+        if pending == Address::ZERO {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::NoPendingAdmin {}));
+        }
+        let caller = self.storage.caller();
+        if pending != caller {
+            return Err(BasePrecompileError::revert(IPolicyRegistry::Unauthorized {}));
+        }
```
refcell, line 186: these feel like they should be methods too. These nits are not blocking just FYI :) Would rather merge nits and follow on

crates/common/precompiles/src/policy/storage.rs
eric-ships: done c947ba27 - created `as_discriminant`

crates/common/precompiles/src/policy/storage.rs
eric-ships: if they are methods, ideally they aren't on PolicyType, but maybe on a new type PackedPolicy or something, we can figure out in a follow-up

crates/common/precompiles/src/policy/storage.rs
eric-ships: removed 78bb1553

crates/common/precompiles/src/policy/storage.rs
eric-ships: refactored 9f067397 - `require_write` is now the single home for the static call guard. `create_policy`, `finalize_update_admin`, and `require_admin` all call it instead of repeating the check inline.

rayyan224 (approved):
LGTM

PR #2798, opened 2026-05-20
fix(registrar): dedup concurrent registrations for the same signer

leopoldjoy:
## Summary

Closes a TOCTOU race in the prover registrar that causes duplicate `registerSigner` transactions to be submitted for the same signer during a rotation. The on-chain `TEEProverRegistry.registerSigner` is idempotent, so state remained correct, but each duplicate tx still paid the full RISC Zero proof-verification gas and re-emitted `SignerRegistered`.

## Background

During a recent prover rotation the registrar submitted two materially-identical `registerSigner` transactions for the same prover signer back-to-back, e.g. https://sepolia.etherscan.io/tx/0xd8d18cc4175bfba937bfad096bf3f52f8ee627a0f8d3b6202ee333d8823d8ac9 and https://sepolia.etherscan.io/tx/0x1cc07245b5870d41ea8d606cd6958fbafa24a52cda3bdd88a937c35290c048e9.

Root cause: `RegistrationDriver::step()` fans `process_instance` futures out with `buffer_unordered(max_concurrency)` (`driver.rs:244`), and the only duplicate guard inside `try_register` is a TOCTOU check against the on-chain `is_registered` mapping (`driver.rs:500`). When two concurrent tasks resolve to the same signer address — which happens naturally during a rotation in which an old EC2 instance and its replacement briefly back the same enclave signing key — both pass the precheck, both generate (potentially identical) proofs, both grab their own nonces from `NonceManager::next_nonce()`, and both reach the mempool.

A comment block immediately above the concurrent fan-out (`driver.rs:228-229`) already flagged this gap as deferred work.

## Fix

Add a process-local `Arc<Mutex<HashSet<Address>>>` of in-flight registrations on `RegistrationDriver`, reserved at the top of `try_register` before the `is_registered` check and released via an RAII `InFlightGuard` on every exit path: success, error, retry-exhaustion, cancellation drop, and panic.

- The lock's critical section is a single `HashSet::insert` / `HashSet::remove`; no awaits are held under the lock.
- The slot is held across the full registration lifecycle (including the ~20 min Boundless proof generation and the tx-manager confirmation wait), so dedup holds both within a single `step()` cycle and across cycles for the same signer.
- The existing `is_registered` precheck stays in place as a backstop for the "registered by a previous registrar process" case.
- `try_register` is called from exactly one site (`driver.rs:462`), so the blast radius is minimal.

## Tests

Three new tests in `driver::tests`:

1. `try_register_concurrent_same_signer_dedups` — two concurrent `process_instance` calls for the same signer collapse to a single `tx_manager.send()` and a single proof. Uses a `YieldingProofProvider` to force the second future to be polled while the first is mid-proof.
2. `try_register_in_flight_slot_released_after_completion` — sequential calls for the same signer both submit txs, proving the slot is released on the success path.
3. `try_register_in_flight_slot_released_after_failure` — a non-retryable `InsufficientFunds` failure on the first call doesn't permanently block the same signer; the second call still reaches the tx manager.

All 124 tests in `base-proof-tee-registrar` pass. `cargo clippy --all-targets` and `cargo fmt --check` are clean.

## Scope notes

- This patch does not touch `submit_deregistration`, which has a similar race shape. Treating that as a separate, follow-up change.
- No on-chain contract changes. An on-chain `if (isRegisteredSigner[…]) return;` guard was considered but offers little value here because the address can only be derived from the proof's public key *after* `NITRO_VERIFIER.verify` runs — i.e. after the expensive work is already paid for.

### DIFF
--- a/crates/proof/tee/registrar/src/driver.rs
+++ b/crates/proof/tee/registrar/src/driver.rs
@@ -5,7 +5,13 @@
 //! to L1 via the [`TxManager`]. Also detects orphaned on-chain signers (those
 //! no longer backed by a healthy instance) and deregisters them.
 
-use std::{collections::HashSet, error::Error, fmt, sync::Arc, time::Duration};
+use std::{
+    collections::HashSet,
+    error::Error,
+    fmt,
+    sync::{Arc, Mutex},
+    time::Duration,
+};
 
 use alloy_primitives::{Address, Bytes, FixedBytes, hex};
 use alloy_sol_types::SolCall;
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

0x00101010 (approved):

PR #2799, opened 2026-05-20
feat(prover-registrar): decouple discovery from proof generation via spawn-and-reap pipeline

leopoldjoy:
## Summary

Refactor the TEE prover registrar's main loop so discovery and proof generation
run concurrently instead of strictly sequentially. Previously, a single
`try_register` call inside the polling loop could block discovery for the
duration of a Boundless proof (~70 min in production), so a freshly-launched
enclave instance would not be picked up until the in-flight proof for an old
instance completed. After this change, new instances are detected on the next
discovery cycle (~30 s) regardless of in-flight proof work.

## Approach

- The polling loop is now a **spawn-and-reap pipeline**: each registerable
  signer becomes a `tokio::task::JoinSet`-owned task; discovery, reconcile,
  and orphan-deregistration run on every cycle independent of in-flight
  proof tasks.
- **Per-task `tokio_util::sync::CancellationToken`** (child of the
  driver-wide cancel) lets `reconcile_proof_tasks` cooperatively cancel
  proof tasks for instances that have vanished from discovery, without
  abandoning in-flight `tx_manager.send()` calls (which would leak a nonce
  — `NonceGuard::Drop` does not roll back).
- **`AttestationProofProvider` trait now takes `&CancellationToken`** so
  both `DirectProver` and `BoundlessProver` can honor cancel cooperatively
  at phase boundaries; the trait documents the survive-on-drop contract
  per backend.
- **`Arc<tokio::sync::Semaphore>` sized from `max_concurrency`** caps
  the number of concurrent proof tasks; an `in_flight: HashSet<Address>`
  dedupes the same signer across overlapping cycles.
- **Reconcile-cancelled tasks return `Ok(())`** and bump both
  `proof_tasks_cancelled` (intent) and `proof_tasks_completed` (terminal).
- Drain on shutdown is **cooperative only** — no `abort_all` — so a
  cancel during the broadcast-but-not-mined window cannot leak a nonce.

## Test Plan

- `cargo test -p base-proof-tee-registrar --lib`: 156 passing (was 121).
- `cargo test -p base-proof-tee-registrar --lib --features metrics`: 156 passing.
- `cargo test -p base-proof-tee-nitro-attestation-prover --lib`: 17 passing.
- `cargo clippy -p base-proof-tee-registrar -p base-proof-tee-nitro-attestation-prover --all-targets -- -D warnings`: clean (with and without `--features metrics`).
- `cargo fmt --all -- --check`: clean.
- 35 new tests, including end-to-end pipeline behavior driven through the
  real `run()` loop via a `GatedProofProvider` harness — covers fan-out,
  scale-up under existing in-flight load, stale-task cancellation,
  per-signer failure isolation, orphan dereg under load, single-flight
  dedupe across overlapping cycles, attestation-by-index pairing, and
  unhealthy-instance grace-window registration.

## Commits

1. `feat(registrar): add proof-task lifecycle metrics and tokio rt/sync features`
2. `feat(registrar): decouple discovery from proof generation via spawn-and-reap loop`
3. `test(registrar): add spawn-and-reap pipeline tests with gated proof harness`
4. `docs(registrar): fix unresolved intra-doc links across crate`
5. `test(registrar): DRY pipeline tests, parametrize over fan-out/scale-up`
6. `test(registrar): cover panic / proof-failure / mixed-cycle pipeline arms`
7. `fix(prover-registrar): close nonce-gap hazards on shutdown / discovery cancel`
8. `fix(prover-registrar): bound proof concurrency, drop duplicate-signer race, thread cancel through provider trait`
9. `fix(prover-registrar): cycle hygiene fixes for reap, reconcile, and try_register`
10. `refactor(prover-registrar): named RegisterableSigner, run(self) value-API, doc fixes`
11. `test(prover-registrar): cover dedupe, attestation indexing, unhealthy-window via run()`
12. `docs(nitro-attestation-prover): fix unresolved rustdoc link in trait`
13. `refactor(prover-registrar): single gauge update, assert RegisterableSigner invariants, Cargo waterfall`
14. `style(prover-registrar): cargo fmt`

## Notes

- macOS-only: `cargo build --features prove` hits a pre-existing
  `risc0-build-kernel-2.0.1` Metal kernel error unrelated to this branch.
  Builds without `prove` and CI Linux builds are clean.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -80,6 +89,128 @@ pub struct DriverConfig {
...
+    pub reachable_count: usize,
+    /// Total instances returned by discovery (reachable + unreachable).
+    pub total_count: usize,
+    /// Whether the cycle satisfies both the majority guard
+    /// (`reachable * 2 > total`) and the cancellation policy
+    /// (token not cancelled). `true` when `total_count` is zero **and**
+    /// the driver is not cancelled — otherwise `false`, so the orphan
+    /// dereg pass is skipped during shutdown to avoid acquiring nonces
+    /// we don't intend to broadcast.
+    pub ok_to_dereg: bool,
```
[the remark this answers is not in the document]
leopoldjoy: Fixed in `fae7e9f12`. The new doc enumerates both `true` branches (`total_count == 0` OR strict majority quorum) and both `false` branches (cancelled OR quorum failure).

crates/proof/tee/nitro-attestation-prover/src/boundless.rs
```diff
@@ -421,7 +422,24 @@ impl BoundlessProver {
...
+    /// proof recovery).
+    async fn generate_proof(
+        &self,
+        attestation_bytes: &[u8],
+        cancel: &CancellationToken,
+    ) -> Result<AttestationProof> {
+        if cancel.is_cancelled() {
+            return Err(ProverError::Boundless("proof generation cancelled before start".into()));
+        }
         let (client, params) = self.build_client_and_params(attestation_bytes).await?;
```
jackchuma, line 447: is there a reason we can't pass the cancellation token through to `build_client_and_params` and `submit_and_wait` too?

crates/proof/tee/registrar/src/traits.rs
```diff
@@ -7,9 +7,10 @@ use crate::{ProverInstance, Result};
 
 /// Discovers active prover instances from the infrastructure layer.
 ///
-/// The primary implementation is [`AwsTargetGroupDiscovery`], which queries
```
jackchuma, line 10: Nit: this comment change is irrelevant to this PR

crates/proof/tee/registrar/README.md
jackchuma, line 1: Nit: the changes in this file are irrelevant to the PR

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -86,6 +89,129 @@ pub struct DriverConfig {
...
+///   attestations are tolerated (some prover endpoints return additional
+///   enclave attestations beyond the active signer count).
+#[derive(Debug, Clone)]
+pub struct RegisterableSigner {
+    /// Source prover instance, retained so per-signer log lines and
+    /// `PendingTask::instance_id` can attribute the spawned task.
+    pub instance: ProverInstance,
+    /// Signer addresses derived from the instance's enclave public keys.
+    /// Each address gets its own spawned proof task.
+    pub addresses: Vec<Address>,
```
jackchuma: Would it be cleaner for each spawned task to be for a single address?

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -743,6 +1083,460 @@ where
...
+                let instance_owned = entry.instance.clone();
+                let attestation = entry.attestations[idx].clone();
+                let task_cancel = signer_cancel.clone();
+                let instance_id = entry.instance.instance_id.clone();
+
+                let handle = tasks.spawn(async move {
+                    driver
+                        .run_proof_task(instance_owned, signer, idx, attestation, task_cancel)
+                        .await
+                });
```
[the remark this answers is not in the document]
leopoldjoy: Intentional — every log line in `try_register` already carries `signer = %signer_address` and `instance = %instance.instance_id` as inline fields (driver.rs:818, 843, 870, etc.), so grep-style log correlation already works without span propagation.

crates/proof/tee/nitro-attestation-prover/src/boundless.rs
leopoldjoy: The `select!` in `try_register` (driver.rs:911) already drops the entire `generate_proof_for_signer` future on cancel — propagating the token deeper would only change "abandon the future" into "return an explicit Err", which gets logged as a task failure rather than a clean cancel. The trade-off is documented on the trait-level # Cancellation contract and the boundless.rs:425 doc. (for Boundless specifically, any in-flight request is obviously recoverable on the next cycle via the deterministic request-id derivation, so we lose nothing by dropping mid-flight.)

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -211,43 +321,151 @@ where
...
+                }
             }
 
+            // Publish the in-flight gauge once per cycle, after every
+            // path that could mutate `pending` has run. `pending.len()`
+            // is stable across the sleep below — finished tasks
+            // accumulate in the `JoinSet` but only enter `pending` via
+            // reconcile / leave it via reap — so a single update here
+            // is observationally equivalent to one per state change.
+            RegistrarMetrics::proof_tasks_pending().set(pending.len() as f64);
```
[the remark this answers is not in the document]
leopoldjoy: `metrics-0.24.3` `IntoF64` is only implemented for `i8/u8/i16/u16/i32/u32/f32/f64/Duration` (see `metrics-0.24.3/src/common.rs:259`) — `usize/u64/i64` are excluded because their f64 conversion is lossy above 2^53. The cast is forced by the API; the precision floor is purely theoretical at our scale (pending count is bounded by signer fleet size).

crates/proof/tee/nitro-attestation-prover/src/direct.rs
```diff
@@ -53,7 +54,25 @@ impl DirectProver {
...
+    /// blocking thread pool until the backend finishes — `spawn_blocking`
+    /// has no abort signal. This is acceptable because the prover has
+    /// no on-chain side effects.
+    async fn generate_proof(
+        &self,
+        attestation_bytes: &[u8],
+        cancel: &CancellationToken,
+    ) -> Result<AttestationProof> {
+        if cancel.is_cancelled() {
+            return Err(ProverError::Risc0("proof generation cancelled before start".into()));
```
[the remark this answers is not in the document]
leopoldjoy: Acknowledged and intentional — the doc comment at lines 57-67 already covers this trade-off; `DirectProver`'s RISC Zero proof has no on-chain side effects, so a "doomed" proof completing on the blocking pool is harmless waste with no correctness impact, and there are no natural checkpoints inside `spawn_blocking` to insert finer-grained checks without invasive backend hacks.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -743,152 +1091,675 @@ where
...
+            if !task.cancel.is_cancelled() {
+                task.cancel.cancel();
+                RegistrarMetrics::proof_tasks_cancelled().increment(1);
+            }
+        }
+        // NOTE: we drain through `join_next_with_id` (not
+        // `JoinSet::shutdown`) so each terminal outcome flows through
+        // `apply_join_outcome` — keeping the `pending` map and the
+        // proof-task metrics consistent at shutdown.
+        while let Some(joined) = tasks.join_next_with_id().await {
```
jackchuma, line 1696: When shutdown is requested while a proof task is awaiting the initial `registry.is_registered(...).await` (or the post-send registry check), this drain waits for the task to finish even though those side-effect-free RPCs do not observe the cancelled child token. A stalled L1 RPC can therefore keep run() from returning; make the registry awaits cancellation-aware before relying on natural task drain for shutdown

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -743,152 +1091,675 @@ where
...
+    async fn drain_proof_tasks(
+        tasks: &mut JoinSet<Result<()>>,
+        pending: &mut HashMap<task::Id, PendingTask>,
+    ) {
+        for task in pending.values() {
+            // Only count tasks we are cancelling here — already-cancelled
+            // tasks (because `reconcile_proof_tasks` cancelled them earlier
+            // this cycle) were counted at intent time and must not double
+            // count under shutdown.
+            if !task.cancel.is_cancelled() {
```
jackchuma: On normal shutdown the parent `DriverConfig::cancel` has already cancelled every child token, so this condition is false for all pending tasks and `proof_tasks_cancelled` is never incremented for shutdown, despite the metric description. This makes shutdown cancellations invisible unless they were already cancelled by reconcile

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -211,43 +331,151 @@ where
...
-    /// deregister orphans.
+    /// Synchronous single-cycle equivalent of [`Self::run`]: discover →
+    /// resolve addresses → register → deregister orphans.
+    ///
+    /// Retained as a test helper for the synchronous-cycle test suite
+    /// (calldata accounting, majority guard, draining/unhealthy
+    /// behaviour). The production path is [`Self::run`], which spawns
+    /// proof generation into a `JoinSet` so discovery is never blocked
+    /// behind a long-running proof.
+    #[cfg(test)]
```
jackchuma: I'm concerned about code smell here. Why are we maintaining the old synchronous behavior for tests?

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -211,43 +331,151 @@ where
...
+    /// uses [`Self::run`].
+    pub async fn run_arc(self: Arc<Self>) -> Result<()> {
         info!(
             poll_interval = ?self.config.poll_interval,
             registry = %self.config.registry_address,
             "starting registration driver"
         );
 
+        let mut tasks: JoinSet<Result<()>> = JoinSet::new();
+        let mut pending: HashMap<task::Id, PendingTask> = HashMap::new();
```
jackchuma: This task tracking looks more complex than it needs to be. We key `pending` by `tokio::task::Id`, then carry a separate `PendingTask` record so reaping can recover the signer, instance ID, and cancellation token later. Since the lifecycle we actually care about is “one registration task per signer,” could this be keyed by `Address` instead?

For example, `HashMap<Address, PendingRegistration>` plus task results that return their signer would avoid the task-id indirection entirely. It would also make duplicate suppression, cancellation, respawn, and metrics easier to reason about because the map key would match the domain invariant directly: at most one active proof task per signer. The current `task::Id -> PendingTask { signer, ... }` shape makes the scheduler feel more generic than the problem requires

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -211,165 +363,159 @@ where
...
+    /// [`Self::drain_proof_tasks`] for the nonce-gap rationale.
+    ///
+    /// # Ownership
+    ///
+    /// Consumes `self` by value so the API matches every other long-lived
+    /// `*_service::run` in the workspace. Internally the driver is wrapped
+    /// in an [`Arc`] and the spawned proof tasks each hold a clone — see
+    /// [`Self::run_arc`] for the underlying loop and the rationale for the
+    /// shared ownership.
+    pub async fn run(self) -> Result<()> {
```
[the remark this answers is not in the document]
leopoldjoy: Acknowledged — this is intentional and already documented in the `# Ownership` doc on `run` (matches every other `*_service::run` in the workspace; the spawn-and-reap loop needs `Arc<Self>` clones in each task).

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -211,165 +363,159 @@ where
...
+
+    /// Underlying registration loop that powers [`Self::run`].
+    ///
+    /// Takes `self: Arc<Self>` directly because each spawned proof task
+    /// owns an `Arc<Self>` clone so the cycle loop can continue to mutate
+    /// `pending` and `tasks` while proofs run for tens of minutes. Tests
+    /// that need to inspect driver state from outside the task can call
+    /// this method directly with their own `Arc` clone; production code
+    /// uses [`Self::run`].
+    pub async fn run_arc(self: Arc<Self>) -> Result<()> {
```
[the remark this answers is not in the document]
leopoldjoy: Skipping — `pub(crate)` would violate AGENTS.md ("No private or pub(crate) types"), and `run_arc` is a legitimate (if rarely needed) production entry point the test harness leans on for `Arc<Self>` inspection. `run` is the convenience wrapper for the common case.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -211,165 +363,159 @@ where
...
-            return Ok(());
-        }
-
-        let registered_signers = self.registry.get_registered_signers().await?;
-
-        if let Err(e) = self.deregister_orphans(&active_signers, &registered_signers).await {
-            warn!(error = %e, "failed to deregister orphan signers");
-            RegistrarMetrics::processing_errors_total().increment(1);
-        }
+        Self::drain_proof_tasks(&mut tasks, &mut pending).await;
```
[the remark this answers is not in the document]
leopoldjoy: Not actionable — the `up` gauge IS managed by the binary crate: `bin/prover-registrar/src/cli.rs:425` sets it to 1 at startup and `cli.rs:582` calls `RegistrarMetrics::record_shutdown()` after the driver returns. Working as designed.

crates/proof/tee/registrar/src/lib.rs
```diff
@@ -19,8 +19,9 @@ pub use discovery::AwsTargetGroupDiscovery;
 mod driver;
 pub use driver::{
     DEFAULT_MAX_CONCURRENCY, DEFAULT_MAX_TX_RETRIES, DEFAULT_TX_RETRY_DELAY_SECS,
-    DEFAULT_UNHEALTHY_REGISTRATION_WINDOW_SECS, DriverConfig, OnchainRevocationCheck,
-    RegistrationDriver,
+    DEFAULT_UNHEALTHY_REGISTRATION_WINDOW_SECS, DiscoveryResolution, DriverConfig,
+    OnchainRevocationCheck, PendingRegistration, RegisterableSigner, RegistrationDriver,
+    ResolveOutcome,
```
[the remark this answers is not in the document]
leopoldjoy: Deferred — AGENTS.md mandates these types be `pub` + re-exported ("No private or pub(crate) types"), so the re-export itself is rule-driven. Making the fields private would require adding constructors and refactoring ~8 in-crate test construction sites for an internal-only type. Reasonable cleanup for a follow-up API-hygiene pass, but out of scope for this PR.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -211,165 +363,159 @@ where
...
+                    if !self.config.cancel.is_cancelled() {
+                        self.reconcile_proof_tasks(&resolution, &mut tasks, &mut pending);
+                    }
 
-            tokio::select! {
-                () = self.config.cancel.cancelled() => {
-                    info!("registration driver received shutdown signal");
-                    break;
+                    if resolution.ok_to_dereg && !self.config.cancel.is_cancelled() {
+                        if let Err(e) = self.run_orphan_dereg(&resolution.active_signers).await {
```
jackchuma: When an instance is discovered but `resolve_instance` fails during a cycle, `reconcile_proof_tasks` preserves any pending proof task for that instance via `unresolved_instance_ids`, but the orphan pass still uses only `resolution.active_signers`. If the preserved task completes and registers the signer before or during this pass, the signer is absent from `active_signers`, so the majority-guarded orphan cleanup can immediately deregister the newly registered signer. Consider adding preserved pending signers to the protected set or skipping orphan cleanup for inconclusive instances.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -743,334 +1045,939 @@ where
...
+                    signer = ?signer,
+                    instance = ?removed.as_ref().map(|t| t.instance_id.as_str()),
+                    "proof task join error (panic or abort)"
+                );
+                RegistrarMetrics::processing_errors_total().increment(1);
+            }
+        }
+    }
+
+    /// Cancels every pending task cooperatively, awaits them to natural
```
jackchuma, line 1651: Nit: it seems that the comments added to this file are overkill in many places

crates/proof/tee/registrar/src/driver.rs
leopoldjoy: Good catch — fixed in `bb1efa4d4`. New `RegistrationDriver::protected_signers(&resolution, &pending)` builds the union and is plumbed through the orphan-dereg call site; `deregister_orphans` arg renamed `protected_signers` to make the contract explicit. Two new tests (`protected_signers_union_blocks_dereg_of_freshly_registered_signer` + empty-pending sanity contrast) cover the union. 167/167 tests pass with metrics feature.

crates/proof/tee/registrar/src/driver.rs
leopoldjoy: Fair — trimmed three of the most extreme offenders in `62ec80380`: dropped the 16-line obituary for the removed `step()` helper (documented a method that no longer exists), and condensed the gauge-publish + second-reap rationales to one sentence each. Load-bearing cancellation/nonce safety comments kept verbatim.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -86,6 +89,140 @@ pub struct DriverConfig {
...
+    /// which is bounded by [`DriverConfig::max_concurrency`] (typically
+    /// <20) so a single-pass linear scan is cheaper than a second
+    /// reverse `task::Id → Address` map.
+    ///
+    /// Also gates the cleanup in the `Ok((id, Ok(signer)))` happy path:
+    /// after reconcile overwrites a stale (cancelled) entry with a
+    /// fresh one for the same signer, the stale task's terminal
+    /// outcome must NOT evict the fresh entry. The `apply_join_outcome`
+    /// success arm checks `pending[signer].task_id == id` before
+    /// removing so a same-signer respawn within a cycle is safe.
```
mw2000: Can we trim this comment?

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -743,2309 +1036,4986 @@ where
...
+    /// AWS CRL), `Ok(false)` if all certificates are clean.
+    ///
+    /// Layer 1 fails open internally: an on-chain RPC/decode error (for
+    /// example, when this binary is deployed against a verifier contract
+    /// that predates the `revokedCerts` selector) is logged, counted, and
+    /// the call falls through to Layer 2 — never silently disabling AWS CRL
+    /// enforcement. The shared cert chain parse happens once up front; if
+    /// it fails, the error propagates and the caller fail-opens uniformly.
+    /// Confirmed revocations at either layer block registration for the
+    /// instance.
```
mw2000: Can we trim this comment as well

crates/proof/tee/registrar/src/driver.rs
leopoldjoy: Trimmed in `bd2970f98` — 14 → 7 lines. Kept both load-bearing bits (the O(n) recovery rationale gated by `max_concurrency`, and the success-arm `task_id`-match invariant); dropped the redundant elaborations.

crates/proof/tee/registrar/src/driver.rs
leopoldjoy: Trimmed in `bd2970f98` — 30 → 18 lines. Kept the two-layer enumeration (on-chain pre-check vs AWS CRL) and the layer-1-fails-open semantics, since both are non-obvious safety contracts; dropped the prose elaborations around CHAIN-4194 history and parse-error fallthrough.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -743,2309 +1036,4986 @@ where
...
+                };
+                debug!(
+                    task_id = ?id,
+                    signer = %signer,
+                    instance = ?removed.as_ref().map(|t| t.instance_id.as_str()),
+                    superseded = removed.is_none(),
+                    "proof task completed",
+                );
+            }
+            Ok((id, Err(e))) => {
```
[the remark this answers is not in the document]
leopoldjoy: Fixed in `bd2970f98`. Extracted the address-keyed match into `RegistrationDriver::remove_if_task_matches` and routed all three arms (`Ok` success, `Ok(_, Err)`, `JoinError`) through it. The `Err`/`JoinError` arms were correct only by virtue of `find_signer_by_task_id` implicitly filtering on the matching id — fine today, fragile under future refactors of the recovery routine. Added `apply_join_outcome_err_arm_preserves_fresh_entry_when_stale_task_fails_for_same_signer` mirroring the success-arm fresh/stale test.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -211,162 +363,147 @@ where
...
         loop {
-            if let Err(e) = self.step().await {
-                warn!(error = %e, "registration step failed");
-                RegistrarMetrics::processing_errors_total().increment(1);
-            }
+            // Reap before discovery so finished tasks don't linger in
+            // `pending` for an entire cycle and (incorrectly) cause
+            // reconcile to skip spawning a replacement on transient
+            // failure (audit finding #9).
+            Self::reap_finished_tasks(&mut tasks, &mut pending);
```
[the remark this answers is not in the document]
leopoldjoy: Yes, intentional. The double reap is cheap (try-only, no await on the JoinSet) and benign: the first reap drains tasks that completed during the previous poll sleep so reconcile sees an up-to-date `pending` map; the second reap (after `discover_and_resolve`) catches tasks that finished during the typically-multi-second discovery RPC fan-out. Reconcile would handle the stale-task case correctly either way via the address-keyed overwrite + `task_id`-match guard, but the first reap shortens the window where reconcile would needlessly spawn a duplicate task for a signer whose proof just completed.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -743,2309 +1036,4986 @@ where
...
+            // explicit — the string is allocated twice either way
+            // because `instance_owned` is moved into the spawned future
+            // while `PendingRegistration` outlives the move.
+            let instance_id = instance_owned.instance_id.clone();
+            let attestation = entry.attestation.clone();
+            let task_cancel = signer_cancel.clone();
+            let signer = entry.signer;
+            let enclave_index = entry.enclave_index;
+
+            let handle = tasks.spawn(async move {
```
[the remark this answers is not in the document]
leopoldjoy: Intended — captured `Arc<Self>` is the spawned-task lifetime contract (same pattern as every other `*_service` in the workspace). The `Arc` keeps `proof_semaphore`, `in_flight_registrations`, and the proof/registry clients alive until the last spawned task drops it, decoupling task lifetime from `run_arc`. Already documented in the `# Ownership` block on `run`.

crates/proof/tee/nitro-attestation-prover/src/boundless.rs
```diff
@@ -421,7 +422,24 @@ impl BoundlessProver {
...
+    /// once started; dropping the future is safe because any submitted
+    /// request remains discoverable on the next call via the
+    /// deterministic request-id derivation (see the module docs on
+    /// proof recovery).
+    async fn generate_proof(
+        &self,
+        attestation_bytes: &[u8],
+        cancel: &CancellationToken,
+    ) -> Result<AttestationProof> {
+        if cancel.is_cancelled() {
```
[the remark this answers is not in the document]
leopoldjoy: Fixed in `c4480a125`. The doc now correctly describes `generate_proof` as honoring `cancel` once at the top (before build), with build/submit/wait uninterruptible thereafter — and points readers to `generate_proof_for_signer` for the per-probe cancel checks. No code change: callers that need finer-grained cancellation already `select!` against the token externally in `try_register`, so adding a second check in `generate_proof` would be redundant and would complicate the "dropping the future is safe" guarantee that the deterministic request-id recovery relies on.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -592,14 +769,88 @@ where
...
+                    signer = %signer_address,
+                    instance = %instance.instance_id,
+                    "task cancelled before acquiring proof permit"
+                );
+                return Ok(());
+            }
+            permit = Arc::clone(&self.proof_semaphore).acquire_owned() => {
+                // `acquire_owned` only errors when the semaphore is
+                // closed; we never call `close()`, so this is unreachable.
+                permit.expect("proof semaphore is never closed")
```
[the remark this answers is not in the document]
leopoldjoy: Fixed in `f1ccc4a27`. Swapped the `expect()` for a `match` that warn-logs and returns `Ok(())` on the `Err` arm — should never fire today (we never `close()` the semaphore), but removes a panic path from the registration driver loop in case a future refactor introduces `close()` or a third-party semaphore wrapper that closes on drop.

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -743,111 +1029,744 @@ where
...
+            // including CRL" — is enforced locally. Today the outer
+            // `run_arc` loop re-checks `cancel.is_cancelled()` before
+            // calling `reconcile_proof_tasks`, so a `Some(..)` here
+            // would still be discarded; keeping it `None` removes the
+            // non-local dependence on that re-check and matches the
+            // CRL-revoked branch below.
+            if self.config.cancel.is_cancelled() {
+                return Ok(ResolveOutcome { addresses, attestations: None });
+            }
+            match self.check_and_revoke_crls(&all_attestations[0], instance).await {
```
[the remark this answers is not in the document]
leopoldjoy: Fixed in `f1ccc4a27`. Switched to `.first().ok_or_else(RegistrarError::ProverClient { .. })?` so the non-empty invariant is locally visible at the CRL-check site. As you noted, the existing `addresses.is_empty()` and `all_attestations.len() < addresses.len()` guards already enforce it — but they live 30+ lines upstream, so the local `.first()` makes `resolve_instance` indexing-panic-free regardless of how those upstream guards evolve.

mw2000 (approved):

PR #2813, opened 2026-05-21
feat(precompiles): add b20_security token variant

rayyan224:
## Summary

- Adds `b20_security` as a new native precompile variant for tokenized securities (equities, ETFs, commodities), following the same pattern as `b20_stablecoin` (`feat/b20stablecoin`)
- Implements the full `IB20Security` surface on top of `IB20`: announcement bracketing, share-ratio accounting, batch mint/burn, security-specific redeem, and security identifier CRUD (ISIN, CUSIP, FIGI, etc.)
- Wires factory creation for the Security variant; Stablecoin creation remains `InvalidVariant` until `feat/b20stablecoin` merges

## New files — `crates/common/precompiles/src/b20_security/`

| File | Purpose |
|---|---|
| `abi.rs` | `IB20Security` sol! interface — errors, events, all security-specific functions |
| `accounting.rs` | `SecurityAccounting` trait extending `TokenAccounting` with share ratio, identifier map, and announcement-id dedup |
| `storage.rs` | `B20SecurityStorage` — slots 0–10 mirror `B20TokenStorage`; slots 11–13 add share ratio, `Mapping<B256, String>` identifiers, `Mapping<B256, bool>` announcement ids |
| `token.rs` | `B20SecurityToken<S, P>` with `in_announcement: bool` reentrancy guard for `announce` |
| `dispatch.rs` | Full dispatch: security-specific calls first, then inherited `IB20` fallthrough; 12 unit tests |
| `precompile.rs` | `B20SecurityPrecompile::create_precompile` |

## Modified files

- **`activation/storage.rs`** — `B20_SECURITY` constant (`keccak256("base.b20_security")`)
- **`b20/precompile.rs`** — removed `install`/`lookup`; routing now in `b20_token_lookup`
- **`factory/storage.rs`** — Security variant creation enabled; test updated to assert creation succeeds
- **`common/test_utils.rs`** — `SecurityAccounting` impl for `InMemoryTokenAccounting`
- **`provider.rs`** — `b20_token_lookup` single registration point (match on `TokenVariant`, calls `create_precompile` directly — avoids double `from_address` decode and works around `set_precompile_lookup` replacing rather than chaining)
- **`lib.rs`** — exports

## Test plan

- [ ] `cargo test -p base-common-precompiles` — 197 tests pass
- [ ] Factory creates a security token, stores ISIN and 1:1 share ratio, initializes `minimumRedeemable`
- [ ] `B20TokenPrecompile` and `B20SecurityPrecompile` each only route their own variant
- [ ] `b20_token_lookup` match is exhaustive — adding Stablecoin later requires a new arm


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/b20_security/storage.rs
```diff
@@ -0,0 +1,205 @@
...
+    fn set_shares_to_tokens_ratio(&mut self, ratio: U256) -> Result<()> {
+        self.shares_to_tokens_ratio.write(ratio)
+    }
+
+    fn security_identifier(&self, key: B256) -> Result<String> {
+        self.security_identifiers.at(&key).read()
+    }
+
+    fn set_security_identifier(&mut self, key: B256, value: String) -> Result<()> {
+        if value.is_empty() {
```
rayyan224, line 308: Why are we deleting here

crates/common/precompiles/src/activation/storage.rs
```diff
@@ -36,6 +36,10 @@ impl ActivationRegistryStorage<'_> {
     pub const POLICY_REGISTRY: B256 =
         b256!("0xb582ebae03f16fee49a6763f78df482fb11ae73f103ed0d330bbe556aa90a43f");
 
+    /// B20 security precompile feature id (`keccak256("base.b20_security")`).
+    pub const B20_SECURITY: B256 =
+        b256!("0x83d32fab502ae0e8bc4352a117767262cb5e47cc8d67a744008ed4ff03fcf5e6");
+
```
refcell, line 38: We already have this on line 23, though I think this format/naming is much nicer. Can we remove the `SECURITIES_TOKEN_CREATION` one?

crates/common/precompiles/src/b20/precompile.rs
```diff
@@ -14,27 +14,7 @@ use crate::{PolicyHandle, TokenVariant, macros::base_precompile};
...
-        TokenVariant::from_address(*address).map(|variant| match variant {
-            TokenVariant::B20 | TokenVariant::Stablecoin | TokenVariant::Security => {
-                Self::create_precompile(*address)
-            }
-        })
-    }
-
     /// Returns a [`DynPrecompile`] that dispatches to the [`B20Token`] logic at `token_address`.
-    ///
-    /// Used by the precompile-lookup fallback to route calls to any B-20 token address.
```
refcell, line 37: Why are we removing this?

crates/common/precompiles/src/provider.rs
```diff
@@ -13,10 +13,24 @@ use revm::{
 };
 
 use crate::{
-    ActivationRegistry, B20TokenPrecompile, BasePrecompileSpec, PolicyRegistryPrecompile,
-    TokenFactory, bls12_381, bn254_pair,
+    ActivationRegistry, B20SecurityPrecompile, B20TokenPrecompile, BasePrecompileSpec,
+    PolicyRegistryPrecompile, TokenFactory, TokenVariant, bls12_381, bn254_pair,
 };
 
+/// Combined lookup for all B-20 token variants (default and security).
```
refcell, line 20: I don't like that we've hoisted this up here. It deviates from the pattern of delegating the installation to the precompile itself, deviating from the other precompiles. If the issue here is not placing this on the B20TokenPrecompile type, I think we should have a separate `lookup.rs` module where we define a slim type to contain installing the lookup. At least then we can keep this provider module slim.

crates/common/precompiles/src/b20/precompile.rs
refcell: I see - we need a different lookup that maps to other precopmiles. left a comment in the providers.rs module for how i think we should handle this.

crates/common/precompiles/src/b20_security/dispatch.rs
```diff
@@ -0,0 +1,1055 @@
...
+};
+use crate::{
+    ActivationRegistryStorage, B20PolicyType, B20TokenRole, Burnable, Configurable,
+    IB20::{self, IB20Calls as C},
+    Mintable, Pausable, Permittable, Policy, Token, Transferable,
+    macros::{decode_precompile_call, deduct_calldata_cost},
+};
+
+/// WAD precision for share ratio arithmetic: 1e18.
+const WAD: U256 = U256::from_limbs([1_000_000_000_000_000_000, 0, 0, 0]);
```
refcell, line 29: Thinking out loud, I wonder if there's a WAD const we can pull in here from some alloy crate rather than defining magic numbers.

refcell: Doesn't look like it :/

refcell: Maybe we can add a test to assert that 10^18 == WAD const?

crates/common/precompiles/src/b20_security/dispatch.rs
```diff
@@ -0,0 +1,1054 @@
...
+    /// Ensures `caller` has `role`, reverting with `AccessControlUnauthorizedAccount` otherwise.
+    fn ensure_role(&self, caller: Address, role: B256) -> base_precompile_storage::Result<()> {
+        if self.accounting.has_role(role, caller)? {
+            Ok(())
+        } else {
+            Err(BasePrecompileError::revert(IB20::AccessControlUnauthorizedAccount {
+                account: caller,
+                neededRole: role,
+            }))
+        }
```
refcell: I think we could clean this up as:

```rust
        if !self.accounting.has_role(role, caller)? {
            Err(BasePrecompileError::revert(IB20::AccessControlUnauthorizedAccount {
                account: caller,
                neededRole: role,
            }))
        }
        Ok(())
```

crates/common/precompiles/src/b20/precompile.rs
rayyan224: It was cause this, would create for all tokens the same variant

crates/common/precompiles/src/provider.rs
rayyan224: Hmm what would be the difference between these two functions ?

rayyan224: Get Precompile By address ?

crates/common/precompiles/src/b20_security/precompile.rs
```diff
@@ -0,0 +1,28 @@
...
+    /// `token_address`.
+    pub fn create_precompile(token_address: Address) -> DynPrecompile {
+        base_precompile!(alloc::format!("B20SecurityToken@{token_address}"), |ctx, calldata| {
+            B20SecurityToken::with_storage_and_policy(
+                B20SecurityStorage::from_address(token_address, ctx),
+                PolicyHandle::new(ctx),
+            )
+            .dispatch(ctx, &calldata)
+        })
+    }
```
refcell, line 27: I love how concise this is now

rayyan224: 💯

refcell (approved):

PR #2824, opened 2026-05-21
refactor(policy): simplify PolicyType enum and fix built-in policy IDs

eric-ships:
[BOP-133](https://linear.app/coinbase/issue/BOP-133)

## Motivation

`PolicyType` had four variants: `ALWAYS_ALLOW`, `ALWAYS_BLOCK`, `ALLOWLIST`, `BLOCKLIST`. The first two were not real types; they were degenerate cases of the latter two with empty member sets. Nobody could create them via `createPolicy`. This made `as_discriminant()` fallible for no reason and required special-casing throughout the storage layer. `PackedPolicy` also stored the type redundantly, since it is already encoded in the high byte of every policy ID.

## What changed

- `PolicyType` reduced to `BLOCKLIST = 0`, `ALLOWLIST = 1`. `as_discriminant()` is now `const fn` and infallible.
- `PackedPolicy` layout: `[255]` exists flag | `[254:160]` reserved | `[159:0]` admin. Type is no longer stored; all callers derive it from `policy_id >> 56`.
- `ALWAYS_ALLOW_ID = 0` (BLOCKLIST, counter 0) and `ALWAYS_BLOCK_ID = (1<<56)|1` (ALLOWLIST, counter 1) now encode their semantic types correctly in their high bytes with no hardcoded special cases.
- Counter starts at 0. `write_builtins` claims counters 0 and 1, sets `next_counter = BUILTIN_POLICY_COUNT (2)`, and is idempotent. `debug_assert_eq!` guards catch future drift between ID constants and enum discriminants.
- `create_policy` now validates the `PolicyType` discriminant is in range, closing a gap where crafted calldata with an out-of-range `uint8` could store a permanently broken policy.
- All five query methods go through storage uniformly. Only `is_authorized` retains builtin fast-paths for the pre-init case where a token policy field defaults to `ALWAYS_ALLOW_ID` before any policy is created.
- Removed dead code: `POLICY_ALWAYS_ALLOW`, `POLICY_ALWAYS_BLOCK`, `is_builtin_policy`.

- **Keeping `ALWAYS_ALLOW`/`ALWAYS_BLOCK` as enum variants**: implies callers can create them, which was never true, and made `as_discriminant()` return `Result` with an unreachable error path.
- **Keeping the type in `PackedPolicy`**: redundant with the ID encoding, and caused a zero-collision between `{type=0, admin=0}` and an unwritten slot.
- **Old `ALLOWLIST = 0` ordering**: required a hardcoded special-case in `get_policy_type` because `ALWAYS_ALLOW_ID = 0` had high byte 0 = ALLOWLIST but semantics of BLOCKLIST. Flipping to `BLOCKLIST = 0` makes both builtin IDs self-describing.

### DIFF
--- a/crates/common/precompiles/src/policy/storage.rs
+++ b/crates/common/precompiles/src/policy/storage.rs
@@ -373,15 +373,6 @@ impl PolicyRegistryStorage<'_> {
     /// Returns the `PolicyType` of `policy_id`.
     pub fn get_policy_type(&self, policy_id: u64) -> Result<PolicyType> {
         Self::require_well_formed(policy_id)?;
-        // Fast-path for built-in IDs: both encode the correct type in their high byte
-        // (ALWAYS_ALLOW_ID=0 → BLOCKLIST=0, ALWAYS_BLOCK_ID=(1<<56)|1 → ALLOWLIST=1),
-        // but the storage slot may be zero before write_builtins() has been called,
-        // but the storage slot may not have the exists flag set before write_builtins() is called,
-        // so we skip the exists check for them.
-        if policy_id == Self::ALWAYS_ALLOW_ID || policy_id == Self::ALWAYS_BLOCK_ID {
-            return PolicyType::try_from(Self::policy_id_type(policy_id))
-                .map_err(|_| BasePrecompileError::enum_conversion_error());
-        }
         let packed = PackedPolicy::from_raw(self.policies.at(&policy_id).read()?);
         if !packed.exists() {
             return Err(BasePrecompileError::revert(IPolicyRegistry::PolicyNotFound {}));
--- a/crates/common/precompiles/src/policy/storage.rs
+++ b/crates/common/precompiles/src/policy/storage.rs
@@ -369,16 +369,6 @@ impl PolicyRegistryStorage<'_> {
         }
     }
 
-    /// Returns the policy ID that would be assigned to the next policy of `policy_type`.
-    ///
-    /// The counter is global across all policy types, so this is a hint only — the counter
-    /// may advance between this query and the subsequent `create_policy` call.
-    pub fn next_policy_id(&self, policy_type: PolicyType) -> Result<u64> {
-        let discriminant = policy_type.as_discriminant()?;
-        let counter = self.next_counter()?;
-        Ok(Self::make_id(discriminant, counter))
-    }
-
     /// Returns `true` if `policy_id` refers to an existing or built-in policy.
     pub fn policy_exists(&self, policy_id: u64) -> Result<bool> {
         Self::require_well_formed(policy_id)?;
--- a/crates/common/precompiles/src/policy/storage.rs
+++ b/crates/common/precompiles/src/policy/storage.rs
@@ -363,12 +363,9 @@ impl PolicyRegistryStorage<'_> {
         }
     }
 
-    /// Returns `true` if `policy_id` refers to an existing or built-in policy.
+    /// Returns `true` if `policy_id` refers to an existing policy.
     pub fn policy_exists(&self, policy_id: u64) -> Result<bool> {
         Self::require_well_formed(policy_id)?;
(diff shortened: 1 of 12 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -157,6 +182,7 @@ impl PolicyRegistryStorage<'_> {
         // charges warm/cold account-read gas before skipping repeated `set_code`.
         if !self.is_initialized()? {
             self.__initialize()?;
+            self.write_builtins()?;
```
eric-ships, line 183: @refcell can you sanity-check if this is how we want to create the 2 policies? between forks, this would only be initialized once and only once right?

crates/common/precompiles/src/policy/storage.rs
```diff
@@ -143,10 +131,30 @@ impl PolicyRegistryStorage<'_> {
...
+            return Ok(());
+        }
+        let builtin = PackedPolicy::new(Address::ZERO).into_u256();
+        // counter=0 → BLOCKLIST → ALWAYS_ALLOW_ID=0
+        self.policies.at_mut(&Self::ALWAYS_ALLOW_ID).write(builtin)?;
+        // counter=1 → ALLOWLIST → ALWAYS_BLOCK_ID=(1<<56)|1
+        self.policies.at_mut(&Self::ALWAYS_BLOCK_ID).write(builtin)?;
+        self.next_counter.write(2)?;
+        Ok(())
+    }
```
eric-ships, line 164: @amiecorso @ilikesymmetry - assuming creating the built-ins with Address::ZERO is fine here and we don't need to explicitly call **renounce_admin()** for eventing

crates/common/precompiles/src/policy/storage.rs
refcell: Yes exactly. We can actually check this by adding tests if we haven't already

crates/common/precompiles/src/policy/storage.rs
ilikesymmetry: I have no idea what a "built-in" means or what you're asking unfortunately. Can you ELI5?

refcell: I think built-in just means that we're effectively hardcoding / wiring these in.

refcell (approved):

crates/common/precompiles/src/policy/storage.rs
ilikesymmetry: Thanks 👍. I think we should emit `PolicyCreated` event and on second glance, we're missing an `admin` param on the event.

PR #2857, opened 2026-05-22
refactor(policy): align IPolicyRegistry errors with base-std interface

eric-ships:
## What

Removes `InvalidPolicyType` and `MalformedPolicyId` from the Rust `IPolicyRegistry` ABI definition. Neither error appears in [`base/base-std`'s `IPolicyRegistry.sol`](https://github.com/base/base-std/blob/main/src/interfaces/IPolicyRegistry.sol), so Solidity callers currently have no way to decode reverts carrying those selectors.

## Behavioral changes

**Write ops with a malformed policy ID** (type byte > 1) previously reverted with `MalformedPolicyId`. Now they reach `require_custom`, read the zero slot (no policy with type > 1 is ever written), and revert with `PolicyNotFound`. This is the correct error: a malformed ID was never created, so it doesn't exist.

**`policyExists` with a malformed ID** previously reverted with `MalformedPolicyId`. Now the zero-slot read makes `exists()` false, so it returns `false` without reverting. This matches the base-std `Never reverts` contract.

**`createPolicy`'s `InvalidPolicyType` guard** was dead code: `PolicyType` is a Solidity enum, and ABI decoding rejects out-of-range discriminants before the function body runs. The guard is removed; the `_` arm in `createPolicyWithAccounts` becomes `unreachable!()`.

**The `_` fallback in `is_authorized`'s match** becomes `unreachable!()`. Any stored policy has type byte 0 or 1 (enforced at creation time), so the `_` arm is unreachable for any ID that passes the `exists()` check.

Replacing `MalformedPolicyId` with `PolicyNotFound` in `require_well_formed` was considered, but the simpler approach is to just remove `require_well_formed` entirely. Write ops still get `PolicyNotFound` (via the zero-slot read), and read ops become non-reverting for malformed IDs, which is what the base-std interface specifies.

### DIFF
--- a/actions/harness/tests/beryl/policy_registry.rs
+++ b/actions/harness/tests/beryl/policy_registry.rs
@@ -238,8 +238,8 @@ async fn policy_registry_action_tests_cover_policy_lifecycle_and_views() {
         0,
         IPolicyRegistry::PolicyAdminStaged {
             policyId: allowlist_id,
-            previousAdmin: BerylTestEnv::alice(),
-            newAdmin: BerylTestEnv::bob(),
+            currentAdmin: BerylTestEnv::alice(),
+            pendingAdmin: BerylTestEnv::bob(),
         }
         .encode_log_data(),
     );
--- a/crates/common/precompiles/src/policy/abi.rs
+++ b/crates/common/precompiles/src/policy/abi.rs
@@ -15,13 +15,11 @@ sol! {
         error Unauthorized();
         error PolicyNotFound();
         error IncompatiblePolicyType();
-        error InvalidPolicyType();
         error ZeroAddress();
         error NoPendingAdmin();
-        error MalformedPolicyId(uint64 policyId);
 
         event PolicyCreated(uint64 indexed policyId, address indexed creator, PolicyType policyType);
-        event PolicyAdminStaged(uint64 indexed policyId, address indexed previousAdmin, address indexed newAdmin);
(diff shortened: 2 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (approved):

refcell (approved):

PR #2883, opened 2026-05-22
fix(precompiles): skip pause check for privileged mint and burn

rayyan224:
## Summary

- `Burnable::burn` and `Mintable::mint` were calling `ensure_not_paused` unconditionally, even when `privileged=true` — blocking factory bootstrap operations when a feature was paused
- Move `ensure_not_paused` inside the `if !privileged` block for both, matching the existing pattern in `Transferable::transfer`
- Remove stale comment in `burn_blocked` claiming burn pauses always halt every path (no longer true since it calls `burn` with `privileged=true`)
- Update existing paused-feature revert tests to use `privileged=false` (with the role granted), and add two new tests confirming privileged callers succeed when the feature is paused

## Test plan

- [ ] `cargo test -p base-common-precompiles` passes (293 tests)
- [ ] `burn_reverts_when_burn_feature_paused` — non-privileged caller with role reverts on paused BURN
- [ ] `privileged_burn_succeeds_when_burn_feature_paused` — privileged caller bypasses paused BURN
- [ ] `mint_reverts_when_mint_feature_paused` — non-privileged caller with role reverts on paused MINT
- [ ] `privileged_mint_succeeds_when_mint_feature_paused` — privileged caller bypasses paused MINT


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (dismissed):

rayyan224:
https://linear.app/coinbase/issue/BOP-149/audit-security-token

crates/common/precompiles/src/common/ops/burnable.rs
```diff
@@ -67,8 +67,6 @@ pub trait Burnable: Token {
             B20Guards::ensure_token_role::<Self>(self, caller, B20TokenRole::BurnBlocked)?;
         }
         B20Guards::ensure_blocked::<Self>(self, from)?;
-        // Intentional asymmetry: BURN_BLOCKED_ROLE replaces BURN_ROLE, but emergency burn pauses
-        // still halt every burn path, including burnBlocked.
         self.burn(caller, from, amount, true)?;
```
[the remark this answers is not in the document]
ilikesymmetry: @rayyan224 I think we still want to constrain `burnBlocked` to pausability as bot implies so we should make a fix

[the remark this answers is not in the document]
ilikesymmetry: Also implies we may be missing a test that `burnBlocked` is prevented when we are paused if this is passing CI

ilikesymmetry (commented):
I'm not sure if we need need this? Token operations are not paused by default so if a token is paused while in factory phase, that means deployer specifically added a `pause` call _**before**_ its mint or burn calls in `initCalls`, which seems more like an integrator choice. The `initCalls` is meant to be convenience as close to 1:1 with if the admin applied the exact same calls after deployment. This change breaks that principle because now we have divergence where `initCalls` allows pausing then minting/burning, but doing this batch outside of factory context would error.

eric-ships:
Good catch, and the design question is worth being explicit about.

The motivation for moving `ensure_not_paused` inside the `!privileged` branch was to match `Transferable::transfer`, which already has this pattern: when `privileged=true`, it bypasses pause, policy checks, and role checks all at once. Before this PR, `burn` was inconsistent with `transfer` in that respect. There was also a specific issue: `burn_blocked` calls `burn(... privileged=true)` to implement its emergency-override semantics, but the old `burn` checked pause unconditionally, meaning BURN pause would block `burn_blocked` even though it is designed as an admin override. The PR aligns all three (mint, burn, transfer) under the same invariant.

Your point about `initCalls` equivalence is valid though. Since tokens are not paused by default, the only scenario where the pause bypass matters during factory init is if the deployer explicitly puts `pause(MINT/BURN)` before `mint/burn` in their `initCalls`. That is an odd ordering, and arguably an error that should fail the same way it would fail post-deployment.

Two paths forward:

**Option A (keep current):** `privileged` is a superset of all guards, not just roles. The `initCalls` context is explicitly a higher-trust context than post-deployment, analogous to a genesis operation. `transfer` already has this semantics and `burn_blocked` relies on it for compliance overrides.

**Option B (ilikesymmetry's preference):** `privileged` bypasses role checks only. Pause remains an unconditional operational halt for all callers. If we take this path we should also update `transfer` to always check pause for consistency, and change `burn_blocked` to pass `privileged` through to the inner `burn` call rather than hardcoding `true` (which currently makes `burn_blocked` immune to BURN pause unconditionally, regardless of whether the `burn_blocked` caller is in a factory context or not).

My read: Option B gives a cleaner invariant (pause means pause for everyone), but it does require touching `transfer` and `burn_blocked` as well. If the intent is that `initCalls` is 1:1 with post-deployment admin behavior, Option B is the right call.

@rayyan224 what is the intended semantics of `privileged` here? Happy to update the implementation whichever way you decide.

crates/common/precompiles/src/b20_security/token.rs
refcell, line 1: This is a large diff now woah

crates/common/precompiles/src/b20_security/token.rs
```diff
@@ -1,13 +1,24 @@
...
 use crate::{
-    Burnable, Configurable, Mintable, Pausable, Permittable, Policy, RoleManaged, Token,
-    Transferable,
+    B20Guards, B20PolicyType, B20TokenRole, Burnable, Configurable, IB20, Mintable, Pausable,
+    Permittable, Policy, RoleManaged, Token, Transferable,
 };
 
+/// WAD precision for share ratio arithmetic: 1e18.
+pub(super) const WAD: U256 = U256::from_limbs([1_000_000_000_000_000_000, 0, 0, 0]);
+
```
refcell, line 21: Think this needs to be `U256::ZERO`, see: https://github.com/base/base/pull/2889

crates/common/precompiles/src/b20_security/token.rs
rayyan224: It's basically moving out the business logic that was generated in dispatch into token.rs

PR #2895, opened 2026-05-23
fix(precompiles): seed policy builtins independently of bytecode init

haardikk21:
## Summary

The lazy-init gate in `PolicyRegistryStorage::create_policy` ties two distinct concerns to the same condition (`is_initialized()` = "is the precompile account bytecode-empty?"):

1. Installing the precompile's marker bytecode (`__initialize`), so Solidity high-level wrappers pass the `EXTCODESIZE > 0` check, and
2. Seeding built-in policies (`write_builtins`), so `ALWAYS_ALLOW_ID` / `ALWAYS_BLOCK_ID` lookups work and `next_counter` starts at the reserved `BUILTIN_POLICY_COUNT`.

This works fine on a real chain because bytecode arrives at the precompile address only via `__initialize()`. It breaks on any harness that pre-warms the precompile account with bytecode independently of EVM execution — concretely, **anvil/foundry**, which writes a sentinel byte at every base precompile address so `EXTCODESIZE` checks in Solidity wrappers pass before the first call dispatches to the precompile.

In that setup `is_initialized()` returns `true` on the first `create_policy` call, the entire init block is skipped, and the built-in policy slots stay empty while `next_counter` starts at `0` — silently violating the policy registry's published invariants.

## Fix

Decouple the two concerns:

- Keep the bytecode init under the existing `is_initialized()` guard (semantics unchanged on a real chain).
- Call `write_builtins()` unconditionally afterwards.

`write_builtins()` already self-gates via `next_counter >= BUILTIN_POLICY_COUNT`, so this only adds a single SLOAD in the steady state and restores correct semantics for any harness with external bytecode pre-warming.

## Verification

Repro & fix verified against [`base-anvil`](https://github.com/base/base-anvil)'s reth v2.2.0 bump PR, running [`base-std`](https://github.com/base/base-std)'s fork test suite:

- Before fix (rev pinned at current main): 437 / 465 passing — 19 of the 26 failures trace to this bug.
- With this patch applied locally (via cargo `[patch]`): expected to clear the 19 PolicyRegistry-related failures.

(Will update with exact post-fix numbers once the patched binary finishes its run.)

## Why not gate on `next_counter == 0` directly?

`write_builtins()` already has that guard internally:

```rust
if self.next_counter.read()? >= Self::BUILTIN_POLICY_COUNT {
    return Ok(());
}
```

So the call site only needs to drop its outer condition, not duplicate the check.

### DIFF
--- a/crates/common/precompiles/src/policy/storage.rs
+++ b/crates/common/precompiles/src/policy/storage.rs
@@ -171,8 +171,13 @@ impl PolicyRegistryStorage<'_> {
         // charges warm/cold account-read gas before skipping repeated `set_code`.
         if !self.is_initialized()? {
             self.__initialize()?;
-            self.write_builtins()?;
         }
+        // Seed built-ins independently of bytecode presence: harnesses (e.g., anvil/foundry forks)
+        // may pre-warm precompile bytecode to satisfy Solidity's `EXTCODESIZE` check, which would
+        // otherwise short-circuit the lazy init above and leave `policies[ALWAYS_ALLOW_ID]` /
+        // `policies[ALWAYS_BLOCK_ID]` unset. `write_builtins` self-gates via `next_counter`, so
+        // the extra SLOAD on every subsequent `create_policy` is a no-op cost.
+        self.write_builtins()?;
 
         let counter = self.next_counter()?;
         let next = counter.checked_add(1).ok_or_else(BasePrecompileError::under_overflow)?;


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (approved):

PR #2949, opened 2026-05-26
feat(proof): configure Boundless offer pricing

0x00101010:
Adds registrar flags for explicit Boundless offer min/max ETH prices and an optional ramp-up period so proof requests can start at higher marketplace prices. Threads the settings through registrar config into the Boundless prover and applies them to request submission while preserving SDK defaults when unset.

### DIFF
--- a/bin/prover-registrar/src/cli.rs
+++ b/bin/prover-registrar/src/cli.rs
@@ -26,7 +26,10 @@ use base_proof_tee_registrar::{
     RegistrarMetrics, RegistrationDriver, RegistryContractClient,
 };
 use base_tx_manager::{BaseTxMetrics, SignerConfig, SimpleTxManager, TxManagerConfig};
-use boundless_market::alloy::signers::local::PrivateKeySigner;
+use boundless_market::{
+    alloy::signers::local::PrivateKeySigner,
+    price_oracle::{Amount, Asset},
+};
 use clap::{Args, Parser, ValueEnum};
 use eyre::WrapErr;
 use tokio_util::sync::CancellationToken;
@@ -193,6 +196,26 @@ struct BoundlessArgs {
     #[arg(long, env = cli_env!("BOUNDLESS_TIMEOUT_SECS"), default_value_t = 600)]
     boundless_timeout_secs: u64,
 
+    /// Minimum Boundless offer price in ETH for each submitted proof request.
+    ///
+    /// Accepts either a plain ETH amount (for example, `0.01`) or an explicit
+    /// ETH amount (for example, `0.01 ETH`). Must be set together with
+    /// `--boundless-max-price-eth`.
+    #[arg(long, env = cli_env!("BOUNDLESS_MIN_PRICE_ETH"))]
+    boundless_min_price_eth: Option<String>,
+
+    /// Maximum Boundless offer price in ETH for each submitted proof request.
+    ///
+    /// Accepts either a plain ETH amount (for example, `0.03`) or an explicit
+    /// ETH amount (for example, `0.03 ETH`). Must be greater than or equal to
+    /// `--boundless-min-price-eth`.
+    #[arg(long, env = cli_env!("BOUNDLESS_MAX_PRICE_ETH"))]
+    boundless_max_price_eth: Option<String>,
+
+    /// Optional duration for the Boundless offer price to ramp from min to max.
+    #[arg(long, env = cli_env!("BOUNDLESS_OFFER_RAMP_UP_PERIOD_SECS"))]
+    boundless_offer_ramp_up_period_secs: Option<u32>,
+
     /// Maximum number of deterministic request-ID slots to probe when
     /// recovering in-flight proofs after an instance rotation.
     #[arg(
@@ -271,6 +294,16 @@ fn parse_image_id(s: &str) -> std::result::Result<[u32; 8], RegistrarError> {
     Ok(id)
 }
 
+/// Parse an ETH-denominated Boundless offer price.
+fn parse_boundless_eth_amount(field: &str, s: &str) -> std::result::Result<Amount, RegistrarError> {
+    let amount = Amount::parse(s, Some(Asset::ETH))
+        .map_err(|e| RegistrarError::Config(format!("{field}: {e}")))?;
+    if amount.asset != Asset::ETH {
+        return Err(RegistrarError::Config(format!("{field}: expected ETH amount")));
+    }
+    Ok(amount)
+}
+
 impl Cli {
     /// Validate the CLI arguments for logical conflicts and parse into a [`RegistrarConfig`].
     pub(crate) fn into_config(self) -> std::result::Result<RegistrarConfig, RegistrarError> {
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

jackchuma (approved):

bin/prover-registrar/src/cli.rs
```diff
@@ -303,6 +336,40 @@ impl Cli {
...
+                        ));
+                    }
+                    (Some(min_price), Some(max_price)) if max_price.value < min_price.value => {
+                        return Err(RegistrarError::Config(
+                            "--boundless-max-price-eth must be greater than or equal to --boundless-min-price-eth"
+                                .into(),
+                        ));
+                    }
+                    _ => {}
+                }
```
leopoldjoy, line 372: The min/max pairing is validated, but `--boundless-offer-ramp-up-period-secs` can still be set in isolation, which would silently override the SDK's recommended (cycle-count-derived) ramp while leaving prices on SDK defaults — likely not the intent given the PR description frames ramp-up as the slope from configured min to configured max. Suggest a follow-up check after the match block:

```suggestion
                }

                if self.boundless.boundless_offer_ramp_up_period_secs.is_some()
                    && (offer_min_price.is_none() || offer_max_price.is_none())
                {
                    return Err(RegistrarError::Config(
                        "--boundless-offer-ramp-up-period-secs requires both --boundless-min-price-eth and --boundless-max-price-eth"
                            .into(),
                    ));
                }
```

Worth adding a matching test alongside the existing `boundless_offer_min_price_requires_max_price` case.

bin/prover-registrar/src/cli.rs
```diff
@@ -193,6 +196,26 @@ struct BoundlessArgs {
...
+    /// ETH amount (for example, `0.01 ETH`). Must be set together with
+    /// `--boundless-max-price-eth`.
+    #[arg(long, env = cli_env!("BOUNDLESS_MIN_PRICE_ETH"))]
+    boundless_min_price_eth: Option<String>,
+
+    /// Maximum Boundless offer price in ETH for each submitted proof request.
+    ///
+    /// Accepts either a plain ETH amount (for example, `0.03`) or an explicit
+    /// ETH amount (for example, `0.03 ETH`). Must be greater than or equal to
+    /// `--boundless-min-price-eth`.
```
leopoldjoy, line 211: In `boundless-market` 1.4.0 (`request_builder/offer_layer.rs:757-761`), `OfferLayer` only runs `max_price_with_gas` when `params.max_price` is `None`. Once `--boundless-max-price-eth` is set, the configured value is used verbatim as the on-chain `maxPrice` with **no** gas-cost buffer added. Operators should be aware so they don't under-bid fulfillment during gas spikes. Suggested doc expansion:

```suggestion
    /// Maximum Boundless offer price in ETH for each submitted proof request.
    ///
    /// Accepts either a plain ETH amount (for example, `0.03`) or an explicit
    /// ETH amount (for example, `0.03 ETH`). Must be greater than or equal to
    /// `--boundless-min-price-eth`.
    ///
    /// Note: when this is set, the Boundless SDK uses the configured value
    /// verbatim as the on-chain `maxPrice` and does NOT add a gas-cost buffer
    /// (which it does when `maxPrice` is left to the SDK to compute). Bake in
    /// headroom for gas-price volatility between submission and fulfillment.
```

crates/proof/tee/nitro-attestation-prover/src/boundless.rs
```diff
@@ -153,6 +163,29 @@ impl BoundlessProver {
         debug.to_ascii_lowercase().contains(NEEDLE)
     }
 
+    /// Applies optional explicit Boundless offer pricing to request params.
+    fn apply_offer_config(&self, params: RequestParams) -> RequestParams {
```
leopoldjoy, line 167: The both-or-neither invariant on `offer_min_price`/`offer_max_price` only lives in the CLI today. If a future caller constructs `BoundlessProver` directly (test, library use, refactor), an asymmetric configuration would silently emit a partial offer — only one side overridden, the other defaulted by the SDK. A `debug_assert!` keeps the type cheap to construct while documenting and catching drift in tests:

```suggestion
    fn apply_offer_config(&self, params: RequestParams) -> RequestParams {
        // Invariant enforced by the CLI: min/max prices must be both-or-neither.
        debug_assert_eq!(
            self.offer_min_price.is_some(),
            self.offer_max_price.is_some(),
            "offer_min_price and offer_max_price must be set together",
        );
```

bin/prover-registrar/src/cli.rs
```diff
@@ -624,6 +697,9 @@ mod tests {
     const TEST_VERIFIER_URL: &str = "https://gateway.pinata.cloud/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi";
     const TEST_IMAGE_ID: &str =
         "0x0100000002000000030000000400000005000000060000000700000008000000";
+    const TEST_BOUNDLESS_MIN_PRICE_ETH: &str = "0.01";
+    const TEST_BOUNDLESS_MAX_PRICE_ETH: &str = "0.03";
+    const TEST_BOUNDLESS_RAMP_UP_PERIOD_SECS: u32 = 30;
```
leopoldjoy, line 702: Nit: add a string-form constant so the test on line 925 can stop hardcoding `"30"`. `boundless_args()` returns `Vec<&'static str>`, so a `'static` str const is the easiest way to keep the two values in sync.

```suggestion
    const TEST_BOUNDLESS_RAMP_UP_PERIOD_SECS: u32 = 30;
    const TEST_BOUNDLESS_RAMP_UP_PERIOD_SECS_STR: &str = "30";
```

bin/prover-registrar/src/cli.rs
```diff
@@ -825,6 +901,83 @@ mod tests {
...
+    #[rstest]
+    fn boundless_offer_pricing_parses_eth_amounts() {
+        let mut args = boundless_args();
+        args.extend([
+            "--boundless-min-price-eth",
+            TEST_BOUNDLESS_MIN_PRICE_ETH,
+            "--boundless-max-price-eth",
+            TEST_BOUNDLESS_MAX_PRICE_ETH,
+            "--boundless-offer-ramp-up-period-secs",
+            "30",
```
leopoldjoy, line 925: Use the new `_STR` constant from above so this literal can't drift from `TEST_BOUNDLESS_RAMP_UP_PERIOD_SECS`:

```suggestion
            TEST_BOUNDLESS_RAMP_UP_PERIOD_SECS_STR,
```

PR #2953, opened 2026-05-27
fix(derive): assert batch type in channel_reader tests

danilaverbena:
## Summary

Two tests in `channel_reader.rs` used `matches!` to check the decoded batch type but never actually asserted on the result:

```rust
// test_next_batch_succeeds  (line 287)
let res = reader.next_batch().await.unwrap();
matches!(res, Batch::Span(_));   // ← bool silently dropped
assert!(reader.next_batch.is_some());

// test_flush_post_holocene  (line 301)
let res = reader.next_batch().await.unwrap();
matches!(res, Batch::Span(_));   // ← bool silently dropped
assert!(reader.next_batch.is_some());
```

`matches!` is a plain macro that expands to a `match` expression returning `bool`. Because `bool` is not `#[must_use]`, the Rust compiler emits no warning when the value is discarded — so these no-op "assertions" compile and pass silently. If `next_batch()` were ever changed to return a `Batch::Single`, both tests would still pass, giving a false sense of correctness.

## Fix

Wrap both call sites with `assert!()`:

```rust
assert!(matches!(res, Batch::Span(_)));
```

No logic change — only the tests are affected.

### DIFF
--- a/crates/consensus/derive/src/sources/blobs.rs
+++ b/crates/consensus/derive/src/sources/blobs.rs
@@ -26,7 +26,7 @@ where
     /// Fetches blobs.
     pub blob_fetcher: B,
     /// The address of the batcher contract.
-    pub batcher_address: Address,
+    pub batch_inbox_address: Address,
     /// Data.
     pub data: Vec<BlobData>,
     /// Whether the source is open.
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #2955, opened 2026-05-27
batcher/encoder: adjust open channel block_start in prune_safe

danilaverbena:
## Problem

`prune_safe()` removes leading safe blocks from the encoder deque and correctly decrements `block_cursor`, and adjusts `encoded_block_range` on every `ReadyChannel`. However, the currently-open channel's `block_start` field is never updated.

`block_start` is used in `close_current_channel()` to compute the channel's encoded block range:

```rust
let encoded_block_range = open.block_start..self.block_cursor;
```

After pruning `N` blocks, `block_cursor` is correctly offset by `N` but `block_start` retains its pre-prune value. The resulting range is stale by `N` positions — it can be empty, or in extreme cases inverted — which causes two downstream problems:

1. **Wrong `encoded_block_range`** on the resulting `ReadyChannel`: the start index is `N` too large, making the range shorter than it should be.
2. **Silent block skip on replay**: when the channel is later invalidated (e.g. confirmed too late), the replay cursor is set from `encoded_block_range.start`:

   ```rust
   self.block_cursor = self.block_cursor.min(replay_from);
   ```

   With an inflated `replay_from`, the encoder does not rewind far enough and the blocks that were in the open channel at the time of pruning are silently skipped on the re-encoding pass, producing a gap in the submitted L2 chain.

## Fix

Decrement `open.block_start` by `prune_count` (saturating) inside `prune_safe()`, alongside the existing `block_cursor` and `ready_channel` adjustments:

```rust
if let Some(ref mut open) = self.current_channel {
    open.block_start = open.block_start.saturating_sub(prune_count);
}
```

## Test

Added `test_prune_safe_adjusts_open_channel_block_start`: constructs a two-channel scenario, calls `prune_safe` while the second channel is still open, and asserts that `block_start` is correctly decremented and that the resulting `encoded_block_range` covers the right blocks.

### DIFF
--- a/crates/batcher/encoder/src/encoder.rs
+++ b/crates/batcher/encoder/src/encoder.rs
@@ -858,6 +858,16 @@ impl BatchPipeline for BatchEncoder {
             ch.encoded_block_range.start = ch.encoded_block_range.start.saturating_sub(prune_count);
             ch.encoded_block_range.end = ch.encoded_block_range.end.saturating_sub(prune_count);
         }
+
+        // Adjust the open channel's block_start so that close_current_channel()
+        // computes `encoded_block_range = block_start..block_cursor` correctly.
+        // Without this, after pruning N blocks the start index stays N too large,
+        // producing a wrong (potentially empty) encoded_block_range. If that channel
+        // is later invalidated, `replay_from` is set too high and those blocks are
+        // silently skipped on the re-encoding pass.
+        if let Some(ref mut open) = self.current_channel {
+            open.block_start = open.block_start.saturating_sub(prune_count);
+        }
     }
 
     fn da_backlog_bytes(&self) -> u64 {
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #2964, opened 2026-05-27
feat(metering): split opcode gas by contract

niran:
## Summary
- key meterBundle opcode gas usage by executed contract address
- include `contractAddress` on each opcode/precompile gas response entry
- add coverage that two contracts executing SSTORE report separate opcode gas rows
- remove the now-unused `revm-inspectors` dependency from `base-metering`

## Why
Tracking opcode calls and gas costs gives us a lever to compute synthetic metering gas with adjusted pricing that reflects actual load. Splitting storage opcode usage by executed contract makes it possible to price storage-heavy behavior against contract-local state size, rather than only using transaction gas or bundle-level state-root timing.

This synthetic gas is not charged to users. It is intended for metering and throttling so Base can keep chain performance healthy. If the per-contract opcode signal is precise enough, it may also let us reduce or remove state-root calculation from `meterBundle`, lowering the latency required to enforce resource throttling.

## Testing
- `cargo test -p base-metering --lib`
- `cargo test -p base-bundles --lib`
- `cargo check -p base-metering`
- `cargo +nightly fmt --check`
- `cargo clippy -p base-metering --lib -- -D warnings`
- `cargo clippy -p base-bundles --lib -- -D warnings`
- `git diff --check`

type=routine
risk=low
impact=sev5

### DIFF
--- a/crates/execution/metering/src/inspector.rs
+++ b/crates/execution/metering/src/inspector.rs
@@ -90,30 +90,24 @@ impl MeteringInspector {
         std::mem::take(&mut self.precompile_gas)
     }
 
-    /// Records nested call/create gas so the parent opcode is charged only its own overhead.
+    /// Subtracts forwarded call/create gas so the parent opcode is charged only its own overhead.
     ///
-    /// The parent opcode's `step_end` sees overhead plus callee execution gas. Nested opcode
-    /// execution is already metered in the callee frame, so subtract it from the parent opcode
-    /// when CALL/CREATE-family opcodes are explicitly metered.
-    fn record_nested_gas_used(
+    /// The parent opcode's `step_end` runs before revm refunds unused child-frame gas, so it sees
+    /// overhead plus the full forwarded frame gas limit. By `call_end`/`create_end`, the parent
+    /// opcode frame has already been popped, so subtract from the accumulated opcode entry.
+    fn subtract_forwarded_gas(
         &mut self,
         contract_address: Address,
         opcode_value: u8,
-        gas_used: u64,
+        gas_limit: u64,
     ) {
         let Some(opcode) = OpCode::new(opcode_value) else { return };
         if !self.metered_opcodes.contains(&opcode) {
             return;
         }
 
-        if let Some(frame) = self
-            .opcode_frames
-            .iter_mut()
-            .rev()
-            .find(|frame| frame.contract_address == contract_address && frame.opcode == opcode)
-        {
-            frame.nested_gas_used = frame.nested_gas_used.saturating_add(gas_used);
-        }
+        let entry = self.opcode_gas.entry((contract_address, opcode)).or_default();
+        entry.gas_used = entry.gas_used.saturating_sub(gas_limit);
     }
 }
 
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/metering/src/inspector.rs
```diff
@@ -65,41 +89,108 @@ impl MeteringInspector {
     pub(crate) fn take_precompile_gas(&mut self) -> HashMap<Address, PrecompileGasUsage> {
         std::mem::take(&mut self.precompile_gas)
     }
+
+    /// Records nested call/create gas so the parent opcode is charged only its own overhead.
+    ///
+    /// The parent opcode's `step_end` sees overhead plus callee execution gas. Nested opcode
+    /// execution is already metered in the callee frame, so subtract it from the parent opcode
+    /// when CALL/CREATE-family opcodes are explicitly metered.
+    fn record_nested_gas_used(
```
wlawt: can we add a dedicated unit test for this fn? 

from claude:

> A test for record_nested_gas_used would verify that when Contract A executes a CALL to Contract B, the gas attributed to A's CALL opcode is only the CALL overhead (stipend, memory expansion, account creation cost) — not the gas B consumed internally.
> Concretely:
> 1. Deploy a "caller" contract and a "callee" contract
> 2. Caller calls Callee, which does an SSTORE (expensive)
> 3. Meter with ["CALL", "SSTORE"]
> 4. Assert:
> - Callee's SSTORE gas is attributed to Callee's address
> - Caller's CALL gas is attributed to Caller's address
> - CALL.gas_used ≈ CALL overhead (~100-2600 gas), not CALL overhead + Callee's full execution (~22k+)
> - CALL.gas_used < SSTORE.gas_used — if nested gas wasn't subtracted, CALL would include SSTORE's cost and be larger
> Without record_nested_gas_used, the CALL opcode's step_end would see gas_before - gas_after which includes everything the callee spent. The subtraction is what isolates the opcode's own cost.
> An inspector-level unit test could also work — mock step(CALL) → call_end(gas_spent=20000) → step_end() and verify the frame's nested_gas_used was subtracted from the delta. Faster than the full harness, more surgical.

wlawt (approved):

PR #2990, opened 2026-05-27
fix: enable storage read tracking in erc7562Tracer

stephancill:
One-line fix to enable storage read tracking in the native `erc7562Tracer`.

## Problem

`accessedSlots.reads` is always empty for SLOAD operations because `from_geth_erc7562_config()` in revm-inspectors starts from `Self::none()` which sets `record_state_diff: false`. The SLOAD handler depends on `step.storage_change`, which is only populated when `record_state_diff` is enabled.

## Fix

Cargo patch on `revm-inspectors` v0.34.3 adding `.set_state_diffs(true)`:
https://github.com/stephancill/revm-inspectors/commit/c8643d7

## Why

ERC-4337 bundlers need `(slot, initial_value)` pairs from SLOAD to enforce ERC-7562 storage access rules. Without this fix, the native `erc7562Tracer` cannot replace custom JS tracers (which are 4-100x slower due to per-opcode JS interpretation). geth's implementation already tracks reads correctly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -284,7 +284,10 @@ base-runtime = { path = "crates/utilities/runtime", default-features = false }
 revm = { version = "34.0.0", default-features = false }
 revm-bytecode = { version = "8.0.0", default-features = false }
 revm-database = { version = "10.0.0", default-features = false }
-revm-inspectors = { version = "0.34.3", default-features = false }
+# NOTE: 0.34.4 includes fix for erc7562Tracer accessedSlots.reads (paradigmxyz/revm-inspectors#451).
+# The fix was only backported to 0.34.x. If upgrading to 0.35.x-0.39.x, request a cherry-pick
+# from the reth team: https://github.com/paradigmxyz/revm-inspectors/commits/v0.34.4/
+revm-inspectors = { version = "0.34.4", default-features = false }
 revm-precompile = { version = "32.0.0", default-features = false }
 revm-primitives = { version = "22.0.0", default-features = false }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -602,3 +602,6 @@ toml = { version = "1.0", default-features = false }
 ark-ff = { version = "0.5.0", default-features = false }
 unsigned-varint = { version = "0.8", default-features = false }
 ark-bls12-381 = { version = "0.5.0", default-features = false }
+
+[patch.crates-io]
+revm-inspectors = { git = "https://github.com/stephancill/revm-inspectors.git", rev = "c8643d7" }
```
refcell: This should be upstreamed into revm-inspectors not inserted as a patch here.

refcell (changes requested):

Cargo.toml
stephancill: yeah didn't mean to open a PR for this, trying to verify it works by deploying it to dev first

stephancill:
Note: the upstream fix ([paradigmxyz/revm-inspectors#451](https://github.com/paradigmxyz/revm-inspectors/pull/451)) was backported to v0.34.4 by the reth team. If base/base upgrades to a newer minor release (e.g. 0.35.x - 0.39.x), the fix may not be present — the reth team offered to cherry-pick to those versions on request. See: https://github.com/paradigmxyz/revm-inspectors/commits/v0.34.4/

haardikk21 (approved):

Cargo.toml
stephancill: merged upstream!

PR #2998, opened 2026-05-28
derive/frame_queue: fix usize underflow in prune() when queue is empty

esazonova:
## Problem

The Holocene-era `prune()` method in `FrameQueue` iterates with:

```rust
while i < self.queue.len() - 1 { ... }
```

When `self.queue` is empty, `self.queue.len()` is `0`, and the subtraction wraps:

- **Debug builds**: Rust panics immediately with an arithmetic-overflow check.
- **Release builds**: `0usize - 1` evaluates to `usize::MAX`, so the `while` condition is trivially true (`0 < usize::MAX`). The loop body then indexes `self.queue[0]`, which panics with an out-of-bounds access.

## How is the empty-queue path reached?

`prune()` is called from `load_frames()` after extending the queue with newly parsed frames:

```rust
self.queue.extend(frames);
self.prune(origin);
```

If `Frame::parse_frames` returns zero frames — e.g. a batch transaction whose payload contains only the `0x00` derivation version prefix and no actual frame bodies — `frames` is empty, the queue remains empty, and `prune()` is called.

## Fix

Rewrite the loop condition to avoid the subtraction entirely:

```rust
// Before (underflows when len == 0)
while i < self.queue.len() - 1 { ... }

// After (safe: trivially false when len == 0 or 1)
while i + 1 < self.queue.len() { ... }
```

The semantics are identical for non-empty queues; the only change is that `i + 1 < 0` is never evaluated as unsigned arithmetic.

## Test

Added `test_prune_empty_queue_no_panic`: activates Holocene, leaves the frame queue empty, and asserts that `prune()` completes without panicking.

### DIFF
--- a/crates/consensus/derive/src/stages/frame_queue.rs
+++ b/crates/consensus/derive/src/stages/frame_queue.rs
@@ -65,7 +65,7 @@ where
         }
 
         let mut i = 0;
-        while i < self.queue.len() - 1 {
+        while i + 1 < self.queue.len() {
             let prev_frame = &self.queue[i];
             let next_frame = &self.queue[i + 1];
             let extends_channel = prev_frame.id == next_frame.id;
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #3006, opened 2026-05-28
Use hash-pinned L2 reads during derivation reset

niran:
## Problem
During the duplicate-blob investigation, base-consensus was observed entering a reset loop where the safe head stayed unchanged while the derivation pipeline failed with L2 BlockNotFound errors for blocks that existed onchain. The vulnerable path was using L2 block numbers to recover parent state during derivation reset and payload-attribute preparation.

L2 block numbers are not a stable identifier across a reorg or transient provider/engine inconsistency. If the local execution provider has not exposed the expected number mapping yet, or the number briefly resolves against a different view of the chain, a number-based lookup can fail or return the wrong block even though the hash-pinned parent that base-consensus is trying to derive from is known. That keeps derivation from making progress, which can keep the batcher channel open longer and contribute to repeated blob submission behavior.

This PR narrows the fix to matching op-node behavior for the affected reads: once base-consensus has a parent block hash, it uses that hash as the source of truth for parent L2 block info and system config reads. This does not mask a deeper provider issue if the execution provider also cannot serve the block by hash, but it removes the avoidable number-aliasing failure mode from the reset and attributes paths.

## Summary
- Fetch parent system config by L2 parent hash when preparing payload attributes.
- Walk reset state backward by parent hash and load the reset system config by hash.
- Add provider and test support for hash-based L2 block/system config reads.

## Tests
- cargo test -p base-consensus-derive -p base-consensus-providers -p base-consensus-node

type=routine
risk=low
impact=sev5

### DIFF
--- a/crates/consensus/derive/src/attributes/stateful.rs
+++ b/crates/consensus/derive/src/attributes/stateful.rs
@@ -75,7 +75,7 @@ where
 
         let mut sys_config = self
             .config_fetcher
-            .system_config_by_number(l2_parent.block_info.number, Arc::clone(&self.rollup_cfg))
+            .system_config_by_hash(l2_parent.block_info.hash, Arc::clone(&self.rollup_cfg))
             .await
             .map_err(Into::into)?;
 
@@ -376,7 +376,7 @@ mod tests {
         let l1_cfg = Arc::new(Sepolia::l1_config());
         let l2_number = 1;
         let mut fetcher = TestSystemConfigL2Fetcher::default();
-        fetcher.insert(l2_number, SystemConfig::default());
+        fetcher.insert_by_hash(B256::ZERO, SystemConfig::default());
         let mut provider = TestChainProvider::default();
         let header = Header::default();
         let hash = header.hash_slow();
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

niran:
I don't think this is the actual root cause, so closing for now.

PR #3013, opened 2026-05-28
ci(release): Split Binary Builds By Artifact

refcell:
## Summary

The failing Create RC run lost communication with the hosted `ubuntu-24.04-arm` runner during the native binary build. Instead of throttling cargo parallelism, this splits the release binary build matrix by artifact.

`build-binaries` now runs one `(target, binary)` job for each release artifact, so the four binaries for each target can build independently and in parallel. Each job still packages, checksums, signs, attests, and uploads the same asset names expected by `baseup`:

- `<binary>-<tag>-<target>.tar.gz`
- `<binary>-<tag>-<target>.tar.gz.sha256`
- `<binary>-<tag>-<target>.tar.gz.asc`

The final publish job still downloads `binaries-*` with `merge-multiple: true` and uploads all release assets to GitHub.

## Validation

- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.7 .github/workflows/build-release.yml .github/workflows/create-rc.yml .github/workflows/publish-release.yml`
- `git diff --check`

PR #3014, opened 2026-05-28
ci(release): Split Binary Builds By Artifact

refcell:
## Summary

Backport of #3013 for `releases/v1.0.0`.

The failing Create RC run lost communication with the hosted `ubuntu-24.04-arm` runner during the native binary build. Instead of throttling cargo parallelism, this splits the release binary build matrix by artifact.

`build-binaries` now runs one `(target, binary)` job for each release artifact, so the four binaries for each target can build independently and in parallel. Each job still packages, checksums, signs, attests, and uploads the same asset names expected by `baseup`:

- `<binary>-<tag>-<target>.tar.gz`
- `<binary>-<tag>-<target>.tar.gz.sha256`
- `<binary>-<tag>-<target>.tar.gz.asc`

The final publish job still downloads `binaries-*` with `merge-multiple: true` and uploads all release assets to GitHub.

## Validation

- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.7 .github/workflows/build-release.yml .github/workflows/create-rc.yml .github/workflows/publish-release.yml`
- `git diff --check origin/releases/v1.0.0...HEAD`

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -15,11 +15,6 @@ on:
 
 env:
   REGISTRY_IMAGE: ghcr.io/${{ github.repository_owner }}/node-reth-dev
-  RELEASE_BINS: |
-    base
-    base-reth-node
-    base-consensus
-    basectl
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


haardikk21:
We do not need a backport since the workflow can be ran off a branch different from its target. We should close this and just do #3013 - also helps avoid an unnecessary RC for a finalized release

PR #2996, opened 2026-05-28
feat(proof): configure boundless offer lock timeout

leopoldjoy:
Adds `--boundless-offer-lock-timeout-secs` to the prover-registrar CLI, threading the value through `BoundlessConfig` and `BoundlessProver` into `OfferParams::lock_timeout`. Mirrors the wiring introduced in #2949 for offer pricing fields.

## Why

`lockTimeout` bounds both:

1. the window in which any prover may **lock** the request (acquiring exclusive fulfillment rights against the stake bond), and
2. the **delivery deadline** by which the locking prover must produce the proof before forfeiting that bond and the request opening to permissionless secondary fulfillment.

Both events fire at the same on-chain instant: `rampUpStart + lockTimeout`.

With `Offer.timeout` left unset, the Boundless SDK derives `Offer.timeout = 2 * lockTimeout` (`boundless-market` `request_builder/offer_layer.rs`), so a `lockTimeout = 600 s` deployment yields a `1200 s` total on-chain request lifetime with a symmetric secondary fallback window.

## Client-side timeout bump

This PR also bumps `--boundless-timeout-secs` default `600 → 1260`.

That flag is a **purely client-side** poll budget on `BoundlessProver::wait_for_request_fulfillment` — it is **not** submitted on-chain. It must exceed the on-chain `Offer.timeout` plus headroom for clock skew, RPC latency, and indexer lag past the on-chain `Fulfilled` event; otherwise the registrar gives up before the secondary fallback window even opens.

`1260 s` covers a `600 s` `lockTimeout` deployment (`1200 s` SDK-derived on-chain timeout) with `~60 s` of headroom.

The docstring on `--boundless-timeout-secs` is also reworded to make the client-side vs. on-chain distinction explicit, since the previous text ("Proof generation timeout in seconds") was ambiguous.

## Deployment

Operators that want the 10-minute lock-timeout behaviour should set:

```
BASE_REGISTRAR_BOUNDLESS_OFFER_LOCK_TIMEOUT_SECS=600
```

When unset, the Boundless SDK falls back to its cycle-count-derived recommendation, matching the behaviour of the other optional flags introduced in #2949.

## Tests

- `boundless_offer_lock_timeout_parses` — CLI flag round-trips into `BoundlessConfig`.
- `boundless_timeout_default_covers_default_lock_timeout` — asserts the `1260 s` default.
[...]

PR #3019, opened 2026-05-28
feat(proof): configure boundless offer bidding-start delay

leopoldjoy:
Adds `--boundless-offer-bidding-start-delay-secs` to the prover-registrar CLI, threading the value through `BoundlessConfig` and `BoundlessProver` into `OfferParams::bidding_start`. Mirrors the wiring introduced in #2949 (offer pricing) and #2996 (lock timeout) for the other offer fields.

## What this controls

`bidding_start` (`Offer.rampUpStart` on-chain) is the moment a request becomes biddable. Before this instant the request is published on-chain but no prover can lock it.

The Boundless SDK auto-derives this value as roughly the guest program's executor time (`cycle_count / 1 MHz`, capped at 1 hour), creating a 'discovery window' that ensures provers have time to find the request and run the executor before bidding opens. This window is visible on the Boundless explorer as the flat period preceding the price ramp.

For our 314M-cycle workload that derived window is ~5 min 15 sec — a substantial chunk of the observed ~21 min total turn-around.

## Why expose it as an explicit flag

Setting `--boundless-offer-bidding-start-delay-secs=0` eliminates the discovery window entirely:

- Bidding opens immediately at request submission.
- The fastest prover can lock as soon as it has executed the program.
- End-to-end submission→fulfillment latency drops to roughly `executor_time + proving_time` (~15 min for our workload, down from ~21 min).

Trade-offs:

- **Higher effective price**: with no flat period, provers don't get a chance to compete on price during a ramp; you'll typically pay closer to `max_price`. (Mitigated by setting `min_price == max_price` anyway in a 'fast lane' configuration.)
- **Smaller prover set**: only provers fast enough to discover, execute, and lock immediately will win these requests. With a longer flat period more provers can compete.
- **Possible no-lock fallthrough**: if no prover wins in time, the request waits through the rest of `lockTimeout` before opening to secondary fulfillment, which is slower than the discovery-window default.

## Implementation note: per-request `now()`

`OfferParams::bidding_start` is an **absolute** Unix timestamp, not a delay. The new field is exposed as a *delay* and converted to an absolute timestamp per request inside `apply_offer_config` (`now_unix_secs() + delay`). This avoids the failure mode where a value captured at registrar startup ends up far in the past after hours of uptime.

The helper `BoundlessProver::now_unix_secs()` is factored out so tests can assert the computed timestamp falls within a `(before, after)` window bracketing the call.

## Combined recipe for fastest turn-around

Together with the offer pricing knobs from #2949 and the lock-timeout knob from #2996, operators can now run a 'fast lane' configuration that skips the auction entirely:

```
BASE_REGISTRAR_BOUNDLESS_OFFER_BIDDING_START_DELAY_SECS=0     # no discovery window
BASE_REGISTRAR_BOUNDLESS_OFFER_RAMP_UP_PERIOD_SECS=0          # no price ramp
BASE_REGISTRAR_BOUNDLESS_MIN_PRICE_ETH=<X>                    # same as max
BASE_REGISTRAR_BOUNDLESS_MAX_PRICE_ETH=<X>                    # high enough to attract fast provers
BASE_REGISTRAR_BOUNDLESS_OFFER_LOCK_TIMEOUT_SECS=<proving_time + buffer>
```

## Tests

- `boundless_offer_bidding_start_delay_parses` — CLI flag round-trips into `BoundlessConfig`.
- `apply_offer_config_sets_bidding_start_alone` — `bidding_start_delay` is enough on its own to emit an offer override (early-return guard updated); resulting `bidding_start` falls within `[before+delay, after+delay]`.
- `apply_offer_config_bidding_start_uses_current_clock` — non-zero delay is offset from the current clock at call time, not captured at startup.
- Existing `apply_offer_config_preserves_default_when_unset` / `apply_offer_config_sets_explicit_prices` extended to cover the new field.

cc #2949 #2996

### DIFF
--- a/bin/prover-registrar/src/cli.rs
+++ b/bin/prover-registrar/src/cli.rs
@@ -226,9 +226,18 @@ struct BoundlessArgs {
     #[arg(long, env = cli_env!("BOUNDLESS_MAX_PRICE_ETH"))]
     boundless_max_price_eth: Option<String>,
 
-    /// Optional duration for the Boundless offer price to ramp from min to max.
-    #[arg(long, env = cli_env!("BOUNDLESS_OFFER_RAMP_UP_PERIOD_SECS"))]
-    boundless_offer_ramp_up_period_secs: Option<u32>,
+    /// Duration in seconds for the Boundless offer price to ramp from
+    /// `--boundless-min-price-eth` to `--boundless-max-price-eth`.
+    ///
+    /// Defaults to `0` so that the max price is offered immediately,
+    /// eliminating the auction "discount window" that the SDK would
+    /// otherwise insert. With `0`, a prover that locks the request
+    /// receives the full max price, which minimises time-to-lock at
+    /// the cost of paying the max price every time. Set to a non-zero
+    /// value to reintroduce a price ramp (typically for cost
+    /// optimisation when latency is less critical).
+    #[arg(long, env = cli_env!("BOUNDLESS_OFFER_RAMP_UP_PERIOD_SECS"), default_value_t = 0)]
+    boundless_offer_ramp_up_period_secs: u32,
 
     /// Maximum time, in seconds, that a prover that locks a request has to
     /// deliver the proof before forfeiting its stake bond and the request
@@ -250,24 +259,21 @@ struct BoundlessArgs {
     /// Delay, in seconds, between request submission and the moment
     /// bidding is allowed to begin (`Offer.rampUpStart`).
     ///
-    /// The Boundless SDK's default for this value is roughly the
-    /// guest program's executor time (`cycles / 1 MHz`, capped at
-    /// 1 hour), giving provers a "discovery window" to see the
-    /// request and run the executor before bidding opens. That
-    /// window appears on the Boundless explorer as the "flat
-    /// period" preceding the price ramp and accounts for several
-    /// minutes of submission-to-fulfillment latency.
-    ///
-    /// Setting this to `0` eliminates the discovery window entirely:
-    /// bidding opens immediately at submission and the fastest prover
-    /// can lock as soon as it has executed the program. This minimises
-    /// end-to-end latency at the cost of higher prices (no time spent
-    /// on the ramp-up) and a smaller set of provers seeing the
-    /// request before it is locked.
-    ///
-    /// Defaults to SDK behaviour when unset.
-    #[arg(long, env = cli_env!("BOUNDLESS_OFFER_BIDDING_START_DELAY_SECS"))]
-    boundless_offer_bidding_start_delay_secs: Option<u64>,
+    /// Defaults to `0` so that bidding opens immediately at submission
+    /// and the fastest prover can lock as soon as it has executed the
+    /// guest program. This eliminates the SDK's default "discovery
+    /// window" (roughly `cycles / 1 MHz`, capped at 1 hour) — visible
+    /// on the Boundless explorer as the "flat period" preceding the
+    /// price ramp — which accounts for several minutes of
+    /// submission-to-fulfillment latency on large workloads. Set to a
+    /// non-zero value to reintroduce a discovery window so more provers
+    /// can see the request before it is locked.
+    #[arg(
+        long,
+        env = cli_env!("BOUNDLESS_OFFER_BIDDING_START_DELAY_SECS"),
+        default_value_t = 0
+    )]
+    boundless_offer_bidding_start_delay_secs: u64,
 
     /// Maximum number of deterministic request-ID slots to probe when
     /// recovering in-flight proofs after an instance rotation.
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

bin/prover-registrar/src/cli.rs
```diff
@@ -226,9 +226,18 @@ struct BoundlessArgs {
...
+    ///
+    /// Defaults to `0` so that the max price is offered immediately,
+    /// eliminating the auction "discount window" that the SDK would
+    /// otherwise insert. With `0`, a prover that locks the request
+    /// receives the full max price, which minimises time-to-lock at
+    /// the cost of paying the max price every time. Set to a non-zero
+    /// value to reintroduce a price ramp (typically for cost
+    /// optimisation when latency is less critical).
+    #[arg(long, env = cli_env!("BOUNDLESS_OFFER_RAMP_UP_PERIOD_SECS"), default_value_t = 0)]
+    boundless_offer_ramp_up_period_secs: u32,
```
0x00101010: does this make sense? if this is set, we lost the price increase capability?

crates/proof/tee/nitro-attestation-prover/src/boundless.rs
```diff
@@ -174,29 +188,44 @@ impl BoundlessProver {
...
             offer.max_price(max_price.clone());
         }
-        if let Some(ramp_up_period) = self.offer_ramp_up_period_secs {
-            offer.ramp_up_period(ramp_up_period);
-        }
+        offer.ramp_up_period(self.offer_ramp_up_period_secs);
         if let Some(lock_timeout) = self.offer_lock_timeout_secs {
             offer.lock_timeout(lock_timeout);
         }
+        offer.bidding_start(
```
jackchuma: When a fresh submission follows a recovery scan/poll or sits behind `submit_lock` / preflight, this stores `rampUpStart` in `RequestParams` before `client.submit_onchain` actually builds and submits the request. Boundless deadlines are computed as `rampUpStart + lockTimeout/timeout`, so an old timestamp consumes the lock/expiry window before the request exists; after a recovery timeout or large preflight, a request with an explicit 600s lock timeout can be submitted already expired or with almost no time for provers. Compute the timestamp immediately before building/submitting the fresh request instead of when the params are first created.

bin/prover-registrar/src/cli.rs
leopoldjoy: Yes that's right. If we prefer, we can leave the ramp up. It's the trade-off between 1 minute more potential latency and paying our max price. Do you have a preference?

crates/proof/tee/nitro-attestation-prover/src/boundless.rs
leopoldjoy: Nice catch! I've added a correction for this here: https://github.com/base/base/pull/3019/changes/ab4328c5e65767f5b46861160b6dd06a6a741d22

bin/prover-registrar/src/cli.rs
leopoldjoy: I've dropped the default here: https://github.com/base/base/pull/3019/changes/7df0f580f41c51817f0bb439f16af8ca877bf59c

crates/proof/tee/nitro-attestation-prover/src/boundless.rs
```diff
@@ -201,6 +237,25 @@ impl BoundlessProver {
         params.with_offer(offer)
```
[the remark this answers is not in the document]
leopoldjoy: Verified against the Boundless SDK source — this is safe. `RequestParams::new()` already initialises `offer: OfferParams::default()` (all fields `None`, since `OfferParams` derives `Default` per `offer_layer.rs:369`), so calling `with_offer` with an all-`None` builder replaces one all-`None` value with another all-`None` value.

The offer layer reads each field with `.or(...)` / `.unwrap_or(...)` fallbacks rather than treating "`with_offer` was called" as a flag — e.g. in `offer_layer.rs:819-820`:

```rust
rampUpStart: params.bidding_start.unwrap_or(ramp_up_start),
rampUpPeriod: params.ramp_up_period.unwrap_or(ramp_up_period),
```

So `None` always means "use the SDK-derived value", regardless of whether `with_offer` was ever invoked. No code change needed.

crates/proof/tee/nitro-attestation-prover/src/boundless.rs
```diff
@@ -174,16 +189,37 @@ impl BoundlessProver {
...
-        }
+    /// Returns the current Unix timestamp in seconds, used to compute
+    /// `Offer.rampUpStart` as `now + offer_bidding_start_delay_secs`
+    /// immediately before each on-chain submission. Factored out so
+    /// tests can bracket calls between `before`/`after` snapshots to
+    /// assert the resulting timestamp falls in the expected window.
+    fn now_unix_secs() -> u64 {
+        std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .expect("system clock before UNIX epoch")
```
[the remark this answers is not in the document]
leopoldjoy: Good catch — fixed in 4f608a52b. Aligned `now_unix_secs` to use `.unwrap_or_default()` to match the existing pattern in `effective_expiry` and `is_journal_fresh`. Chose fail-soft over panic for the registrar since a pre-epoch clock would result in `bidding_start = delay` (rejected downstream) rather than crashing the whole process — preferable for a long-running service.

0x00101010 (approved):

PR #3052, opened 2026-05-29
fix(devnet): run setup containers as host uid:gid

meyer9:
## Summary

Every `Devnet Tests` job in the merge queue has been failing since `2026-05-19 ~18:19 UTC` with all devnet tests reporting:

```
Failed to read L2 genesis
Caused by: Permission denied (os error 13)
```

(see e.g. job [`78325699089`](https://github.com/base/base/actions/runs/26584261333/job/78325699089) on the queue commit for #3002, summary `7/27 tests run: 1 passed, 6 failed`). `Devnet Tests` is not a required check, so `main` has continued to advance with this broken.

## Root cause

The setup container (`Dockerfile.devnet`) has no `USER` directive and runs as `root`. `SetupContainer::generate_l1_genesis` and `SetupContainer::deploy_l2_contracts` start it with the host output directory bind-mounted in, so the generated `l2/genesis.json`, `rollup.json`, `el/genesis.json`, JWT, etc. are written to the host as `root:root`. The test harness in `devnet/src/smoke.rs:289-293` then reads those files back from the host:

```rust
std::fs::read(l2_deployment.genesis_path()).wrap_err("Failed to read L2 genesis")?;
```

On the `BasePerfRunnerGroup` self-hosted Linux runners the harness runs as an unprivileged user, so the read fails with `EACCES`. nextest's `--max-fail` then trips after 6 failures and the remaining 20 devnet tests are skipped (only `accounts::test_mnemonic_derivation`, which is a pure unit test, ever passes).

This was always latent. It started erroring on `2026-05-19` because something in the runner image changed (the only seven repo commits between the last green run and the first red one are all in `crates/consensus/**`, `crates/common/precompiles/**`, or `bin/basectl/**` — none touch `Dockerfile.devnet`, the setup scripts, mounts, or any UID/perm code, so this is environmental).

## Fix

Pin the setup container to the host `uid:gid` via testcontainers' `with_user`. The setup scripts (`setup-l1.sh`, `setup-l2.sh`) only need to `mkdir` inside the mounted output dir, `mktemp -d` in `/tmp`, and run pre-installed binaries — none of that needs root, and the templates / scripts in the image are world-readable, so the container runs cleanly as any UID.

Linux-only: Docker Desktop on macOS/Windows already remaps bind-mount ownership to the host user, so forcing `--user` there fights the VM and is unnecessary for local dev.

## Files

| File | Change |
|------|--------|
| `devnet/src/setup/container.rs` | Add `host_user_spec()` helper (Linux-gated `getuid`/`getgid`), apply `with_user` to both `GenericImage` builders |
| `devnet/Cargo.toml` | Add `libc` as a Linux-only target dep (already a workspace dep) |

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -24,7 +24,7 @@ jobs:
         just build::affected-ci "origin/${{ github.base_ref || 'main' }}"
       clippy_command: >-
         just check::clippy-affected-ci "origin/${{ github.base_ref || 'main' }}"
-      run_devnet: false
+      run_devnet: true
       run_sigsegv: false
       save_rust_cache: false
       test_command: >-
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8615,6 +8615,7 @@ dependencies = [
  "indicatif 0.18.4",
  "jsonrpsee",
  "k256",
+ "libc",
  "libp2p",
  "nanoid",
  "rand 0.9.4",
--- a/devnet/Cargo.toml
+++ b/devnet/Cargo.toml
@@ -88,6 +88,9 @@ serde_json = { workspace = true, features = ["std"] }
 serde = { workspace = true, features = ["derive", "std"] }
 testcontainers = { workspace = true, features = ["blocking", "host-port-exposure"] }
 
+[target.'cfg(target_os = "linux")'.dependencies]
+libc.workspace = true
+
 [dev-dependencies]
 # cli
 clap = { workspace = true, features = ["derive"] }
--- a/devnet/src/l2/in_process_consensus.rs
+++ b/devnet/src/l2/in_process_consensus.rs
@@ -17,6 +17,7 @@ use alloy_rpc_types_engine::JwtSecret;
 use alloy_signer_local::PrivateKeySigner;
 use base_builder_core::test_utils::get_available_port;
 use base_common_genesis::RollupConfig;
+use nanoid::nanoid;
 use base_consensus_disc::LocalNode;
 use base_consensus_node::{
     EngineConfig, L1ConfigBuilder, NetworkConfig, NodeMode, RollupNodeBuilder, SequencerConfig,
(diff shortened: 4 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

.github/workflows/ci-pr.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         just build::affected-ci "origin/${{ github.base_ref || 'main' }}"
       clippy_command: >-
         just check::clippy-affected-ci "origin/${{ github.base_ref || 'main' }}"
-      run_devnet: false
+      run_devnet: true
```
danyalprout, line 27: Let's not run this on PRs. It's quite slow and doesn't provide too much signal.

Happy to revisit post Depot to check the speed

meyer9: I'm trying to test this workflow since it's failing on main. This should be removed before merging.

PR #3058, opened 2026-05-29
fix(batcher/comp): always flush shadow on write to catch cumulative fullness

kozemyrpol:
## Summary

Fixes a silent frame-overflow bug in `ShadowCompressor`.

## Root Cause

`ShadowCompressor::write()` is supposed to detect when the cumulative compressed output would exceed `target_output_size` and mark the compressor as full. The implementation only flushed the shadow compressor and ran the fullness check when a **single** write's uncompressed size exceeded the target:

```rust
let mut newbound = data.len() as u64;
if newbound > self.config.target_output_size {   // ← only for large writes
    self.shadow.flush()?;
    newbound = self.shadow.len() as u64 + CLOSE_OVERHEAD_ZLIB;
    if newbound > self.config.target_output_size {
        self.is_full = true;
        ...
    }
}
```

Writes whose individual uncompressed size was **below** `target_output_size` skipped the flush and the check entirely. Many small batches that collectively compressed to more than `target_output_size` bytes were never detected — the compressor kept accepting data, and the resulting L1 frame was silently larger than intended.

The struct's own docstring already says *"The first compression buffer is flushed on every write"*, but the code did not honour this.

## Fix

Flush the shadow compressor after **every** write and derive `newbound` from the shadow's actual compressed length unconditionally:

```rust
self.shadow.write(data)?;
self.shadow.flush()?;   // always
let newbound = self.shadow.len() as u64 + CLOSE_OVERHEAD_ZLIB;

if newbound > self.config.target_output_size {
    self.is_full = true;
    if self.compressor.len() > 0 {
        return Err(CompressorError::Full);
    }
}
```

The first-write exception (single block larger than target is still allowed through so the frame encoder can split it) is unchanged.

## Tests Added

- `test_small_writes_eventually_trigger_full` — repeated small writes below the target eventually trigger `is_full`; before the fix this looped forever.
- `test_large_first_write_allowed_second_rejected` — first-write exception still works.
- `test_reset_clears_full_flag_and_bound` — `reset()` restores initial state correctly.

### DIFF
--- a/crates/batcher/comp/src/shadow.rs
+++ b/crates/batcher/comp/src/shadow.rs
@@ -72,21 +72,31 @@ impl CompressorWriter for ShadowCompressor {
             return Err(CompressorError::Full);
         }
 
-        // Write to the shadow compressor.
+        // Write to the shadow compressor and always flush it to obtain an
+        // up-to-date compressed-size estimate.
+        //
+        // The previous implementation only flushed the shadow when a single
+        // write's *uncompressed* size exceeded `target_output_size`.  That
+        // missed the common case where many small writes collectively push the
+        // total compressed output past the target: `is_full` was never set,
+        // the compressor kept accepting data, and the resulting frame was
+        // silently larger than `target_output_size` on-chain.
+        //
+        // Flushing on every write matches the docstring ("the shadow buffer is
+        // flushed on every write") and ensures that `bound` always reflects
+        // the cumulative compressed-size upper bound.
         self.shadow.write(data)?;
+        self.shadow.flush()?;
+        let newbound = self.shadow.len() as u64 + CLOSE_OVERHEAD_ZLIB;
 
-        // The new bound increases by the length of the compressed data.
-        let mut newbound = data.len() as u64;
         if newbound > self.config.target_output_size {
-            // Don't flush the buffer if there's a chance we're over the size limit.
-            self.shadow.flush()?;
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #3114, opened 2026-06-01
feat(snapshotter): use BLAKE3 to diff static file chunks

### LINKED ISSUE #2960: snapshotter: upload skip uses size only, but manifest already carries BLAKE3 hashes
While reading `crates/infra/snapshotter/src/upload.rs` I noticed the `DiffBySize` strategy decides whether to re-upload a static file chunk by comparing only the local and remote sizes:

```rust
UploadStrategy::DiffBySize => {
    if let Some(&remote_size) = remote_static_files.get(&file_name) {
        if remote_size == local_size {
            debug!(file = %file_name, size = local_size, "skipping static file (size matches)");
            skipped += 1;
            continue;
        }
        ...
    }
}
```

The reason this caught my eye is that `snapshot.rs` already computes a BLAKE3 hash for every file as part of archive creation:

```rust
struct HashingReader<R> {
    inner: R,
    hasher: blake3::Hasher,
    bytes_read: u64,
}

// ...

manifest_entries.push(FileEntry {
    path: planned.relative_path.to_string_lossy().to_string(),
    size: reader.bytes_read,
    blake3: reader.finalize(),
});
```

That hash is written into `manifest.json` and uploaded to S3 with every run, so the integrity info is there — it's just not consulted on subsequent runs.

A few concrete cases where size-only skip is the wrong call:

1. **Silent local corruption.** A finalized static file chunk gets corrupted on disk (bit rot, fs error, partial write after a crash) without changing its size. On the next snapshot run, the chunk's BLAKE3 in the new manifest will diverge from the data the consumer eventually reads back from S3 — because the corrupted local chunk was skipped and the old remote one stays. The two artifacts published in the same run no longer agree.

2. **Manifest <-> data drift across runs.** Same root cause as (1), but observed by a snapshot consumer downloading the latest `manifest.json` + chunks and validating per-file BLAKE3. With size-only skip, a single corrupt local copy can keep the published manifest hash and the published chunk out of sync indefinitely.

3. **Determinism check on the producer side.** If the operator wants to re-run a snapshot to confirm reproducibility, size-only skip silently masks any drift in the data that doesn't change archive size.

## Possible directions

A few options, in roughly increasing complexity:

- **Compare against manifest BLAKE3.** Fetch the previous `manifest.json` from S3, look up each chunk's recorded BLAKE3, compare with the freshly computed BLAKE3 from `HashingReader`. This reuses the work already being done and adds no new hashing.
- **Store BLAKE3 as S3 object metadata.** Set `x-amz-meta-blake3` on upload, fetch via `head_object` during diff. Avoids a manifest round-trip but adds one HEAD per chunk.
- **Document the tradeoff.** If the asymmetry is intentional (size is cheap, BLAKE3 verification is the consumer's job), the `DiffBySize` doc comment and the README probably want to call that out right now the BLAKE3 work in `snapshot.rs` looks like it should be load-bearing for skip decisions and it isn't.

Happy to open a PR for the manifest-based variant if that's the preferred direction.

Pulsator01:
## Summary

  Replaces `UploadStrategy::DiffBySize` with `DiffByHash`, comparing per-file BLAKE3 hashes (already computed
  during chunk generation and stored in `manifest.json`) instead of file sizes when deciding whether to skip
  re-uploading a static-file chunk archive. Size-only comparison could silently skip a chunk whose contents
  changed without changing size; BLAKE3 closes that hole and reuses work that was already being done.

  ## Changes

  - `crates/infra/snapshotter/src/snapshot.rs`: add `ChunkFilename` helper (parse/format
  `{component}-{start}-{end}.tar.zst`) and `SnapshotManifest::chunk_hashes_for_file()` to extract sorted per-file
   BLAKE3s for a given chunk archive.
  - `crates/infra/snapshotter/src/upload.rs`: rename `DiffBySize` → `DiffByHash`; add
  `fetch_previous_manifest()`; change `upload()` to take `&local_manifest` + `Option<&remote_manifest>` instead
  of a remote-size map. Falls through to re-upload on first run or any missing entry.
  - `crates/infra/snapshotter/src/orchestrator.rs`: fetch previous manifest from S3, parse the freshly written
  local manifest from disk, pass both into `upload()`.
  - `crates/infra/snapshotter/src/lib.rs`: re-export `ChunkFilename`.
  - `crates/infra/snapshotter/tests/e2e_test.rs`: update existing `upload()` callsites for new signature; add
  coverage for skip-on-blake3-match, re-upload-on-mismatch-even-when-size-matches, and
  first-run-uploads-everything.

  No manifest schema change. No behavior change for `AlwaysUpload`. `compute_skip_ranges` still uses the existing
   remote static-file listing.

  Closes #2960

  ## Test plan

  - [ ] `cargo build -p base-snapshotter` clean
  - [ ] `cargo clippy -p base-snapshotter -- -D warnings` clean
  - [ ] `cargo test -p base-snapshotter` (15 lib unit + 10 e2e tests) passes

### DIFF
--- a/crates/infra/snapshotter/src/lib.rs
+++ b/crates/infra/snapshotter/src/lib.rs
@@ -14,7 +14,7 @@ mod container;
 pub use container::{ContainerManager, DockerContainerManager};
 
 mod snapshot;
-pub use snapshot::{OutputFileChecksum, SnapshotGenerator, SnapshotManifest};
+pub use snapshot::{ChunkFilename, OutputFileChecksum, SnapshotGenerator, SnapshotManifest};
 
 mod upload;
 pub use upload::{SnapshotUploader, UploadStrategy};
--- a/crates/infra/snapshotter/src/orchestrator.rs
+++ b/crates/infra/snapshotter/src/orchestrator.rs
@@ -9,7 +9,9 @@ use anyhow::{Context, Result, bail};
 use tracing::{error, info};
 
 use crate::{
-    SnapshotterConfig, container::ContainerManager, snapshot::SnapshotGenerator,
+    SnapshotterConfig,
+    container::ContainerManager,
+    snapshot::{SnapshotGenerator, SnapshotManifest},
     upload::SnapshotUploader,
 };
 
@@ -98,12 +100,18 @@ impl<C: ContainerManager> Snapshotter<C> {
 
         info!(remote_files = remote_static_files.len(), "fetched remote static file listing");
 
+        let remote_manifest = self.uploader.fetch_previous_manifest().await?;
+        info!(
+            has_remote_manifest = remote_manifest.is_some(),
+            "fetched previous manifest for blake3 diff"
+        );
+
         let source_datadir = self.config.source_datadir.clone();
         let output_dir_for_gen = run_output_dir.clone();
         let chain_id = self.config.chain_id;
         let block = self.config.block;
         let blocks_per_file = self.config.blocks_per_file;
-        let remote_for_gen = remote_static_files.clone();
+        let remote_for_gen = remote_static_files;
 
         let files = tokio::task::spawn_blocking(move || {
             SnapshotGenerator::generate_manifest(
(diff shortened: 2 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

haardikk21 (approved):

PR #3120, opened 2026-06-01
feat(PRIV-1914): replace op-deployer with base/contracts forge scripts in devnet

jowparks:
### Summary

TLDR: 
Deploys via `base/contracts` in devnet instead of op-deployer, deploy with custom commit with:
`CONTRACTS_COMMIT=abc123 just devnet up`

Replace the hardcoded op-deployer binary in the devnet with direct `forge script` calls to [base/contracts](https://github.com/base/contracts), enabling deployment of custom contracts instead of the OP Stack contracts embedded in op-deployer. This is a prerequisite for the privacy appchain work which requires modified contracts.

Closes PRIV-1914

### Changes

**Docker infrastructure:**
- `Dockerfile.devnet`: Replace op-deployer download with a contracts-builder stage that clones base/contracts at a pinned commit, installs Solidity dependencies, and compiles with forge. Final image uses foundry base (Ubuntu) with `USER root` for runtime tool installation.
- `docker-compose.yml`: Add `patch-genesis-hash` init service that queries the L2 EL node for the real genesis block hash and patches `rollup.json` before consensus nodes start.
- `Justfile`: Add `_ensure-sp1-elfs` recipe to auto-install SP1 toolchain and build ELFs before devnet up. Forward `CONTRACTS_COMMIT` env var as `--build-arg`.

**New scripts:**
- `deploy-config.json.template`: Parameterized deploy config for base/contracts, filled via `envsubst` from devnet-env variables.
- `extract-artifacts.sh`: Extracts `l1-addresses.json` from forge deployment output and assembles `genesis.json` from L2Genesis state dump.
- `assemble-rollup-config.sh`: Builds `rollup.json` from L1 RPC state, deployed contract addresses, on-chain SystemConfig values, and protocol parameters. Carefully matches the `RollupConfig` struct which uses `deny_unknown_fields`.
- `L2GenesisDevnet.s.sol`: Thin Solidity wrapper that reads deploy-config + L1 addresses, constructs the `L2Genesis.Input` struct, runs L2Genesis, and dumps state via `vm.dumpState()`.
- `patch-genesis-hash.sh`: Waits for L2 EL to boot, queries genesis block hash, patches rollup.json and rollup-conductor.json.

**Rewritten:**
- `setup-l2.sh`: 5-step pipeline replacing op-deployer: generate deploy-config → forge script SystemDeploy → L2GenesisDevnet → extract artifacts → assemble rollup config. Post-processing (Azul/Beryl activation timestamps, activation admin) preserved unchanged.

**Rust deployer updates:**
- Remove `OP_DEPLOYER_IMAGE` constant, update `DeployerContainer` to use `devnet-setup:local` image.
- Rename `l2_intent.rs` → `deploy_config.rs`, generates deploy-config JSON instead of intent TOML.
- Add rollup config deserialization validation tests (positive + deny_unknown_fields negative test).

**Other:**
- Update SP1 ELF manifest hashes.
- Delete deprecated `l2-intent.toml.template`.

### Testing

- `just devnet up` boots successfully with L2 producing blocks at 2-second cadence
- L1 contracts verified deployed: OptimismPortalProxy, SystemConfigProxy, L1StandardBridgeProxy
- Rollup config deserialization unit tests added
- Shell scripts pass `shellcheck`

type=routine
risk=low
impact=sev5

### Notes for Reviewers
- `CONTRACTS_COMMIT` ARG in Dockerfile.devnet is pinned to `c8f0238ec570a446e01b6290bd2413ef7704fafd`. Override with `CONTRACTS_COMMIT=<sha> just devnet up`.
- `FOUNDRY_SCRIPT_EXECUTION_PROTECTION=false` is set on forge script calls because base/contracts `SystemDeploy.s.sol` uses `address(this)` which forge v1.7+ blocks by default.
- The `patch-genesis-hash` docker-compose service bridges the genesis hash chicken-and-egg problem: the hash is only known after reth initializes from genesis.json, but rollup.json needs it before consensus starts.
- base/contracts `L2Genesis.s.sol` takes an `Input` struct rather than reading from deploy-config. `L2GenesisDevnet.s.sol` bridges this.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

etc/systems/src/images.rs
```diff
@@ -2,6 +2,3 @@
 
 /// Docker image for Reth.
 pub const RETH_IMAGE: &str = "ghcr.io/paradigmxyz/reth:v1.11.4";
```
jowparks, line 4: 🔴 **Compile error** — `OP_DEPLOYER_IMAGE` was removed here but `lib.rs:40` still re-exports it:
```rust
pub use images::{OP_DEPLOYER_IMAGE, RETH_IMAGE};
```
Needs to be:
```rust
pub use images::RETH_IMAGE;
```

etc/systems/src/config/deploy_config.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use alloy_primitives::Address;
+
+use super::accounts;
+
+/// Generates the L2 deploy configuration as JSON.
+///
+/// The output matches the field schema expected by `base/contracts` forge scripts
+/// (i.e. `DeployConfig.s.sol` / `deploy-config/local.json`).
+pub fn deploy_config_json(l1_chain_id: u64, l2_chain_id: u64) -> String {
```
jowparks: This function and `etc/scripts/devnet/templates/deploy-config.json.template` produce identical JSON with the same 40+ fields. If the schema changes (new field in base/contracts), both must be updated in lockstep with no compile-time safety net.

At minimum, add a cross-reference comment in both locations. Better: a test that validates the Rust output against the template structure.

etc/scripts/devnet/setup-l2.sh
```diff
@@ -229,16 +234,9 @@ else
 fi
 
 echo "Writing rollup-conductor.json (base fields stripped for op-conductor compatibility)..."
-jq 'del(.base)' "$OUTPUT_DIR/rollup.json" >"$OUTPUT_DIR/rollup-conductor.json"
+jq 'del(.base, .granite_channel_timeout)' "$OUTPUT_DIR/rollup.json" >"$OUTPUT_DIR/rollup-conductor.json"
```
jowparks, line 281: Previously only `.base` was stripped. Why is `granite_channel_timeout` now also deleted? If op-conductor uses `deny_unknown_fields` too and *does* know about this field, removing it could break things in the other direction.

Worth a comment explaining the rationale (e.g., op-conductor's `RollupConfig` doesn't include `granite_channel_timeout`).

etc/scripts/devnet/setup-l2.sh
```diff
@@ -74,77 +74,82 @@ echo "L1 genesis timestamp: $L1_TIMESTAMP"
...
-# Extract L2 Genesis and Rollup Config
-# =============================================================================
+echo "--- Step 4: Extracting artifacts ---"
+L2_GENESIS_TIMESTAMP="$L1_TIMESTAMP" \
+CONTRACTS_DIR=/contracts \
+OUTPUT_DIR="$OUTPUT_DIR" \
+L1_RPC_URL="$L1_RPC_URL" \
+L1_CHAIN_ID="$L1_CHAIN_ID" \
+L2_CHAIN_ID="$L2_CHAIN_ID" \
+FOUNDRY_SCRIPT_EXECUTION_PROTECTION=false \
```
jowparks: nit: `FOUNDRY_SCRIPT_EXECUTION_PROTECTION` isn't used by `extract-artifacts.sh` — it only does `jq`/`cast` work, no forge invocations. Passing it here is misleading. Remove from this call site.

etc/docker/Dockerfile.devnet
```diff
@@ -33,14 +24,48 @@ RUN git clone https://github.com/protolambda/eth2-val-tools.git --branch "${ETH2
     test "$(git rev-parse HEAD)" = "${ETH2_VAL_TOOLS_COMMIT}"
 RUN go build -o /go/bin/eth2-val-tools .
 
-FROM public.ecr.aws/docker/library/alpine:3.21.3
+# Build contracts using Foundry
+# The foundry image is Ubuntu-based and multi-arch (amd64/arm64)
+FROM ghcr.io/foundry-rs/foundry:latest AS contracts-builder
```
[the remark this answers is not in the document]
jowparks: Fixed — pinned to `v1.7.1` in b8e33d5bf.

etc/scripts/devnet/assemble-rollup-config.sh
```diff
@@ -0,0 +1,236 @@
+#!/bin/bash
+set -e
```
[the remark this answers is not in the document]
jowparks: Fixed — added `set -euo pipefail` to all three scripts in b8e33d5bf.

etc/systems/src/images.rs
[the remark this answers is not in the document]
jowparks: Fixed — updated `lib.rs:40` to `pub use images::RETH_IMAGE;` in 13c5db240.

jowparks: Fixed in 13c5db240.

etc/systems/src/config/deploy_config.rs
jowparks: Added cross-reference comments to both files in 13c5db240.

etc/scripts/devnet/setup-l2.sh
jowparks: Added comment explaining that op-conductor's RollupConfig doesn't recognize `granite_channel_timeout`. Fixed in 13c5db240.

etc/scripts/devnet/setup-l2.sh
jowparks: Removed from Step 4 call site in 13c5db240.

etc/systems/src/deployer/base_deployer.rs
```diff
@@ -107,68 +106,70 @@ impl DeployerContainer {
...
             .with_env_var("DEPLOYER_KEY", self.deployer_key_hex())
-            .with_copy_to(INTENT_PATH, intent_toml.into_bytes())
+            .with_env_var("DEPLOYER_ADDR", format_address(self.deployer_address()?))
+            .with_env_var("SEQUENCER_ADDR", format_address(self.roles.sequencer))
+            .with_env_var("BATCHER_ADDR", format_address(self.roles.batcher))
+            .with_env_var("PROPOSER_ADDR", format_address(self.roles.proposer))
+            .with_env_var("CHALLENGER_ADDR", format_address(self.roles.challenger))
+            .with_env_var("OUTPUT_DIR", OUTPUT_DIR)
+            .with_env_var("TEMPLATE_DIR", "/templates")
+            .with_copy_to(DEPLOY_CONFIG_PATH, deploy_config.into_bytes())
```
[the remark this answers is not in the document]
jowparks: Removed `deploy_config()`, `DEPLOY_CONFIG_PATH`, `replace_address()`, `with_copy_to`, and unused `config` self-import in 13c5db240.

etc/systems/src/lib.rs
```diff
@@ -18,7 +18,7 @@ pub use config::{
     ANVIL_ACCOUNT_0, ANVIL_ACCOUNT_1, ANVIL_ACCOUNT_2, ANVIL_ACCOUNT_3, ANVIL_ACCOUNT_4,
     ANVIL_ACCOUNT_5, ANVIL_ACCOUNT_6, ANVIL_ACCOUNT_7, ANVIL_ACCOUNT_8, ANVIL_ACCOUNT_9, Account,
     BATCHER, BUILDER, CHALLENGER, DEPLOYER, PROPOSER, SEQUENCER, TEST_MNEMONIC, anvil_addresses,
-    l1_beacon_config_yaml, l1_el_genesis, l1_el_genesis_json, l2_intent_toml,
+    deploy_config_json, l1_beacon_config_yaml, l1_el_genesis, l1_el_genesis_json,
```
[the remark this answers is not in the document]
jowparks: Fixed in 13c5db240.

etc/scripts/devnet/templates/deploy-config.json.template
```diff
@@ -0,0 +1,54 @@
+# NOTE: This template produces the same JSON structure as the Rust function
+# deploy_config_json() in etc/systems/src/config/deploy_config.rs.
+# If the schema changes, both must be updated in lockstep.
```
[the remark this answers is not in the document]
jowparks: Fixed in ce52e78d0 — removed `#` comments from the JSON template, moved cross-reference note into `setup-l2.sh`.

crates/proof/succinct/elf/manifest.toml
```diff
@@ -17,7 +17,7 @@
 
 [[elfs]]
 name = "range-elf-embedded"
```
jowparks: I believe this was out of date on `main`? Would like a double check on this to make sure I am not doing something wrong.

etc/systems/src/config/deploy_config.rs
```diff
@@ -0,0 +1,107 @@
...
+
+/// Generates the L2 deploy configuration as JSON.
+///
+/// NOTE: This produces the same JSON structure as
+/// `etc/scripts/devnet/templates/deploy-config.json.template`.
+/// If the schema changes, both must be updated in lockstep.
+///
+/// The output matches the field schema expected by `base/contracts` forge scripts
+/// (i.e. `DeployConfig.s.sol` / `deploy-config/local.json`).
+pub fn deploy_config_json(l1_chain_id: u64, l2_chain_id: u64) -> String {
```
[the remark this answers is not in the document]
jowparks: Fixed — removed `deploy_config.rs` entirely. The template is the single source of truth.

etc/scripts/devnet/setup-l2.sh
```diff
@@ -74,77 +74,83 @@ echo "L1 genesis timestamp: $L1_TIMESTAMP"
...
-  exit 1
-fi
-
-echo "op-deployer state.json created successfully"
-
-# =============================================================================
-# Extract L2 Genesis and Rollup Config
-# =============================================================================
+echo "--- Step 4: Extracting artifacts ---"
+L2_GENESIS_TIMESTAMP="$L1_TIMESTAMP" \
```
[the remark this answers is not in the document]
jowparks: Fixed — `setup-l2.sh` now uses `set -euo pipefail`. Also added defaults for `BUILDER_P2P_KEY`, `BUILDER_ENODE_ID`, `SEQ1_P2P_KEY`, `SEQ2_P2P_KEY` so `-u` doesn't break when called without devnet-env.

etc/systems/src/deployer/base_deployer.rs
```diff
@@ -107,68 +105,54 @@ impl DeployerContainer {
...
             .with_env_var("DEPLOYER_KEY", self.deployer_key_hex())
-            .with_copy_to(INTENT_PATH, intent_toml.into_bytes())
+            .with_env_var("DEPLOYER_ADDR", format_address(self.deployer_address()?))
+            .with_env_var("SEQUENCER_ADDR", format_address(self.roles.sequencer))
+            .with_env_var("BATCHER_ADDR", format_address(self.roles.batcher))
+            .with_env_var("PROPOSER_ADDR", format_address(self.roles.proposer))
+            .with_env_var("CHALLENGER_ADDR", format_address(self.roles.challenger))
+            .with_env_var("OUTPUT_DIR", OUTPUT_DIR)
+            .with_env_var("TEMPLATE_DIR", "/templates")
             .with_mount(Mount::bind_mount(output_dir, OUTPUT_DIR));
```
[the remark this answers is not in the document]
jowparks: Added defaults for the missing P2P vars directly in `setup-l2.sh` (lines 18-21) rather than passing from the Rust side. This way both the docker-compose and Rust test paths work — `setup-l2.sh` is self-contained with sensible defaults.

crates/proof/succinct/elf/manifest.toml
jowparks: Correct — the manifest was stale on `main` because `Cargo.lock` changed (dependency version updates in recent commits) but the manifest wasn't rebuilt. Rebuilt with `--no-docker-cache` for a clean reproducible hash. The aggregation ELF hash is unchanged; only `range-elf-embedded` drifted.

etc/systems/src/config/mod.rs
```diff
@@ -15,6 +15,4 @@ mod l1_genesis;
 /// L1 execution layer genesis configuration generators.
 pub use l1_genesis::{l1_el_genesis, l1_el_genesis_json};
 
-mod l2_intent;
-/// L2 intent configuration generator for op-deployer.
-pub use l2_intent::l2_intent_toml;
+/// L1 deploy configuration generator for base/contracts forge scripts.
```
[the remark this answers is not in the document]
jowparks: Fixed in b16a09389 — removed the dangling doc comment.

etc/systems/src/deployer/base_deployer.rs
```diff
@@ -107,68 +105,54 @@ impl DeployerContainer {
...
-        let image = GenericImage::new(image_name, image_tag)
-            .with_entrypoint("sh")
+        let image = GenericImage::new("devnet-setup", "local")
             .with_wait_for(WaitFor::exit(ExitWaitStrategy::default().with_exit_code(0)));
 
-        let cmd = vec!["-c".to_string(), script];
         let output_dir = self.output_dir.to_string_lossy().to_string();
         let mut request = image
-            .with_cmd(cmd)
+            .with_cmd(["setup-l2.sh"])
```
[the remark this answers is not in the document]
jowparks: Fixed in b16a09389 — added `.with_entrypoint("/bin/bash")` and changed cmd to full path `/usr/local/bin/setup-l2.sh`.

etc/docker/docker-compose.yml
```diff
@@ -171,6 +171,20 @@ services:
...
+    image: devnet-setup:local
+    container_name: patch-genesis-hash
+    depends_on:
+      base-builder:
+        condition: service_healthy
+    volumes:
+      - ../../.devnet/l2/configs:/configs
+    environment:
+      - L2_RPC_URL=http://base-builder:${L2_BUILDER_HTTP_PORT}
+      - CONFIG_DIR=/configs
```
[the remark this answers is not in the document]
jowparks: Fixed in a8be5eebc — added `patch-genesis-hash: service_completed_successfully` as a dependency to `base-cl-bootnode`. No circular dependency: the chain is `setup-l2` → `base-builder` (healthy) → `patch-genesis-hash` → `base-cl-bootnode`.

etc/scripts/devnet/templates/deploy-config.json.template
jackchuma, line 1: I wonder if this should be pulled from base/contracts instead? Seems like this could easily diverge from what's used in the other repo. although if we're pinning to a specific contracts commit then less of a concern

etc/scripts/devnet/assemble-rollup-config.sh
```diff
@@ -0,0 +1,236 @@
...
+
+# ===========================================================================
+# 5. Compute batch_inbox_address from l2ChainId
+# ===========================================================================
+echo ""
+echo "--- Computing batch_inbox_address ---"
+
+# Solidity: address(uint160(uint256(keccak256(abi.encode("]]]]", _chainId)))))
+# Steps: ABI-encode string + uint256, keccak256, take last 20 bytes as address
+ENCODED=$(cast abi-encode "f(string,uint256)" "]]]]" "$L2_CHAIN_ID")
```
leopoldjoy: This recomputes `batch_inbox_address` using the legacy `keccak(abi.encode("]]]]", chainId))` formula, but `SystemDeploy` initializes `SystemConfig.batchInbox()` with `Types.chainIdToBatchInboxAddress`, so please read `batchInbox()` from the deployed `SystemConfigProxy` instead of deriving it here.

etc/scripts/devnet/extract-artifacts.sh
```diff
@@ -0,0 +1,220 @@
...
+    jovianTime:          0,
+    optimism: {
+      eip1559Elasticity:        6,
+      eip1559Denominator:       50,
+      eip1559DenominatorCanyon: 250
+    }
+  },
+  nonce:         "0x0",
+  timestamp:     $genesis_timestamp,
+  extraData:     "0x",
```
leopoldjoy: Because `jovianTime` is set to `0`, genesis `extraData` needs the Jovian EIP-1559/min-base-fee payload rather than `0x`, otherwise the first child block computes base fee from invalid parent extra data.

etc/scripts/devnet/L2GenesisDevnet.s.sol
```diff
@@ -0,0 +1,54 @@
...
+import { L2Genesis } from "scripts/L2Genesis.s.sol";
+import { DeployConfig } from "scripts/deploy/DeployConfig.s.sol";
+
+/// @title L2GenesisDevnet
+/// @notice Thin wrapper that reads deploy-config + deployed L1 addresses,
+///         runs L2Genesis.run(Input), and dumps the resulting state.
+contract L2GenesisDevnet is Script {
+    function run() public {
+        // Read deploy config
+        DeployConfig cfg = new DeployConfig();
```
leopoldjoy, line 15: Creating `DeployConfig` and `L2Genesis` with `new` before `vm.dumpState` leaves those helper contracts in the dumped alloc, so the generated L2 genesis includes noncanonical script-only accounts.

etc/scripts/devnet/setup-l2.sh
```diff
@@ -70,81 +74,89 @@ L1_TIMESTAMP=$(echo "$L1_GENESIS" | jq -r '.timestamp')
...
+# Step 2: Deploy L1 contracts via forge script
+# Contracts are pre-compiled at /contracts/ (from Dockerfile.devnet).
+# --slow flag sends transactions one at a time for devnet reliability.
+echo ""
+echo "--- Step 2: Deploying L1 contracts ---"
+(
+  cd /contracts
+  FOUNDRY_SCRIPT_EXECUTION_PROTECTION=false \
+    DEPLOY_CONFIG_PATH="$WORKDIR/deploy-config/devnet.json" \
+    forge script scripts/deploy/SystemDeploy.s.sol:SystemDeploy \
```
leopoldjoy, line 146: This currently treats `base/contracts`'s `SystemDeploy` script as the deployment execution path (even though those scripts were just for testing), but Base production deployments use `contract-deployments`, so I would suggest either routing privacy-appchain deployment artifacts through a `contract-deployments`-style task / template or explicitly scoping this path to devnet-only testing with parity checks.

etc/docker/Dockerfile.devnet
```diff
@@ -33,14 +24,48 @@ RUN git clone https://github.com/protolambda/eth2-val-tools.git --branch "${ETH2
     test "$(git rev-parse HEAD)" = "${ETH2_VAL_TOOLS_COMMIT}"
 RUN go build -o /go/bin/eth2-val-tools .
 
-FROM public.ecr.aws/docker/library/alpine:3.21.3
+# Build contracts using Foundry
+# The foundry image is Ubuntu-based and multi-arch (amd64/arm64)
+FROM ghcr.io/foundry-rs/foundry:v1.7.1 AS contracts-builder
 
-RUN apk add --no-cache bash jq openssl curl gettext
+ARG CONTRACTS_COMMIT=c8f0238ec570a446e01b6290bd2413ef7704fafd
```
leopoldjoy, line 31: The default contracts pin is behind local `base/contracts` main, is there any particular reason it's set to this commit? Normally this would be set to the commit hash of a recent release.

etc/docker/Dockerfile.devnet
```diff
@@ -33,14 +24,48 @@ RUN git clone https://github.com/protolambda/eth2-val-tools.git --branch "${ETH2
...
-# Copy the built binaries
+# git is pre-installed in the foundry image (Ubuntu 22.04)
+RUN git config --global user.email "docker@build" && \
+    git config --global user.name "docker"
+
+WORKDIR /contracts
+RUN git clone https://github.com/base/contracts.git . && \
+    git checkout "${CONTRACTS_COMMIT}"
+
+RUN forge install \
```
leopoldjoy: Hard-coding dependency revisions independently of `CONTRACTS_COMMIT` creates a second dependency-pinning scheme from `contract-deployments`, so custom contract commits can compile against the wrong library set.

etc/docker/Justfile
```diff
@@ -118,7 +118,8 @@ build-image target='client' profile='dev':
 
 # Prepares Docker images needed for system tests
 pull-images:
-    docker build -t devnet-setup:local -f etc/docker/Dockerfile.devnet .
+    docker build -t devnet-setup:local -f etc/docker/Dockerfile.devnet \
```
leopoldjoy: This Justfile path passes `CONTRACTS_COMMIT`, but the Rust `SetupContainer` builder still invokes `docker build` without this arg and reuses any existing `devnet-setup:local` tag, so system tests can silently run against a different contracts checkout than the one requested.

etc/systems/src/deployer/base_deployer.rs
```diff
@@ -107,68 +105,54 @@ impl DeployerContainer {
...
-        let intent_toml = self.intent_toml()?;
-        let script = deploy_script();
-        let (image_name, image_tag) = OP_DEPLOYER_IMAGE
-            .rsplit_once(':')
-            .ok_or_else(|| eyre!("op-deployer image tag is missing"))?;
+        std::fs::create_dir_all(&self.output_dir).wrap_err("failed to create output directory")?;
 
-        let image = GenericImage::new(image_name, image_tag)
-            .with_entrypoint("sh")
+        let image = GenericImage::new("devnet-setup", "local")
```
leopoldjoy: Switching this wrapper to the local `devnet-setup:local` image without sharing `SetupContainer`'s build and startup-timeout path makes direct `DeployerContainer` callers fail unless the image was prebuilt and the forge deployment exits within the default timeout.

etc/scripts/devnet/L2GenesisDevnet.s.sol
```diff
@@ -0,0 +1,54 @@
...
+            baseFeeVaultRecipient: cfg.baseFeeVaultRecipient(),
+            baseFeeVaultMinimumWithdrawalAmount: cfg.baseFeeVaultMinimumWithdrawalAmount(),
+            baseFeeVaultWithdrawalNetwork: cfg.baseFeeVaultWithdrawalNetwork(),
+            l1FeeVaultRecipient: cfg.l1FeeVaultRecipient(),
+            l1FeeVaultMinimumWithdrawalAmount: cfg.l1FeeVaultMinimumWithdrawalAmount(),
+            l1FeeVaultWithdrawalNetwork: cfg.l1FeeVaultWithdrawalNetwork(),
+            operatorFeeVaultRecipient: cfg.operatorFeeVaultRecipient(),
+            operatorFeeVaultMinimumWithdrawalAmount: cfg.operatorFeeVaultMinimumWithdrawalAmount(),
+            operatorFeeVaultWithdrawalNetwork: cfg.operatorFeeVaultWithdrawalNetwork(),
+            fork: 5, // JOVIAN — all forks active at genesis
```
leopoldjoy: The raw `fork: 5` ordinal is coupled to the current contracts enum and already sits below `LATEST_FORK = INTEROP`, so use the checked-out contracts' `Fork` symbol or constant instead of a magic number.

etc/systems/src/l2/stack.rs
```diff
@@ -118,6 +118,30 @@ impl L2Stack {
             .await
             .wrap_err("Failed to start in-process builder")?;
 
+        // Patch the L2 genesis hash in the rollup config if it's a zero placeholder.
+        // This happens when the devnet uses base/contracts forge scripts (which can't
+        // compute the genesis hash at deploy time). The real hash is only known after
+        // reth initializes from genesis.json.
+        let rollup_config = {
+            let mut cfg = rollup_config;
+            if cfg.genesis.l2.hash == B256::ZERO {
```
leopoldjoy: Patching the zero L2 genesis hash only in memory leaves `L2DeploymentOutput::rollup_config_path()` and any downstream artifact consumers with the unpatched zero hash, so persist the fetched hash or reject stale artifacts.

etc/scripts/devnet/setup-l2.sh
```diff
@@ -78,81 +81,114 @@ L1_TIMESTAMP=$(echo "$L1_GENESIS" | jq -r '.timestamp')
...
-echo "Configuring intent.toml for devnet..."
-
-# Convert L2 chain ID to hex (0x prefixed, 32 bytes padded)
-L2_CHAIN_ID_HEX=$(printf "0x%064x" $L2_CHAIN_ID)
+# Step 1: Generate deploy-config.json from contracts' local.json
+# Uses contracts' own config as the source of truth, overlaying only the
+# devnet-specific addresses and chain IDs.  This avoids maintaining a
+# separate template that can diverge from the contracts repo.
+echo ""
+echo "--- Step 1: Generating deploy-config.json ---"
```
jowparks, line 110: Using template from contracts repo

etc/scripts/devnet/setup-l2.sh
jowparks: Agreed makes sense. This is for local/devnet usage only

PR #3127, opened 2026-06-01
fix(snapshotter): guard against non-aligned start in chunk_hashes_for_file

### LINKED ISSUE #3126: snapshotter: chunk index computation missing alignment guard in chunk_hashes_for_file
While reviewing the changes landed in #3114 I noticed `chunk_hashes_for_file` in `crates/infra/snapshotter/src/snapshot.rs` computes the chunk index with integer division:

```rust
pub fn chunk_hashes_for_file(&self, filename: &str) -> Option<Vec<String>> {
    let (component, start, _end) = ChunkFilename::parse(filename)?;
    let value = self.components.get(&component)?;
    let meta: ChunkedComponentMeta = serde_json::from_value(value.clone()).ok()?;
    // index = start / blocks_per_file  ← no alignment check
```

If `start` is not a multiple of `meta.blocks_per_file`, the division silently truncates and the wrong chunk's hashes are returned. The upload loop then compares those hashes against the remote manifest entry for a *different* chunk, potentially producing a false-positive match and skipping an upload that should have happened.

Current filenames are always generated with aligned `start` values, so this doesn't trigger in practice today. But there's no enforcement at the boundary a filename constructed outside the normal generation path (e.g. in a test, a manual upload, or a future code path) could silently corrupt the diff decision without any warning.

The defensive fix is a one-liner before the division:

```rust
if start % meta.blocks_per_file != 0 {
    return None;
}
```

This matches the spirit of the existing `ChunkFilename::parse` validation (which already rejects structural malformations) and makes the alignment assumption explicit and enforced rather than implicit.

Happy to send a PR for this.

Sertug17:
Fixes #3126.

Follow-up to #3114. `chunk_hashes_for_file` uses integer division to compute the chunk index (`start / blocks_per_file`), but never checks that `start` is actually aligned to `blocks_per_file`. A non-aligned `start` silently truncates to the wrong index, returning hashes for a different chunk and potentially producing a false-positive match in the upload diff loop.

The fix adds a one-liner guard immediately after the existing `blocks_per_file == 0` check:

```rust
if start % meta.blocks_per_file != 0 {
    return None;
}
```

This returns `None` for any filename whose `start` isn't a clean multiple of `blocks_per_file`, causing the upload loop to fall through to a re-upload rather than skipping on a wrong hash match. It matches the spirit of the existing structural validation in `ChunkFilename::parse` and makes the alignment assumption explicit.

Current production filenames are always generated with aligned values, so this has no effect on the happy path. It only fires on malformed filenames that today would silently produce wrong results.

Three-line addition, no other changes.

### DIFF
--- a/crates/infra/snapshotter/src/snapshot.rs
+++ b/crates/infra/snapshotter/src/snapshot.rs
@@ -83,6 +83,9 @@ impl SnapshotManifest {
         if meta.blocks_per_file == 0 {
             return None;
         }
+        if start % meta.blocks_per_file != 0 {
+            return None;
+        }
         let chunk_index = usize::try_from(start / meta.blocks_per_file).ok()?;
         let entries = meta.chunk_output_files.get(chunk_index)?;
         if entries.is_empty() {


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #3188, opened 2026-06-02
fix(b20-asset): emit effective multiplier (WAD) in MultiplierUpdated when zero is passed

### LINKED ISSUE #3187: b20_asset: update_multiplier(0) emits MultiplierUpdated{multiplier: 0} but effective multiplier is WAD
While reading `crates/common/precompiles/src/b20_asset/` I noticed a mismatch between the emitted event value and the effective multiplier when `update_multiplier(0)` is called.

**The code path:**

`update_multiplier` in `token.rs`:
```rust
pub fn update_multiplier(&mut self, caller: Address, new_multiplier: U256, privileged: bool) -> Result<()> {
    self.ensure_operator_role(caller, privileged)?;
    self.accounting_mut().set_multiplier(new_multiplier)?;
    self.accounting_mut().emit_event(
        IB20Asset::MultiplierUpdated { multiplier: new_multiplier }.encode_log_data(),
    )
}
```

`set_multiplier` writes the value directly to storage with no validation.

But `multiplier()` getter (generated by the `AssetAccounting` macro in `precompile-macros/src/accounting.rs`) returns WAD when storage holds zero:

```rust
fn multiplier(&self) -> Result<U256> {
    let multiplier = self.asset.multiplier()?;
    Ok(if multiplier.is_zero() { Self::WAD } else { multiplier })
}
```

**The inconsistency:**

When `update_multiplier(0)` is called:
1. Storage is written with `0`
2. `MultiplierUpdated { multiplier: 0 }` is emitted
3. But all subsequent reads via `multiplier()` return `WAD` (1e18)
4. `to_scaled_balance` and `to_raw_balance` both use the getter, so they operate with `WAD` not `0`

Off-chain indexers and consumers observing `MultiplierUpdated { multiplier: 0 }` would conclude the multiplier is `0`, but the token actually behaves as if it is `WAD`.

**Possible fixes:**

1. Reject `new_multiplier == 0` in `update_multiplier` with an `InvalidMultiplier` error makes the zero-means-WAD sentinel an internal implementation detail that callers never observe.
2. Emit `WAD` in the event when `new_multiplier == 0` keeps the setter permissive but makes the event accurate.
3. Document the zero-means-WAD behavior explicitly in the natspec so callers know `0` is an alias for `WAD`.

Happy to send a PR for whichever direction is preferred.

Sertug17:
Fixes #3187.

`update_multiplier(0)` stored zero in the asset slot and emitted `MultiplierUpdated { multiplier: 0 }`, but the `multiplier()` getter treats zero as a sentinel for WAD (1e18):

```rust
Ok(if multiplier.is_zero() { Self::WAD } else { multiplier })
```

So all arithmetic (`to_scaled_balance`, `to_raw_balance`) operated with WAD while the emitted event said `0`. Off-chain indexers and consumers would read a multiplier of `0` but the token would behave as if it were `1e18`.

This PR emits the effective value (`WAD`) when `new_multiplier == 0`, so the event always reflects what arithmetic will actually use. The storage write and the getter behavior are unchanged only the event payload is corrected.

No ABI change. One line behavior fix in `update_multiplier`.

### DIFF
--- a/crates/common/precompiles/src/b20_asset/token.rs
+++ b/crates/common/precompiles/src/b20_asset/token.rs
@@ -215,8 +215,16 @@ impl<S: AssetAccounting, P: Policy> B20AssetToken<S, P> {
     ) -> Result<()> {
         self.ensure_operator_role(caller, privileged)?;
         self.accounting_mut().set_multiplier(new_multiplier)?;
+        // Emit the effective multiplier: zero is the storage sentinel for WAD,
+        // so callers and off-chain indexers always see the value that arithmetic
+        // will actually use. See #3187.
+        let effective = if new_multiplier.is_zero() {
+            B20AssetStorage::WAD
+        } else {
+            new_multiplier
+        };
         self.accounting_mut().emit_event(
-            IB20Asset::MultiplierUpdated { multiplier: new_multiplier }.encode_log_data(),
+            IB20Asset::MultiplierUpdated { multiplier: effective }.encode_log_data(),
         )
     }
 


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #3200, opened 2026-06-03
fix(consensus): speed up reset anchor selection

niran:
## Summary

This is a discussion PR for reducing base-consensus reset latency after shallow L1 reorgs.

It changes derivation reset anchor selection from a one-L2-block-at-a-time walk to:

- a validated SafeDB hint when SafeDB is configured
- an exponential backward search plus binary search fallback when no usable hint exists
- the same channel-timeout/system-config correctness condition as the existing reset path

It also threads SafeDB into the engine reset path as a read-only hint source and adds walk-depth fields to sync-start logs.

## Incident context

On 2026-06-02 at 7:28 AM PDT, Base mainnet produced five consecutive empty L2 blocks after a 1-block Ethereum L1 reorg. The reorg detection itself was immediate: base-consensus logged an L1 parent-hash mismatch at `2026-06-02T14:28:04.444581Z`.

The delay appears to have been in the reset path after detection. Logs from the active host show `sync_start` walking the reset safe cursor backward from about L2 `46810567` toward about `46809969`, with sampled messages spanning roughly 10.3 seconds. During that window, the builder logged late forkchoice updates and produced five empty blocks.

The working hypothesis from `docs/blog/2026-06-02-mainnet-l1-reorg-reset-delay.md` is that reset reloaded forkchoice from the EL labels and used an older EL-labeled safe head, even though the engine actor had recently observed a newer in-memory safe head around `46810502`. One plausible code-level explanation is that some safe-head advancement can happen through in-memory consolidation paths without an FCU that updates reth's `safe` label. If reset then trusts the older EL label, it may perform a large serial parent walk before it can resume derivation/sequencing.

This PR does not try to prove or fix every part of that hypothesis. It gives us a concrete starting point for one mitigation: keep reset correctness unchanged, but make the expensive derivation reset anchor lookup sublinear and allow SafeDB to provide a validated jump target.

## Design

`ResetSignal` now carries an optional `l2_reset_anchor_hint`. The engine processor computes the current reset cutoff from the selected safe head's L1 origin and the configured channel timeout. If SafeDB is enabled, it queries `safe_head_at_l1(cutoff_l1)` and passes the returned L2 block ID as a hint.

The derivation pipeline treats the hint as untrusted:

- it fetches the hinted L2 block locally
- it checks the hash matches
- it checks the block satisfies the existing reset condition: `block.l1_origin + channel_timeout <= safe_head.l1_origin`
- if any check fails, it ignores the hint and searches normally

The fallback search preserves existing semantics but avoids the linear scan:

- exponential search backward from the selected safe head until it finds a block satisfying the reset condition or genesis
- binary search forward to find the highest block satisfying that same condition
- system config is still fetched from the selected reset block

This intentionally does not change the channel-timeout invariant. We still rewind far enough to preserve open-channel and system-config correctness.

## Why this is useful

The incident logs suggest the reset path walked hundreds of L2 parents through serial RPC calls. With this change, the no-SafeDB path should perform O(log N) L2 block lookups for the same target. If SafeDB is enabled and contains a useful entry, reset can start near the target and validate the candidate before doing any narrowing.

This also makes reset behavior easier to discuss and observe:

- SafeDB hint success/failure is logged
- sync-start unsafe/safe walk depths are logged
- derivation reset behavior remains covered by the same correctness condition as before

## Non-goals

- This does not enable SafeDB in Helm or production configs.
- This does not change SafeDB schema or RPC behavior.
- This does not yet prefer the engine actor's newer in-memory safe head over the EL safe label.
- This does not yet suspend/cancel sequencer builds while reset is in progress.

Those are likely separate follow-ups if the team agrees with the direction.

## Tests

- `cargo check -p base-consensus-derive -p base-consensus-engine -p base-consensus-node`
- `cargo test -p base-consensus-derive --lib initial_reset`
- `cargo test -p base-consensus-node --lib reset_anchor_hint_queries_safedb_at_channel_timeout_cutoff`
- `cargo test -p base-consensus-derive -p base-consensus-engine -p base-consensus-node`
- `cargo +nightly fmt`
- `git diff --check`

type=routine
risk=low
impact=sev5

### DIFF
--- a/crates/consensus/derive/src/pipeline/core.rs
+++ b/crates/consensus/derive/src/pipeline/core.rs
@@ -65,6 +65,7 @@ where
     async fn initial_reset(
         &mut self,
         l2_safe_head: L2BlockInfo,
+        l2_reset_anchor_hint: Option<BlockNumHash>,
     ) -> PipelineResult<(BlockNumHash, SystemConfig)>
     where
         <P as BatchValidationProvider>::Error: Into<PipelineErrorKind>,
@@ -73,22 +74,9 @@ where
         let l1_origin_ts_lower_bound =
             l2_safe_head_ts.saturating_sub(self.rollup_config.max_sequencer_drift(l2_safe_head_ts));
         let channel_timeout = self.rollup_config.channel_timeout(l1_origin_ts_lower_bound);
-        let l1_origin_number = l2_safe_head.l1_origin.number;
-        let mut current = l2_safe_head;
-
-        loop {
-            if current.block_info.number == self.rollup_config.genesis.l2.number {
-                break;
-            }
-            if current.l1_origin.number + channel_timeout <= l1_origin_number {
-                break;
-            }
-            current = self
-                .l2_chain_provider
-                .l2_block_info_by_number(current.block_info.number - 1)
-                .await
-                .map_err(Into::into)?;
-        }
+        let current = self
+            .find_initial_reset_block(l2_safe_head, l2_reset_anchor_hint, channel_timeout)
+            .await?;
 
         let system_config = self
             .l2_chain_provider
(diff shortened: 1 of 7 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

niran:
Closing this out along the lines of the related #2698 discussion: https://github.com/base/base/pull/2698#discussion_r3350484475

The key overlap is the `find_starting_forkchoice` safe-head walk. #2698 added the `is_labeled_safe` stop condition, so the safe search now stops when the cursor reaches the EL-labeled safe head instead of continuing until it is behind the sequencing window, finalized, or genesis. That directly targets the behavior we were trying to investigate here: during the five missed blocks ([46810570](https://basescan.org/block/46810570)-[46810574](https://basescan.org/block/46810574)), we appear to have walked back to finalized instead of safe.

PR #3207, opened 2026-06-03
fix(precompile-storage): enforce bounds in VecHandler Index/IndexMut (BOP-273)

rayyan224:
## Summary

`VecHandler`'s `Index` and `IndexMut` impls computed element slot addresses from `data_slot + index` without reading `len_slot`, allowing writes to arbitrary storage slots beyond the logical vector length. `SetHandler` was the only internal caller and relied on this unchecked path in its `write()` implementation.

## Root cause

```rust
// Before — no bounds check, any index silently accepted:
fn index(&self, index: usize) -> &Self::Output {
    self.cache.get_or_insert(&index, || Self::compute_handler(..., index))
}
```

`at()` correctly read `len_slot` and returned `None` on OOB; the operator impls did not.

## Changes

**`vec.rs`**
- Removed `Index` and `IndexMut` impls from `VecHandler` — they cannot return `Result`, making them incompatible with precompile error handling (panics crash the node rather than reverting the call)
- Added `pub(crate) at_with_len(index, len) -> Option<&T::Handler>` and `at_mut_with_len(index, len) -> Option<&mut T::Handler>`: bounds-checked via pure integer comparison against a caller-supplied len — zero extra sloads

**`set.rs`**
- All `self.values[i]` call sites replaced with `at_with_len` / `at_mut_with_len` + `.ok_or_else(position_invariant_violated)?` — invariant violations now return `BasePrecompileError::Fatal` rather than panicking
- `SetHandler::write` rewrites all elements directly via `at_mut_with_len(index, new_len)` then updates `len_slot` once — eliminates the O(k) extra sloads/sstores that `push()` was performing per new element
- Added `rstest` parameterized test covering grow, shrink, and equal-size rewrites, asserting both that new elements are present and old elements are absent

## Gas improvement

| Operation | Before | After |
|---|---|---|
| `[]` per element in loop | +1 warm sload (len_slot) | 0 extra sloads |
| Growing set by k elements | k sloads + k sstores to len_slot | 0 extra sloads + 1 sstore total |

## Test plan

- [x] `cargo test -p base-precompile-storage` — 142 tests pass
- [x] `test_set_write_len_slot_updated` — rstest cases for grow / shrink / equal-size
- [x] `test_vec_at_with_len_oob_returns_none` — confirms `at_with_len` bounds check
- [x] `test_vec_at_oob_returns_none` — confirms `at()` unchanged behaviour

Closes BOP-273


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompile-storage/src/types/vec.rs
```diff
@@ -254,27 +253,44 @@ where
...
-    fn index(&self, index: usize) -> &Self::Output {
+    /// Returns a handler for the element at `index`, bounds-checked against a caller-supplied
+    /// `len` (no `sload` — pure integer comparison).
+    ///
+    /// Callers must supply a `len` freshly read from `self.len()` and not invalidated by a
+    /// subsequent `push`/`pop`. Returns `Err(Fatal)` if `index >= len`.
+    #[inline]
+    pub(crate) fn at_with_len(&self, index: usize, len: usize) -> Result<&T::Handler<'a>> {
+        if index >= len {
+            return Err(BasePrecompileError::Fatal(
```
eric-ships, line 268: to be clear, this path would only be callable within precompile, right? else it might make sense to `return Err(BasePrecompileError::array_oob());` which would translate to `Panic(PanicKind::ArrayOutOfBounds)` in solidity?

eric-ships (dismissed):
LGTM

crates/common/precompile-storage/src/types/vec.rs
rayyan224: Yea it would bubble up

crates/common/precompile-storage/src/types/set.rs
```diff
@@ -276,28 +277,31 @@ where
...
-
         for i in new_len..old_len {
-            self.values[i].delete()?;
+            self.values.at_mut_with_len(i, old_len)?.delete()?;
+        }
+
+        if new_len != old_len {
+            Slot::<U256>::new(self.values.len_slot(), self.address, self.storage)
+                .write(U256::from(new_len))?;
         }
```
refcell, line 296: Ah nice change

refcell: Can we add a test asserting we didn't touch the slot?

rayyan224: Sorry wdum touch slot ? we have tests underneat already testing growing shrinking and equal ?

refcell (approved):

PR #3185, opened 2026-06-02
refactor (basectl): Move basectl TUI routes under monitor subcommand

Pulsator01:
Moved the TUI under a new `basectl monitor` subtree, reserving top-level `basectl` for future non-interactive operational CLI commands.

  ## Specific changes

  - Bare `basectl` now prints help and exits 0 instead of opening the TUI
  - `basectl monitor` opens the TUI Home view
  - All TUI views moved under `monitor`:
    - `basectl monitor conductor`
    - `basectl monitor da`
    - `basectl monitor flashblocks`
    - `basectl monitor command-center`
    - `basectl monitor upgrades`
    - `basectl monitor config`
  - Short aliases (`c`, `f`, `d`, `cc`, `co`, `u`) moved under `monitor` accordingly
  - `basectl flashblocks` remains top-level and streams JSONL directly. The post-restructure `flashblocks` namespace exists only for the JSONL stream, so the originally-required `--json` flag was dropped — a bool that can only ever be `true` was a code smell with nothing to disambiguate
  - `basectl flashblocks --help` shows an `after_help` pointer to `basectl monitor flashblocks` for users looking for the TUI view
  - Updated `etc/docker/Justfile` devnet conductor recipe to use the new path
  - Updated `bin/basectl/README.md`

  No underlying TUI behavior, views, config loading, background services, or keybindings were changed.

  ## Tests

  - `cargo check -p basectl`
  - `cargo build -p basectl`
  - `cargo clippy -p basectl --all-targets -- -D warnings`
  - `just --justfile etc/docker/Justfile --dry-run conductor`

  ### Manual parser checks

  - Bare `basectl` exits 0 and prints help
  - `basectl monitor --help` shows all moved TUI commands and aliases
  - `basectl flashblocks` streams JSONL (same output as the pre-PR `basectl flashblocks --json` behavior)
  - `basectl flashblocks --json` is rejected by clap as an unknown argument
  - `basectl flashblocks --help` reads cleanly with the `after_help` pointer
  - Old top-level TUI paths (e.g. `basectl conductor`) are correctly rejected
  - All short aliases under `monitor` resolve correctly
  - `basectl monitor --json` is correctly rejected

PR #3209, opened 2026-06-03
refactor(basectl): split rpc.rs into per-service modules

Pulsator01:
## Summary

  Splits `crates/infra/basectl/src/rpc.rs` (~1800 lines) into per-service modules under `crates/infra/basectl/src/rpc/`:

  - `rpc/admin.rs` — sequencer admin RPCs (start/stop/pause/unpause)
  - `rpc/conductor.rs` — HA conductor cluster RPCs and poller
  - `rpc/el.rs` — EL block-fetch and tx-decoding helpers
  - `rpc/flashblocks.rs` — flashblocks WS client
  - `rpc/l1.rs` — L1 blob watcher and system-config fetch (absorbs the former `l1_client.rs`)
  - `rpc/p2p.rs` — placeholder for Phase 2 PR 2.3
  - `rpc/rollup.rs` — safe-head / validator / proofs pollers

  `l1_client.rs` is deleted; its contents fold into `rpc/l1.rs`. The crate's public API is preserved — `lib.rs` re-exports from `rpc::` cover the same set of items as before.

  This is **mostly relocation**: free functions stay free functions, signatures unchanged, no callsite rewrites. The only structural touch is restoring the private `fetch_raw_block_info` helper (originally in `rpc.rs`) as `pub(super)` in `rpc/el.rs` so `rpc/l1.rs` can share it instead of duplicating the fetch-and-sum logic.

  Two intentional semantic fixes in `rpc/l1.rs` are included alongside the relocation:
  - The subscription loop now guards the current-block fetch with `block_num > last_block.unwrap_or(0)`, preventing a duplicate send when the subscription fires with the same block number as the initial latest-block fetch at startup.
  - The initial latest-block send before the subscription loop now propagates a channel-closed error (returns early) instead of using `let _ =
  ...` fire-and-forget, making it consistent with the loop body.

  Stacked on #3185 (basectl rearch).

  ## Test plan

  - [x] `cargo build -p basectl-cli -p basectl` clean
  - [x] `cargo clippy -p basectl-cli -p basectl --all-targets -- -D warnings` clean
  - [x] `cargo test -p basectl-cli -p basectl` — 29 passed, 0 failed
  - [x] `basectl monitor` and all subviews still open (regression check on the touched modules)


crates/infra/basectl/src/rpc/rollup.rs
```diff
@@ -0,0 +1,414 @@
...
+) -> Option<LatestProposal> {
+    let game_type = respected_type?;
+    let count = total_games.filter(|&c| c > 0)?;
+
+    // Scan backwards from the most recent game (max 50 games).
+    let scan_start = count - 1;
+    let scan_end = count.saturating_sub(50);
+
+    for idx in (scan_end..=scan_start).rev() {
+        let game = factory.gameAtIndex(alloy_primitives::U256::from(idx)).call().await.ok()?;
```
Pulsator01: **Bug: `find_latest_proposal` aborts the entire scan on any transient RPC error**

`ok()?` propagates `None` out of the *function* on the first failing `gameAtIndex` call — not just out of that loop iteration. A single transient network error or contract revert at any index causes the proofs view to show no latest proposal, silently dropping the entire backward scan of up to 50 games.

```suggestion
        let Ok(game) = factory.gameAtIndex(alloy_primitives::U256::from(idx)).call().await else {
            continue;
        };
```

Using `continue` skips a flaky index and keeps scanning older games, matching the intent of the backward scan.

crates/infra/basectl/src/rpc/l1.rs
```diff
@@ -0,0 +1,265 @@
...
+                provider.get_block_by_number(BlockNumberOrTag::Number(gap_num)).full().await
+            {
+                let info = extract_l1_block_info(&block, batcher_address);
+                if result_tx.send(info).await.is_err() {
+                    return Ok(());
+                }
+            }
+        }
+
+        if let Ok(Some(block)) =
```
Pulsator01: **Bug: duplicate L1 block sent when subscription fires for the same block as the initial fetch**

After the initial latest-block fetch at startup sets `last_block = Some(N)`, if the subscription's first header carries `block_num = N` (same block, delivered immediately after being mined), the gap loop `(N+1)..N` is empty (correct), but lines 181–188 then unconditionally re-fetch and re-send block N — the same block that was just sent at startup. Block N ends up in the TUI's L1 history twice.

Fix: guard the unconditional send with a check against `last_block`:
```suggestion
        if block_num > last_block.unwrap_or(0) {
            if let Ok(Some(block)) =
                provider.get_block_by_number(BlockNumberOrTag::Number(block_num)).full().await
            {
                let info = extract_l1_block_info(&block, batcher_address);
                if result_tx.send(info).await.is_err() {
                    return Ok(());
                }
            }
        }
```

crates/infra/basectl/src/rpc/admin.rs
```diff
@@ -0,0 +1,187 @@
...
+            for enode in &peers.el_enodes {
+                let r: Result<bool, _> =
+                    ClientT::request(&el_client, "admin_addPeer", rpc_params![enode]).await;
+                if matches!(r, Ok(true)) {
+                    el_ok += 1;
+                }
+            }
+        }
+
+        if cl_ok != peers.cl_addrs.len() || el_ok != peers.el_enodes.len() {
```
Pulsator01: **Bug: `unpause_sequencer_node` false-fails when `el_rpc` is `None` but `peers.el_enodes` is non-empty**

The EL reconnect loop is guarded by `if let Some(ref el_rpc) = node.el_rpc`, so when `el_rpc` is `None`, `el_ok` stays 0. If `peers.el_enodes` is non-empty (populated during a prior `pause_sequencer_node` call when `el_rpc` was `Some`), the check `el_ok != peers.el_enodes.len()` fires and `anyhow::bail!` returns an error even though no reconnection was possible and none was attempted. The operator sees a false failure toast.

The guard should skip the check (or set `el_ok = peers.el_enodes.len()` before entering the block) when `el_rpc` is `None`:
```suggestion
        if cl_ok != peers.cl_addrs.len() || (node.el_rpc.is_some() && el_ok != peers.el_enodes.len()) {
```

crates/infra/basectl/src/rpc/l1.rs
```diff
@@ -0,0 +1,265 @@
...
+    let mut stream = sub.into_stream();
+
+    let _ = mode_tx.send(L1ConnectionMode::WebSocket).await;
+
+    let mut last_block: Option<u64> = None;
+
+    if let Ok(Some(block)) = provider.get_block_by_number(BlockNumberOrTag::Latest).full().await {
+        let info = extract_l1_block_info(&block, batcher_address);
+        last_block = Some(block.header.number);
+        let _ = result_tx.send(info).await;
```
Pulsator01: **Inconsistency: initial block send silently discards channel-closed error**

The initial `result_tx.send` uses `let _ = ...` (error ignored), while every send in the subscription loop correctly checks `.is_err()` and returns. If the receiver is dropped in the window between WS connection and the loop starting (e.g., during TUI teardown), the initial block is silently lost with no log entry, whereas the loop would have returned cleanly.

```suggestion
        if result_tx.send(info).await.is_err() {
            return Ok(());
        }
```

haardikk21 (approved):

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #3217, opened 2026-06-03
feat: make beacon API optional

dguenther:
Makes beacon API configuration optional so calldata-only / beaconless devnet modes can run without an L1 beacon endpoint. The base-anvil L3 devnet wiring is split into the stacked follow-up PR #3219.

Targeting `privacy` for now, since this may be temporary and replaced with alt-DA

### DIFF
--- a/bin/prover/nitro-host/src/cli.rs
+++ b/bin/prover/nitro-host/src/cli.rs
@@ -156,7 +156,7 @@ impl ServerArgs {
         let config = ProverConfig {
             l1_eth_url: self.server.l1_eth_url,
             l2_eth_url: self.server.l2_eth_url,
-            l1_beacon_url: self.server.l1_beacon_url,
+            l1_beacon_url: Some(self.server.l1_beacon_url),
             l2_chain_id: self.server.l2_chain_id,
             rollup_config,
             l1_config,
@@ -218,7 +218,7 @@ impl LocalArgs {
         let prover_config = ProverConfig {
             l1_eth_url: self.server.l1_eth_url,
             l2_eth_url: self.server.l2_eth_url,
-            l1_beacon_url: self.server.l1_beacon_url,
+            l1_beacon_url: Some(self.server.l1_beacon_url),
             l2_chain_id: self.server.l2_chain_id,
             rollup_config,
             l1_config,
--- a/bin/prover/nitro-host/src/cli.rs
+++ b/bin/prover/nitro-host/src/cli.rs
@@ -75,7 +75,11 @@ struct ProverServerArgs {
 
     /// L1 beacon API URL.
     #[arg(long, env = "L1_BEACON_URL")]
-    l1_beacon_url: String,
+    l1_beacon_url: Option<String>,
+
+    /// Run proving without an L1 beacon API (calldata-only / appchain mode).
+    #[arg(long, env = "L1_CALLDATA_ONLY")]
+    l1_calldata_only: bool,
 
     /// L2 chain ID.
     #[arg(long, env = "L2_CHAIN_ID")]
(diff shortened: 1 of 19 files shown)


bin/prover/zk/src/cli.rs
```diff
@@ -157,6 +157,14 @@ impl ZkArgs {
         let repo = ProofRequestRepo::new(pool);
         info!("database connection initialized");
 
+        // Normalize an absent/blank beacon address to `None` so the prover runs in
+        // calldata-only mode.
+        let configured_beacon_url = self
```
hughy: Do we need to do the same kind of validation for beacon URL vs. calldata-only that the other CLI commands have? This will default to calldata-only if there is no beacon URL while the others explicitly require some calldata-only flag

hughy (approved):

bin/prover/zk/src/cli.rs
dguenther: Taking a look at this, good spot! I can't remember if I forgot to update this one, or if there was a reason for it

bin/prover/nitro-host/src/cli.rs
```diff
@@ -75,7 +75,11 @@ struct ProverServerArgs {
 
     /// L1 beacon API URL.
     #[arg(long, env = "L1_BEACON_URL")]
-    l1_beacon_url: String,
+    l1_beacon_url: Option<String>,
+
+    /// Run proving without an L1 beacon API (calldata-only / appchain mode).
+    #[arg(long, env = "L1_CALLDATA_ONLY")]
+    l1_calldata_only: bool,
```
and-cb, line 82: I'm not fully convinced this argument should be needed, but anyway... if we switch to a compile-time feature in the future, this won't be relevant anymore so LGTM for now

crates/consensus/cli/src/l1.rs
```diff
@@ -45,9 +56,78 @@ impl Default for L1ClientArgs {
...
+impl L1ClientArgs {
+    /// Validates that exactly one derivation mode is selected, making the choice explicit so a
+    /// main-chain node cannot silently run without a beacon API.
+    ///
+    /// - `--l1.beacon <url>` (no `--l1.calldata-only`): beacon mode (main Base chain).
+    /// - `--l1.calldata-only` (no `--l1.beacon`): calldata-only mode (appchain).
+    /// - neither: error - the operator must state intent.
+    /// - both: error - conflicting flags.
+    pub fn validate(&self) -> eyre::Result<()> {
+        match (self.l1_beacon.is_some(), self.l1_calldata_only) {
```
and-cb: Other code in this PR does some stripping of the URL (so that e.g. passing a blank string is equivalent to not passing any argument). Should the same be done here?

and-cb (approved):

crates/consensus/cli/src/l1.rs
```diff
@@ -45,9 +61,114 @@ impl Default for L1ClientArgs {
...
+                "conflicting L1 derivation modes: --l1.beacon and --l1.calldata-only are mutually \
+                 exclusive"
+            )),
+        }
+    }
+
+    /// Logs the active L1 derivation mode at startup. Call after [`Self::validate`], which has
+    /// already vetted the beacon URL.
+    pub fn log_derivation_mode(&self) {
+        match self.l1_beacon().ok().flatten() {
```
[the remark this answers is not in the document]
dguenther: Good catch on the .ok() swallowing the error. In the current code this can't actually mislead: the only caller is node.rs, which calls validate()? on the line immediately before log_derivation_mode(), and validate() runs l1_beacon()?, so any unparseable URL is rejected before we ever log. follow.rs doesn't call this method at all.

I'd rather not change a logging helper to return Result for a state that can't be reached, but you're right that the precondition is load-bearing. The function doc already notes "Call after validate(), which has already vetted the beacon URL"

crates/proof/host/src/host.rs
```diff
@@ -191,10 +191,19 @@ impl Host {
...
-        ))
-        .await;
+        let blob_provider = match self
+            .config
+            .prover
+            .l1_beacon_url
+            .as_deref()
+            .map(str::trim)
+            .filter(|url| !url.is_empty())
+        {
```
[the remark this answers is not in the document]
dguenther: Agreed this is redundant with the upstream CLI/env normalization. It's intentional defense-in-depth at the boundary where the blob provider is actually constructed, so `create_providers` doesn't depend on every caller having normalized first. It's the last line of defense and it's cheap.
 
On extracting a shared helper: the three sites live in different crates (`base-consensus-cli`, `base-proof-succinct-host-utils`, and here in `base-proof-host`), so deduping three lines would mean introducing a shared dependency just for this. I don't think the coupling is worth it for the size of the duplication.

crates/proof/succinct/utils/host/src/fetcher.rs
```diff
@@ -118,6 +118,24 @@ pub fn get_rpcs_from_env() -> RPCConfig {
...
+        (None, false) => {
+            panic!("provide L1_BEACON_RPC for blob-backed proving or set L1_CALLDATA_ONLY=true")
+        }
+        (Some(_), true) => {
+            panic!("L1_BEACON_RPC and L1_CALLDATA_ONLY=true are mutually exclusive")
+        }
+    }
+}
+
 /// The info to fetch for a block.
```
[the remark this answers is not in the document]
dguenther: Fair point on the inconsistency with the CLIs. The reason it panics here is local consistency: `l1_beacon_rpc_from_env` is called only from `get_rpcs_from_env`, which already panics on every env misconfiguration: `expect("L1_RPC must be set")`, `expect("L2_RPC must be set")`, and the `expect("L1_BEACON_RPC must be a valid URL")` on the line right above these arms. Returning `Result` from just the beacon case would actually make this function less internally consistent until the whole `get_rpcs_from_env` signature is converted.

 Worth noting the panic messages here are explicit and actionable (e.g. "provide L1_BEACON_RPC for blob-backed proving or set L1_CALLDATA_ONLY=true"), so it's not a bare backtrace. I'd prefer to keep it consistent with the rest of the function for now.

PR #3222, opened 2026-06-04
chore(prover-service-client): add retrying proof requester

leopoldjoy:
## What

Adds S1 from the proof consolidation plan: a shared retry/backoff wrapper for proof requester operations.

- Adds `ProofRequesterRetryConfig` with exponential backoff defaults.
- Adds `RetryingProofRequester`, an opt-in wrapper around `Arc<dyn ProofRequesterProvider>`.
- Retries requester RPCs when `ProverServiceClientError::is_retryable()` returns true:
  - `prove_block_range`
  - `get_proof`
  - `list_proofs`
- Leaves existing `ProofRequesterClient` behavior unchanged; consumers can opt in by wrapping it.

## Why

This centralizes retry/backoff behavior in `base-prover-service-client` so proposer, challenger, and future callers do not each implement local retry loops.

## Test plan

- `cargo +nightly fmt --all --check`
- `cargo test -p base-prover-service-client`
- `cargo clippy -p base-prover-service-client --all-targets -- -D warnings`

### DIFF
--- a/crates/proof/prover-service/client/src/config.rs
+++ b/crates/proof/prover-service/client/src/config.rs
@@ -10,6 +10,8 @@ use thiserror::Error;
 use tracing::debug;
 use url::Url;
 
+use crate::ProofRequesterRetryConfig;
+
 /// Errors that can occur during prover-service client configuration validation.
 #[derive(Clone, Debug, Error, Eq, PartialEq)]
 pub enum ProverServiceClientConfigError {
@@ -54,6 +56,7 @@ pub struct ProverServiceClientConfig {
     request_timeout: Duration,
     poll_interval: Duration,
     max_wait: Duration,
+    retry: ProofRequesterRetryConfig,
 }
 
 impl ProverServiceClientConfig {
@@ -73,6 +76,7 @@ impl ProverServiceClientConfig {
             request_timeout: Self::DEFAULT_REQUEST_TIMEOUT,
             poll_interval: Self::DEFAULT_POLL_INTERVAL,
             max_wait: Self::DEFAULT_MAX_WAIT,
+            retry: ProofRequesterRetryConfig::default(),
         }
     }
 
@@ -96,6 +100,14 @@ impl ProverServiceClientConfig {
         self.max_wait
     }
 
+    /// Return the retry configuration applied by [`ProofRequesterClient`] when calling
+    /// requester JSON-RPC methods.
+    ///
+    /// [`ProofRequesterClient`]: crate::ProofRequesterClient
+    pub const fn retry_config(&self) -> ProofRequesterRetryConfig {
+        self.retry
+    }
+
     /// Set the per-request timeout used by the JSON-RPC HTTP client.
     pub const fn with_request_timeout(mut self, request_timeout: Duration) -> Self {
         self.request_timeout = request_timeout;
@@ -111,6 +123,15 @@ impl ProverServiceClientConfig {
     /// Set the maximum time to wait for proof completion.
     pub const fn with_max_wait(mut self, max_wait: Duration) -> Self {
         self.max_wait = max_wait;
+        self
+    }
+
+    /// Set the retry configuration applied by [`ProofRequesterClient`] when calling
+    /// requester JSON-RPC methods.
+    ///
+    /// [`ProofRequesterClient`]: crate::ProofRequesterClient
+    pub const fn with_retry_config(mut self, retry: ProofRequesterRetryConfig) -> Self {
+        self.retry = retry;
         self
     }
 
--- a/crates/proof/prover-service/client/src/config.rs
+++ b/crates/proof/prover-service/client/src/config.rs
@@ -36,6 +36,9 @@ pub enum ProverServiceClientConfigError {
     /// The configured poll interval is greater than the maximum wait duration.
     #[error("poll interval must be less than or equal to max wait")]
     PollIntervalExceedsMaxWait,
+    /// The configured retry initial delay is greater than the retry max delay.
+    #[error("retry initial delay must be less than or equal to retry max delay")]
+    RetryInitialDelayExceedsMaxDelay,
 }
 
 /// Errors that can occur when building a prover-service client.
@@ -164,6 +167,17 @@ impl ProverServiceClientConfig {
             return Err(ProverServiceClientConfigError::PollIntervalExceedsMaxWait);
         }
 
+        // Validate retry sub-config. `max_attempts == 0` is intentionally accepted and
+        // documented as equivalent to one attempt by `ProofRequesterRetryConfig`, matching
+        // the workspace's `base_proof_rpc::config::RetryConfig` convention; the broader
+        // API hardening (private fields + construction-time validation) is tracked as
+        // S2 in the consolidation plan. We do reject `initial_delay > max_delay` here so
+        // misconfiguration surfaces as a config error rather than a silent clamp inside
+        // the backoff builder.
+        if self.retry.initial_delay > self.retry.max_delay {
+            return Err(ProverServiceClientConfigError::RetryInitialDelayExceedsMaxDelay);
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/prover-service/client/src/retry.rs
```diff
@@ -0,0 +1,354 @@
...
+pub const DEFAULT_PROOF_REQUESTER_INITIAL_DELAY: Duration = Duration::from_millis(100);
+
+/// Default maximum retry delay for requester RPC operations.
+pub const DEFAULT_PROOF_REQUESTER_MAX_DELAY: Duration = Duration::from_secs(10);
+
+/// Exponential backoff configuration for proof requester retries.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ProofRequesterRetryConfig {
+    /// Maximum number of retry attempts.
+    pub max_attempts: u32,
```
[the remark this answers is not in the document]
leopoldjoy: Pushing back on the rename: the existing `base_proof_rpc::config::RetryConfig` in `crates/proof/rpc/src/config.rs` already uses the same `max_attempts: u32` → `with_max_times(...)` mapping, and renaming here would diverge from that workspace convention. Clarified the semantics on the field doc instead so callers aren't surprised: total calls in an exhausted run = `max_attempts + 1`.

crates/proof/prover-service/client/src/retry.rs
jackchuma, line 1: Can we work this directly into the existing ProofRequesterProvider instead of adding another wrapper?

jackchuma: (I recognize we need this for the worker provider too. I can spin up a PR for that tomorrow)

leopoldjoy: Good call, I've folded the retries directly into `ProofRequesterClient` so every consumer (proposer, challenger, worker, CLI) of the requester surface picks them up via `ProverServiceClientConfig::with_retry_config(...)`. This should compose cleanly with your `ProverWorkerProvider` rework — I can also mirror the same `ProverServiceClientConfig::retry_config()` plumbing on the worker side if useful, or leave it for your PR.

crates/proof/prover-service/client/src/requester.rs
```diff
@@ -70,7 +95,22 @@ impl ProofRequesterClient {
             session_id = ?request.proof.session_id,
             "proving block range"
         );
-        Ok(self.inner.prove_block_range(request).await?)
+        (|| {
```
jackchuma, line 109: When `request.proof.session_id` is `None`, the server generates a new UUID for each `prove_block_range` call, so a retry after a timeout or dropped response can enqueue the same proof multiple times under different session IDs. Either generate / fill a client-side session ID before entering the retry loop or skip retries for requests that do not provide an idempotency key

crates/proof/prover-service/client/src/retry.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// Default initial retry delay for requester RPC operations.
+pub const DEFAULT_PROOF_REQUESTER_INITIAL_DELAY: Duration = Duration::from_millis(100);
+
+/// Default maximum retry delay for requester RPC operations.
+pub const DEFAULT_PROOF_REQUESTER_MAX_DELAY: Duration = Duration::from_secs(10);
+
+/// Exponential backoff configuration for proof requester retries.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ProofRequesterRetryConfig {
```
jackchuma, line 21: Not blocking for this PR, but there are similar retry configs already defined in the codebase. I think a future cleanup PR to extract these into a small shared `retry` crate would be good

crates/proof/prover-service/client/src/requester.rs
leopoldjoy: Went with the "skip retry when session_id.is_none()" guard plus a regression test. This way the change is one line and composes cleanly with making `session_id` required at the protocol layer.

crates/proof/prover-service/client/src/retry.rs
leopoldjoy: Agreed. This makes sense as a separate cleanup PR alongside `base_proof_rpc::config::RetryConfig` and the TEE submitter backoff config; will track it as a follow-up.

crates/proof/prover-service/client/src/retry.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Maximum number of retries performed after the initial call. The total number of
+    /// requester calls in a fully exhausted run is therefore `max_attempts + 1`. Zero is
+    /// treated as one (a single retry); this matches the existing
+    /// `base_proof_rpc::config::RetryConfig` convention in the workspace.
+    pub max_attempts: u32,
+    /// First delay after a retryable requester failure.
+    pub initial_delay: Duration,
+    /// Maximum delay between retry attempts.
+    pub max_delay: Duration,
+}
```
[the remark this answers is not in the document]
leopoldjoy: Pushing back on this for #3222: the same pub-field convention is already used by `base_proof_rpc::config::RetryConfig` (cited as workspace precedent earlier in this PR), so locking the API down here would diverge from that. Tracked the broader API-hardening (private fields + normalized accessors / construction-time validation) as part of S2 in the consolidation plan — extracting these duplicated retry configs into a shared crate is the right venue for it, and we'll handle it there alongside `base_proof_rpc::config::RetryConfig` and the TEE submitter backoff config.

crates/proof/prover-service/client/src/config.rs
```diff
@@ -54,6 +56,7 @@ pub struct ProverServiceClientConfig {
     request_timeout: Duration,
     poll_interval: Duration,
     max_wait: Duration,
+    retry: ProofRequesterRetryConfig,
```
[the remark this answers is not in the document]
leopoldjoy: Good catch — addressed in c1ce7b29c by adding a `RetryInitialDelayExceedsMaxDelay` variant and validating `retry.initial_delay <= retry.max_delay` in `ProverServiceClientConfig::validate()`, alongside two new unit tests covering the rejection and the documented `max_attempts = 0` acceptance.

Keeping `max_attempts == 0` accepted (it's documented as equivalent to one attempt, matching workspace's `base_proof_rpc::config::RetryConfig` convention — see [discussion_r3352829047](https://github.com/base/base/pull/3222#discussion_r3352829047)). Validating in `ProofRequesterRetryConfig::new()` itself would require breaking the `const fn new` API; the broader hardening (private fields + construction-time validation, also raised in [discussion_r3358746246](https://github.com/base/base/pull/3222#discussion_r3358746246)) is tracked as S2 in the consolidation plan and will land alongside the other duplicated retry configs in the workspace.

jackchuma (approved):

PR #3244, opened 2026-06-05
feat(prover-service): implement worker job api server

mw2000:
Wire `ProverWorkerApiServer` (getNextProof/heartbeat/submitProof) onto `ProverServiceServer`, mapping database claim/lock outcomes to JSON-RPC error codes and adding a tunable `WorkerApiConfig` for lock durations and reclaim budget.

### DIFF
--- a/crates/proof/prover-service/db/src/repo.rs
+++ b/crates/proof/prover-service/db/src/repo.rs
@@ -553,22 +553,6 @@ impl ProofRequestRepo {
         row.as_ref().map(row_to_proof_job).transpose()
     }
 
-    /// Fetch the stored `result_payload` for a canonical `session_id`, if any.
-    async fn stored_result_payload(&self, session_id: &str) -> Result<Option<serde_json::Value>> {
-        let row = sqlx::query(
-            r#"
-            SELECT result_payload
-            FROM proof_requests
-            WHERE COALESCE(session_id, id::text) = $1
-            "#,
-        )
-        .bind(session_id)
-        .fetch_optional(&self.pool)
-        .await?;
-
-        Ok(row.and_then(|row| row.get::<Option<serde_json::Value>, _>("result_payload")))
-    }
-
     /// Extend the lock for the currently owned worker proof job (`heartbeat`).
     pub async fn heartbeat_proof_job(&self, req: HeartbeatProofJob) -> Result<HeartbeatOutcome> {
         let session_id = canonical_session_id(&req.session_id)
@@ -645,8 +629,7 @@ impl ProofRequestRepo {
             && existing.lock_id == Some(req.lock_id)
             && existing.worker_id.as_deref() == Some(req.worker_id.as_str())
         {
-            let stored = self.stored_result_payload(&session_id).await?;
-            return Ok(if stored.as_ref() == Some(&result_payload) {
+            return Ok(if existing.result_payload.as_ref() == Some(&result_payload) {
                 SubmitProofOutcome::Completed(existing)
             } else {
                 SubmitProofOutcome::ResultConflict { job: existing }
@@ -1903,6 +1886,7 @@ fn row_to_proof_job(row: &sqlx::postgres::PgRow) -> Result<ProofJob> {
         claimed_at: row.get("claimed_at"),
         last_heartbeat_at: row.get("last_heartbeat_at"),
         error_message: base.error_message,
+        result_payload: base.result_payload,
         created_at: base.created_at,
         updated_at: base.updated_at,
         completed_at: base.completed_at,
--- a/crates/proof/prover-service/db/src/repo.rs
+++ b/crates/proof/prover-service/db/src/repo.rs
@@ -553,6 +553,22 @@ impl ProofRequestRepo {
         row.as_ref().map(row_to_proof_job).transpose()
     }
 
+    /// Fetch the stored `result_payload` for a canonical `session_id`, if any.
+    async fn stored_result_payload(&self, session_id: &str) -> Result<Option<serde_json::Value>> {
+        let row = sqlx::query(
+            r#"
+            SELECT result_payload
+            FROM proof_requests
+            WHERE COALESCE(session_id, id::text) = $1
+            "#,
+        )
+        .bind(session_id)
+        .fetch_optional(&self.pool)
+        .await?;
+
+        Ok(row.and_then(|row| row.get::<Option<serde_json::Value>, _>("result_payload")))
+    }
+
     /// Extend the lock for the currently owned worker proof job (`heartbeat`).
     pub async fn heartbeat_proof_job(&self, req: HeartbeatProofJob) -> Result<HeartbeatOutcome> {
         let session_id = canonical_session_id(&req.session_id)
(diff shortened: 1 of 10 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/prover-service/src/server/mod.rs
```diff
@@ -46,8 +88,17 @@ impl ProverServiceServer {
         repo: ProofRequestRepo,
         manager: ProofRequestManager,
         max_proof_retries: i32,
+        worker_queue: WorkerQueueConfig,
```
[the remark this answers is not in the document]
mw2000: this should be fine since the call site in `bin/prover/zk/cli.rs` is from the `zk-prover-service` and not the new `prover-service`

crates/proof/prover-service/src/server/mod.rs
```diff
@@ -12,30 +12,72 @@ use jsonrpsee::{
...
+pub struct WorkerApiConfig {
+    /// Lock duration applied when a worker requests `0` seconds.
+    pub default_lock_duration_seconds: u32,
+    /// Upper bound a worker-requested lock duration is clamped to.
+    pub max_lock_duration_seconds: u32,
+}
+
+impl WorkerApiConfig {
+    /// Default worker API tuning.
+    pub const DEFAULT: Self = Self::new(300, 3600);
```
jackchuma: Nit: could this just be inlined directly in the `default()` impl below?

crates/proof/prover-service/src/server/mod.rs
```diff
@@ -12,30 +12,72 @@ use jsonrpsee::{
...
+/// JSON-RPC server implementing the requester and worker API traits.
 #[derive(Clone)]
 pub struct ProverServiceServer {
     repo: ProofRequestRepo,
     manager: ProofRequestManager,
-    /// Shared `retry_count` cap with [`crate::worker::StatusPoller`] (same as `retry_or_fail_stuck_request`).
+    /// Shared `retry_count` cap with [`crate::worker::StatusPoller`].
     max_proof_retries: i32,
+    /// Worker job lock-duration tuning.
+    worker: WorkerApiConfig,
```
jackchuma: Nit: field name of `worker_config` or `worker_cfg` would make more sense

jackchuma: Also should there be a single `ServerConfig` struct that folds both into one `cfg` ?

crates/proof/prover-service/src/server/worker_api.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+/// Build a non-retryable ownership rejection and log it for diagnostics.
+fn reject_ownership(method: &str, session_id: &str, reason: &str) -> ErrorObjectOwned {
+    warn!(method = %method, session_id = %session_id, reason = %reason, "rejected worker ownership request");
+    failed_precondition(format!("{method} rejected for session_id {session_id}: {reason}"))
+}
+
+#[cfg(test)]
+mod tests {
```
jackchuma, line 277: Nit: I think you can reduce duplication in these tests by using `rstest` as used elsewhere in the codebase

crates/proof/prover-service/src/server/worker_api.rs
```diff
@@ -0,0 +1,288 @@
...
+            .map_err(|e| invalid_argument(format!("{e}")))?;
+        let lock_id = parse_lock_id(&request.lock_id)?;
+
+        let outcome = self
+            .repo
+            .complete_claimed_proof_job(CompleteClaimedProofJob {
+                session_id,
+                lock_id,
+                worker_id: request.worker_id.clone(),
+                result: request.result,
```
jackchuma, line 180: Validate proof result before completing job. When a worker has a valid lock for one proof type, this forwards the submitted `ProofResult` unchanged, so a buggy or misconfigured worker can complete a TEE job with a compressed/SNARK result, or a ZK job with the wrong VM. `complete_claimed_proof_job` will store that payload and mark the request `SUCCEEDED`, after which requester-side adapters reject the terminal result as the wrong variant; validate the result variant and capability against the claimed job before completing it

crates/proof/prover-service/src/server/worker_api.rs
```diff
@@ -0,0 +1,288 @@
...
+            )),
+            SubmitProofOutcome::StaleLock(_) => Err(reject_ownership(
+                "submit_proof",
+                &request.session_id,
+                "lock is held by another worker or has been rotated",
+            )),
+            SubmitProofOutcome::Expired(_) => {
+                Err(reject_ownership("submit_proof", &request.session_id, "lock has expired"))
+            }
+            SubmitProofOutcome::Terminal(_) => Err(reject_ownership(
```
jackchuma, line 233: Make `submitProof` idempotent after a lost response. If the first `submitProof` updates the row but the HTTP response is dropped or times out, `ProverWorkerClient::submit_proof` retries the same request; the retry sees the job as terminal and this branch returns `FAILED_PRECONDITION`, turning a successful submission into a client-visible worker lease failure. Since the client treats submit as retryable, an identical retry for the same worker/lock/result should return the completed job instead of rejecting it

crates/proof/prover-service/src/server/worker_api.rs
mw2000: `submitProof` now validates the submitted `ProofResult` against the claimed job before completion

crates/proof/prover-service/src/server/worker_api.rs
mw2000: `submitProof` now treats an identical retry after a successful completion as idempotent

crates/proof/prover-service/db/src/repo.rs
```diff
@@ -630,19 +605,58 @@ impl ProofRequestRepo {
...
         let session_id = canonical_session_id(&req.session_id)
             .map_err(|e| sqlx::Error::InvalidArgument(e.to_string()))?;
+
+        // Read first to classify ownership state and detect idempotent retries.
+        let Some(existing) = self.get_proof_job_by_canonical_session_id(&session_id).await? else {
+            return Ok(SubmitProofOutcome::NotFound);
+        };
+
+        // Idempotent retry by the owning worker/lock; a differing payload conflicts.
+        if existing.job_status == ProofJobStatus::Succeeded
```
jackchuma, line 621: Re-check idempotent submit after the `UPDATE` loses the race. When a worker retries `submitProof` while the original submission is still in flight, this early idempotency check can read the row as `CLAIMED` before the first request commits; the later `UPDATE` then waits/rechecks, returns no row because the first request already set `SUCCEEDED`, and the fallback classifies it as `Terminal` instead of comparing the stored `result_payload`. That makes a same-worker/same-lock/same-result retry return a non-retryable failed precondition even though the proof was accepted, so the idempotent comparison needs to be repeated after the failed `UPDATE` path as well

jackchuma (approved):
left one more minor comment that could be addressed in a follow up

PR #3258, opened 2026-06-05
fix(precompile-storage): propagate SSTORE refunds through IntoPrecompileResult (BOP-289)

eric-ships:
[BOP-289](https://linear.app/coinbase/issue/BOP-289/9-generic-precompile-success-conversion-drops-accumulated-sstore)

## Motivation

Successful precompile calls that clear previously nonzero storage slots accumulate an EIP-3529 gas refund in the storage provider via `refund_gas`. The `StorageCtx::success_output` path correctly copies this refund into `PrecompileOutput::gas_refunded`. The generic `IntoPrecompileResult::into_precompile_result` path did not: it constructed a `PrecompileOutput::new(gas, bytes, state_gas)` and returned it unchanged, silently dropping the accumulated refund.

This is reachable in production. Operations that zero out storage slots before returning through the generic dispatch path include B20 balance zeroing on transfer, allowance zeroing on approval, full burns, feature deactivation, and policy admin finalization/renounce. All of those dispatch paths used `IntoPrecompileResult`, so callers overpaid gas whenever those slots were cleared.

## What changed

- `IntoPrecompileResult::into_precompile_result` gains a `gas_refunded: i64` parameter. On the success arm, `out.gas_refunded` is set before returning. The error arm is unchanged (refunds are not meaningful on revert/halt).
- All six dispatch call sites (`policy`, `activation`, `b20_factory`, `b20_asset`, `b20_stablecoin`, `activation::assert_activated`) pass `ctx.gas_refunded()` or `self.storage.gas_refunded()` as the new argument.
- Three regression tests in `error.rs` assert that the refund propagates on success, is zero when zero, and is not set on the error path.

An alternative was to replace all `IntoPrecompileResult` call sites with `ctx.success_output(encode_ok(res))`, which already handles refunds. That would have worked but requires the ctx to be in scope at every call site, which is not always the case (e.g. `assert_activated` holds a `StorageCtx` indirectly via `self.storage`, not a local binding). Threading `gas_refunded` through the existing signature is a smaller, more contained change that does not require restructuring the dispatch functions.

### DIFF
--- a/crates/common/precompile-storage/src/error.rs
+++ b/crates/common/precompile-storage/src/error.rs
@@ -131,10 +131,14 @@ impl BasePrecompileError {
 /// Extension trait to convert `Result<T, BasePrecompileError>` into a [`PrecompileResult`].
 pub trait IntoPrecompileResult<T> {
     /// Converts `self` into a [`PrecompileResult`] using `encode_ok` for the success path.
+    ///
+    /// `gas_refunded` is copied into `PrecompileOutput::gas_refunded` on success so revm can
+    /// apply EIP-3529 refund accounting at the transaction level. Pass `ctx.gas_refunded()`.
     fn into_precompile_result(
         self,
         gas: u64,
         state_gas: u64,
+        gas_refunded: i64,
         encode_ok: impl FnOnce(T) -> Bytes,
     ) -> PrecompileResult;
 }
@@ -144,10 +148,15 @@ impl<T> IntoPrecompileResult<T> for Result<T> {
         self,
         gas: u64,
         state_gas: u64,
+        gas_refunded: i64,
         encode_ok: impl FnOnce(T) -> Bytes,
     ) -> PrecompileResult {
         match self {
-            Ok(res) => Ok(PrecompileOutput::new(gas, encode_ok(res), state_gas)),
+            Ok(res) => {
+                let mut out = PrecompileOutput::new(gas, encode_ok(res), state_gas);
+                out.gas_refunded = gas_refunded;
+                Ok(out)
+            }
             Err(err) => err.into_precompile_result(gas, state_gas),
         }
     }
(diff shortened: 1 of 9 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (approved):

Sertug17:
The EIP 2200 stipend guard removed from `sstore` in `evm.rs` was added in #3285 to prevent warm dirty SSTOREs (~200 gas) from succeeding inside a 2300 gas stipend call. With this PR moving refund propagation to `result_output`, is the stipend guard being intentionally dropped or does it need to be re added elsewhere?

Specifically: without the guard, a precompile `sstore` call inside a `.transfer()` (2300 gas stipend) can still succeed where the EVM SSTORE opcode would have halted. Happy to be wrong if the new refund propagation path covers this just flagging for reviewer awareness.

Maiwandzadran (approved):

cb-heimdall:
Review Error for Maiwandzadran @ 2026-06-08 04:17:53 UTC

Sertug17:
Opened #3299 to track the latent gap flagged by the bot: `Snapshot` does not include `gas_refunded`, so `checkpoint_revert` leaves stale refunds after rollback. Happy to send a PR once vouched.

PR #3273, opened 2026-06-06
feat(proposer): rework proving pipeline

leopoldjoy:
## Summary

Reworks the proposer proving pipeline into the planned concurrent dispatcher / collector architecture.

The pipeline now runs two cooperative tasks per session:

- **Dispatcher loop:** recovers the latest on-chain game, walks forward by `block_interval`, and dispatches proof requests up to the safe head so the proposer can catch up from a deficit.
- **Collector loop:** independently derives prover-service sessions from canonical L2 output roots, polls completed proofs, and submits proposals sequentially in on-chain order.

Submit failures restart both loops from fresh chain-derived recovery state. This keeps the dispatcher and collector independent while ensuring failure handling returns to the canonical on-chain tip.

```text
┌──────────┐     ┌──────────────────┐
│ RECOVER  │ ──> │ DISPATCH LOOP    │ ──> prover service
│ (cached) │     └──────────────────┘
│          │     ┌──────────────────┐
│          │ ──> │ COLLECT LOOP     │ ──> L1 submitter
└──────────┘     └──────────────────┘
```

## Behavior

- Dispatcher can fan out proof requests for every eligible target up to safe head, instead of limiting the proposer to one in-flight proof.
- Collector preserves sequential submission order, but drains immediately-ready proof backlogs in one tick instead of sleeping `poll_interval` between successful submits.
- Normal prover-service sessions remain deterministic and root-derived so restarts can rediscover in-flight work.
- Terminal prover-service `Failed` sessions are re-dispatched so prover-service actually requeues them.
- `SubmitAction::Discard` no longer skips the target. It dispatches retry-specific sessions so `L1OriginTooOld` / `InvalidSigner` can obtain a fresh TEE proof instead of looping on the old `Succeeded` session.
- Discard retries are capped by `max_retries`; exhaustion clears retry state and drops the recovery cache.
- Missing retry-specific sessions are treated as dispatch-needed and replaced with a fresh retry session.
- Request-build failures from transient L1/L2 RPC reads are logged as `build_failed` dispatch outcomes and do not burn the proof retry budget.
- Submit failures that invalidate progress (`RootMismatch`, invalid parent game, timeout, transient submit failure) request a session restart so both loops re-walk from chain state.
- `GameAlreadyExists` is treated as success-like: gauges advance and recovery is refreshed to discover the existing game.
- Task exits, task errors, and task panics restart the pipeline session instead of terminating `run()`.

## Observability

- Removed obsolete in-memory queue metrics that no longer apply to the final task structure.
- Added `proof_dispatch_total{outcome="build_failed"}` for request-build failures that never reached prover-service.
- `last_proposed_block` reflects the recovered on-chain tip on every iteration.
- `last_collected_block` advances on successful inline submit / success-like `GameAlreadyExists` handling.
- `PipelineConfig.submit_timeout` controls the inline validation + submit budget.

## Tests

Added or updated coverage for:

- dispatcher fan-out up to safe head
- collector backlog draining without `poll_interval` lag
- prover-service `Failed` re-dispatch
- discard retry-specific sessions
- discard retry exhaustion
- missing retry-session replacement
- submit-failure restart signaling
- task panic restart handling
- build-error no-retry-budget behavior
- inline submit success and failure mappings
- recovery cache / forward-walk behavior

Verification run locally:

- `cargo test -p base-proposer --lib` -> 121 passed
- `cargo test -p base-proposer --lib --features metrics` -> 123 passed

### DIFF
--- a/crates/proof/proposer/src/pipeline.rs
+++ b/crates/proof/proposer/src/pipeline.rs
@@ -488,122 +488,135 @@ where
         if cursor.is_none_or(|c| c.l2_block_number < recovered.l2_block_number) {
             *cursor = Some(recovered);
         }
-        let Some(current) = *cursor else { return };
+        while let Some(current) = *cursor {
+            let target_block = match self.next_target_block(current.l2_block_number) {
+                Some(block) => block,
+                None => break,
+            };
 
-        let target_block = match self.next_target_block(current.l2_block_number) {
-            Some(block) => block,
-            None => return,
-        };
-
-        if target_block > safe_head {
-            debug!(
-                current_block = current.l2_block_number,
-                target_block,
-                safe_head,
-                "Safe head below collection target, waiting for L2 head to advance"
-            );
-            Metrics::pipeline_retries().set(retry_counts.values().sum::<u32>() as f64);
-            return;
-        }
-
-        let poll = if let Some(session_id) = retry_sessions.get(&target_block).cloned() {
-            self.proof_collector.poll_session(target_block, session_id).await
-        } else {
-            self.proof_collector.poll(target_block).await
-        };
-
-        match poll {
-            TargetPoll::Ready { session_id, proof } => {
-                info!(
+            if target_block > safe_head {
+                debug!(
+                    current_block = current.l2_block_number,
                     target_block,
-                    session_id = %session_id,
-                    "Proof ready, submitting inline"
+                    safe_head,
+                    "Safe head below collection target, waiting for L2 head to advance"
                 );
-                Metrics::proof_collection_total(Metrics::COLLECTION_OUTCOME_READY).increment(1);
-                match self.submit_inline(target_block, &current, proof, retry_counts, cache).await {
-                    SubmitEffect::Submitted => {
-                        retry_sessions.remove(&target_block);
-                        discard_retry_counts.remove(&target_block);
-                        if let Some(cached) = cache.as_ref() {
-                            *cursor = Some(cached.state);
+                break;
+            }
+
+            let poll = if let Some(session_id) = retry_sessions.get(&target_block).cloned() {
+                self.proof_collector.poll_session(target_block, session_id).await
+            } else {
+                self.proof_collector.poll(target_block).await
+            };
+
+            match poll {
+                TargetPoll::Ready { session_id, proof } => {
+                    info!(
+                        target_block,
+                        session_id = %session_id,
+                        "Proof ready, submitting inline"
+                    );
+                    Metrics::proof_collection_total(Metrics::COLLECTION_OUTCOME_READY).increment(1);
+                    match self
+                        .submit_inline(target_block, &current, proof, retry_counts, cache)
+                        .await
+                    {
+                        SubmitEffect::Submitted => {
+                            retry_sessions.remove(&target_block);
+                            discard_retry_counts.remove(&target_block);
+                            if let Some(cached) = cache.as_ref() {
+                                *cursor = Some(cached.state);
+                                if cached.state.l2_block_number > current.l2_block_number {
+                                    continue;
+                                }
+                            }
+                            break;
+                        }
+                        SubmitEffect::Restart => {
+                            request_restart(restart_tx, "submit failure").await;
+                            break;
+                        }
+                        SubmitEffect::Redispatch { claimed_l2_output_root } => {
+                            self.dispatch_discard_retry(
+                                target_block,
+                                &current,
+                                claimed_l2_output_root,
+                                retry_counts,
+                                cache,
+                                DiscardRetryState {
+                                    counts: discard_retry_counts,
+                                    sessions: retry_sessions,
+                                },
+                            )
+                            .await;
+                            break;
                         }
                     }
-                    SubmitEffect::Restart => {
-                        request_restart(restart_tx, "submit failure").await;
-                    }
-                    SubmitEffect::Redispatch { claimed_l2_output_root } => {
-                        self.dispatch_discard_retry(
-                            target_block,
-                            &current,
-                            claimed_l2_output_root,
-                            retry_counts,
-                            cache,
-                            DiscardRetryState {
-                                counts: discard_retry_counts,
-                                sessions: retry_sessions,
-                            },
-                        )
-                        .await;
-                    }
                 }
-            }
-            TargetPoll::Pending { session_id, status } => {
-                debug!(
-                    target_block,
-                    session_id = %session_id,
-                    ?status,
-                    "Proof pending, waiting for prover service"
-                );
-            }
-            TargetPoll::NotFound { session_id, claimed_l2_output_root } => {
-                debug!(
-                    target_block,
-                    session_id = %session_id,
-                    claimed_l2_output_root = %claimed_l2_output_root,
-                    "No prover-service session for target, waiting for dispatcher"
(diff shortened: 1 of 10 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -120,130 +141,108 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+    /// Returns the block interval the collector was configured with.
+    pub const fn block_interval(&self) -> u64 {
+        self.block_interval
+    }
 
-        targets
+    /// Returns the rollup-output fetch concurrency the collector was configured with.
+    pub const fn output_fetch_concurrency(&self) -> usize {
+        self.output_fetch_concurrency
     }
```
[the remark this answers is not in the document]
leopoldjoy: Done in e2cee1e6c — both `block_interval()` and `output_fetch_concurrency()` accessors removed, along with the now-unused fields on `ProofCollector` itself (the constructors no longer take them). Confirmed no in-tree callers before deleting.

refcell (dismissed):

crates/proof/proposer/src/pipeline.rs
```diff
@@ -97,72 +100,13 @@ struct CachedRecovery {
...
-        // means the transaction already landed on L1.  Aborting the task
-        // would prevent `handle_submit_result` from recording the
-        // `last_proposed_block` metric and performing proper state cleanup.
-        // The task will finish with `Success` or `GameAlreadyExists`, and
-        // `handle_submit_result` will clear `submitting` and update metrics.
-    }
-}
-
-/// The parallel proving pipeline.
+/// The self-driving proving pipeline.
```
jackchuma: Nit: the term "self-driving" is mentioned multiple times and it doesn't make sense to me

crates/proof/proposer/src/pipeline.rs
```diff
@@ -293,600 +233,420 @@ where
...
-                    || state.submitting == Some(cursor))
-            {
-                last_skipped = Some(cursor);
-                cursor = match cursor.checked_add(self.config.driver.block_interval) {
-                    Some(c) => c,
-                    None => return Ok((plans, output_blocks)),
-                };
-            }
+        Metrics::safe_head().set(safe_head as f64);
+        // Reflect the on-chain proposer tip on every iteration. Without this,
```
jackchuma: Nit: this 5 line comment is not needed

crates/proof/proposer/src/pipeline.rs
```diff
@@ -293,600 +233,420 @@ where
...
-                self.handle_proof_failure(plan.target_block, error, state);
+            TargetPoll::Failed { session_id, error } => {
+                warn!(
+                    target_block,
+                    session_id = %session_id,
+                    error = %error,
+                    "Prover service reported failed session"
+                );
+                Metrics::proof_collection_total(Metrics::COLLECTION_OUTCOME_FAILED).increment(1);
+                self.handle_proof_failure(target_block, error, retry_counts, cache);
```
jackchuma: When a prover-service session reaches `Failed`, this path only increments the local retry counter and then the next iteration polls the same terminal `Failed` session again. The prover service requeues a failed request only when `prove_block_range` is submitted again with the same session id, so transient backend/worker failures will exhaust the pipeline retry budget and drop the cache without ever starting another proof for the target

crates/proof/proposer/src/pipeline.rs
```diff
@@ -293,600 +233,420 @@ where
...
+            Ok(Err(SubmitAction::Discard(error))) => {
+                submit_timer.disarm();
                 Metrics::errors_total(error.metric_label()).increment(1);
+                Metrics::discarded_targets_total().increment(1);
+                // The prover-service session for this target is keyed
+                // deterministically on the canonical output root, so the
+                // next poll would return the same `Ready` proof and
+                // re-discard it. Mark the target as discarded so subsequent
+                // iterations skip polling until the chain advances past it.
+                discarded_targets.insert(target_block);
```
jackchuma: Do not wait for discarded targets to advance themselves

When `SubmitAction::Discard` occurs, e.g. `L1OriginTooOld` or `InvalidSigner`, marking the target as discarded makes later `step()` calls return before polling or dispatching while `recovered.l2_block_number` is still behind that same target. In a single-proposer deployment the chain cannot advance past the target without a successful submission, so one discarded proof permanently halts progress instead of obtaining a fresh proof

jackchuma (commented):
I'm not a huge fan of this large orchestration idea. It took hours for me to wrap my head around it

- We definitely cannot implement this in a way that only handles one proof at a time. If we fall behind we will never catch up
- I still think the original brainstormed idea of 2 parallel async tasks handling proof dispatching + proof collecting is a superior design. Will be much simpler
- I understand the "limiting the blast radius" thoughts backing this PR up, but my view is it's over engineering for a problem that would need manual intervention anyways if we ran into it (i.e. if a parent game were nullified). Also under normal operations (like if the system is current), the dual async idea will only have one proof in flight too

crates/proof/proposer/src/pipeline.rs
```diff
@@ -293,573 +259,638 @@ where
...
-        // Compute targets synchronously while we still hold an immutable view of
-        // `state` so the collector itself can run without borrowing pipeline state
-        // across await points.
-        let targets = self.proof_collector.collectable_targets(recovered, safe_head, |target| {
-            state.proved.contains_key(&target) || state.submitting == Some(target)
-        });
+        cancel: CancellationToken,
+        restart_tx: mpsc::Sender<String>,
+    ) -> Result<()> {
+        let mut cache: Option<CachedRecovery> = None;
```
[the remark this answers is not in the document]
leopoldjoy: Leaving this as-is intentionally: dispatcher and collector cursors/caches represent different progress (requested vs landed), and sharing them would recouple the loops for only an idempotent RPC-load optimization.

leopoldjoy:
Closing in favour of multiple smaller, broken-up PRs, namely https://github.com/base/base/pull/3339, https://github.com/base/base/pull/3340, and https://github.com/base/base/pull/3341

PR #3285, opened 2026-06-06
fix(precompile-storage): enforce EIP-2200 gas-stipend guard in sstore

eric-ships:
[BOP-291](https://linear.app/coinbase/issue/BOP-291/7-native-sstore-path-omits-eip-2200-gas-stipend-guard)

## Motivation

EIP-2200 specifies that an SSTORE must halt with out-of-gas if the frame's remaining gas is at or below the call stipend (2300). This is the reentrancy sentry that Solidity's `.transfer()` relies on: by forwarding only 2300 gas, a contract guarantees the recipient cannot perform state-changing SSTOREs. The native precompile `sstore` path skipped this check, checking only for static-call violations before mutating storage.

The practical exploit requires deliberate setup: a transaction that first warms and dirties a target slot, then makes a direct call into a stateful precompile (e.g., B20 `approve`) with remaining gas at or below 2300. At that point, a warm dirty rewrite costs only ~200 gas, so the write succeeds where the EVM SSTORE opcode would have halted. The attack is not trivial but the preconditions are realistic for allowance-style B20 flows.

The more important reason to fix this is EVM semantic parity. As the B20 ecosystem grows, contracts will be written assuming the 2300-gas invariant holds across all state-changing operations on Base. Breaking that assumption silently is the kind of divergence that produces serious audit findings 6 months post-launch. The fix is a one-line guard.

## What changed

Added an EIP-2200 stipend sentry to `EvmPrecompileStorageProvider::sstore`: returns `OutOfGas` when `gas.remaining() <= gas_params.call_stipend()` before the storage write, mirroring revm's opcode-level guard.

Placing the guard at the dispatch layer (in the `base_precompile!` macro) rather than in `sstore` was considered. That would catch the boundary earlier but would require threading gas state into the macro and would not protect callers that invoke `sstore` through the `PrecompileStorageProvider` trait directly. Placing it in `sstore` is more defensive and matches where revm enforces it at the opcode level.

### DIFF
--- a/crates/common/precompile-storage/src/evm.rs
+++ b/crates/common/precompile-storage/src/evm.rs
@@ -173,6 +173,14 @@ impl PrecompileStorageProvider for EvmPrecompileStorageProvider<'_> {
         if self.is_static {
             return Err(BasePrecompileError::StaticCallViolation);
         }
+        // EIP-2200: if remaining gas is at or below the call stipend (2300), halt with
+        // out-of-gas. This is the reentrancy sentry that Solidity's `.transfer()` relies on:
+        // forwarding only 2300 gas guarantees the recipient cannot perform state-changing
+        // SSTOREs. Without this guard, a warm-dirty rewrite (~200 gas) would succeed where
+        // the EVM SSTORE opcode would have halted, breaking the 2300-gas invariant.
+        if self.gas.remaining() <= self.gas_params.call_stipend() {
+            return Err(BasePrecompileError::OutOfGas);
+        }
         let s = self
             .internals
             .sstore(address, key, value)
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (approved):

PR #3339, opened 2026-06-09
feat(proposer): add collector polling support

leopoldjoy:
## Summary

Adds proposer collector/prover-service support needed for the concurrent proving pipeline split.

This keeps the existing `CollectedProof` / `collect()` API used by the current pipeline, while adding the single-target polling API needed by the final wiring PR.

## Changes

- Add `ProverServiceClientError::is_not_found()` for JSON-RPC NotFound classification.
- Add `TargetPoll` with `Ready`, `Pending`, `NotFound`, `Failed`, and `Unknown` outcomes.
- Add `ProofCollector::poll(target_block)` for deterministic root-derived sessions.
- Add `ProofCollector::poll_session(target_block, session_id)` for retry-specific sessions.
- Add retry-session NotFound handling so missing retry sessions can be re-dispatched.
- Add proof adapter helpers for root-derived and discard retry-specific TEE session ids.
- Update the test prover-service mock to return JSON-RPC NotFound for unknown sessions and support failed-session tests.

## Verification

### DIFF
--- a/crates/proof/proposer/src/proof_adapter.rs
+++ b/crates/proof/proposer/src/proof_adapter.rs
@@ -46,6 +46,33 @@ impl ProofRequesterDispatcher {
         request: PrimitiveProofRequest,
     ) -> Result<DispatchedProof, ProposerError> {
         let request = ProposerProofAdapter::tee_prove_block_range_request(request, self.tee_kind);
+        self.dispatch_prepared(request).await
+    }
+
+    /// Submits a TEE proof request under an explicit session id.
+    ///
+    /// Used for discard retries. The normal proposer session id is keyed only
+    /// by claimed output root so restarts can rediscover in-flight work, but a
+    /// discarded `Succeeded` session cannot be requeued by replaying that same
+    /// id. A retry-specific id lets the proposer obtain a genuinely fresh TEE
+    /// proof for the same output root.
+    pub async fn dispatch_tee_with_session_id(
+        &self,
+        request: PrimitiveProofRequest,
+        session_id: String,
+    ) -> Result<DispatchedProof, ProposerError> {
+        let request = ProposerProofAdapter::tee_prove_block_range_request_with_session_id(
+            request,
+            self.tee_kind,
+            session_id,
+        );
+        self.dispatch_prepared(request).await
+    }
+
+    async fn dispatch_prepared(
+        &self,
+        request: ProveBlockRangeRequest,
+    ) -> Result<DispatchedProof, ProposerError> {
         let response = self
             .requester
             .prove_block_range(request)
@@ -97,12 +124,37 @@ impl ProposerProofAdapter {
         ProofSessionId::derive(Self::SESSION_NAMESPACE, Self::tee_session_label(tee_kind), root)
     }
 
+    /// Derives a TEE proof retry session id for a discarded proof.
+    pub fn tee_discard_retry_session_id(
+        request: &PrimitiveProofRequest,
+        tee_kind: TeeKind,
+        attempt: u32,
+    ) -> String {
+        let label = Self::tee_session_label(tee_kind);
+        let l1_head_number = request.l1_head_number.to_be_bytes();
+        let attempt = attempt.to_be_bytes();
+        ProofSessionId::derive_from_components(
+            Self::SESSION_NAMESPACE,
+            label,
+            &[request.claimed_l2_output_root.as_slice(), &l1_head_number, &attempt],
+        )
+    }
+
     /// Builds a prover-service request for a TEE proposal proof.
     pub fn tee_prove_block_range_request(
         request: PrimitiveProofRequest,
         tee_kind: TeeKind,
     ) -> ProveBlockRangeRequest {
         let session_id = Self::tee_session_id(&request, tee_kind);
+        Self::tee_prove_block_range_request_with_session_id(request, tee_kind, session_id)
+    }
+
+    /// Builds a prover-service request for a TEE proposal proof with a caller-supplied session id.
+    pub const fn tee_prove_block_range_request_with_session_id(
+        request: PrimitiveProofRequest,
+        tee_kind: TeeKind,
+        session_id: String,
+    ) -> ProveBlockRangeRequest {
         ProveBlockRangeRequest {
             proof: ProofRequest {
                 session_id,
--- a/crates/proof/proposer/src/proof_adapter.rs
+++ b/crates/proof/proposer/src/proof_adapter.rs
@@ -307,6 +307,38 @@ mod tests {
     }
 
     #[test]
+    fn tee_discard_retry_session_id_differs_from_root_session() {
+        let request = test_request(B256::repeat_byte(0xaa));
+
+        assert_ne!(
+            ProposerProofAdapter::tee_discard_retry_session_id(&request, TeeKind::AwsNitro, 1),
+            ProposerProofAdapter::tee_session_id(&request, TeeKind::AwsNitro),
+        );
+    }
+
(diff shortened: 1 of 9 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proposer/src/proof_adapter.rs
```diff
@@ -97,12 +124,37 @@ impl ProposerProofAdapter {
...
     pub fn tee_prove_block_range_request(
         request: PrimitiveProofRequest,
         tee_kind: TeeKind,
     ) -> ProveBlockRangeRequest {
         let session_id = Self::tee_session_id(&request, tee_kind);
+        Self::tee_prove_block_range_request_with_session_id(request, tee_kind, session_id)
+    }
+
+    /// Builds a prover-service request for a TEE proposal proof with a caller-supplied session id.
+    pub const fn tee_prove_block_range_request_with_session_id(
```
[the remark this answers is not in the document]
leopoldjoy: Leaving this as-is: this compiles and clippy requires `const fn` here because it only moves the already-allocated `String` into the request; it does not allocate in a const context.

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -246,16 +292,108 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+
+        self.response_to_poll(target_block, session_id, output.output_root, response)
+    }
+
+    /// Polls a specific prover-service session for `target_block`.
+    ///
+    /// Used after a discard retry is dispatched under a retry-specific session
+    /// id. The canonical root is still fetched so a terminal failed session can
+    /// be re-dispatched with a complete proof request.
+    pub async fn poll_session(&self, target_block: u64, session_id: String) -> TargetPoll {
```
mw2000, line 359: not-blocking: 

Could `poll_session()` query `get_proof()` before fetching the canonical output root?

For retry-session polling we already have the exact `session_id`, and the common `Pending` path does not need `claimed_l2_output_root`. As written, a transient rollup RPC failure prevents us from observing that the retry session is still pending, failed, missing, or ready, and it adds a rollup RPC to every pending retry poll.

The root is only needed when returning outcomes that carry `claimed_l2_output_root` (`NotFound` and `Failed`). Maybe `poll_session()` can fetch it lazily after `get_proof()` shows that one of those outcomes is needed, while allowing `Pending` and successful `Ready` to return without the extra rollup call.

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -104,6 +160,14 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
         )
     }
 
+    /// Creates a TEE proof collector for single-target AWS Nitro polling.
+    pub fn target_poller_aws_nitro(
+        proof_requester: Arc<dyn ProofRequesterProvider>,
+        rollup_client: Arc<R>,
+    ) -> Self {
+        Self::new(proof_requester, rollup_client, 0, 1, TeeKind::AwsNitro)
```
mw2000, line 168: `target_poller_aws_nitro()` constructs this type with a zero interval, and `collectable_targets()` advances with `checked_add(self.block_interval)`, so an accidental call can loop forever. 

Returning an empty vec for zero from `collectable_targets` would make this safe.

leopoldjoy: Done: https://github.com/base/base/pull/3339/commits/cecb3903bc98dccc68c32b8a21beecf91e1bb109

crates/proof/proposer/src/proof_collector.rs
leopoldjoy: Thanks! Done: https://github.com/base/base/pull/3339/commits/cecb3903bc98dccc68c32b8a21beecf91e1bb109

mw2000 (dismissed):

crates/proof/prover-service/client/src/error.rs
```diff
@@ -60,6 +63,16 @@ impl ProverServiceClientError {
...
     }
 
+    /// Returns `true` when the prover service responded that the session does not exist.
+    ///
+    /// The proposer's self-driving loop uses this to distinguish "no session yet,
+    /// dispatch needed" from other transient or terminal errors.
+    #[must_use]
+    pub fn is_not_found(&self) -> bool {
+        let Self::RpcTransport(JsonRpcClientError::Call(call)) = self else { return false };
+        call.code() == Self::ERROR_NOT_FOUND
```
jackchuma: Disambiguate session `NotFound` before redispatching. When get_proof finds an existing `Succeeded` request whose stored result is missing, the prover-service server returns the same `-32004` code via `not_found("proof result not available")` as it uses for a missing session. With this helper returning true for every `-32004`, `poll()`/`poll_session()` report `TargetPoll::NotFound`, so callers will redispatch/re-poll the same completed session instead of surfacing a failed/unknown result, which can leave that target stuck. Please distinguish the actual `Proof request not found` case (or use a separate server code) before driving dispatch from it

crates/proof/proposer/src/proof_collector.rs
jackchuma, line 1: This file is becoming way over complicated in my opinion. Why can't it be a single minimal loop that does the following?

1. Derive next expected block (last known + 600)
2. Query get_proof from prover service
3. If proof exists:
      1. Submit onchain
      2. Update last known block
4. If proof does not exist:
      1. Sleep, and try again

crates/proof/prover-service/client/src/error.rs
leopoldjoy: Good catch, done

crates/proof/proposer/src/proof_collector.rs
leopoldjoy: Fair concern — this PR is just the stateless polling adapter, and the later wiring PR (#3341) is where that minimal loop shape becomes visible in the main proposer flow.

jackchuma: Why not set up the minimal loop in this PR? I guess I'm not grasping how a "stateless polling adapter" fits in here

crates/proof/proposer/src/proof_adapter.rs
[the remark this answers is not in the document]
leopoldjoy: This compiles on our toolchain and clippy requires const here; the function moves an already-allocated String and does not allocate in const context.

jackchuma (approved):

PR #3340, opened 2026-06-09
feat(proposer): add proof dispatcher orchestration

leopoldjoy:
## Summary

Adds a dedicated `proof_dispatcher` module for proposer TEE proof request construction, dispatch, and dispatcher-side orchestration.

This keeps request construction and safe-head fan-out reviewable independently from collector behavior and final pipeline wiring.

## Changes

- Add `ProofDispatcherConfig` for static proof request parameters.
- Add `ProofDispatcher<L1, L2, R>` with AWS Nitro construction and canonical output-root access.
- Add root-derived proof dispatch via `dispatch_for`.
- Add discard retry-specific dispatch via `dispatch_discard_retry`.
- Add dispatcher runtime/state types for cursor tracking, retry accounting, and safe-head fan-out.
- Return explicit dispatch outcomes distinguishing accepted dispatch, request-build failure, prover-service dispatch failure, and retry exhaustion.
- Record `build_failed` dispatch outcomes without burning proof retry budget.
- Add focused dispatcher unit tests for root-derived sessions, retry-specific sessions, and fan-out through safe head.

## Verification

### DIFF
--- a/crates/proof/proposer/src/proof_dispatcher.rs
+++ b/crates/proof/proposer/src/proof_dispatcher.rs
@@ -1,0 +1,261 @@
+//! Proof request construction and dispatch helpers for proposer TEE proofs.
+
+use std::sync::Arc;
+
+use alloy_primitives::{Address, B256};
+use base_proof_primitives::ProofRequest;
+use base_proof_rpc::{L1Provider, L2Provider};
+use base_prover_service_client::ProofRequesterProvider;
+use tracing::info;
+
+use crate::{
+    driver::RecoveredState,
+    error::ProposerError,
+    proof_adapter::{DispatchedProof, ProofRequesterDispatcher, ProposerProofAdapter},
+    proof_collector::ProofCollector,
+};
+
+/// Static parameters needed to build proposer proof requests.
+#[derive(Debug, Clone, Copy)]
+pub struct ProofDispatcherConfig {
+    /// Address of the proposer that will submit the proof on-chain.
+    pub proposer_address: Address,
+    /// Number of L2 blocks between intermediate output root checkpoints.
+    pub intermediate_block_interval: u64,
+    /// Expected TEE enclave image hash.
+    pub tee_image_hash: B256,
+}
+
+/// Outcome of attempting to dispatch a proof request.
+#[derive(Debug)]
+pub enum ProofDispatchAttempt {
+    /// The request was accepted by prover-service.
+    Accepted(DispatchedProof),
+    /// The request could not be built from local RPC data.
+    BuildFailed(ProposerError),
+    /// The request reached prover-service but dispatch failed.
+    DispatchFailed(ProposerError),
+}
+
+/// Builds and dispatches proposer TEE proof requests.
+#[derive(Clone)]
+pub struct ProofDispatcher<L1, L2>
+where
+    L1: L1Provider,
+    L2: L2Provider,
+{
+    dispatcher: ProofRequesterDispatcher,
+    l1_client: Arc<L1>,
+    l2_client: Arc<L2>,
+    config: ProofDispatcherConfig,
+}
+
+impl<L1, L2> std::fmt::Debug for ProofDispatcher<L1, L2>
+where
+    L1: L1Provider,
+    L2: L2Provider,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ProofDispatcher")
+            .field("dispatcher", &self.dispatcher)
+            .field("config", &self.config)
+            .finish_non_exhaustive()
+    }
+}
+
+impl<L1, L2> ProofDispatcher<L1, L2>
+where
+    L1: L1Provider + 'static,
+    L2: L2Provider + 'static,
+{
+    /// Creates an AWS Nitro TEE proof dispatcher.
+    pub fn aws_nitro(
+        proof_requester: Arc<dyn ProofRequesterProvider>,
+        l1_client: Arc<L1>,
+        l2_client: Arc<L2>,
+        config: ProofDispatcherConfig,
+    ) -> Self {
+        Self {
+            dispatcher: ProofRequesterDispatcher::aws_nitro(proof_requester),
+            l1_client,
+            l2_client,
+            config,
+        }
+    }
+
+    /// Returns the inner prover-service dispatcher.
+    pub const fn requester_dispatcher(&self) -> &ProofRequesterDispatcher {
+        &self.dispatcher
+    }
+
+    /// Builds a proof request for `target_block` using `recovered` as the agreed parent.
+    pub async fn build_request(
+        &self,
+        target_block: u64,
+        recovered: &RecoveredState,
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proposer/src/proof_dispatcher.rs
```diff
@@ -0,0 +1,539 @@
...
+        recovered: &RecoveredState,
+        claimed_l2_output_root: B256,
+    ) -> ProofDispatchAttempt {
+        let request =
+            match self.build_request(target_block, recovered, claimed_l2_output_root).await {
+                Ok(request) => request,
+                Err(error) => return ProofDispatchAttempt::BuildFailed(error),
+            };
+
+        match self.dispatcher.dispatch_tee(request).await {
```
jackchuma, line 209: Reject mismatched prover session IDs. When prover-service returns a `session_id` that does not equal the deterministic ID for this request, this branch still reports `Accepted`; `tick` then advances `state.cursor`, while `ProofCollector` later polls the deterministic ID derived from the claimed root. In that mismatch scenario the accepted work is never found and the target is not requeued, which is why the existing pipeline path treats mismatched IDs as dispatch failures

crates/proof/proposer/src/proof_dispatcher.rs
```diff
@@ -0,0 +1,539 @@
...
+        let count = self.retry_counts.entry(target).or_insert(0);
+        *count += 1;
+        if *count >= max_retries {
+            error!(
+                target_block = target,
+                attempts = *count,
+                error = %error,
+                "Proof failed after max retries, dropping cached recovery"
+            );
+            self.retry_counts.remove(&target);
```
jackchuma, line 430: Rewind the cursor after retry exhaustion. When a target that was already accepted by `tick` later exhausts proof retries via `handle_proof_failure`, this removes only the retry counter and returns `false`. The caller will drop/recover onchain state, but `tick` keeps any higher `state.cursor` because it only replaces the cursor when recovery advances, so if recovery returns the same/lower block the failed target is skipped permanently instead of being dispatched again

crates/proof/proposer/src/proof_dispatcher.rs
mw2000: +1 on this comment from Jack

crates/proof/proposer/src/proof_dispatcher.rs
leopoldjoy: Done: https://github.com/base/base/pull/3340/changes/12b7af2e8e704648f5e73936364ef03fca646bd6

crates/proof/proposer/src/proof_dispatcher.rs
leopoldjoy: True! Fixed: https://github.com/base/base/pull/3340/changes/12b7af2e8e704648f5e73936364ef03fca646bd6

crates/proof/proposer/src/proof_dispatcher.rs
```diff
@@ -0,0 +1,631 @@
...
+        &self,
+        state: &mut ProofDispatcherState,
+        recovered: RecoveredState,
+        safe_head: u64,
+        runtime: ProofDispatcherRuntimeConfig,
+        cancel: &CancellationToken,
+    ) -> ProofDispatcherTickResult {
+        state.retry_counts.retain(|&target, _| target > recovered.l2_block_number);
+
+        if state.cursor.is_none_or(|cursor| cursor.l2_block_number < recovered.l2_block_number) {
```
jackchuma: Reset dispatcher cursor when recovery rewinds. If the caller refreshes recovery to an earlier block than `state.cursor` (for example after a recovery-cache reset or L1 reorg), this condition leaves the cursor at the old higher dispatched target. The loop then computes the next target from that stale cursor and skips every range between the new recovered block and the old cursor, so no fresh proof requests are sent for the new recovery state. The cursor should be reset when recovery moves backward or otherwise changes identity, not only when it moves forward

jackchuma (approved):

crates/proof/proposer/src/proof_dispatcher.rs
```diff
@@ -0,0 +1,631 @@
...
+    L1: L1Provider,
+    L2: L2Provider,
+    R: RollupProvider,
+{
+    fn clone(&self) -> Self {
+        Self {
+            dispatcher: self.dispatcher.clone(),
+            l1_client: Arc::clone(&self.l1_client),
+            l2_client: Arc::clone(&self.l2_client),
+            rollup_client: Arc::clone(&self.rollup_client),
```
mw2000, line 117: Good call on making them Arc cloneable 😄

crates/proof/proposer/src/proof_dispatcher.rs
leopoldjoy: Fixed: https://github.com/base/base/pull/3340/changes/50f90e0fff6937597d24a8c167c967b90c12a454

crates/proof/proposer/src/proof_dispatcher.rs
```diff
@@ -0,0 +1,670 @@
...
+                ProofDispatchOutcome::Accepted => {
+                    current.l2_block_number = target_block;
+                    current.output_root = claimed_l2_output_root;
+                    state.cursor = Some(current);
+                }
+                ProofDispatchOutcome::RetryExhausted => {
+                    drop_recovery_cache = true;
+                    break;
+                }
+                ProofDispatchOutcome::Skipped => break,
```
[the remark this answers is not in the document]
leopoldjoy: This is intentional. `BuildFailed` means we could not construct / accept the request for the current cursor target, so `tick` must not advance to later targets and create a dispatch gap. The dispatcher fan-out is sequential by target: a later request should only be sent after the previous target is accepted and the cursor advances. Build failures also do not burn retry budget because nothing reached prover-service; the next tick retries from the same cursor.

mw2000 (approved):

PR #3344, opened 2026-06-09
fix(chainspec,activation): close zero-address admin backdoor in Beryl activation registry

eric-ships:
## Motivation

`ActivationRegistryStorage::admin()` maps both `None` (no admin configured) and `Some(Address::ZERO)` (zero address explicitly set) to the same return value. A chain misconfigured with `activation_admin_address == Some(Address::ZERO)` therefore creates a backdoor: any deposit transaction whose recovered sender is `Address::ZERO` matches the admin check and can freely toggle Beryl activation state.

## What changed

Two layers of defense:

**Config-time.** `ChainConfig.activation_admin_address` is now a required `Address` rather than `Option<Address>`. Since every Base chain schedules Beryl, an admin was always required — the `Option` was misleading and allowed `Some(Address::ZERO)` to pass without a type error. Chain spec construction (`try_from_genesis`, `try_from_chainspec`, `try_build`) now also rejects `Address::ZERO` explicitly with a new `ZeroActivationAdminAddress` error, alongside the existing check for a missing admin.

**Runtime.** `set_activated` now rejects any caller whose address is `Address::ZERO` before reaching the admin comparison. This ensures that even if a misconfiguration bypasses the config-time guard, a zero-address caller cannot activate or deactivate features.

## What was considered

The `admin()` helper — which maps `None` to `Address::ZERO` for Solidity null-address conventions — is left unchanged. Once the config-time guard is in place, `Some(Address::ZERO)` can no longer reach production, eliminating the ambiguity at the source. The runtime guard is kept as defense-in-depth: it is cheap and protects against any future code path that might construct a chain spec without going through the validated constructors.

### DIFF
--- a/crates/common/chains/src/config.rs
+++ b/crates/common/chains/src/config.rs
@@ -100,7 +100,7 @@ pub struct ChainConfig {
     /// Activation registry admin address.
     ///
     /// Required and non-zero for all Base chains: every Base chain has Beryl scheduled, and
-    /// Beryl's activation registry precompile needs an admin at genesis. Address::ZERO is
+    /// Beryl's activation registry precompile needs an admin at genesis. `Address::ZERO` is
     /// rejected by chainspec validation.
     pub activation_admin_address: Address,
 
--- a/crates/common/chains/src/config.rs
+++ b/crates/common/chains/src/config.rs
@@ -98,7 +98,11 @@ pub struct ChainConfig {
     /// Unsafe block signer address.
     pub unsafe_block_signer: Option<Address>,
     /// Activation registry admin address.
-    pub activation_admin_address: Option<Address>,
+    ///
+    /// Required and non-zero for all Base chains: every Base chain has Beryl scheduled, and
+    /// Beryl's activation registry precompile needs an admin at genesis. Address::ZERO is
+    /// rejected by chainspec validation.
+    pub activation_admin_address: Address,
 
     // Gas limits
     /// Maximum gas limit for L2 blocks.
@@ -369,7 +373,7 @@ const MAINNET: ChainConfig = ChainConfig {
     protocol_versions_address: address!("8062abc286f5e7d9428a0ccb9abd71e50d93b935"),
 
     unsafe_block_signer: Some(address!("Af6E19BE0F9cE7f8afd49a1824851023A8249e8a")),
-    activation_admin_address: Some(address!("331C9d37BbcebBC9dfAf98FBE3C5B8A39Dd6E771")),
+    activation_admin_address: address!("331C9d37BbcebBC9dfAf98FBE3C5B8A39Dd6E771"),
 
     max_gas_limit: 105_000_000,
     prune_delete_limit: 20_000,
@@ -443,7 +447,7 @@ const SEPOLIA: ChainConfig = ChainConfig {
     protocol_versions_address: address!("79add5713b383daa0a138d3c4780c7a1804a8090"),
 
     unsafe_block_signer: Some(address!("b830b99c95Ea32300039624Cb567d324D4b1D83C")),
-    activation_admin_address: Some(address!("5Be7Dd3678e999D5F7bC508c413db239F7D4Ac59")),
+    activation_admin_address: address!("5Be7Dd3678e999D5F7bC508c413db239F7D4Ac59"),
 
     max_gas_limit: 45_000_000,
     prune_delete_limit: 10_000,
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Sertug17:
The two-layer fix looks correct. One observation on the admin() helper (line 60): it still maps None to Address::ZERO, which is semantically "no admin configured"  distinct from Some(Address::ZERO) which is now rejected at config time. The set_activated path doesn't use admin(), so this PR is safe. However, admin() is a public method and any future caller that passes its return value into an address comparison could silently conflate "not configured" with "zero address explicitly set". A doc comment on admin() noting that Address::ZERO here means "no admin" (not a valid admin address) would prevent that confusion down the line.

Sertug17:
Confirmed the use alloy_primitives::Address; inside the two test function bodies (beryl_chain_config_with_zero_activation_admin_is_rejected and beryl_builder_with_zero_activation_admin_is_rejected) violates the import placement rule. Moving Address to the existing module-level import in mod tests would fix both.

Sertug17:
Confirmed the Beryl fork condition !matches!(hardforks.fork(BaseUpgrade::Beryl), ForkCondition::Never) is evaluated twice, once for the None check and once for the zero check. Extracting it into a let beryl_scheduled = ... local above both guards would eliminate the duplication and ensure the two checks always use the same condition.

Sertug17:
Format check is failing on the latest commit  looks like cargo fmt wasn't run before pushing. Might need a quick rustfmt pass on the refactored function.

Sertug17:
The doc comment on validate_beryl_activation_admin (line 235) still says "include an activation registry admin address" without mentioning the non zero constraint. Since the function now validates both conditions (non None and non zero), updating the doc to reflect that would keep it accurate.

crates/common/precompiles/src/activation/storage.rs
```diff
@@ -119,6 +124,9 @@ impl ActivationRegistryStorage<'_> {
         }
 
         let caller = self.storage.caller();
+        if caller.is_zero() {
```
rayyan224, line 127: Question, how can the caller be zero ?

Sertug17: In normal EVM execution, msg.sender can't be Address::ZERO for a real transaction. This is a defense-in-depth guard for the misconfiguration scenario described in the PR if activation_admin_address is somehow set to Some(Address::ZERO), the zero-address admin would match any zero-address caller and create a backdoor. The config-time validation in validate_beryl_activation_admin is the primary fix, but this runtime guard ensures the precompile boundary stays safe even if that validation is ever bypassed by a future code path.

crates/common/chains/src/config.rs
```diff
@@ -98,7 +98,11 @@ pub struct ChainConfig {
     /// Unsafe block signer address.
     pub unsafe_block_signer: Option<Address>,
     /// Activation registry admin address.
-    pub activation_admin_address: Option<Address>,
+    ///
+    /// Required and non-zero for all Base chains: every Base chain has Beryl scheduled, and
+    /// Beryl's activation registry precompile needs an admin at genesis. Address::ZERO is
+    /// rejected by chainspec validation.
+    pub activation_admin_address: Address,
```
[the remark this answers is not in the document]
eric-ships: All three cases are already addressed in the current branch:

1. boot.rs:249 uses `.map` (not `.and_then`) so it correctly returns `Option<Address>`.
2. boot.rs:438 wraps with `Some(chain_config.activation_admin_address)`.
3. succinct/utils/client/src/boot.rs:88 wraps with `Some(ChainConfig::MAINNET.activation_admin_address)`.

The comment was generated against an earlier commit.

crates/common/precompiles/src/activation/storage.rs
rayyan224: Ok sounds good thanks !

rayyan224 (approved):

PR #3345, opened 2026-06-09
fix(b20-variant): make decimals() return Option<u8> to prevent asset misuse

eric-ships:
[BOP-323](https://linear.app/coinbase/issue/BOP-323)

## Motivation

`B20Variant::decimals()` previously returned `6` for both `Stablecoin` and `Asset` variants, despite asset decimals being a per-token value supplied at creation time via `B20AssetInit::decimals`. The factory correctly used `init.decimals` (not the variant constant) when emitting the `B20Created` event for asset tokens, but the misleading method signature invited future callers to assume `6` is always correct for any B-20 token address.

## What changed

- `variant.rs`: `decimals()` now returns `Option<u8>`. Returns `Some(6)` for `Stablecoin` and `None` for `Asset`, making the absence of a fixed constant explicit at the type level.
- `accounting.rs`: The proc-macro `expand_token` call site changed from `.map_or(0, crate::B20Variant::decimals)` to `.and_then(|v| v.decimals()).unwrap_or(0)` to match the new return type. Semantics are unchanged for all existing variants.
- `storage.rs`: The `init_stablecoin` call site changed to `.expect("stablecoin has fixed 6-decimal precision")` to surface the invariant explicitly.
- `storage.rs` (tests): Added `b20created_asset_event_emits_token_specific_decimals`, which creates an asset token with `decimals: 12` and asserts the emitted `B20Created` event carries `12`, not any variant constant.

An alternative was to rename the method to `stablecoin_decimals()` and keep the `u8` return type. This was rejected because it would still leave a `Self::Asset` match arm that either returns a wrong value or panics, and callers would need to know to only call it on the right variant. Returning `Option<u8>` encodes the constraint in the type system and forces all call sites to handle the `None` case explicitly.

### DIFF
--- a/crates/common/precompile-macros/src/accounting.rs
+++ b/crates/common/precompile-macros/src/accounting.rs
@@ -20,6 +20,13 @@ fn expand_token(input: DeriveInput) -> syn::Result<TokenStream> {
     require_field(&input, "b20")?;
     let has_asset = has_field(&input, "asset");
     let name = input.ident;
+    // This branch is only generated for token structs that do NOT have an `asset` field —
+    // i.e. stablecoin token structs. Asset token structs always have an `asset` field and
+    // take the `has_asset` branch above (which delegates to `AssetAccounting::decimals` and
+    // reads per-token decimals from storage). Therefore `B20Variant::Asset` is structurally
+    // unreachable in the generated code below; `B20Variant::Asset.decimals()` returning `None`
+    // cannot occur here. The only `None` case from `from_address` is an unrecognized (non-B20)
+    // address, which `unwrap_or(0)` handles the same way the original `map_or(0, ...)` did.
     let decimals_impl = if has_asset {
         quote! { crate::AssetAccounting::decimals(self) }
     } else {
--- a/crates/common/precompile-macros/src/accounting.rs
+++ b/crates/common/precompile-macros/src/accounting.rs
@@ -27,7 +27,8 @@ fn expand_token(input: DeriveInput) -> syn::Result<TokenStream> {
             Ok(crate::B20Variant::from_address(
                 ::base_precompile_storage::ContractStorage::address(self),
             )
-            .map_or(0, crate::B20Variant::decimals))
+            .and_then(|v| v.decimals())
+            .unwrap_or(0))
         }
     };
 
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Sertug17:
In `accounting.rs`, the fallback from `map_or(0, decimals)` to `and_then(|v| v.decimals()).unwrap_or(0)` now silently returns `0` for both "variant not found" and "Asset variant" cases. Since an accounting token backed by an `Asset` B20 would have per-token decimals stored in init params rather than the variant, is it possible for this code path to be reached with an `Asset` variant in practice? If so, returning `0` silently might mask a misconfiguration  would `expect()` or `unreachable!()` be more appropriate here to surface the error early?

Sertug17:
Thanks for the clarification the added comment makes the invariant clear. The unreachability of B20Variant::Asset in the stablecoin branch is now explicit, which was exactly the concern. LGTM.

refcell (approved):

PR #3346, opened 2026-06-09
fix(precompiles): reject high-s signatures in EIP-2612 permit

eric-ships:
[BOP-334](https://linear.app/coinbase/issue/BOP-334)

## Motivation

`PermitArgs::recover_signer` accepted both canonical low-s signatures and their high-s malleated twins because `recover_address_from_prehash` internally normalizes s before recovery. This means the pair `(r, s, v)` and `(r, N-s, v XOR 1)` both recover the same address, widening the valid signature set. While the nonce structure prevents replay, accepting high-s violates the `OpenZeppelin` ECDSA convention that wallets and tooling typically assume.

## What changed

- Added a high-s guard in `PermitArgs::recover_signer`: if `sig.normalize_s().is_some()` (i.e. `s > N/2`), the function returns `Err(InvalidSigner)` before attempting recovery. `normalize_s()` is alloy's own half-order check, so no hardcoded constant is needed.
- Added `permit_args_recover_signer_rejects_high_s`: signs with k256 (which always produces canonical low-s), malleates to `s_high = N - s_low` with flipped `v`, and asserts that `recover_signer` returns `Err(InvalidSigner)`.
- Added `permit_args_recover_signer_accepts_canonical_low_s`: confirms the un-malleated signature is still accepted after the guard is in place.

The alternative approach of hardcoding the secp256k1 half-order `0x7fffffffffffffffffffffffffffffff5d576e7357a4501ddfe92f46681b20a0` as a constant and comparing directly was rejected in favor of `normalize_s()`, which delegates the boundary check to alloy's own implementation and keeps the fix anchored to the same source of truth the library uses internally.

### DIFF
--- a/crates/common/precompiles/src/common/ops/permittable.rs
+++ b/crates/common/precompiles/src/common/ops/permittable.rs
@@ -360,6 +360,8 @@ mod tests {
 
     /// Regression test for BOP-334: high-s signatures must be rejected even though they would
     /// recover the same owner address as the canonical low-s counterpart.
+    /// The complementary case — canonical low-s is still accepted — is covered by
+    /// `permit_args_recover_signer_returns_owner`.
     #[test]
     fn permit_args_recover_signer_rejects_high_s() {
         let token = make_token();
@@ -387,19 +389,6 @@ mod tests {
             malleated.recover_signer(signing_hash).unwrap_err(),
             BasePrecompileError::revert(IB20::InvalidSigner { signer: Address::ZERO, owner })
         );
-    }
-
-    /// Companion to `permit_args_recover_signer_rejects_high_s`: confirms the un-malleated
-    /// canonical low-s signature is still accepted after the high-s guard is in place.
-    #[test]
-    fn permit_args_recover_signer_accepts_canonical_low_s() {
-        let token = make_token();
-        let owner = owner_address();
-        let args = signed_permit_args(&token, owner, SPENDER, U256::from(100u64), U256::MAX);
-        let domain_sep = domain_separator_for_token(&token, CHAIN_ID);
-        let signing_hash = args.signing_hash(domain_sep, U256::ZERO);
-
-        assert_eq!(args.recover_signer(signing_hash).unwrap(), owner);
     }
 
     // ---- Permittable ----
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/common/ops/permittable.rs
```diff
@@ -88,6 +92,15 @@ impl PermitArgs {
...
         let sig =
             alloy_primitives::Signature::from_scalars_and_parity(self.r, self.s, odd_y_parity);
+
+        // Reject high-s signatures: normalize_s() returns Some iff s > N/2.
+        if sig.normalize_s().is_some() {
+            return Err(BasePrecompileError::revert(IB20::InvalidSigner {
+                signer: Address::ZERO,
+                owner: self.owner,
+            }));
+        }
```
refcell, line 102: I'm not sure we actually wanted this. @roger-bai-coinbase suggested it should actually behave this way.

refcell (commented):
Need @roger-bai-coinbase to take a look

roger-bai-coinbase:
I'm fine with including the change, but base-std specifically does not do this check as ecrecover does not do this: https://github.com/base/base-std/blob/main/test/lib/mocks/MockB20.sol#L566

There is no attack vector as we use nonces

crates/common/precompiles/src/common/ops/permittable.rs
eric-ships: Agree - closing PR

PR #3348, opened 2026-06-09
fix(ci): pin contracts commit for devnet-setup build

jowparks:
Resolve the HEAD of `base/contracts#l3-changes` at build time and pass it as `CONTRACTS_COMMIT` build-arg to the devnet-setup Docker build.

This ensures the devnet-setup image always uses the correct contracts commit from the l3-changes branch.


robriks (approved):
nice work, auto fetching the latest commit for deploys is very convenient !

PR #3351, opened 2026-06-09
fix(chainspec,precompiles): store activation registry admin in genesis state

eric-ships:
## Motivation

The activation registry admin address was carried as chain spec metadata outside of consensus state. Two nodes sharing the same genesis hash could hold different admin values if configured differently, causing them to disagree on who can call `activate`/`deactivate`. The same issue affected fault proofs: `BootInfo` loaded the admin from a static binary lookup not committed to the preimage oracle, so native and proof execution could diverge on authorization.

The fix: write the admin into the activation registry genesis alloc so any two nodes with the same genesis state root are guaranteed to agree on the admin without out-of-band configuration.

## What changed

- `ActivationRegistryStorage` gains an `admin: Address` storage field. `set_activated` reads the admin from storage via `self.admin.read()` instead of accepting it as a parameter.
- `TryFrom<&ChainConfig>` writes the admin address into the genesis alloc when Beryl is scheduled. The genesis hash is recomputed with `seal_slow` when an admin is injected.
- `activation_admin_address` is removed from the runtime call chain: `dispatch()`, `inner()`, `#[precompile(args(...))]`, `ActivationRegistry::install()`, `BasePrecompiles`, `BaseEvmFactory`, `ZkvmBaseEvmFactory`, and `BaseZkvmPrecompiles`. The field is kept on `BaseChainSpec`/`BaseChainSpecBuilder` for genesis construction only.
- **Breaking (Zeronet only):** The Zeronet genesis hash changes. A devnet reset is required. Mainnet and Sepolia are unaffected.

## What was considered

**Partial fix via `PerChainConfig::marshal_binary()`:** Including the admin in the proof config hash would fix the fault-proof side without touching genesis state. Rejected because it does not close the peer-to-peer consensus split and does not prevent authorization divergence between peers.

**base-std slot coordination:** The `admin` slot must match what base-std independently defines. The slot assignment is deterministic from field declaration order in the `#[contract]` macro: `admin` declared second after `features` produces `NAMESPACE_ROOT+1`. A code comment points to the constant for future base-std alignment.

### DIFF
--- a/crates/common/evm/src/factory.rs
+++ b/crates/common/evm/src/factory.rs
@@ -1,5 +1,4 @@
 use alloy_evm::{Database, EvmEnv, EvmFactory, precompiles::PrecompilesMap};
-use alloy_primitives::Address;
 use revm::{
     Context, Inspector,
     context::{BlockEnv, TxEnv},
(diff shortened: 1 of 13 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

refcell (commented):
I'm not sure I like this change... This is writing to storage right? This is a large change that I don't think buys us much.

rayyan224:
I don't think we need this let's close

PR #3353, opened 2026-06-09
feat: add basectl doctor diagnostics

Pulsator01:
## Summary
- Add `basectl doctor` as a read-only diagnostic command for node operators.
- Compose existing block, sync-status, and p2p helpers with new doctor checks for network/config health.
- Add graceful degradation for restricted RPC methods, concurrent diagnostic fetches, and readable pretty/JSON output.
## Test Plan
- [x] Unit tests pass: `cargo test -p basectl-cli`
- [x] Binary tests pass: `cargo test -p basectl`
- [x] Clippy passes: `cargo clippy -p basectl-cli -p basectl --all-targets -- -D warnings`
- [x] Formatting passes: `cargo +nightly fmt --all -- --check`
- [x] Manual testing completed against Sepolia EL/CL RPC endpoints
- [x] Edge cases considered: missing CL RPC, restricted p2p methods, missing `reth.toml`, wrong chain ID, unavailable peer counts

### DIFF
--- a/crates/infra/basectl/src/rpc/el.rs
+++ b/crates/infra/basectl/src/rpc/el.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 use alloy_consensus::{Transaction, transaction::SignerRecoverable};
 use alloy_eips::{
@@ -7,7 +7,9 @@ use alloy_eips::{
 };
 use alloy_primitives::{Address, B256, Bytes};
 use alloy_provider::{Network, Provider, ProviderBuilder, network::TransactionResponse};
+use alloy_rpc_client::RpcClient;
 use alloy_rpc_types_eth::BlockNumberOrTag;
+use alloy_transport_http::Http;
 use anyhow::{Context, Result, anyhow};
 use base_common_consensus::BaseTxEnvelope;
 use base_common_network::Base;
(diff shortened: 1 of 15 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Sertug17:
The threshold validation cleanly addresses the unreachable warn tier issue validate_thresholds rejecting warn >= fail for both head lag and safe recency pairs, with tests covering the inverted and equal cases. One small note: the guard lives in the binary's run() rather than in DoctorThresholds construction, so the invariant holds for the CLI path but a library caller constructing DoctorThresholds directly could still set warn >= fail and silently lose the Warn tier. If DoctorThresholds is part of the public API surface, enforcing warn < fail at its constructor would make the invariant total rather than CLI only. Either way the CLI fix is correct.

Sertug17:
Confirming the missing module doc: the library-side crates/infra/basectl/src/doctor.rs starts directly with use std::{...} while the sibling modules (rpc/p2p.rs, rpc/pods.rs, output/json.rs, etc.) and even the binary side bin/basectl/src/doctor.rs all open with a //! doc comment. Adding something like //! Diagnostic checks and report types for `basectl doctor`. would keep the crate's module-doc convention consistent.

crates/infra/basectl/src/rpc/el.rs
```diff
@@ -45,6 +47,36 @@ pub async fn fetch_block(
...
+/// Fetches the latest L2 block number via `eth_blockNumber` with an explicit request timeout.
+pub async fn fetch_l2_block_number(rpc: &Url) -> Result<u64> {
+    connect_l2_provider(rpc)
+        .await?
+        .get_block_number()
+        .await
+        .with_context(|| format!("fetching eth_blockNumber from {rpc}"))
+}
+
+async fn connect_l2_provider(rpc: &Url) -> Result<impl Provider<Base>> {
```
wlawt, line 68: can we check if there's existing helper functions like these across the codebase? and, if it we could re-use those cleanly

Sertug17: config.rs:332-334 (MonitoringConfig::detect_name_from_rpc) has a similar pattern ProviderBuilder::new().connect() followed by get_chain_id()  but without the explicit 10s request timeout or Base network typing that connect_l2_provider adds. The timeout is important for the doctor use case where a hung RPC should not stall the entire diagnostic run. One option would be to promote connect_l2_provider into a shared helper (e.g. in rpc/mod.rs) and have both config.rs and doctor reuse it, so the timeout and network typing are consistent across the crate.

wlawt (approved):

PR #3360, opened 2026-06-09
fix(b20-asset): propagate system errors through announce inner calls unchanged

eric-ships:
[BOP-327](https://linear.app/coinbase/issue/BOP-327)

## Motivation

`B20AssetToken::announce` dispatches each `internalCalls` entry through `inner_with_privilege`. The error handler previously used `map_err(|_| ...)`, which discarded the original error and unconditionally wrapped every failure, including system errors (`OutOfGas`, `Fatal`, `Panic`, `SlotOverflow`), as `InternalCallFailed`. System errors represent EVM-level conditions that must abort the entire call frame; substituting a user-level revert breaks that invariant and can mask critical failures from callers.

## What changed

- In the `announce` inner-call dispatch loop in `dispatch.rs`, replaced the wildcard `map_err(|_| ...)` with an `is_system_error()` guard: system errors propagate unchanged, non-system failures are still wrapped as `InternalCallFailed`.
- Added two regression tests:
  - `announce_inner_system_error_propagates_unchanged`: triggers an arithmetic overflow inside `toScaledBalance` (a `Panic(UnderOverflow)` system error) and asserts it surfaces directly rather than as `InternalCallFailed`.
  - `announce_inner_ordinary_revert_wraps_as_internal_call_failed`: triggers an access-control revert from a `mint` call without `MINT_ROLE` and asserts it is still wrapped as `InternalCallFailed`.

## What was considered

`B20FactoryStorage::map_init_call_error` already handles this correctly: it passes system errors through verbatim and wraps only non-system failures. This fix adopts the same `is_system_error()` guard. The factory helper also preserves non-empty `Revert(bytes)` intact rather than wrapping them, but `announce` intentionally keeps the `InternalCallFailed` wrapping for ordinary reverts, since the call field already captures the offending calldata for debugging.

### DIFF
--- a/crates/common/precompiles/src/b20_asset/dispatch.rs
+++ b/crates/common/precompiles/src/b20_asset/dispatch.rs
@@ -448,8 +448,14 @@ impl<S: AssetAccounting, P: Policy> B20AssetToken<S, P> {
             if call_bytes[..4] == IB20Asset::announceCall::SELECTOR {
                 return Err(BasePrecompileError::revert(IB20Asset::AnnouncementInProgress {}));
             }
-            self.inner_with_privilege(ctx, call_bytes, privileged).map_err(|_| {
-                BasePrecompileError::revert(IB20Asset::InternalCallFailed { call: call.clone() })
+            self.inner_with_privilege(ctx, call_bytes, privileged).map_err(|err| {
+                if err.is_system_error() {
+                    err
+                } else {
+                    BasePrecompileError::revert(IB20Asset::InternalCallFailed {
+                        call: call.clone(),
+                    })
+                }
             })?;
         }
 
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

rayyan224:
Don't we need to do this for other internal calls also ? I think factory does some ?

rayyan224 (approved):

eric-ships:
> Don't we need to do this for other internal calls also ? I think factory does some ?

Will follow up on this

eric-ships:
The factory already handles this correctly via `map_init_call_error` (b20_factory/storage.rs), which has the `is_system_error()` guard and also passes non-empty reverts through unchanged. The only call site with the blanket `map_err(|_| ...)` pattern was the `announce` loop in `b20_asset/dispatch.rs`, which this PR fixes.

PR #3362, opened 2026-06-09
fix(b20-factory): enforce nonpayable guard on all factory selectors

eric-ships:
## Motivation

The B20 factory declares `createB20` and its view functions as nonpayable, but the native precompile did not enforce this. Any call with ETH attached would succeed and the ETH would be permanently locked at the factory address with no way to recover it.

## What changed

The factory now rejects any call that arrives with ETH attached. If a caller sends value to `createB20`, `getB20Address`, `isB20`, or `isB20Initialized`, the call reverts immediately with a `NonPayable()` error before any other logic runs — matching how Solidity handles nonpayable functions.

The `NonPayable()` error is also now part of the factory's published ABI, so tooling (block explorers, trace decoders, front-ends) can surface a meaningful name instead of a raw revert byte string.

## What was considered

A shared hook in the `base_precompile!` macro could enforce this for every precompile at once. That approach was deferred — it would require auditing every precompile to confirm it's nonpayable and is not currently deployed with live funds. The targeted fix is lower risk.

### DIFF
--- a/crates/common/precompiles/src/b20_factory/abi.rs
+++ b/crates/common/precompiles/src/b20_factory/abi.rs
@@ -32,6 +32,9 @@ sol! {
 
         // ── Errors ───────────────────────────────────────────────────────────
 
+        /// ETH was sent to a nonpayable factory function.
+        error NonPayable();
+
         /// A token already exists at the address derived from `(variant, msg.sender, salt)`.
         error TokenAlreadyExists(address token);
 
--- a/crates/common/precompiles/src/b20_factory/dispatch.rs
+++ b/crates/common/precompiles/src/b20_factory/dispatch.rs
@@ -2,7 +2,7 @@
 
 use alloy_primitives::Bytes;
 use alloy_sol_types::SolCall;
-use base_precompile_storage::StorageCtx;
+use base_precompile_storage::{BasePrecompileError, StorageCtx};
 use revm::precompile::PrecompileResult;
 
 use crate::{
@@ -26,6 +26,12 @@ impl<'a> B20FactoryStorage<'a> {
     where
         O: PrecompileCallObserver,
     {
+        // All factory selectors are nonpayable: reject any call that attaches ETH.
+        // The guard fires before calldata-cost deduction so value-bearing calls pay
+        // zero gas before reverting, matching Solidity's nonpayable semantics.
+        if !ctx.call_value().is_zero() {
+            return ctx.error_result(BasePrecompileError::revert(IB20Factory::NonPayable {}));
+        }
         let mut recorder =
             BerylCallRecorder::start(observer.clone(), BerylMetricLabels::factory_call(calldata));
         if let Err(error) = recorder.deduct_calldata_gas(ctx, calldata) {
--- a/crates/common/precompile-storage/src/evm.rs
+++ b/crates/common/precompile-storage/src/evm.rs
@@ -31,6 +31,7 @@ use crate::{
 pub struct EvmPrecompileStorageProvider<'a> {
     internals: alloy_evm::EvmInternals<'a>,
     caller: Address,
+    call_value: U256,
     gas: Gas,
     gas_params: GasParams,
     is_static: bool,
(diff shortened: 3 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/b20_factory/abi.rs
```diff
@@ -32,6 +32,9 @@ sol! {
 
         // ── Errors ───────────────────────────────────────────────────────────
 
+        /// ETH was sent to a nonpayable factory function.
+        error NonPayable();
```
rayyan224, line 36: Do we need this in base std ?

crates/common/precompiles/src/b20_factory/dispatch.rs
```diff
@@ -26,6 +26,12 @@ impl<'a> B20FactoryStorage<'a> {
     where
         O: PrecompileCallObserver,
     {
+        // All factory selectors are nonpayable: reject any call that attaches ETH.
+        // The guard fires before calldata-cost deduction so value-bearing calls pay
+        // zero gas before reverting, matching Solidity's nonpayable semantics.
+        if !ctx.call_value().is_zero() {
+            return ctx.error_result(BasePrecompileError::revert(IB20Factory::NonPayable {}));
```
rayyan224, line 33: Would there not be other functions that are not payable. In b20 and policy registry. Maybe we can add to the solidity Abi an extension for payable and then check it in the dispatch Wdyt ?

rayyan224 (approved):

cb-heimdall:
Review Error for rayyan224 @ 2026-06-10 02:57:19 UTC

rayyan224 (commented):
Overall lgtm

crates/common/precompiles/src/b20_factory/abi.rs
eric-ships: Yes -- any client that calls a factory function with ETH attached will get a revert with this error selector. If `NonPayable` isn't in base-std, callers won't be able to decode the revert reason. It belongs in the factory's public interface spec.

crates/common/precompiles/src/b20_factory/dispatch.rs
eric-ships: You're right -- b20_asset, b20_stablecoin, and policy_registry all have nonpayable functions too and currently have no guard. The Solidity ABI payable-annotation approach would require decoding the call first and then checking per-function, which is more complex.

Since ALL functions in each precompile are nonpayable (none of them should ever accept ETH), the simpler fix is to add the same dispatch-level `call_value().is_zero()` guard at the top of each precompile's dispatcher -- same pattern as this PR. That catches value-bearing calls before any gas is charged without needing per-function logic. Happy to add those in a follow-up PR if you'd prefer to keep this one scoped to the factory.

PR #3369, opened 2026-06-09
fix(b20-factory): meter keccak gas in factory address derivation dispatch

eric-ships:
## Motivation

When the factory computes the deterministic address for a B20 token, it hashes `(caller, salt)` using keccak256. This hash happened outside the EVM gas meter, so callers could invoke `getB20Address` or trigger address derivation via `createB20` at effectively zero gas cost for the cryptographic work — a metering gap compared to normal EVM keccak operations.

## What changed

`getB20Address` and `createB20` now charge keccak256 gas before deriving the token address. For a recognized token variant, 42 gas is deducted (30 base + 12 for 64 bytes of input). For an unrecognized variant, no hash occurs so no charge is applied.

The underlying `compute_address` helper is unchanged — it stays pure so tests and offline tooling can use it without an EVM context. The gas charge happens at the dispatch level, one layer above.

## What was considered

Reusing the hash from the gas-charging step to avoid double-computing it would require inlining the address assembly logic in dispatch, coupling two pieces of code that should stay separate. The redundant CPU hash costs nothing observable at scale.

### DIFF
--- a/crates/common/precompile-storage/src/hashmap.rs
+++ b/crates/common/precompile-storage/src/hashmap.rs
@@ -358,7 +358,7 @@ impl HashMapStorageProvider {
         self.counter_keccak256
     }
 
-    /// Resets the SLOAD/SSTORE counters (test-utils only).
+    /// Resets the SLOAD/SSTORE/keccak256 counters (test-utils only).
     pub const fn reset_counters(&mut self) {
         self.counter_sload = 0;
         self.counter_sstore = 0;
--- a/crates/common/precompile-storage/src/hashmap.rs
+++ b/crates/common/precompile-storage/src/hashmap.rs
@@ -353,7 +353,7 @@ impl HashMapStorageProvider {
         self.gas_deducted
     }
 
-    /// Returns the number of times [`PrecompileStorageProvider::keccak256`] was called.
+    /// Returns the number of times [`PrecompileStorageProvider::metered_keccak256`] was called.
     pub const fn counter_keccak256(&self) -> u64 {
         self.counter_keccak256
     }
--- a/crates/common/precompile-storage/src/hashmap.rs
+++ b/crates/common/precompile-storage/src/hashmap.rs
@@ -185,7 +185,7 @@ impl PrecompileStorageProvider for HashMapStorageProvider {
         Ok(())
     }
 
-    fn keccak256(
+    fn metered_keccak256(
         &mut self,
         data: &[u8],
     ) -> core::result::Result<alloy_primitives::B256, BasePrecompileError> {
--- a/crates/common/precompile-storage/src/hashmap.rs
+++ b/crates/common/precompile-storage/src/hashmap.rs
@@ -29,6 +29,7 @@ pub struct HashMapStorageProvider {
     counter_sload: u64,
     counter_sstore: u64,
     gas_deducted: u64,
+    counter_keccak256: u64,
     snapshots: Vec<Snapshot>,
     gas_params: GasParams,
     state_gas_used: u64,
@@ -69,6 +70,7 @@ impl HashMapStorageProvider {
             counter_sload: 0,
             counter_sstore: 0,
             gas_deducted: 0,
+            counter_keccak256: 0,
             gas_params: GasParams::default(),
             state_gas_used: 0,
             gas_refunded: 0,
@@ -181,6 +183,14 @@ impl PrecompileStorageProvider for HashMapStorageProvider {
     fn deduct_gas(&mut self, gas: u64) -> Result<(), BasePrecompileError> {
         self.gas_deducted = self.gas_deducted.saturating_add(gas);
         Ok(())
+    }
+
+    fn keccak256(
+        &mut self,
+        data: &[u8],
+    ) -> core::result::Result<alloy_primitives::B256, BasePrecompileError> {
+        self.counter_keccak256 += 1;
+        Ok(alloy_primitives::keccak256(data))
     }
 
     fn deduct_state_gas(&mut self, gas: u64) -> Result<(), BasePrecompileError> {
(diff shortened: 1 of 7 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/b20_factory/dispatch.rs
```diff
@@ -47,6 +47,12 @@ impl<'a> B20FactoryStorage<'a> {
             IB20Factory::IB20FactoryCalls::createB20(call) => {
                 let caller = ctx.caller();
                 let variant = B20Variant::from_abi(call.variant);
+                // Charge keccak gas for address derivation only when the variant is valid.
+                // Invalid variants are rejected in create_b20_with_observer before
+                // compute_address runs, so no keccak hash occurs on the revert path.
+                if variant.is_some() {
+                    ctx.keccak256(&(caller, call.salt).abi_encode())?;
```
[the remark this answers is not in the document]
eric-ships: Fixed for `getB20Address`: added `compute_address_from_hash` to `B20Variant` and now capture the `B256` returned by `ctx.keccak256` and pass it directly, eliminating the second encode and hash.

For `createB20` the hash is buried inside `create_b20_with_observer`, which would need a bigger refactor to thread the pre-computed address through. The existing comment already acknowledges that tradeoff, so left it as is.

crates/common/precompile-storage/src/hashmap.rs
```diff
@@ -26,6 +26,7 @@ pub struct HashMapStorageProvider {
     is_static: bool,
     counter_sload: u64,
     counter_sstore: u64,
+    counter_keccak256: u64,
```
rayyan224, line 32: @eric-ships hasmap.rs, is in memory for testing we need to add this to evm.rs

crates/common/precompiles/src/b20_factory/variant.rs
```diff
@@ -128,7 +128,14 @@ impl B20Variant {
     /// Returns the address and the 9-byte hash tail embedded in the address.
     pub fn compute_address(self, creator: Address, salt: B256) -> (Address, [u8; 9]) {
         let hash = keccak256((creator, salt).abi_encode());
```
rayyan224, line 130: I think this function should be removed right, since it dosent' use context kecak version ?

crates/common/precompiles/src/b20_factory/storage.rs
```diff
@@ -1339,4 +1339,77 @@ mod tests {
             "B20Created.decimals must equal init.decimals, not any variant constant"
         );
     }
+
+    /// Regression test: `getB20Address` and `createB20` must meter the
```
rayyan224: Sorry for the nit can we remove these comments ?

crates/common/precompiles/src/b20_factory/variant.rs
rayyan224: Or this should take in the context and use the gas metered keccak

crates/common/precompiles/src/b20_factory/storage.rs
rayyan224: Not really a regression if we are not live xD

crates/common/precompile-storage/src/hashmap.rs
```diff
@@ -174,6 +176,14 @@ impl PrecompileStorageProvider for HashMapStorageProvider {
         Ok(())
     }
 
+    fn keccak256(
```
rayyan224: can we rename to metered_keccak256()

crates/common/precompile-storage/src/hashmap.rs
eric-ships: These counters are test-only observability hooks on the in-memory mock; the production path verifies keccak via gas deduction, not call counts.

crates/common/precompiles/src/b20_factory/variant.rs
eric-ships: No need - the function is intentionally ctx-free for offline/test use.

`compute_address` is used in 15+ call sites across tests, benchmarks (`benches/base_precompiles.rs`), the succinct client, and load tests, all of which have no `StorageCtx` available and need offline address derivation. Adding a `ctx` parameter would break all those.

The on-chain dispatch path is already correctly metered: both `getB20Address` and `createB20` call `ctx.metered_keccak256()` and pass the returned hash directly to `compute_address_from_hash`, so `compute_address` is never reached from dispatch.

crates/common/precompile-storage/src/hashmap.rs
eric-ships: fixed [f9f33b9](https://github.com/base/base/pull/3369/commits/f9f33b96b25ce60d032c5365677019695d91c066)

crates/common/precompiles/src/b20_factory/variant.rs
rayyan224: Hmm ok if it's for test use it should be in fixtures or some type of test utils ? It's a nit maybe we can remove latter

rayyan224 (approved):

PR #3381, opened 2026-06-10
fix(precompiles): enforce nonpayable guard on b20 token and policy registry dispatchers

eric-ships:
[BOP-324](https://linear.app/coinbase/issue/BOP-324)

## Motivation

PR #3362 added a nonpayable guard to the B20 factory dispatcher but left the three remaining precompile dispatchers (B20 asset tokens, B20 stablecoin tokens, policy registry) without the same protection. All selectors on those precompiles are also nonpayable, so ETH-bearing calls should be rejected consistently before any gas is charged.

## What changed

- `b20_asset/dispatch.rs`: added `call_value().is_zero()` guard at the top of `dispatch_with_observer`, before `BerylCallRecorder::start`.
- `b20_stablecoin/dispatch.rs`: same guard.
- `policy/dispatch.rs`: same guard.
- `common/abi.rs` (`IB20`): added `error NonPayable()` to the shared B20 interface so both token variants expose the error by name.
- `policy/abi.rs` (`IPolicyRegistry`): added `error NonPayable()`.

The guard fires before calldata-cost deduction, matching Solidity nonpayable semantics and the behavior introduced in #3362.

Putting `NonPayable` in `IB20Asset` and `IB20Stablecoin` separately was considered, but both dispatchers already import the shared `IB20` interface, so adding it there avoids duplication. Policy registry has its own independent interface so it gets its own error declaration.

### DIFF
--- a/crates/common/precompiles/src/b20_asset/dispatch.rs
+++ b/crates/common/precompiles/src/b20_asset/dispatch.rs
@@ -36,6 +36,9 @@ impl<S: AssetAccounting, P: Policy> B20AssetToken<S, P> {
     where
         O: PrecompileCallObserver,
     {
+        if !ctx.call_value().is_zero() {
+            return ctx.error_result(BasePrecompileError::revert(IB20::NonPayable {}));
+        }
         let mut recorder =
             BerylCallRecorder::start(observer.clone(), BerylMetricLabels::b20_asset_call(calldata));
         if let Err(error) = recorder.deduct_calldata_gas(ctx, calldata) {
--- a/crates/common/precompiles/src/b20_asset/dispatch.rs
+++ b/crates/common/precompiles/src/b20_asset/dispatch.rs
@@ -36,11 +36,12 @@ impl<S: AssetAccounting, P: Policy> B20AssetToken<S, P> {
     where
         O: PrecompileCallObserver,
     {
-        if !ctx.call_value().is_zero() {
-            return ctx.error_result(BasePrecompileError::revert(IB20::NonPayable {}));
-        }
         let mut recorder =
             BerylCallRecorder::start(observer.clone(), BerylMetricLabels::b20_asset_call(calldata));
+        if !ctx.call_value().is_zero() {
+            return recorder
+                .record_base_error_result(ctx, BasePrecompileError::revert(IB20::NonPayable {}));
+        }
         if let Err(error) = recorder.deduct_calldata_gas(ctx, calldata) {
             return recorder.record_base_error_result(ctx, error);
         }
@@ -489,8 +490,7 @@ mod tests {
         ActivationFeature, ActivationRegistryStorage, AssetAccounting, B20AssetStorage,
         B20AssetToken, B20TokenRole, BerylErrorKind, IB20, IB20Asset, InMemoryPolicy,
         InMemoryTokenAccounting, NoopPrecompileCallObserver, PrecompileCallMetric,
-        PrecompileCallObserver, PrecompileCallOutcome, PrecompileCallStatus, Token,
-        TokenAccounting,
+        PrecompileCallObserver, PrecompileCallOutcome, PrecompileCallStatus, Token, TokenAccounting,
     };
 
     type TestAssetToken = B20AssetToken<InMemoryTokenAccounting, InMemoryPolicy>;
--- a/crates/common/precompiles/src/b20_asset/dispatch.rs
+++ b/crates/common/precompiles/src/b20_asset/dispatch.rs
@@ -480,7 +480,7 @@ mod tests {
     use std::sync::{Arc, Mutex};
 
     use alloy_primitives::{Address, Bytes, U256};
-    use alloy_sol_types::{SolCall, SolEvent};
+    use alloy_sol_types::{SolCall, SolError, SolEvent};
     use base_precompile_storage::{
         BasePrecompileError, HashMapStorageProvider, Result, StorageCtx,
     };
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/b20_factory/dispatch.rs
rayyan224: We know all of this let's remove the comments

crates/common/precompile-storage/src/hashmap.rs
rayyan224, line 335: let's remove also

Sertug17:
The nonpayable guard placement is consistent across all three dispatchers call_value().is_zero() fires before BerylCallRecorder::start and before calldata gas deduction, so ETH-bearing calls revert at zero gas cost matching Solidity nonpayable semantics. One minor observation: because the guard returns via ctx.error_result() before the recorder is initialized, nonpayable reverts will not appear in BerylCallRecorder metrics. This is likely intentional (no gas was consumed so there is nothing to meter), but worth confirming the observability team does not need visibility into rejected payable calls for monitoring or alerting purposes.

crates/common/precompiles/src/b20_asset/dispatch.rs
```diff
@@ -36,6 +36,9 @@ impl<S: AssetAccounting, P: Policy> B20AssetToken<S, P> {
     where
         O: PrecompileCallObserver,
     {
+        if !ctx.call_value().is_zero() {
```
rayyan224: What i'm getting afraid here, is that we are making it blatantly throw on payable I guess it's fine, for now, since no functions are payable

rayyan224:
https://github.com/base/base/pull/3381#issuecomment-4666690782 This is a good point I think we should have

Sertug17:
Right, the blanket guard at the dispatcher entry is correct today because every selector on these precompiles is nonpayable. If a payable selector is ever introduced, the guard would need to move below selector decoding so it can be applied per function rather than blanket same pattern Solidity uses when mixing payable and nonpayable functions in a single contract. For now the dispatch level placement avoids per-selector branching and keeps the revert at zero gas cost.

crates/common/precompiles/src/policy/dispatch.rs
```diff
@@ -609,4 +615,26 @@ mod tests {
...
+        let out = StorageCtx::enter(&mut storage, |ctx| {
+            PolicyRegistryStorage::new(ctx).dispatch(ctx, &calldata)
+        })
+        .expect("dispatch must not fatally error");
+
+        assert!(out.is_revert());
+        assert_eq!(
+            out.bytes,
+            alloy_primitives::Bytes::from(
+                alloy_sol_types::SolError::abi_encode(&IPolicyRegistry::NonPayable {})
```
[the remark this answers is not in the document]
Sertug17: Confirming the consistency nit: the asset, stablecoin, and factory nonpayable tests all use the concise .abi_encode() pattern while this policy test uses the fully qualified SolError::abi_encode(&...) form. Aligning to the shorter form keeps the test style uniform across all four dispatchers.

rayyan224 (approved):

cb-heimdall:
Review Error for rayyan224 @ 2026-06-10 05:56:39 UTC

crates/common/precompile-storage/src/hashmap.rs
eric-ships: Removed.

crates/common/precompiles/src/b20_factory/dispatch.rs
eric-ships: Removed.

crates/common/precompiles/src/policy/dispatch.rs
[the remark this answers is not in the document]
eric-ships: Fixed -- switched to `Bytes::from(IPolicyRegistry::NonPayable {}.abi_encode())` to match the other dispatchers.

crates/common/precompiles/src/b20_asset/dispatch.rs
eric-ships: Correct -- since none of the token functions are payable in any variant, the guard is a correctness guarantee rather than a gating mechanism. Agreed it's fine.

rayyan224 (approved):

PR #3386, opened 2026-06-10
docs(precompile-storage): document why load_account is sufficient in with_account_info

eric-ships:
## Motivation

`with_account_info` was calling `load_account`, which leaves `info.code` unpopulated for accounts with bytecode. An investigation into whether this matters for any caller determined it does not: every call site only reads `code_hash` (via `is_empty_code_hash()` or directly) — never `info.code`. This PR adds a comment making that invariant explicit so future authors know `load_account_code` is not needed here.

## What changed

Added a comment to `with_account_info` documenting that `load_account` is intentionally used rather than `load_account_code`, and why: all callers only need `code_hash`, which is always eagerly populated, so fetching bytecode from the database would be unnecessary overhead.

No behavioral change.

### DIFF
--- a/crates/common/precompile-storage/src/evm.rs
+++ b/crates/common/precompile-storage/src/evm.rs
@@ -557,7 +557,7 @@ mod tests {
 
         // Build a database that stores the account's bytecode so the journal can fetch it.
         let mut db = CacheDB::new(EmptyDB::default());
-        db.insert_account_info(addr, AccountInfo::new(U256::ZERO, 0, code_hash, code.clone()));
+        db.insert_account_info(addr, AccountInfo::new(U256::ZERO, 0, code_hash, code));
 
         let mut ctx = EthEvmContext::new(db, SpecId::AMSTERDAM);
         let mut observed_code: Option<Bytecode> = None;
--- a/crates/common/precompile-storage/src/evm.rs
+++ b/crates/common/precompile-storage/src/evm.rs
@@ -10,7 +10,10 @@ use alloc::string::ToString;
 use alloy_evm::precompiles::PrecompileInput;
 use alloy_primitives::{Address, B256, Log, LogData, U256};
 use revm::{
-    context::{Block, journaled_state::JournalCheckpoint},
+    context::{
+        Block,
+        journaled_state::{JournalCheckpoint, account::JournaledAccountTr},
+    },
     context_interface::cfg::GasParams,
     interpreter::gas::{Gas, KECCAK256, KECCAK256WORD, LOG},
     primitives::keccak256,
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompile-storage/src/evm.rs
```diff
@@ -128,13 +131,17 @@ impl PrecompileStorageProvider for EvmPrecompileStorageProvider<'_> {
         address: Address,
         f: &mut dyn FnMut(&AccountInfo),
     ) -> Result<()> {
-        // Extract is_cold and clone AccountInfo before releasing the internals borrow.
+        // Load the account AND its bytecode so that `info.code` is always hydrated.
+        // Using `load_account` alone leaves `code: None` for accounts with real bytecode
+        // because the journal only populates `code_hash` lazily; `load_account_code`
+        // forces the cold-code fetch from the database, matching upstream behaviour and
+        // ensuring cold-code-load gas is charged (PSRC #23).
```
refcell: Do we actually use the account code? It doesn't look like it

refcell: Need to check the callback functions

eric-ships: You're right that the two current production callers (`is_b20_initialized` and the factory collision check) both read `info.is_empty_code_hash()`, which depends on `code_hash` — populated by `load_account` already, not `code`.

The fix is still correct for two reasons:
1. `with_account_info` is a public trait method; any future caller that reads `info.code` (e.g. an opcode emulator, a debugger) would silently get `None` without this change
2. The upstream tempo implementation explicitly hydrates code here, and aligning with upstream keeps our semantics consistent

Happy to add a comment clarifying this if useful.

eric-ships: Checked all call sites -- both `storage_ctx.rs:387` and `b20_factory/storage.rs:81,120` only read `is_empty_code_hash()`. No caller reads `info.code` today, so the fix is defensive/forward-looking rather than fixing a current bug.

refcell (approved):

PR #3402, opened 2026-06-10
fix(precompile-macros): reject duplicate args() option regardless of argument count

eric-ships:
[BOP-358](https://linear.app/coinbase/issue/BOP-358)

## Motivation

The `#[precompile]` attribute macro parser accepted duplicate `args` options when the first occurrence was empty (`args(), args()` or `args(), args(x: u8)`). Every other option (`id`, `storage`, `macro_path`, `install`) uses the `reject_duplicate` helper which keys off whether a value has been assigned. The `args` branch instead guarded with `!args.is_empty()`, which misses the case where the first `args()` parses zero arguments and leaves the vec empty. The second occurrence would then silently overwrite the first.

## What changed

- Introduced an `args_seen: bool` flag in `PrecompileConfig::parse` that is set on the first encounter of the `args` key, regardless of how many arguments it contains.
- The duplicate check now tests `args_seen` instead of `!args.is_empty()`, matching the presence-tracking pattern used for all other options.
- Added two regression tests: one for `args(), args()` and one for `args(), args(x: u8)`.

Wrapping `args` in an `Option<Vec<PrecompileArg>>` would also work and would unify the representation with the other fields, but it would require touching downstream code that reads `config.args` directly as a `Vec`. The `args_seen` flag achieves the same detection with the smallest diff and no caller changes.

### DIFF
--- a/crates/common/precompile-storage/src/hashmap.rs
+++ b/crates/common/precompile-storage/src/hashmap.rs
@@ -132,10 +132,8 @@ impl PrecompileStorageProvider for HashMapStorageProvider {
         }
         // EIP-2200 stipend guard — mirrors EvmPrecompileStorageProvider::sstore.
         // Only enforced when gas_remaining is explicitly set (opt-in for tests).
-        if let Some(remaining) = self.gas_remaining {
-            if remaining <= self.gas_params.call_stipend() {
-                return Err(BasePrecompileError::OutOfGas);
-            }
+        if self.gas_remaining.is_some_and(|r| r <= self.gas_params.call_stipend()) {
+            return Err(BasePrecompileError::OutOfGas);
         }
         let old = self.internals.get(&(address, key)).copied().unwrap_or(U256::ZERO);
         self.counter_sstore += 1;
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

eric-ships:
Closing -- PR was opened against the wrong head branch. Reopening with the correct branch ericliu/bop-358-34-args-duplicate-fix.

PR #3417, opened 2026-06-10
fix(precompile-storage): route slot-derivation keccak256 through gas meter (BOP-348)

rayyan224:
## Summary

Fixes BOP-348 (Cantina finding #29): storage layout helpers were computing mapping and dynamic-data slots using `alloy_primitives::keccak256` directly, bypassing the gas-charging `StorageCtx::metered_keccak256` path. Precompile calls could force slot-derivation hashes without paying the corresponding keccak gas.

- **`StorageOps::metered_keccak256`** added as a required method (no default) — every implementor must be explicit so unmetered keccak can never silently slip through. `Slot`/`TransientOps` delegate to `StorageCtx::metered_keccak256`; `PackedSlot` returns `unreachable!()` (it only services packable primitives, never dynamic types)
- **`MappingHandler::at/at_mut`** now return `Result<&Handler>` and compute the child slot via `storage.metered_keccak256` inside the `get_or_try_insert` factory — gas charged only on cache miss, not on every access
- **`calc_data_slot`** in `vec.rs` and `bytes_like.rs` takes `&S: StorageOps` and returns `Result<U256>`; `VecHandler` cached accessors (`at`, `at_with_len`, `at_mut_with_len`) compute the data slot inside the factory closure for the same cache-miss-only semantics
- **`String::mapping_key_buffer`** override added (raw bytes, no 32-byte padding) — required because `MappingHandler` now calls `mapping_key_buffer` + `metered_keccak256` instead of the old `mapping_slot` shortcut
- **`Set<T>::Storable::delete`** uses `storage.metered_keccak256` for position mapping slot lookup
- **`Index`/`IndexMut` on `MappingHandler`** removed — they cannot propagate `Result`; callers use `.at(key)?` / `.at_mut(key)?`
- **Derive macros** (`accounting.rs`, `storable.rs`) updated to propagate `?` from the now-fallible `at`/`at_mut` calls
- **`StorageKey::mapping_slot`** marked `#[doc(hidden)]`; production slot derivation no longer calls it

## Test plan

- [ ] `cargo test -p base-precompile-storage` — 171 tests, all pass
- [ ] `cargo test -p base-common-precompiles` — 380 tests, all pass
- [ ] `cargo clippy -p base-precompile-storage -p base-common-precompiles -p base-precompile-macros` — clean

Closes BOP-348


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompile-storage/src/types/vec.rs
```diff
@@ -201,16 +205,18 @@ where
     }
 
     /// Returns a handler for the element at the given index, or `None` if out of bounds.
+    ///
```
rayyan224: remove comment

crates/common/precompile-storage/src/types/vec.rs
```diff
@@ -201,16 +205,18 @@ where
...
+    /// Keccak gas is charged only on cache miss (first access for a given index).
     pub fn at(&self, index: usize) -> Result<Option<&T::Handler<'a>>> {
         if index >= self.len()? {
             return Ok(None);
         }
-        let (data_start, address, storage) = (self.data_slot(), self.address, self.storage);
```
rayyan224, line 218: We are changing from using len_slot ?  I think we should still have data start / slot ?

Sertug17:
Confirming the unreachable!() concern in PackedSlot::metered_keccak256: since the function already returns Result, an Err(Fatal(...)) would preserve the same invariant (dynamic types must not reach this path) without risking an unconditional process abort if a future refactor accidentally violates it. This is the same defense in depth pattern applied in #3344 where the runtime guard was kept even though config time validation should prevent the bad state.

Sertug17:
Also noting finding #5 from the bot: HandlerCache's struct doc at types/mod.rs:25 still says "Enables Index implementations on handlers" but this PR removes the Index/IndexMut impls from MappingHandler, making that description stale. Updating it to reflect the current purpose (fallible get_or_try_insert cache for Result-returning slot derivations) would keep the doc accurate.

Sertug17:
Confirming the missing // SAFETY annotation: get_or_insert (line 58) and the first deref in get_or_try_insert (line 87) both document the invariant, but the second deref at line 91 does not. The justification is identical  append-only BTreeMap with stable Box heap address  so the same annotation should be present for audit consistency.

PR #3415, opened 2026-06-10
fix(precompile-storage): replace saturating/unchecked slot arithmetic with checked_add (BOP-356)

rayyan224:
## Summary

- Cantina audit finding #33 (Low severity, BOP-356): slot offset arithmetic in `base-precompile-storage` used `saturating_add` (which silently aliases two distinct slots to `U256::MAX` on overflow) and unchecked `+` (which panics in debug builds)
- Replace all slot additions with `checked_add`, returning `BasePrecompileError::SlotOverflow` in `Result`-returning functions; use `checked_add().expect("slot overflow")` in constructors and `Index`/`IndexMut` impls where the trait contract requires an infallible signature
- `Slot::new_at_offset` and `Slot::new_at_loc` now return `Result<Self>` instead of `Self`

## Files changed

- `types/slot.rs` — `new_at_offset`, `new_at_loc`: `saturating_add` → `checked_add().ok_or(SlotOverflow)?`, return type changed to `Result<Self>`
- `types/vec.rs` — all five slot-offset sites in free functions; `compute_handler` uses `expect` (infallible cache closure)
- `types/set.rs` — `Storable::delete` uses `?`; `SetHandler::new` constructor uses `expect`; `new` marked `const`
- `types/array.rs` — `compute_handler` uses `expect` (`Index`/`IndexMut` must be infallible)
- `types/bytes_like.rs` — all three chunk-loop sites use `?`

## Test plan

- [x] `cargo test -p base-precompile-storage` — 171 tests pass
- [x] `cargo clippy -p base-precompile-storage` — clean

PR #3427, opened 2026-06-10
fix(precompile-storage): complete slot arithmetic checked-add migration

eric-ships:
## Motivation

The initial fix in #3415 replaced `saturating_add` and unchecked `+` with `checked_add().ok_or(SlotOverflow)?` across most slot arithmetic in `base-precompile-storage`, but missed three sites. This PR completes the migration.

## Changes

**`VecHandler::truncate` (two sites -- the #3415 bot finding)**

Two bare `+` expressions in the packed-element branch of `truncate()` that were added by #3384 and not caught by #3415:

```rust
// before
let boundary_slot_addr = data_start + U256::from(new_len / elems_per_slot);
let slot_addr = data_start + U256::from(slot_idx);

// after
let boundary_slot_addr = data_start
    .checked_add(U256::from(new_len / elems_per_slot))
    .ok_or(BasePrecompileError::SlotOverflow)?;
```

**`VecHandler::try_compute_handler` (renamed from `compute_handler`)**

Returns `Result<T::Handler>` instead of panicking. All callers that return `Result` now use `?`. Call sites inside `HandlerCache` closures use two new methods added to `HandlerCache`:

- `get_or_try_insert` -- fallible shared-ref variant of `get_or_insert`
- `get_or_try_insert_mut` -- fallible mut variant of `get_or_insert_mut`

**`SetHandler::new`**

Remains `const fn`. The `positions` slot (`base_slot + 1`) now uses `checked_add().expect()` with an invariant comment (base slots are keccak256-derived and cannot be `U256::MAX`), replacing the previous `wrapping_add`. `StorableType::handle` for `Set<T>` is an infallible trait method so `Result` propagation is structurally impossible here.

**`ArrayHandler` -- replaced `Index`/`IndexMut` with `at()`/`at_mut()`**

`Index` and `IndexMut` require infallible return types, so they silently panicked on slot overflow. Replaced with `at(&self, index: usize) -> Result<Option<&Handler>>` and `at_mut(&mut self, index: usize) -> Result<Option<&mut Handler>>` which propagate `SlotOverflow` as an error. All call sites updated.

## Validation

- `cargo check -p base-precompile-storage`
- `cargo clippy -p base-precompile-storage --all-targets -- -D warnings`
- `cargo test -p base-precompile-storage` (175 tests pass)
- `cargo +nightly fmt --all -- --check`
- `cargo check -p base-precompile-storage --no-default-features` (no_std)

### DIFF
--- a/crates/common/precompile-storage/src/types/array.rs
+++ b/crates/common/precompile-storage/src/types/array.rs
@@ -73,6 +73,10 @@ impl<'a, T: StorableType, const N: usize> ArrayHandler<'a, T, N> {
         )
     }
 
+    // Called from `Index` and `IndexMut` impls, which return `&T` / `&mut T` — not `Result`.
+    // The `Index` trait is infallible by design, so error propagation is impossible here.
+    // Overflow is unreachable in practice: `base_slot` is a keccak256 output (never U256::MAX)
+    // and `N` is a compile-time constant bounded by the `storable_arrays!` macro (N ≤ 32).
     #[inline]
     fn compute_handler(
         base_slot: U256,
@@ -83,12 +87,16 @@ impl<'a, T: StorableType, const N: usize> ArrayHandler<'a, T, N> {
         let (slot, layout_ctx) = if T::BYTES <= 16 {
             let location = packing::calc_element_loc(index, T::BYTES);
             (
-                base_slot.checked_add(U256::from(location.offset_slots)).expect("slot overflow"),
+                base_slot
+                    .checked_add(U256::from(location.offset_slots))
+                    .expect("slot overflow: unreachable for keccak-derived slots with N ≤ 32"),
                 LayoutCtx::packed(location.offset_bytes),
             )
         } else {
             (
-                base_slot.checked_add(U256::from(index * T::SLOTS)).expect("slot overflow"),
+                base_slot
+                    .checked_add(U256::from(index * T::SLOTS))
+                    .expect("slot overflow: unreachable for keccak-derived slots with N ≤ 32"),
                 LayoutCtx::FULL,
             )
         };
--- a/crates/common/precompile-storage/src/types/array.rs
+++ b/crates/common/precompile-storage/src/types/array.rs
@@ -4,12 +4,10 @@
 //! - **Base slot**: Arrays start directly at `base_slot` (not at keccak256)
 //! - Small elements (`T::BYTES` ≤ 16) are packed; larger elements use full slots.
 
-use core::ops::{Index, IndexMut};
-
 use alloy_primitives::{Address, U256};
 
 use crate::{
-    error::Result,
+    error::{BasePrecompileError, Result},
     packing,
     provider::{Handler, LayoutCtx, Storable, StorableType},
     types::{HandlerCache, Slot},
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompile-storage/src/types/array.rs
```diff
@@ -83,12 +87,16 @@ impl<'a, T: StorableType, const N: usize> ArrayHandler<'a, T, N> {
...
+                    .checked_add(U256::from(location.offset_slots))
+                    .expect("slot overflow: unreachable for keccak-derived slots with N ≤ 32"),
                 LayoutCtx::packed(location.offset_bytes),
             )
         } else {
             (
-                base_slot.checked_add(U256::from(index * T::SLOTS)).expect("slot overflow"),
+                base_slot
+                    .checked_add(U256::from(index * T::SLOTS))
+                    .expect("slot overflow: unreachable for keccak-derived slots with N ≤ 32"),
```
rayyan224: I don't think we should expect anymore ? it should just error and return option

crates/common/precompile-storage/src/types/array.rs
```diff
@@ -61,58 +59,52 @@ impl<'a, T: StorableType, const N: usize> ArrayHandler<'a, T, N> {
...
         index: usize,
-    ) -> T::Handler<'a> {
+    ) -> Result<T::Handler<'a>> {
         let (slot, layout_ctx) = if T::BYTES <= 16 {
             let location = packing::calc_element_loc(index, T::BYTES);
             (
-                base_slot.checked_add(U256::from(location.offset_slots)).expect("slot overflow"),
+                base_slot
+                    .checked_add(U256::from(location.offset_slots))
+                    .ok_or(BasePrecompileError::SlotOverflow)?,
```
refcell, line 96: Awesome

crates/common/precompile-storage/src/types/set.rs
```diff
@@ -112,6 +112,7 @@ where
         storage: crate::StorageCtx<'a>,
     ) -> Self::Handler<'a> {
         SetHandler::new(slot, address, storage)
+            .expect("slot overflow: keccak-derived base slots cannot be U256::MAX")
```
refcell: Shoot now it looks like we panic here. Maybe this means `StorableType` trait has to bubble up an error? Not sure how ugly this is going to get..

crates/common/precompile-storage/src/types/set.rs
```diff
@@ -157,13 +157,14 @@ where
 {
     /// Creates a new handler for the set at the given base slot.
     pub const fn new(base_slot: U256, address: Address, storage: crate::StorageCtx<'a>) -> Self {
+        // positions_slot = base_slot + 1. Base slots are keccak256-derived and never U256::MAX,
+        // so overflow is unreachable in practice.
+        let positions_slot = base_slot
+            .checked_add(U256::ONE)
+            .expect("slot overflow: base slot is keccak256-derived and cannot be U256::MAX");
```
rayyan224, line 164: I think here we should return a fatal error

eric-ships: **base_slot** is always a keccak256-derived address.

For U256::MAX to be a valid base slot would require a keccak256 preimage attack - effectively impossible. 

Making **StorableType::handle** return **Result** to handle this would break the trait for every storage type just to guard against a path that can't occur.

Keeping .expect() with the invariant comment is the right tradeoff here. cc @refcell

rayyan224 (approved):

cb-heimdall:
Review Error for rayyan224 @ 2026-06-11 00:31:45 UTC

PR #3436, opened 2026-06-10
Build zk-prover-postgres image with baked migrations

jowparks:
## Summary
- Adds `etc/docker/Dockerfile.zk-prover-postgres` baking `crates/proof/zk/db/migrations` into `postgres:17-alpine`.
- Publishes via existing l3 Docker CI as `ghcr.io/base/node-reth-dev:zk-prover-postgres-{sha,l3}`.
- Updates local compose files to build/use the image instead of bind-mounting migrations.

## Why
privacy-enclave (and other Odin/Codeflow deploys) cannot bind-mount migration files. Downstream repos should mirror this image rather than vendoring SQL.

## Test plan
- [x] `docker build -f etc/docker/Dockerfile.zk-prover-postgres .`
- [ ] After merge: confirm `docker-l3` workflow publishes `zk-prover-postgres-l3`
- [ ] Run `scripts/bump-docker-mirror-digests.sh` in privacy-enclave to pin digest

### DIFF
--- a/.github/workflows/docker-l3.yml
+++ b/.github/workflows/docker-l3.yml
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [client, base, builder, consensus, batcher, zk-prover]
+        target: [client, base, builder, consensus, batcher, zk-prover, zk-prover-postgres]
 
     runs-on: ubuntu-24.04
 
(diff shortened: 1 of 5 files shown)


hughy (approved):

jowparks:
Not needed for l3

PR #3439, opened 2026-06-10
feat(proposer): add collector orchestration

leopoldjoy:
## Summary

Adds collector-side orchestration on top of the deterministic polling support and dispatcher abstraction.

The collector orchestrator owns polling, sequential inline submit, ready-backlog draining, retry-specific sessions, and restart decisions so final pipeline wiring can remain a thin lifecycle coordinator.

## Changes

- Add `ProofCollectorOrchestrator` for collector-side polling and submit orchestration.
- Add collector runtime/state types for cursor tracking, retry accounting, discard retry state, and retry sessions.
- Move inline submit behavior into collector orchestration, including submit timeout handling and restart/drop-cache decisions.
- Re-dispatch terminal failed sessions through the dispatcher abstraction.
- Dispatch retry-specific sessions for discarded proofs and cap discard retries.
- Replace missing retry-specific sessions with fresh retry sessions.
- Add a recovery-provider hook used after successful submits and `GameAlreadyExists` recovery.

## Verification

### DIFF
--- a/crates/proof/proposer/src/proof_collector.rs
+++ b/crates/proof/proposer/src/proof_collector.rs
@@ -1,4 +1,4 @@
-//! Polls the prover service for the proof of a single derived target block.
+//! Polls and orchestrates prover-service collection for proposer TEE proofs.
 //!
 //! The [`ProofCollector`] does not care about the proof dispatcher. It assumes that
 //! the eligible block range either has, or will have, a `prove_block_range` request
@@ -14,20 +14,107 @@
 //! dispatch state, the collector can rediscover and complete sessions across
 //! proposer restarts.
 
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, sync::Arc, time::Duration};
 
 use alloy_primitives::B256;
+use async_trait::async_trait;
 use base_proof_primitives::ProofResult;
-use base_proof_rpc::RollupProvider;
+use base_proof_rpc::{L1Provider, L2Provider, RollupProvider};
 use base_prover_service_client::{ProofRequesterProvider, ProverServiceClientError};
 use base_prover_service_protocol::{GetProofRequest, GetProofResponse, ProofStatus, TeeKind};
 use futures::{StreamExt, stream};
-use tracing::debug;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, error, info, warn};
 
 use crate::{
-    driver::RecoveredState, error::ProposerError, metrics::Metrics,
+    driver::RecoveredState,
+    error::ProposerError,
+    metrics::Metrics,
     proof_adapter::ProposerProofAdapter,
+    proof_dispatcher::{ProofDispatchAttempt, ProofDispatcher},
+    proof_submitter::{ProofSubmitter, SubmitAction},
 };
+
+/// Cached result from the last successful recovery walk.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ProofRecoveryCache {
+    /// Factory `game_count` at the time of the walk.
+    pub game_count: u64,
+    /// Recovered on-chain state from the walk.
+    pub state: RecoveredState,
+}
+
+/// Recovery hook used by collector orchestration after successful submissions.
+#[async_trait]
+pub trait ProofCollectorRecoveryProvider: Send + Sync {
+    /// Refreshes the recovery cache and returns the latest on-chain state.
+    async fn recover_latest_state(
+        &self,
+        cache: &mut Option<ProofRecoveryCache>,
+    ) -> Result<RecoveredState, ProposerError>;
+}
+
+/// Runtime settings for collector orchestration.
+#[derive(Debug, Clone, Copy)]
+pub struct ProofCollectorRuntimeConfig {
+    /// Number of L2 blocks between output proposals.
+    pub block_interval: u64,
+    /// Maximum proof failures or discard retries before dropping recovery state.
+    pub max_retries: u32,
+    /// Maximum duration for a single inline submit attempt.
+    pub submit_timeout: Duration,
+}
+
+/// Mutable collector-side orchestration state.
+#[derive(Debug, Default)]
+pub struct ProofCollectorState {
+    /// Latest block the collector has submitted through.
+    pub cursor: Option<RecoveredState>,
+    /// Per-target proof/dispatch retry counts.
+    pub retry_counts: HashMap<u64, u32>,
+    /// Per-target discard retry counts.
+    pub discard_retry_counts: HashMap<u64, u32>,
+    /// Active retry-specific prover-service sessions by target block.
+    pub retry_sessions: HashMap<u64, String>,
+}
+
+/// Result of one collector tick.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProofCollectorTickResult {
+    /// The collector made progress or reached a natural wait point.
+    Continue,
+    /// The owning pipeline session should restart from recovered chain state.
+    Restart,
+}
+
+/// Result of an inline submit attempt.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProofSubmitEffect {
+    /// The proof was submitted or the game already existed.
+    Submitted,
+    /// The pipeline should restart before collecting more proofs.
+    Restart,
+    /// The submitted proof should be replaced by a retry-specific session.
+    Redispatch {
+        /// Claimed L2 output root for the discarded proof.
+        claimed_l2_output_root: B256,
+    },
+}
+
+/// Owns collector-side polling, sequential submit, and retry-session orchestration.
+pub struct ProofCollectorOrchestrator<L1, L2, R, Recovery>
+where
+    L1: L1Provider,
+    L2: L2Provider,
+    R: RollupProvider,
+    Recovery: ProofCollectorRecoveryProvider,
+{
+    collector: ProofCollector<R>,
+    dispatcher: ProofDispatcher<L1, L2, R>,
+    submitter: ProofSubmitter<L1, R>,
+    recovery: Arc<Recovery>,
+    runtime: ProofCollectorRuntimeConfig,
+}
 
 /// Outcome of polling the prover service for a single target block.
 ///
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,6 +657,513 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+            || {
+                error!(
+                    current_block,
+                    block_interval = self.runtime.block_interval,
+                    "Overflow computing next target block"
+                );
+                None
+            },
+            Some,
+        )
```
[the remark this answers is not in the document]
leopoldjoy: Fixed in `648443981`: `next_target_block` now returns `None` when `block_interval == 0`.

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,6 +657,518 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+                    {
+                        ProofSubmitEffect::Submitted => {
+                            state.retry_sessions.remove(&target_block);
+                            state.discard_retry_counts.remove(&target_block);
+                            if let Some(cached) = cache.as_ref() {
+                                state.cursor = Some(cached.state);
+                                if cached.state.l2_block_number > current.l2_block_number {
+                                    continue;
+                                }
+                            }
```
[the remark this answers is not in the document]
leopoldjoy: Intentional: the collector drains consecutive ready proofs to catch up without one poll interval per proof, while checking cancellation between iterations.

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,6 +657,518 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+        error: ProposerError,
+        max_retries: u32,
+        cache: &mut Option<ProofRecoveryCache>,
+    ) -> bool {
+        Metrics::errors_total(error.metric_label()).increment(1);
+        Metrics::proof_retries_total().increment(1);
+
+        let count = self.retry_counts.entry(target).or_insert(0);
+        *count += 1;
+        if *count >= max_retries {
```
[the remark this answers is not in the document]
leopoldjoy: Intentional: `max_retries == 0` means zero retries allowed; production defaults use a positive value.

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,6 +657,518 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+            }
+            ProofDispatchAttempt::DispatchFailed(error) => {
+                Metrics::proof_dispatch_total(Metrics::DISPATCH_OUTCOME_FAILED).increment(1);
+                if count_dispatch_failure {
+                    state.handle_proof_failure(
+                        target_block,
+                        error,
+                        self.runtime.max_retries,
+                        cache,
+                    );
```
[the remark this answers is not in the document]
leopoldjoy: Intentional: retry exhaustion drops the recovery cache so the next tick re-walks chain state; it does not permanently suppress the target.

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,18 +657,696 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+        &self,
+        state: &mut ProofCollectorState,
+        cache: &mut Option<ProofRecoveryCache>,
+        recovered: RecoveredState,
+        safe_head: u64,
+        cancel: &CancellationToken,
+    ) -> ProofCollectorTickResult {
+        state.prune_recovered(recovered.l2_block_number);
+
+        if state.cursor.is_none_or(|cursor| cursor.l2_block_number < recovered.l2_block_number) {
```
jackchuma: Reset the collector cursor when recovery moves backward. If a later recovery pass returns a lower L2 block than the current cursor (for example after an L1 reorg, cache invalidation, or invalid-parent restart), this condition keeps the old cursor because it only updates when recovery advances. The loop then starts from `cursor + block_interval` and skips the targets between the newly recovered block and the stale cursor, so subsequent submissions can build on parents that are no longer recovered onchain

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,18 +657,696 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+                    target_block,
+                    session_id = %dispatched.session_id,
+                    attempt,
+                    "Discard retry proof request accepted by prover service"
+                );
+                Metrics::proof_dispatch_total(Metrics::DISPATCH_OUTCOME_ACCEPTED).increment(1);
+                state.discard_retry_counts.insert(target_block, attempt);
+                state.retry_sessions.insert(target_block, dispatched.session_id);
+            }
+            ProofDispatchAttempt::BuildFailed(error) => {
```
jackchuma: Avoid resubmitting proofs after discard retry build failures. When a proof is discarded and building the discard retry fails due to a transient L1/L2 RPC error, this branch removes any retry session and returns success without recording that the root-derived proof must be avoided. On the next tick `retry_sessions` is empty, so the collector polls the original succeeded root session and submits the same discarded proof again instead of retrying the dispatch, which can loop discarded/reverting submissions until a retry request is finally accepted

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,18 +657,696 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+                TargetPoll::Failed { session_id, claimed_l2_output_root, error } => {
+                    warn!(
+                        target_block,
+                        session_id = %session_id,
+                        error = %error,
+                        "Prover service reported failed session, re-dispatching"
+                    );
+                    Metrics::proof_collection_total(Metrics::COLLECTION_OUTCOME_FAILED)
+                        .increment(1);
+                    if state.handle_proof_failure(
```
jackchuma: Count a terminal failed session only once per retry attempt. A terminal failed prover session is counted before the replacement request is successfully scheduled. If the immediate retry dispatch cannot be built or sent, the next tick polls the same terminal session and increments the count again, so a transient dispatch/build outage can exhaust `max_retries` and drop recovery without any additional proof attempts actually running

leopoldjoy: Done: https://github.com/base/base/pull/3439/commits/85f720358c76dd5bdd484dbd6f79a804b2f2a9c5

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,18 +663,736 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+                        error = %error,
+                        "Prover service reported failed session, re-dispatching"
+                    );
+                    Metrics::proof_collection_total(Metrics::COLLECTION_OUTCOME_FAILED)
+                        .increment(1);
+                    let already_counted = state
+                        .counted_failed_sessions
+                        .get(&target_block)
+                        .is_some_and(|id| id == &session_id);
+                    if already_counted
```
mw2000, line 933: I think this can still get stuck if the replacement dispatch keeps failing: once the failed session is marked counted, future ticks skip the retry budget path, so we never make it to Restart.

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,18 +663,736 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+            ProofDispatchAttempt::Accepted(dispatched) => {
+                info!(
+                    target_block,
+                    session_id = %dispatched.session_id,
+                    attempt,
+                    "Discard retry proof request accepted by prover service"
+                );
+                Metrics::proof_dispatch_total(Metrics::DISPATCH_OUTCOME_ACCEPTED).increment(1);
+                state.discard_retry_counts.insert(target_block, attempt);
+                state.retry_sessions.insert(target_block, dispatched.session_id);
```
jackchuma, line 1232: Preserve retry session ids across dispatch errors. For discard retries, the session id includes the L1 head number, but the orchestrator only records `dispatched.session_id` after a successful RPC response. If `prove_block_range` times out after the prover service accepted the request, the accepted session id is lost; the next tick can rebuild the same attempt with a newer L1 head and create a different session, leaving the first proof unpolled and potentially repeating until retry exhaustion

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -570,18 +663,736 @@ impl<R: RollupProvider + 'static> ProofCollector<R> {
...
+            }
+            ProofDispatchAttempt::BuildFailed(error) => {
+                warn!(target_block, error = %error, "Failed to build discard retry proof request");
+                Metrics::proof_dispatch_total(Metrics::DISPATCH_OUTCOME_BUILD_FAILED).increment(1);
+                state.retry_sessions.remove(&target_block);
+            }
+            ProofDispatchAttempt::DispatchFailed(error) => {
+                Metrics::proof_dispatch_total(Metrics::DISPATCH_OUTCOME_FAILED).increment(1);
+                if count_dispatch_failure {
+                    if !state.handle_proof_failure(
```
jackchuma: Clear discard state when retry dispatch exhausts. When a discard retry dispatch failure exhausts `max_retries`, `handle_proof_failure` only clears `retry_counts` and the recovery cache. Because `pending_discard_roots` was inserted above and is not removed here, any caller that keeps `ProofCollectorState` across the requested restart will immediately redispatch this exhausted target on the next tick with a fresh budget, instead of dropping the retry state like the pre-check exhaustion path does

jackchuma (approved):
a few non-blocking comments that can be addressed in a follow up

mw2000 (approved):

crates/proof/proposer/src/proof_collector.rs
leopoldjoy: Fixed: https://github.com/base/base/pull/3439/commits/d9a4604394bab212de4fd916b1572b977d98c869

crates/proof/proposer/src/proof_collector.rs
leopoldjoy: Addressed: https://github.com/base/base/pull/3439/commits/435f091e07d34b9eb650238ea3f96dd2628c60f7

crates/proof/proposer/src/proof_collector.rs
leopoldjoy: https://github.com/base/base/pull/3439/commits/435f091e07d34b9eb650238ea3f96dd2628c60f7

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -14,21 +14,114 @@
...
+}
+
+/// Recovery hook used by collector orchestration after successful submissions.
+#[async_trait]
+pub trait ProofCollectorRecoveryProvider: Send + Sync {
+    /// Refreshes the recovery cache and returns the latest on-chain state.
+    async fn recover_latest_state(
+        &self,
+        cache: &mut Option<ProofRecoveryCache>,
+    ) -> Result<RecoveredState, ProposerError>;
```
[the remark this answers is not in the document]
leopoldjoy: Keeping the `RecoveredState` return: the submit path now uses it to advance the collector cursor after successful recovery.

mw2000 (approved):

PR #3458, opened 2026-06-11
[backport] fix(precompiles): Cap B20 Account Balances

refcell:
> [!NOTE]
> This is a backport of #3457 into `releases/v1.1.0`.

## Summary

This reserves the upper 128 bits of each B20 account balance slot by capping balances at 2^128 - 1. Mint and transfer paths now validate the recipient's resulting balance before writing state and return a dedicated MaxBalanceExceeded error when the cap would be crossed. The focused unit tests cover the exact boundary and the revert path without changing burn or allowance behavior.

### DIFF
--- a/crates/common/precompiles/src/common/ops/transferable.rs
+++ b/crates/common/precompiles/src/common/ops/transferable.rs
@@ -2,7 +2,7 @@ use alloy_primitives::{Address, B256, U256};
 use alloy_sol_types::SolEvent;
 use base_precompile_storage::{BasePrecompileError, Result};
 
-use crate::{B20Guards, B20PolicyType, IB20, Token, TokenAccounting};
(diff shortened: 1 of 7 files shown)


crates/common/precompiles/src/common/ops/transferable.rs
```diff
@@ -54,12 +54,16 @@ pub trait Transferable: Token {
                 needed: amount,
             }));
         }
+        if from == to {
+            return self
+                .accounting_mut()
+                .emit_event(IB20::Transfer { from, to, amount }.encode_log_data());
+        }
```
refcell, line 61: Do we want this?

refcell: It should be just harmless, returning early saving us ops

PR #3476, opened 2026-06-11
feat(upgrade-signal): add dynamic hardfork scheduling foundation

PelleKrab:
## Summary

Adds the shared foundation for consuming the L1 upgrade signal and materializing dynamic hardfork activation overrides at runtime.

This PR does not yet wire the signal into execution or consensus services. It introduces the shared types, readers, validation, and application primitives that follow-up PRs will integrate into those startup and runtime flows.

## Changes

- add a process-local runtime hardfork override registry and shared activation sink API in base-common-genesis
- make rollup hardfork checks and execution chainspec fork views consult runtime activation overrides when present
- add the base-upgrade-signal crate for:
  - reading hardfork timestamps and protocol versions from the L1 contract
  - validating malformed signals and unsupported protocol versions
  - tracking live signal state and metrics
  - applying validated schedules to runtime destinations

## Testing

- cargo fmt --check --all
- cargo test -p base-upgrade-signal -p base-common-genesis -p base-common-chains -p base-execution-chainspec

## Manual Validation

- just devnet up

## Follow-ups

- wire the upgrade signal into execution startup and runtime flows
- wire the upgrade signal into consensus startup and runtime flows
- add docs and devnet/e2e coverage


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/chainspec/src/spec.rs
```diff
@@ -408,23 +669,23 @@ impl EthChainSpec for BaseChainSpec {
...
     fn latest_fork_id(&self) -> ForkId {
-        self.inner.latest_fork_id()
+        self.runtime_chain_spec().latest_fork_id()
     }
 
     fn fork_filter(&self, head: Head) -> ForkFilter {
-        self.inner.fork_filter(head)
+        self.runtime_chain_spec().fork_filter(head)
     }
 }
```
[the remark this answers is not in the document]
PelleKrab: Seems like too much to add in this pr, possible retroactive pr.

crates/common/genesis/src/rollup.rs
```diff
@@ -182,78 +189,139 @@ macro_rules! rollup_fork_methods {
...
+        }
+    }
+
     rollup_fork_methods! {
         is_regolith_active,
         is_first_regolith_block,
-        [hardforks.regolith_time],
+        [hardfork_activation_timestamp("regolith")],
         "Regolith",
         implies is_canyon_active;
```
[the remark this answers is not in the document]
PelleKrab: Valid, but in the current state this inefficiency is acceptable. It may be worth a revisit in the future.

crates/utilities/upgrade-signal/src/metrics.rs
```diff
@@ -0,0 +1,80 @@
...
+            Self::record_signal(signal);
+        }
+    }
+
+    /// Records all metrics derived from a successfully read signal.
+    pub fn record_signal(signal: &UpgradeSignal) {
+        Self::init();
+        let hardfork_id = signal.hardfork_id.clone();
+
+        Self::activation_timestamp(hardfork_id.clone()).set(signal.activation_timestamp as f64);
```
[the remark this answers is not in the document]
PelleKrab: dtm

crates/utilities/upgrade-signal/README.md
```diff
@@ -0,0 +1,17 @@
...
+and live signal changes.
+
+Three graduated rollout modes are supported:
+
+- **metrics-only** — observe signals and record metrics without applying them
+- **startup-apply** — pin activation timestamps into the chain spec at node startup
+- **runtime-admin** — write live overrides into `RuntimeHardForkRegistry` so fork checks reflect
+  contract-sourced signals without a node restart
+
+See [`docs/guides/UPGRADE_SIGNAL.md`](../../../docs/guides/UPGRADE_SIGNAL.md) for the feature
```
refcell: I'm not sure this resolves to this repo

crates/utilities/upgrade-signal/README.md
```diff
@@ -0,0 +1,17 @@
+# `base-upgrade-signal`
+
+Shared utilities for reading network upgrade activation signals from L1.
+
+The crate reads an L1 contract interface and decodes the announced activation timestamp and minimum
+node protocol version for each configured hardfork ID. Metrics are recorded for both startup reads
```
refcell:
```suggestion
node protocol version for each configured upgrade ID. Metrics are recorded for both startup reads
```

crates/execution/chainspec/src/spec.rs
```diff
@@ -285,6 +288,251 @@ impl BaseChainSpec {
...
+            "Osaka" | "Azul" => Some("azul"),
+            "Beryl" => Some("beryl"),
+            "Cobalt" => Some("cobalt"),
+            _ => None,
+        }
+    }
+
+    /// Returns a runtime hardfork activation condition for an execution fork.
+    pub fn runtime_fork_condition<H: Hardfork + ?Sized>(&self, fork: &H) -> Option<ForkCondition> {
+        let hardfork_id = Self::runtime_hardfork_id(fork)?;
```
refcell: We should be able to remove `runtime_hardfork_id` and just do:

```rust
let upgrade_id = fork.id();
```

crates/common/genesis/src/chain/upgrade.rs
refcell, line 1: Can we use the `upgrade` terminology consistent rather than `hardfork`

crates/execution/chainspec/src/spec.rs
```diff
@@ -285,6 +288,251 @@ impl BaseChainSpec {
...
+                hardforks.insert(EthereumHardfork::Osaka, condition);
+                hardforks.insert(BaseUpgrade::Azul, condition);
+            }
+            "beryl" | "baseberyl" | "v2" => {
+                hardforks.insert(BaseUpgrade::Beryl, condition);
+            }
+            "cobalt" | "basecobalt" | "v3" => {
+                hardforks.insert(BaseUpgrade::Cobalt, condition);
+            }
+            _ => return false,
```
refcell: This issue with this matching is every time we add a new upgrade, we'll have to come here and add a new branch. We've tried to manage this by consolidating upgrade logic in the `base-common-chains` crate. 

I'm thinking something that could work here is having `BaseUpgrade` define a `FromStr` method. More broadly though, I'd like to avoid using strings where possible. It's easy to typo and is similar to having magic numbers across your codebase. That is, the chance of inconsistency is higher since we're not leveraging rust's native strong typing.

crates/utilities/upgrade-signal/src/config.rs
refcell, line 1: I think we should break this module up. Ideally modules are <300 LOC for readability. Since there are many types here we can trivially break out types.

crates/utilities/upgrade-signal/src/config.rs
```diff
@@ -0,0 +1,609 @@
...
+pub const DEFAULT_UPGRADE_SIGNAL_READ_BACKOFF: Duration = Duration::from_secs(2);
+
+/// Node protocol version supported by this binary for contract-backed upgrade signals.
+///
+/// Contract schedules with a higher minimum protocol version are rejected before any timestamp is
+/// applied. Bump this with the node software that fully implements the next dynamic upgrade.
+pub const DEFAULT_UPGRADE_SIGNAL_NODE_PROTOCOL_VERSION: u64 = 7;
+
+/// Default hardfork IDs read from the L1 upgrade signal contract.
+pub const DEFAULT_UPGRADE_SIGNAL_HARDFORK_IDS: &[&str] = HardForkConfig::CONTRACT_HARDFORK_IDS;
```
refcell: I'm not sure we need this constant, it seems to just be a re-export for `HardForkConfig::CONTRACT_HARDFORK_IDS` from `base-common-genesis`. We should be able to just use that crate directly where needed.

crates/utilities/upgrade-signal/src/config.rs
```diff
@@ -0,0 +1,609 @@
...
+pub const DEFAULT_UPGRADE_SIGNAL_READ_ATTEMPTS: u32 = 3;
+
+/// Default backoff between L1 upgrade signal schedule read attempts.
+pub const DEFAULT_UPGRADE_SIGNAL_READ_BACKOFF: Duration = Duration::from_secs(2);
+
+/// Node protocol version supported by this binary for contract-backed upgrade signals.
+///
+/// Contract schedules with a higher minimum protocol version are rejected before any timestamp is
+/// applied. Bump this with the node software that fully implements the next dynamic upgrade.
+pub const DEFAULT_UPGRADE_SIGNAL_NODE_PROTOCOL_VERSION: u64 = 7;
```
refcell: One way we can make our export surface slimmer is grouping constants into a unit struct. We try to follow this pattern across the repo where possible.

e.g.

```rust
pub struct UpgradeSignalDefaults;

impl UpgradeSignalDefaults {
    //// Default wall-clock interval used to check whether another L1 block polling window has elapsed.
    pub const POLL_INTERVAL: Duration = Duration::from_secs(12);

    ...
}
```

Then, usage is just `UpgradeSignalDefaults::POLL_INTERVAL`, and in `lib.rs` we only need to do one export instead of listing out each constant.

```rust
pub use config::UpgradeSignalDefaults;
```

crates/utilities/upgrade-signal/src/config.rs
```diff
@@ -0,0 +1,609 @@
...
+
+    /// Returns a copy of `schedule` filtered to the hardfork IDs that may be applied locally.
+    pub fn application_schedule(&self, schedule: &UpgradeSignalSchedule) -> UpgradeSignalSchedule {
+        schedule.filtered_to_hardfork_ids(&self.apply_hardfork_ids)
+    }
+
+    /// Returns true if this node supports the minimum protocol version attached to `signal`.
+    pub fn supports_signal_protocol_version(&self, signal: &UpgradeSignal) -> bool {
+        signal.protocol_version <= self.node_protocol_version
+    }
```
refcell: Nice this is clean and simple 👍

crates/utilities/upgrade-signal/src/config.rs
```diff
@@ -0,0 +1,609 @@
...
+}
+
+/// CLI argument for the L1 RPC endpoint used by standalone execution nodes.
+#[derive(Debug, Clone, Default, PartialEq, Eq, clap::Args)]
+pub struct UpgradeSignalL1RpcArgs {
+    /// L1 execution RPC URL used to read the upgrade signal contract.
+    #[arg(long = "upgrade-signal.l1-rpc", env = "BASE_NODE_UPGRADE_SIGNAL_L1_RPC")]
+    pub upgrade_signal_l1_rpc: Option<Url>,
+}
```
refcell: It'd be nice if we could default to using the same L1 rpc url as defined by the execution cli args, otherwise most users will just have to pass the same url in twice. This definitely doesn't have to be done in this PR as it probably requires a bit of refactoring.

refcell: Maybe we could add a `// TODO` comment above this type though noting this ^

crates/utilities/upgrade-signal/src/config.rs
```diff
@@ -0,0 +1,609 @@
...
+            UpgradeSignalConfig::new(address!("0000000000000000000000000000000000000001"), "azul");
+        let minimum_protocol_version = config.node_protocol_version + U256::from(1);
+
+        assert!(matches!(
+            config.validate_signal_protocol_version(&signal(minimum_protocol_version)).unwrap_err(),
+            UpgradeSignalError::UnsupportedProtocolVersion { .. }
+        ));
+    }
+
+    #[test]
```
refcell: Some of this testing is good candidates for table testing with `rstest` since we can parameterize the upgrade id (i.e. "azul"). `rstest` also helps keep our modules a bit more concise in LOC

crates/utilities/upgrade-signal/src/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    #[error("{context}: {error}")]
+    Decode {
+        /// Decode context.
+        context: &'static str,
+        /// Decode error string.
+        error: String,
+    },
+    /// The activation timestamp does not fit in a `u64`.
+    #[error("activation timestamp {0} does not fit in u64")]
+    TimestampOverflow(String),
```
refcell: Think this can be a `U256` right since that's the value parsed in `contract.rs`

crates/utilities/upgrade-signal/src/error.rs
```diff
@@ -0,0 +1,99 @@
...
+        "upgrade signal for {0} has an activation timestamp but no minimum node protocol version"
+    )]
+    MissingProtocolVersion(String),
+    /// The contract requires a newer node protocol version than this binary supports.
+    #[error(
+        "upgrade signal for {hardfork_id} requires node protocol version {minimum_protocol_version}, but this binary supports {node_protocol_version}"
+    )]
+    UnsupportedProtocolVersion {
+        /// Hardfork ID whose signal required a newer protocol version.
+        hardfork_id: String,
```
refcell: Let's stick to `upgrade` rather than `hardfork` terminology

crates/utilities/upgrade-signal/src/contract.rs
```diff
@@ -0,0 +1,264 @@
...
+        event TimestampSet(string indexed hardforkId, uint256 timestamp);
+
+        /// Emitted when a protocol version is set for a hardfork ID.
+        event ProtocolVersionSet(string indexed hardforkId, uint256 protocolVersion);
+
+        /// Returns the activation timestamp for `hardforkId`.
+        function getTimestamp(string hardforkId) external view returns (uint256);
+
+        /// Returns the minimum node protocol version for `hardforkId`.
+        function getProtocolVersion(string hardforkId) external view returns (uint256);
```
refcell, line 31: How are we recording protocol versions as `U256`? Right now node software is released as a semver. How do we expect to translate this?

It's worth having sections in the crate `README.md` to outline What and How Protocol Versions and Upgrade Timestamps work I think in this regard.

crates/utilities/upgrade-signal/src/runtime.rs
refcell, line 1: This module is worth breaking up as well

crates/utilities/upgrade-signal/src/runtime.rs
```diff
@@ -0,0 +1,537 @@
...
+}
+
+/// Runtime schedule validation context shared by execution and consensus refresh paths.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct UpgradeSignalRuntimeValidation {
+    /// Whether positive Beryl signals require an execution activation admin address.
+    pub require_activation_admin_for_beryl: bool,
+    /// Execution activation admin address for the L2 chain, when known.
+    pub activation_admin_address: Option<Address>,
+}
```
refcell: I don't think we need this. The admin activation address will be moved into L2 state space such that it's removed from the rollup config and validated in execution land.

crates/common/genesis/src/rollup.rs
```diff
@@ -182,78 +189,139 @@ macro_rules! rollup_fork_methods {
...
 
 impl RollupConfig {
+    /// Returns this rollup config's runtime-aware hardfork activation for a hardfork ID.
+    pub fn hardfork_activation(&self, hardfork_id: &str) -> Option<HardForkActivation> {
+        #[cfg(feature = "std")]
+        if let Some(activation) =
+            RuntimeHardForkRegistry::activation(self.l2_chain_id.id(), hardfork_id)
+        {
+            return Some(activation);
+        }
```
refcell: We can't gate this on `std` since this needs to be re-used in the proof system. The `RuntimeHardForkRegistry` needs to be `no_std`

crates/common/genesis/src/chain/mod.rs
```diff
@@ -4,7 +4,12 @@ mod addresses;
 pub use addresses::AddressList;
 
 mod hardfork;
-pub use hardfork::{HardForkConfig, HardforkConfig};
+#[cfg(feature = "std")]
```
refcell: rm this std feature flag

crates/common/genesis/src/lib.rs
```diff
@@ -28,7 +28,12 @@ pub use system::{
 };
 
 mod chain;
-pub use chain::{AddressList, HardForkConfig, HardforkConfig, Roles};
+#[cfg(feature = "std")]
```
refcell: ibid

refcell (dismissed):
Left a bunch of comments as a first pass. This PR has great structure. Most of the code is in the right place and is very clean. I'm slightly concerned with having upgrades denoted as strings but there might not be a nice way around this. We also need to be careful with gating logic behind `std` feature flags since that logic then won't apply for `no_std` proofs which will result in divergent behavior.

crates/utilities/upgrade-signal/README.md
PelleKrab: Oh that is part of pr4, let me pick that out.

crates/utilities/upgrade-signal/src/config.rs
PelleKrab: Yeah that was from testing, but I thought it would also be a good moat to stop user from accidentally turning this on. I do worry about peoples agents turning it dynamic runtime upgrades on, without their knowledge.

crates/common/genesis/src/lib.rs
PelleKrab: New latin phrase to use :)

crates/common/genesis/src/chain/upgrade.rs
PelleKrab: I think I made all the appropriate changes lemme know if you see any more.

crates/common/genesis/src/rollup.rs
```diff
@@ -129,29 +132,34 @@ impl EthereumHardforks for RollupConfig {
...
-                self.hardforks.isthmus_time,
-                self.hardforks.jovian_time,
+                self.hardfork_activation_timestamp("canyon"),
+                self.hardfork_activation_timestamp("ecotone"),
+                self.hardfork_activation_timestamp("fjord"),
+                self.hardfork_activation_timestamp("granite"),
+                self.hardfork_activation_timestamp("holocene"),
+                self.hardfork_activation_timestamp("isthmus"),
+                self.hardfork_activation_timestamp("jovian"),
             ])
```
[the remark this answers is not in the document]
PelleKrab: not in this pr

crates/execution/chainspec/src/spec.rs
```diff
@@ -285,6 +288,195 @@ impl BaseChainSpec {
...
+
+        if let Some(execution_hardfork) = contract_upgrade.execution_hardfork() {
+            hardforks.insert(execution_hardfork, condition);
+        }
+        if let Some(base_upgrade) = contract_upgrade.base_upgrade() {
+            hardforks.insert(base_upgrade, condition);
+        }
+
+        true
+    }
```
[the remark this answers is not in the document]
PelleKrab: will be addressed in future prs.

Ghostop:
nice

crates/common/chains/src/upgrade.rs
```diff
@@ -43,6 +46,65 @@ impl BaseUpgrade {
...
+            Self::Fjord => "fjord",
+            Self::Granite => "granite",
+            Self::Holocene => "holocene",
+            Self::Isthmus => "isthmus",
+            Self::Jovian => "jovian",
+            Self::Azul => "azul",
+            Self::Beryl => "beryl",
+            Self::Cobalt => "cobalt",
+        }
+    }
```
refcell: I don't think we need this. The alloy `hardfork!` macro looks like it provides a [`name` method](https://docs.rs/alloy-hardforks/latest/alloy_hardforks/trait.Hardfork.html#required-methods) that we can use.

crates/common/chains/src/upgrade.rs
```diff
@@ -43,6 +46,65 @@ impl BaseUpgrade {
...
+            "Granite" => Some(Self::Granite),
+            "Holocene" => Some(Self::Holocene),
+            "Prague" | "Isthmus" => Some(Self::Isthmus),
+            "Jovian" => Some(Self::Jovian),
+            "Osaka" | "Azul" => Some(Self::Azul),
+            "Beryl" => Some(Self::Beryl),
+            "Cobalt" => Some(Self::Cobalt),
+            _ => None,
+        }
+    }
```
refcell: I'm not a huge fan of this matching, especially since it's case sensitive. Let's at the minimum match `name.to_lowercase()` and also strip any whitespace so we have less of a casing footgun. Otherwise, I'm fine not having this be a blocker. Let's just add a `// TODO:` comment to remove this method later if possible.

crates/common/chains/src/upgrade.rs
```diff
@@ -143,11 +205,64 @@ impl BaseUpgrade {
...
+
+impl ContractUpgrade {
+    /// Returns the canonical contract upgrade ID.
+    pub const fn id(self) -> &'static str {
+        match self {
+            Self::Upgrade(upgrade) => upgrade.id(),
+            Self::Delta => "delta",
+            Self::PectraBlobSchedule => "pectra_blob_schedule",
+        }
+    }
```
refcell: This can extend alloy hardfork's `name` method instead

crates/common/chains/src/upgrades.rs
```diff
@@ -87,61 +87,47 @@ impl Upgrades for RollupConfig {
         match fork {
             BaseUpgrade::Bedrock => ForkCondition::Block(0),
             BaseUpgrade::Regolith => self
-                .hardforks
-                .regolith_time
+                .hardfork_activation_timestamp("regolith")
```
refcell: Can't we do `BaseUpgrade::Regolith`?

crates/common/genesis/src/chain/hardfork.rs
```diff
@@ -118,6 +304,155 @@ impl Display for HardForkConfig {
...
+        "fjord",
+        "granite",
+        "holocene",
+        "pectra_blob_schedule",
+        "isthmus",
+        "jovian",
+        "azul",
+        "beryl",
+        "cobalt",
+    ];
```
refcell: This leaves another place we need to add upgrade variants to whenever we add a new upgrade. If we want a list of upgrades, we should have a `BaseUpgrades::all()` list that collects all the hardfork/upgrade variants

crates/common/genesis/src/chain/hardfork.rs
```diff
@@ -118,6 +304,155 @@ impl Display for HardForkConfig {
...
+            "holocene" => Some("holocene"),
+            "pectrablobschedule" => Some("pectra_blob_schedule"),
+            "isthmus" => Some("isthmus"),
+            "jovian" => Some("jovian"),
+            "azul" | "baseazul" | "v1" => Some("azul"),
+            "beryl" | "baseberyl" | "v2" => Some("beryl"),
+            "cobalt" | "basecobalt" | "v3" => Some("cobalt"),
+            _ => None,
+        }
+    }
```
refcell: This should live on `BaseUpgrades`

crates/common/genesis/src/rollup.rs
```diff
@@ -182,78 +190,137 @@ macro_rules! rollup_fork_methods {
...
+            return Some(activation);
+        }
+
+        self.hardforks.activation(hardfork_id)
+    }
+
+    /// Returns this rollup config's runtime-aware activation timestamp for a hardfork ID.
+    pub fn hardfork_activation_timestamp(&self, hardfork_id: &str) -> Option<u64> {
+        self.hardfork_activation(hardfork_id).and_then(UpgradeActivation::timestamp)
+    }
```
refcell: I still don't really see why we need to be using strings here. We should be able to use the `BaseUprades` enum so we don't have string literals all over with different casing.

Using strings doesn't give us the same guarantees as using an enum since it's checked by the rust compiler at compile time to at least be a valid variant.

crates/utilities/upgrade-signal/src/config/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+mod schedule;
+pub use schedule::UpgradeSignalConfig;
+
+mod types;
+pub use types::{UpgradeSignalBlockTag, UpgradeSignalMode, UpgradeSignalStartupMode};
+
+/// Default values used by the upgrade signal reader and runtime applier.
+#[derive(Debug)]
+pub struct UpgradeSignalDefaults;
```
refcell, line 19: Nice, thank you

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -0,0 +1,164 @@
...
+
+/// Configuration for reading hardfork IDs from an L1 upgrade signal contract.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UpgradeSignalConfig {
+    /// L1 upgrade signal contract or proxy address.
+    pub contract_address: Address,
+    /// Hardfork IDs to pass to the contract.
+    pub hardfork_ids: Vec<String>,
+    /// Hardfork IDs allowed to mutate local fork schedules.
+    pub apply_hardfork_ids: Vec<String>,
```
refcell: I think we should only be storing the actual hardfork enum type, not strings.

crates/utilities/upgrade-signal/src/config/tests.rs
refcell, line 1: Let's inline these into the modules rather than having a standalone tests.rs module. As in, each `.rs` module file has a `#[cfg(test)] mod tests` at the bottom of the file containing its associated tests.

crates/utilities/upgrade-signal/src/runtime/tests.rs
refcell, line 1: Same here, break up into individual tests mods inside each module file - `refresher.rs`, `sink.rs`, etc

refcell (commented):
Starting to look very good. My main concern is the use of strings wired throughout upgrade matching since it's too easy to have inconsistent casing, typos, whitespace, etc

crates/utilities/upgrade-signal/src/contract.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Reads one hardfork signal using a previously observed L1 block ID.
+    pub async fn read_signal_at_l1_block(
+        &self,
+        hardfork_id: ContractUpgrade,
+        l1_block_number: u64,
+        l1_block: BlockId,
+    ) -> Result<UpgradeSignal, UpgradeSignalError> {
+        let (timestamp_output, version_output) = try_join(
+            self.call_at_block(
+                IUpgradeSignal::getTimestampCall { hardforkId: hardfork_id.to_string() },
```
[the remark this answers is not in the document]
PelleKrab: Will be fixed when we remove strings for hardforks.

crates/common/chains/src/upgrade.rs
```diff
@@ -1,4 +1,7 @@
-use alloy_hardforks::{ForkCondition, hardfork};
+use core::str::FromStr;
+
+use alloy_hardforks::{EthereumHardfork, ForkCondition, hardfork};
+pub use base_common_genesis::ContractUpgrade;
```
refcell: Let's just `use` this here.
```suggestion
use base_common_genesis::ContractUpgrade;
```

crates/common/genesis/src/chain/upgrade.rs
```diff
@@ -30,6 +37,313 @@ impl BaseUpgradeConfig {
     }
 }
 
+/// Contract-backed Base upgrade IDs.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub enum ContractUpgrade {
```
refcell: I still don't think we need / want this type. It's a duplicate of `BaseUpgrade` that I think you're not able to get around because of the dependency graph. Let me put up a PR into your branch to try to help out here.

refcell: https://github.com/base/base/pull/3620

refcell (approved):

PR #3489, opened 2026-06-12
feat(common): State Backed Activation Admin

refcell:
## Summary

This adds a Cobalt-gated state-backed admin slot to the activation registry while preserving the static chain-config admin as the Beryl replay fallback. The activation registry now supports setAdmin(address) once state-backed admin storage is enabled, and Cobalt execution reads the stored admin when present. The tests cover fallback behavior, admin rotation, Beryl rejection of state rotation, and the installed EVM precompile path.

### DIFF
--- a/crates/common/precompiles/src/activation/storage.rs
+++ b/crates/common/precompiles/src/activation/storage.rs
@@ -195,9 +195,6 @@ impl ActivationRegistryStorage<'_> {
         }
 
         let caller = self.storage.caller();
-        if caller.is_zero() {
-            return Err(BasePrecompileError::revert(IActivationRegistry::Unauthorized { caller }));
-        }
         let admin = self.authorized_admin(admin_config)?;
         if caller != admin {
             return Err(BasePrecompileError::revert(IActivationRegistry::Unauthorized { caller }));
--- a/crates/common/precompiles/src/activation/storage.rs
+++ b/crates/common/precompiles/src/activation/storage.rs
@@ -164,16 +164,12 @@ impl ActivationRegistryStorage<'_> {
             return Err(BasePrecompileError::revert(IActivationRegistry::ZeroAdminAddress {}));
         }
 
-        let caller = self.storage.caller();
-        let current_admin = self.authorized_admin(admin_config)?;
-        if caller != current_admin {
-            return Err(BasePrecompileError::revert(IActivationRegistry::Unauthorized { caller }));
-        }
+        let caller = self.require_admin_caller(admin_config)?;
 
         self.__initialize()?;
         self.admin.write(new_admin)?;
         self.emit_event(IActivationRegistry::AdminChanged {
-            previousAdmin: current_admin,
+            previousAdmin: caller,
             newAdmin: new_admin,
             caller,
         })?;
@@ -194,11 +190,7 @@ impl ActivationRegistryStorage<'_> {
             return Err(BasePrecompileError::revert(IActivationRegistry::StaticCallNotAllowed {}));
         }
 
-        let caller = self.storage.caller();
-        let admin = self.authorized_admin(admin_config)?;
-        if caller != admin {
-            return Err(BasePrecompileError::revert(IActivationRegistry::Unauthorized { caller }));
-        }
+        let caller = self.require_admin_caller(admin_config)?;
 
         let current_activated_state = self.features.at(&feature).read()?;
 
@@ -240,6 +232,16 @@ impl ActivationRegistryStorage<'_> {
             return Err(BasePrecompileError::revert(IActivationRegistry::Unauthorized { caller }));
         }
         Ok(admin)
+    }
+
+    /// Reverts unless the current caller is the effective activation admin.
+    pub fn require_admin_caller(&self, admin_config: ActivationAdminConfig) -> Result<Address> {
+        let caller = self.storage.caller();
+        let admin = self.authorized_admin(admin_config)?;
+        if caller != admin {
+            return Err(BasePrecompileError::revert(IActivationRegistry::Unauthorized { caller }));
+        }
+        Ok(caller)
     }
 }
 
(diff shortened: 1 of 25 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/activation/storage.rs
```diff
@@ -127,9 +198,7 @@ impl ActivationRegistryStorage<'_> {
         if caller.is_zero() {
             return Err(BasePrecompileError::revert(IActivationRegistry::Unauthorized { caller }));
         }
-        let Some(admin) = activation_admin_address else {
-            return Err(BasePrecompileError::revert(IActivationRegistry::Unauthorized { caller }));
-        };
+        let admin = self.authorized_admin(admin_config)?;
```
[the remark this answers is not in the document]
refcell: Removed the duplicate `caller.is_zero()` guard from `set_activated`; caller/admin validation now flows through `authorized_admin`, matching `set_admin`.

crates/common/precompiles/src/activation/storage.rs
```diff
@@ -93,29 +137,56 @@ impl ActivationRegistryStorage<'_> {
...
-        self.set_activated(feature, false, activation_admin_address)
+        if self.storage.is_static() {
+            return Err(BasePrecompileError::revert(IActivationRegistry::StaticCallNotAllowed {}));
+        }
+        if !admin_config.state_enabled {
+            return Err(BasePrecompileError::revert(
+                IActivationRegistry::AdminStorageNotEnabled {},
+            ));
+        }
+        if new_admin.is_zero() {
```
rayyan224, line 163: NIT: Can we have reusable function called require_admin_caller(&self, admin_config: ActivationAdminConfig). I think with that, we can condildate the smae logic for set_admin and set_activated

Sertug17:
This PR is the natural evolution of the admin hardening in #3344. The redundant caller.is_zero() guard in set_activated (bot finding #1) was originally added there as defense-in-depth when the admin came from chain-spec config. Now that authorized_admin centralizes the resolution (reading from storage with chain-spec fallback), the standalone is_zero check is genuinely dead code. Rayyan's suggestion to extract require_admin_caller would clean this up naturally both set_admin and set_activated would delegate authorization to the same helper, and the is_zero guard would live in exactly one place inside that helper rather than being duplicated.

rayyan224:
Can we have a test that 
Beryl never writes the admin slot. Assert stored_admin() == ZERO after a (rejected) setAdmin on Beryl.

crates/common/precompiles/src/activation/storage.rs
Sertug17: +1 this would also naturally absorb the redundant caller.is_zero() guard in set_activated that the bot flagged, since the check would live in exactly one place inside the helper.

rayyan224:
I think we should have a test that the old activation admin works, post beryl up untill the swap. For replay protection as there would be a delay here

Sertug17:
Good call. The critical window is between Cobalt activation and the first setAdmin call. During that gap, authorized_admin should fall back to the chain config admin since the storage slot is still ZERO. A test covering the full lifecycle would pin this: (1) Beryl-era activate with chain config admin succeeds, (2) Cobalt activates but no setAdmin yet chain config admin still authorized, (3) setAdmin rotates to new admin old chain config admin is now rejected, new admin succeeds. That covers the replay protection invariant: the chain config fallback is the authority until an explicit storage write supersedes it.

crates/common/precompiles/src/metrics.rs
```diff
@@ -421,6 +421,7 @@ impl BerylMetricLabels {
                 Cow::Borrowed("checkActivated")
             }
             Some(IActivationRegistry::adminCall::SELECTOR) => Cow::Borrowed("admin"),
+            Some(IActivationRegistry::setAdminCall::SELECTOR) => Cow::Borrowed("setAdmin"),
```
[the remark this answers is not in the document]
refcell: Fixed in b27d0c196: `AdminStorageNotEnabled` and `ZeroAdminAddress` now classify as `invalid_input` in `from_revert_bytes`, with unit coverage in `revert_bytes_are_classified`.

crates/common/precompiles/src/activation/storage.rs
```diff
@@ -93,46 +137,60 @@ impl ActivationRegistryStorage<'_> {
...
+            return Err(BasePrecompileError::revert(IActivationRegistry::ZeroAdminAddress {}));
+        }
+
```
eric-ships, line 166: Should we guard against the new admin being the same as the current admin?

```rust
if new_admin == self.admin(admin_config)? {
    return Ok(());  // or revert with a SameAdminAddress error
}
```

eric-ships (approved):

PR #3499, opened 2026-06-12
feat(load-tests): add standalone ERC20 transfer load test

rayyan224:
## What

Adds a load-test workload that drives **standard ERC20 `transfer(...)` traffic** against a real, deployed Solidity contract.

The `erc20` workload type already existed (it produced `transfer(to, amount)` calldata), but there was **no token setup** — every sender held a zero balance, so transfers reverted and the run measured reverts rather than transfers. This adds the missing setup phase, mirroring the existing B-20 flow.

## How

- **Setup phase** (`runner/erc20.rs`): `needs_erc20_setup()` + `setup_erc20_tokens()` deploy a standard OpenZeppelin ERC20 (`LoadTestERC20`) via a `CREATE` tx (or reuse a configured address), mint `erc20_mint_amount` to every sender, resolve the address back into the workload, and rebuild the generator.
- **Contract**: `LoadTestERC20.sol` — OpenZeppelin `ERC20` with an open `mint` and a no-arg constructor. Bytecode is sourced via `base-test-utils` `sol!` bindings (`LoadTestERC20::BYTECODE`), matching the repo's existing convention (no committed build artifacts).
- **Config**: ERC20 `contract` is now optional (auto-deploy when omitted); new `erc20_mint_amount` knob; new `examples/erc20-devnet.yaml`.
- **Ergonomics**: `just load-test run` builds the forge artifacts if missing, so the command stays self-sufficient.

Senders transfer in a ring (`i -> i+1`), so every account both sends and receives — balances circulate and don't drain across a run.

## Testing

- `cargo test -p base-load-tests` — all pass, incl. new config round-trip tests (optional/explicit `contract`, mint-amount default), a guard that the shipped example parses, and a deploy-bytecode-nonempty check.
- `cargo clippy -p base-load-tests -p base-load-tester-bin -p base-test-utils --all-targets` — clean.
- Manual end-to-end run against a live network still pending (requires a funded `FUNDER_KEY`): `just load-test run erc20-devnet`.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Sertug17:
*@rayyan224* nice PR, the ERC20 workload fills a gap in the load test suite. One thing I noticed about the ring transfer pattern that might be worth thinking about:

> In `setup_erc20_tokens`, tokens are minted to all senders in a batch. If the batch fails partway through (say sender 50/100 gets a revert due to gas or a pre-deployed contract's mint cap), the first 49 senders hold tokens but the last 51 don't. The ring pattern (i -> i+1) then enters an inconsistent state: sender 49's `transfer(sender_50, …)` succeeds, but sender 50's `transfer(sender_51, …)` fails because sender 50 has no balance. The result is intermittent failures that look like network drops rather than a setup asymmetry, making debugging harder.

> *Two suggestions:*
> (a) Track mint failures by index and exclude those senders from the generator, so the ring only includes successfully funded accounts.
> (b) At minimum, log the number of successful mints vs configured senders before entering the ring, so a mismatch is visible in the run output without digging through individual tx traces.

> Separately, the pre deployed contract case (`contract: Some(…)`) mints on every `run()` call. If the same contract is reused across runs, balances snowball sender balances increase by `erc20_mint_amount` each run, which changes the load profile over time. Consider either (a) skipping the mint when the address is pre-configured (user manages balances externally), or (b) documenting that repeated runs against the same contract require resetting state.

Sertug17:
@rayyan224 thanks for the quick iteration the `Option<Address>` pattern, pre-flight guard, and mint-failure tracking are good additions. A few remaining items:

**Must fix:**

1. **`flashblocks_ws` placeholder in `erc20-devnet.yaml`** `ws://<fb-ws-url>/ws` uses angle brackets which are forbidden host code points per the WHATWG URL spec. Since `flashblocks_ws` is typed as `Option<Url>`, `to_load_config()` will fail when it tries to parse this URL, breaking the `shipped_erc20_devnet_example_parses` test. Should be `ws://localhost:7111` to match `devnet.yaml`.

2. **`base-test-utils` as a runtime dependency**  this crate is named "test utilities" but added under `[dependencies]`, pulling alloy contract/signer/genesis transitive deps into the production binary. Only `LoadTestERC20::BYTECODE` is used consider inlining the `sol!` binding directly in `base-load-tests`, or splitting contract bindings into a separate non-test-named crate.

**Consider:**

3. **Pre deployed contract mint behavior** When `contract: Some(addr)` is provided, `setup_erc20_tokens()` still calls `mint()` on the user supplied address. An arbitrary ERC20 may not have an open `mint(address,uint256)`, causing all mints to revert with a confusing error. Additionally, if the same contract is reused across runs, balances snowball. Consider either (a) skipping the mint phase when a contract address is pre-configured (the user is responsible for balances), or (b) adding a single test mint `eth_call` before the batch to verify the contract supports minting, with a clear error message otherwise.

If you'd like, I'm happy to pick up any of these items. Just say the word and I'll open a follow up PR to fix them.

wlawt (approved):

PR #3511, opened 2026-06-12
fix(precompile-macros): handle binary/octal/suffixed integer literals in parse_slot_value

eric-ships:
[BOP-388](https://linear.app/coinbase/issue/BOP-388)

## Motivation

`parse_slot_value` called `int.to_string()` on a `LitInt`, which preserves Rust base prefixes (`0b`, `0o`) and type suffixes (`u64`, `usize`, etc.). The function then only checked for a `0x` prefix before passing the string to `U256::from_str_radix` with radix 10. This caused valid Rust integer literals like `0b101`, `0o17`, and `42u64` to be incorrectly rejected with a misleading "Invalid slot number" error.

Cantina finding: https://cantina.xyz/code/c75e122d-bfec-4090-8ae7-2f591169e2b1/findings/36

## What changed

- Strip any Rust integer type suffix (`u8`, `u16`, `u32`, `u64`, `u128`, `usize`, `i8`, `i16`, `i32`, `i64`, `i128`, `isize`) from the raw `to_string()` output before inspecting the prefix.
- Dispatch to the correct `from_str_radix` radix based on prefix: `0x`/`0X` hex (16), `0b`/`0B` binary (2), `0o`/`0O` octal (8).
- Filter underscores before calling `from_str_radix` (ruint does not accept them in non-decimal inputs).
- Use `base10_digits()` for the decimal fallback, which already strips suffixes and underscore separators.
- Update the doc comment on `parse_slot_value` to list all four supported bases.
- Add unit tests for `0b101`, `0o17`, `42u64`, `0x2au64`, plain `42`, and `0x2a`.

An alternative approach was to use `int.base10_digits()` for decimal and call `syn`'s own `LitInt::suffix()` to get the suffix, then inspect the prefix separately. However, `LitInt` does not expose `suffix()` in a way that makes prefix detection clean for non-decimal bases. The approach of stripping known suffixes from the raw `to_string()` result and dispatching on prefix is explicit and covers all cases without relying on undocumented syn internals.

### DIFF
--- a/crates/common/precompile-macros/src/utils.rs
+++ b/crates/common/precompile-macros/src/utils.rs
@@ -413,4 +413,28 @@ mod tests {
         let lit: Lit = parse_quote!(0x2au64);
         assert_eq!(parse_slot_value(&lit).unwrap(), U256::from(42u64));
     }
+
+    #[test]
+    fn test_parse_slot_value_hex_with_underscores() {
+        let lit: Lit = parse_quote!(0xFF_FF);
+        assert_eq!(parse_slot_value(&lit).unwrap(), U256::from(0xFFFFu64));
+    }
+
+    #[test]
+    fn test_parse_slot_value_binary_with_underscores() {
+        let lit: Lit = parse_quote!(0b1010_0101);
+        assert_eq!(parse_slot_value(&lit).unwrap(), U256::from(0b10100101u64));
+    }
+
+    #[test]
+    fn test_parse_slot_value_decimal_with_underscores() {
+        let lit: Lit = parse_quote!(1_000);
+        assert_eq!(parse_slot_value(&lit).unwrap(), U256::from(1000u64));
+    }
+
+    #[test]
+    fn test_parse_slot_value_octal_with_underscores() {
+        let lit: Lit = parse_quote!(0o7_7);
+        assert_eq!(parse_slot_value(&lit).unwrap(), U256::from(0o77u64));
+    }
 }
--- a/crates/common/precompile-macros/src/utils.rs
+++ b/crates/common/precompile-macros/src/utils.rs
@@ -16,19 +16,47 @@ type ExtractedAttributes = (Option<U256>, Option<U256>, Option<NamespaceInfo>);
 /// Parses a slot value from a literal.
 ///
 /// Supports:
-/// - Integer literals: decimal (`42`) or hexadecimal (`0x2a`)
+/// - Integer literals in any Rust base:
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

eric-ships:
Added tests for underscore-separated literals in 1b7a4e163:

- `0xFF_FF` → `0xFFFF` (hex)
- `0b1010_0101` → `0b10100101` (binary)
- `0o7_7` → `0o77` (octal)
- `1_000` → `1000` (decimal)

Sertug17:
@eric-ships nice fix the suffix stripping + prefix dispatch approach is clean. I noticed the Clippy and Action Tests CI runs are failing. Looking at the code, a couple of things that might be causing them:

Potential Clippy issues:
1. unwrap_or(raw.as_str()) raw is String, .as_str() deref is unnecessary here since strip_suffix already borrows from raw. Could trigger clippy::needless_borrow.
2. Test assertions like U256::from(42u64) the explicit u64 suffix is redundant when 42 already coerces through U256::from. In U256::from(0b10100101u64), the binary literal with u64 suffix might also cause a parsing ambiguity with parse_quote!.

Action Tests failure could be related to the test compilation if parse_quote!(0b10100101u64) parses the binary + suffix combination differently than expected.

If you'd like, I'm happy to open a followup PR to fix these just let me know.

PR #3513, opened 2026-06-12
fix(policy-registry): enforce batch size cap before ABI decode (BOP-391/PSRC-29)

eric-ships:
## Bug

`createPolicyWithAccounts`, `updateAllowlist`, and `updateBlocklist` each carry an `address[]` parameter that is subject to a 64-account cap (`MAX_ACCOUNTS_PER_BATCH`), but the cap was only enforced after a full ABI decode that allocates the entire `Vec<Address>` on the heap. A caller could therefore trigger heap allocation of an arbitrarily large address array before receiving `BatchSizeTooLarge`.

## Fix

Added `check_batch_calldata_length` in `dispatch.rs` that reads the raw 32-byte array-length word at calldata offset `[100..132]` (4-byte selector + 3 x 32-byte head words) before any ABI decode occurs. If the encoded length exceeds `MAX_ACCOUNTS_PER_BATCH` (64), the function returns `BatchSizeTooLarge` immediately without allocating. The check is applied in `inner()` for all three affected selectors via a single selector match.

Also promotes `MAX_ACCOUNTS_PER_BATCH` to `pub` so it is accessible from dispatch and test modules without `pub(crate)` workarounds.

## References

- Cantina finding: https://cantina.xyz/code/73a6cf79-adb2-4d94-9e8b-3a5fdef7f64e/findings/29
- Linear ticket: https://linear.app/coinbase/issue/BOP-391

### DIFF
--- a/crates/common/precompiles/src/policy/dispatch.rs
+++ b/crates/common/precompiles/src/policy/dispatch.rs
@@ -1,6 +1,6 @@
-use alloy_primitives::Bytes;
+use alloy_primitives::{Bytes, U256};
 use alloy_sol_types::SolCall;
-use base_precompile_storage::{IntoPrecompileResult, StorageCtx};
+use base_precompile_storage::{BasePrecompileError, IntoPrecompileResult, StorageCtx};
 use revm::precompile::PrecompileResult;
 
 use super::{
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Sertug17:
> Clean fix @eric-ships. The offset calculation at 100..132 is correct for all three selectors given the shared param layout (selector + 3 × 32-byte head words + array length). The pre-decode guard eliminates the OOM vector cleanly.
> 
>Minor observation: the calldata.get(..4) None branch returns Ok(()) silently a sub-4-byte call passes through to the ABI decoder, which will produce a different error (likely a decode failure). Not a security concern since no allocation happens at that point, but worth noting.

eric-ships:
Good catch on the sub-4-byte path. That branch is intentional — a call shorter than 4 bytes can't match any ABI selector, so returning `Ok(())` is safe: the precompile dispatcher will either not invoke the handler at all (no selector match) or the subsequent ABI decode will produce the expected error. No allocation happens before that point regardless, which is what the guard is protecting against.

Sertug17:
---

*Thanks for the explanation @eric-ships the sub-4-byte path makes sense, no allocation before that point so `Ok(())` is safe.*

*A couple of observations while studying the code:*

*1. check_batch_calldata_length runs on every dispatch*

Currently the guard is called at the top of `inner()`, before the selector match. This means every call to the policy registry regardless of selector pays the cost of reading the first 4 bytes and doing three selector comparisons. For non-batch selectors (which are the common case), this is wasted work.

A slightly cleaner approach would be to inline the check inside the three match arms in `inner()`:

```rust
C::createPolicyWithAccounts(call) => {
    Self::check_batch_calldata_length(calldata)?;
    // ... existing logic
}
```

This makes it self-documenting that only those three selectors are gated, and removes the overhead entirely.

*2. U256 comparison for a small constant*

Since `MAX_ACCOUNTS_PER_BATCH = 64`, the comparison `U256::from_be_slice(len_word) > U256::from(64)` constructs two U256 values on the stack for what is effectively a single-byte comparison. A simpler check:

```rust
if len_word[..31].iter().any(|&b| b != 0) || len_word[31] > Self::MAX_ACCOUNTS_PER_BATCH as u8 {
```

This avoids U256 construction entirely. (Minor point not blocking.)

*3. Edge case test coverage*

Currently no tests cover:
- A batch selector with truncated calldata (< 132 bytes) falls through to ABI decode with a generic error
- Sub-4-byte calldata with bytes that happen to start matching a batch selector prefix

Some focused edge case tests would guard against regressions in the early return paths.

Overall the fix is clean and addresses the Cantina finding well. Just some thoughts for hardening.

---

eric-ships:
Thanks for the detailed review. Addressing each point:

**1. check_batch_calldata_length placement** — good point. Moving it inside the three match arms makes the gate self-documenting and removes the per-call overhead for non-batch selectors. Will refactor.

**2. U256 for MAX_ACCOUNTS_PER_BATCH comparison** — agreed, the byte-slice check is cleaner. Will switch to `len_word[..31].iter().any(|&b| b != 0) || len_word[31] > Self::MAX_ACCOUNTS_PER_BATCH as u8`.

**3. Edge case test coverage** — will add tests for truncated calldata (< 132 bytes) on batch selectors and sub-4-byte calldata paths.

Will push a follow-up commit.

Sertug17:
@eric-ships I looked more carefully and found what looks like an off by 4 bug rather than an offset-manipulation issue.

    The code reads the array length from 100..132, but if the dynamic offset is 0x60 (96) as the standard ABI encoding produces, the actual length word lives at 96..128. The 100..132 range reads:

    - bytes[100..128] = last 28 bytes of the length word (all zeros for values ≤ 2^32)
    - bytes[128..132] = first 4 bytes of the first array element, not the length at all

    This means the check's U256::from_be_slice(len_word) is not comparing the array length against MAX_ACCOUNTS_PER_BATCH (64) it's comparing a value formed from the tail of the length word plus the first address bytes. For any reasonable array (addresses starting with 0x00...), this value will be well under 64, so the guard always passes, even for oversized arrays.

    The fix should read from the actual offset location, or at minimum use the correct constant 96..128 instead of 100..132. This also means the test dispatch_batch_selectors_reject_oversized_length_before_decode at line ~228 isn't testing what it thinks it writes the oversized byte at calldata[131], which is past the actual length word location.

Sertug17:
@eric-ships My apologies I need to walk back the off-by-4 claim. The offset [100..132] is correct. I was misreading the ABI encoding: the dynamic offset 0x60 (96) is relative to the start of the tuple (calldata[4]), not calldata[0], so 4 + 96 = 100 puts the array length at [100..132]. The fix and the test are both right.

    For the other two points (placement and comparison), you mentioned you'd address them. Are those still planned?

Issue #3274, opened 2026-06-06
bug(flashblocks): receipts logs data mismatch between builder and execution


0xsinhvu:
I've already raised this on Discord. There appears to be a consistency issue where replaying transactions from a flashblock produces different receipt/execution results than those observed in the final block published about two seconds later, despite the transaction ordering remaining unchanged
Node version: `>=0.9.0`
Ref:  https://discord.com/channels/1067165013397213286/1511214452958036140

ptinnng3:
Yeah, this logs inconsistency in flashblocks receipts looks like a real bug

ron137:
This is downstream of [#300](https://github.com/base/base/pull/300). Since receipts were dropped from the flashblock payload (v0.3.0), every consumer rebuilds logs/receipts by re-executing the flashblock txs locally, and that re-execution doesn't reproduce what the builder did.

On a follower node, comparing local execution against the diff's own gas_used/receipts_root at each flashblock boundary:

- tx set + order, parent hash, basefee, timestamp, and spec id all match the confirmed block
- ~23% of flashblocks still diverge, and rebuilding from the confirmed parent reproduces it, so it's not a stale parent, a cache, or a timing artifact It splits into two classes:

1. Gas-only, concentrated on refund-heavy txs - repeatedly the ERC-4337 EntryPoint (0x0000000071727De2…032, v0.7 handleOps) and aggregators like 1inch. The consumer computes ~8–14k more gas than the builder, logs byte-identical. Breaks the checksum but harmless to log consumers.
2. Content-drift on swaps - same pool address and topics, different data (sqrtPrice/amounts). logs_bloom only covers address+topics, so consumers can't even detect this, and anything serving pendingLogs hands out a swap that never lands on-chain. This is the one that actually bites.

The "more gas, only on refund-heavy txs" asymmetry looks like a gas-refund / warm-cold (or operator-fee) accounting difference between op-rbuilder's per-flashblock execution and a plain re-execution — the two paths aren't byte-identical for a narrow set of tx classes.

There's no real consumer-side fix while receipts are off the wire, and content-drift isn't even detectable with the current checksums. Putting the sequencer receipts back in the payload (revert #300) removes the re-execution path entirely. If that's not on the table, it'd help to know whether op-rbuilder disables per-flashblock state roots or uses a different refund/operator-fee path than canonical execution - that's the most likely seam.

Same root cause as #2790, from the trading side.

Sertug17:
[...]

PR #3521, opened 2026-06-13
feat(proposer): enable attaching TEE proofs to existing dispute games

leopoldjoy:
This PR enhances the proposer by enabling it to dynamically attach fresh TEE proofs to already-existing matching dispute games via the `AggregateVerifier.verifyProposalProof` method. This improvement ensures that the proposer can effectively handle scenarios where a corresponding game already exists, transitioning from solely treating the `GameAlreadyExists` condition as a recovery signal.

### Key Changes

- Implemented the logic for attaching new TEE proofs to existing games.
- Updated unit tests to validate the new functionality, ensuring all proposed proofs are correctly processed.
- Verified configurations and improved error handling for signer validation.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Sertug17:
Nice work @leopoldjoy. The `attach_existing_game_proof` path is cleanly structured.

One observation: on the L1 head mismatch branch (line 373), you return `Err(SubmitAction::GameAlreadyExists)`. But this isn't really a "game already exists" case the game exists with a *different* L1 head, which suggests the proposer's chain state has diverged from the game's anchor. Returning `GameAlreadyExists` here means the caller's existing recovery logic handles it, but the classification is misleading. Consider a separate `SubmitAction::L1HeadMismatch` variant so callers can distinguish "duplicate submission" from "chain state divergence" they may warrant different recovery strategies (e.g. the latter might need a reorg check).

Also curious: any thoughts on the flashblocks execution divergence fix I've been working on (#3274)? The approach wires `FlashblockAccessList` from builder to consumer to warm the DB cache before re-execution. Would appreciate your perspective if you have time (#3525).

Sertug17:
---

"Nice progress @leopoldjoy the test coverage for AlreadyProven classification looks good now.

One edge case I'm thinking about in the attach_existing_game_proof flow: the L1 head mismatch returns SubmitAction::GameAlreadyExists, which is the same error used when a game with identical parameters already exists. But in the L1 head mismatch case, the existing game will never go away every retry will find the same game and hit the same mismatch. If the recovery path treats GameAlreadyExists as a transient condition and retries submit() on the next tick, the proposer enters an infinite retry loop where it always finds the same game, always fails the L1 head check, and never progresses to newer proposals.

A separate SubmitAction::L1HeadMismatch variant (as noted above) would let the recovery logic distinguish 'duplicate submission, skip to next proposal' from 'chain state divergence, investigate before proceeding.' Alternatively, even a discard action would break the loop the proposer would drop the current proposal and move on to the next target block, which is strictly better than retrying forever.

Separately, the make_submitter test helper uses game_type: 0 while the existing-game tests use TEST_GAME_TYPE: 42. Not a bug, but worth documenting why if someone adds a game-type-specific test using make_submitter, they'll never find an existing game."

---

crates/proof/proposer/src/proof_submitter.rs
```diff
@@ -317,6 +344,85 @@ where
...
+                    target_block,
+                    game_address = %game_address,
+                    "Proof signer is invalid on-chain, discarding proof to re-prove"
+                );
+                Metrics::tee_signer_invalid_total().increment(1);
+                Err(SubmitAction::Discard(e))
+            }
+            Err(e) => {
+                attach_timer.disarm();
+                Err(SubmitAction::Failed(e))
```
jackchuma, line 436: Handle stale L1 origins when attaching proofs. When `verifyProposalProof` returns `L1OriginTooOld` for an existing matching game with no TEE proof, this catch-all maps it to `SubmitAction::Failed`. The collector then restarts and retries the same expired proof, unlike the create path which treats the same contract error as a discard/re-prove condition, so this race can leave the proposer stuck retrying an unattachable proof instead of recovering with fresh state/proof data

crates/proof/proposer/src/proof_submitter.rs
```diff
@@ -317,6 +344,85 @@ where
...
+        })?;
+        if tee_prover != Address::ZERO {
+            info!(
+                target_block,
+                game_address = %game_address,
+                tee_prover = %tee_prover,
+                "Existing dispute game already has a TEE proof"
+            );
+            return Ok(());
+        }
```
jackchuma: `tee_prover` preflight duplicates the `ProofAlreadyVerified` handling at L399. Call `verify_proposal_proof` directly; delete the success-precheck test and `tee_prover_map`

crates/proof/proposer/src/proof_submitter.rs
Sertug17: ---

"Good points @jackchuma. On the L1OriginTooOld catch-all (#1): this is the same class of issue as the L1 head mismatch I noted earlier routing distinct failure modes through SubmitAction::Failed means the recovery path can't distinguish 'retry with fresh state' from 'this game is permanently unattachable'. A separate action variant (or at minimum, mapping L1OriginTooOld to Discard in the attach path, matching the create path's behavior) would prevent the stuck-retry loop.

On the tee_prover preflight (#2): agreed it's redundant now that ProofAlreadyVerified is handled in the match. Removing it also eliminates the extra RPC call on the happy path."

---

Sertug17:
Follow up from our discussion. Opened #3551 covering the stuck-retry loop scenarios (L1HeadMismatch -> GameAlreadyExists, L1OriginTooOld -> Failed) and the redundant tee_prover preflight.

@jackchuma @leopoldjoy I filed this separately since these are pre-existing logic bugs in the attach path that can stall the proposer. The fix patterns we discussed (SubmitAction::L1HeadMismatch variant, mapping L1OriginTooOld -> Discard, removing the preflight) should resolve all three.

Sertug17:
Thanks for addressing the L1OriginTooOld path. The L1HeadMismatch → GameAlreadyExists is the remaining one still an infinite retry risk if the game's L1 head never matches. Any plans to add a SubmitAction::L1HeadMismatch variant or map it to Discard?

Sertug17:
@leopoldjoy @jackchuma quick follow-up on the remaining L1HeadMismatch → GameAlreadyExists this is the last stuck retry vector.

    Problem: attach_existing_game_proof returns Err(SubmitAction::GameAlreadyExists) when the existing game's L1 head doesn't match the proposer's l1_head. The recovery path treats GameAlreadyExists as transient and retries submit() → finds the same game → same mismatch → infinite loop. The proposer never advances to newer proposals.

    Fix (minimal): Map to SubmitAction::Discard instead. Discard already exists as a variant on SubmitAction — no enum changes needed. Rough diff:

    rust
    // In attach_existing_game_proof, L1 head mismatch branch:
    // Before:
    if game_l1_head != l1_head {
        warn!(
            target_block,
            game_address = %game_address,
            game_l1_head = %game_l1_head,
            "Existing dispute game has a different L1 head than the proof; cannot attach"
        );
        return Err(SubmitAction::GameAlreadyExists);
    }

    // After:
    if game_l1_head != l1_head {
        warn!(
            target_block,
            game_address = %game_address,
            game_l1_head = %game_l1_head,
            "Existing dispute game has a different L1 head than the proof; discarding proposal"
        );
        return Err(SubmitAction::Discard(ProposerError::L1HeadMismatch {
            expected: l1_head,
            actual: game_l1_head,
        }));
    }


    (If ProposerError::L1HeadMismatch doesn't exist as a variant, a string-based Custom/Other works too — the key change is the action variant, not the inner error type.)

    This is the only remaining item from #3551 since you already handled L1OriginTooOld → Failed cleanup and the tee_prover preflight removal nice work on those.

crates/proof/proposer/src/proof_submitter.rs
leopoldjoy: Done: https://github.com/base/base/pull/3521/commits/dac85ac65249a39e3282db66987ed855ad9905ae

crates/proof/proposer/src/proof_submitter.rs
leopoldjoy: https://github.com/base/base/pull/3521/commits/dac85ac65249a39e3282db66987ed855ad9905ae

crates/proof/proposer/src/proof_submitter.rs
```diff
@@ -262,6 +272,23 @@ where
...
+            .map_err(|e| {
+                SubmitAction::Failed(ProposerError::Contract(format!(
+                    "matching game lookup failed: {e}"
+                )))
+            })?;
+
+        if existing_game != Address::ZERO {
+            return self
+                .attach_existing_game_proof(existing_game, aggregate_proposal, target_block)
+                .await;
```
jackchuma, line 289: Handle post-lookup GameAlreadyExists races. This only attaches when the matching game exists before the preflight `games()` call. If `games()` returns zero and another submitter creates the same game before our `createWithInitData` transaction lands, the code falls into the `GameAlreadyExists` branch below, which just recovers and advances; when the racing game was initialized with a ZK proof, the collected TEE proof is never attached and the game remains ZK-only. Re-looking up the game and calling `attach_existing_game_proof` on `GameAlreadyExists` would close this race

crates/proof/proposer/src/proof_submitter.rs
```diff
@@ -442,3 +543,211 @@ where
...
+        let result = submitter
+            .submit(&proof_result(TEST_BLOCK_INTERVAL), TEST_BLOCK_INTERVAL, Address::ZERO)
+            .await;
+
+        assert!(result.is_ok());
+        assert_eq!(*output.created.lock().unwrap(), 0);
+        assert_eq!(*output.verified.lock().unwrap(), vec![game_address]);
+    }
+
+    #[tokio::test]
```
jackchuma, line 729: shrink: three copy-pasted existing-game attach result tests. One `#[rstest]` over `ProofAlreadyVerified`, `L1OriginTooOld`, and `InvalidSigner`; keep only the L1-head mismatch test separate

crates/proof/proposer/src/output_proposer.rs
```diff
@@ -415,6 +497,32 @@ mod tests {
         ExpectedClassification::GameAlreadyExists,
         "raw data contains selector"
     )]
+    #[case::rpc_with_already_proven_selector_hex(
```
jackchuma: shrink: four `AlreadyProven` classifier cases repeat the transport matrix already covered by other errors. Keep one revert-data case and one RPC-string case

crates/proof/proposer/src/output_proposer.rs
```diff
@@ -547,6 +659,14 @@ mod tests {
         assert_eq!(err.is_game_already_exists(), expected);
     }
 
+    #[rstest]
```
jackchuma: delete: direct test for `is_proof_already_verified()` just repeats a one-line `matches!` wrapper. Existing attach-path test covers the only behavior that matters

jackchuma (approved):
Left some minor nits that can be addressed in a follow up

crates/proof/proposer/src/output_proposer.rs
leopoldjoy: Done: https://github.com/base/base/pull/3521/commits/c0051bd1d9d597ca61ccebca91f16ba33c34721f

crates/proof/proposer/src/output_proposer.rs
leopoldjoy: https://github.com/base/base/commit/c0051bd1d9d597ca61ccebca91f16ba33c34721f

crates/proof/proposer/src/proof_submitter.rs
leopoldjoy: https://github.com/base/base/commit/c0051bd1d9d597ca61ccebca91f16ba33c34721f

crates/proof/proposer/src/proof_submitter.rs
leopoldjoy: https://github.com/base/base/commit/c0051bd1d9d597ca61ccebca91f16ba33c34721f

jackchuma (approved):

Issue #3299, opened 2026-06-08
precompile storage: HashMapStorageProvider::checkpoint_revert does not restore gas_refunded, leaving stale refunds after rollback


Sertug17:
In `crates/common/precompile-storage/src/hashmap.rs`, `Snapshot` captures `internals` and `events` but not `gas_refunded`:

```rust
struct Snapshot {
    internals: HashMap<(Address, U256), U256>,
    events: HashMap<Address, Vec<LogData>>,
    // gas_refunded missing
}
```

`checkpoint_revert` restores `internals` and `events` from the snapshot, but `self.gas_refunded` is not touched:

```rust
fn checkpoint_revert(&mut self, checkpoint: JournalCheckpoint) {
    if let Some(snapshot) = self.snapshots.drain(checkpoint.journal_i..).next() {
        self.internals = snapshot.internals;
        self.events = snapshot.events;
        // gas_refunded is not restored
    }
}
```

`gas_refunded` was added in PR #3258. A test sequence of checkpoint → sstore(nonzero→zero) → revert leaves `gas_refunded` with a stale accumulated refund even though the storage write was rolled back. This would cause tests that rely on checkpoint_revert to observe incorrect refund values.

Fix: add `gas_refunded: i64` to `Snapshot` and restore it in `checkpoint_revert`.

This was flagged as a latent gap in the #3258 review.

Sertug17:
I'd like to take this on. The fix is straightforward and mirrors the pattern already used by revm's journal in `EvmPrecompileStorageProvider`:

1. Add `gas_refunded: i64` to the `Snapshot` struct at `hashmap.rs:38`
2. Capture `self.gas_refunded` in `checkpoint()` (line 218): `Snapshot { internals: ..., events: ..., gas_refunded: self.gas_refunded }`
3. Restore it in `checkpoint_revert()` after restoring `internals` and `events`

This was flagged by the bot in the #3258 review and applies directly to test correctness any checkpoint → sstore(nonzero→zero) → revert sequence would otherwise leave a stale refund in the counter.

Could a maintainer assign this to me and vouch so I can open the PR? Happy to walk through the approach first if useful @eric-ships @refcell.

vsrocoo:
checkpoint_revert should also restore gas_refunded to avoid stale values

Sertug17:
"Fixed by #3534"

PR #3534, opened 2026-06-14
feat(eip8130): AccountConfiguration storage reader (state layer)

chunter-cb:
## Summary

Adds `base-execution-eip8130-state` — a native, read-only mirror of the EIP-8130 `AccountConfiguration` system contract's storage. This is the **state layer** of the validation pipeline: where #3467 does the stateless *Authenticate* step (verify signature → resolve `actorId`), this reads the on-chain account configuration the stateful *Authorize* step needs.

## What it does

Models the deployed contract's storage layout — plain sequential slots (no ERC-7201 namespace), with `account` as the inner mapping key (the contract's ERC-7562 requirement):

```solidity
mapping(bytes32 actorId => mapping(address account => ActorConfig)) _actorConfig;     // slot 0
mapping(bytes32 actorId => mapping(address account => bytes32))     _policyCommitment; // slot 1
mapping(bytes32 actorId => mapping(address account => address))     _policyManager;    // slot 2
mapping(address account => AccountState)                            _accountState;     // slot 3
```

…using the same `base-precompile-storage` `#[contract]` + `Mapping` abstraction the `NonceManager` precompile uses, so the reader runs over **any** `PrecompileStorageProvider`: `HashMapStorageProvider` (tests), the live EVM journal (inclusion), and later a `StateProvider` adapter (mempool admission) — the shared mempool/execution reader that doesn't exist yet.

Exposes the contract's storage views natively, mirroring each 1:1:
- `get_actor_config` → `ActorConfig { authenticator, scope, expiry, policy_type }`
- `is_actor` (incl. the implicit-EOA self-actor rule), `get_policy`
- `get_change_sequences`, `is_initialized`, `is_locked`, `get_lock_status`

## Packing fidelity

`ActorConfig` (`address,uint8,uint48,uint8`) and `AccountState` (`uint64,uint64,uint40,uint16`) each pack into one slot. Because `base-precompile-storage` has no `uint48`/`uint40` primitive, the packed slot is read as a raw `U256` and unpacked manually (documented byte offsets). Tests seed slots via **independent bit-offset packing** (`field << bitOffset`) and assert the byte-offset unpacker recovers each field — a real cross-check of the layout against Solidity, not a self-consistent round-trip.

## Scope (what this PR is not)

- No *Authorize* branching logic yet (implicit-EOA / explicit / delegate discharge against state) — that's the next PR, consuming this + #3467's `DispatchOutcome`.
- No mempool / EVM wiring yet.

## Test plan

- [x] `cargo test -p base-execution-eip8130-state` — 7 tests: per-field slot-position unpacking for `ActorConfig` and `AccountState`, empty/absent reads, the `is_actor` predicate (bound / revoked / implicit-EOA / non-self), `get_policy` gated vs ungated, lock-status states (locked / unlock-initiated / never-locked), and `self_actor_id` left-alignment.
- [x] `cargo clippy -p base-execution-eip8130-state --all-targets --all-features -- -D warnings` clean
- [x] `cargo fmt -- --check` clean

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4509,6 +4509,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "base-execution-eip8130-state"
+version = "0.0.0"
+dependencies = [
+ "alloy-primitives",
+ "base-common-consensus",
+ "base-precompile-macros",
+ "base-precompile-storage",
+ "revm",
+]
+
+[[package]]
 name = "base-execution-evm"
 version = "0.0.0"
 dependencies = [
--- a/crates/execution/eip8130-state/Cargo.toml
+++ b/crates/execution/eip8130-state/Cargo.toml
@@ -1,0 +1,38 @@
+[package]
+name = "base-execution-eip8130-state"
+description = "EIP-8130 AccountConfiguration system-contract storage reader (native)"
+
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+exclude.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+# alloy
+alloy-primitives.workspace = true
+
+# base
+base-precompile-macros.workspace = true
+base-precompile-storage = { workspace = true }
+base-common-consensus = { workspace = true, default-features = false }
+
+# revm
+revm.workspace = true
+
+[dev-dependencies]
+base-precompile-storage = { workspace = true, features = ["test-utils"] }
+
+[features]
+default = [ "std" ]
+std = [
+	"alloy-primitives/std",
+	"base-precompile-storage/std",
+	"revm/std",
+]
+test-utils = [ "base-precompile-storage/test-utils" ]
--- a/crates/execution/eip8130-state/Cargo.toml
+++ b/crates/execution/eip8130-state/Cargo.toml
@@ -19,7 +19,7 @@ alloy-primitives.workspace = true
 
 # base
 base-precompile-macros.workspace = true
-base-precompile-storage = { workspace = true }
+base-precompile-storage.workspace = true
 base-common-consensus = { workspace = true, default-features = false }
 
 # revm
--- a/crates/execution/eip8130-state/Cargo.toml
+++ b/crates/execution/eip8130-state/Cargo.toml
@@ -32,6 +32,7 @@ base-precompile-storage = { workspace = true, features = ["test-utils"] }
 default = [ "std" ]
 std = [
 	"alloy-primitives/std",
+	"base-common-consensus/std",
 	"base-precompile-storage/std",
 	"revm/std",
 ]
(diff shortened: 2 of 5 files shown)


yog45-co (approved):

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

xenoliss (approved):

Sertug17:
Nice work @chunter-cb. The Snapshot + checkpoint_revert coverage is thorough for the fields that were included.

    Two gaps I noticed that follow the same pattern as the original gas_refunded issue (#3299):

    1. caller is not captured in Snapshot

    replace_caller() is part of the PrecompileStorageProvider trait and can change mid-execution, but caller is absent from the Snapshot struct:


    p.set_caller(alice);
    let cp = p.checkpoint();
    p.replace_caller(bob);
    p.checkpoint_revert(cp);
    // p.caller() → returns bob, not alice


    If any test path calls replace_caller() inside a checkpoint region and then reverts, the caller stays wrong. For the EIP-8130 authorize step where caller identity matters, this could cause subtle false test results.

    2. is_static is not captured in Snapshot

    Same pattern — set_static() mutates it, but checkpoint_revert doesn't restore it:


    p.set_static(false);
    let cp = p.checkpoint();
    p.set_static(true);
    p.checkpoint_revert(cp);
    // p.is_static() → returns true, not false


    Precompile methods gate access on is_static(). A test that sets static context inside a checkpoint and then reverts won't see the static flag restored, potentially causing false positives (operation allowed when it should be blocked) or false negatives.

    The fix is straightforward same pattern as the existing fields:

    rust
    struct Snapshot {
        internals: ...,
        caller: Address,
        is_static: bool,
        // ... existing fields
    }


    These could be filed as separate issues or addressed in a follow up. Happy to open issues if that's preferred.

PR #3541, opened 2026-06-15
perf: encoder step-path component benchmarks and channel-close characterization

refcell:
## Summary

Autonomous perf-autopilot run results — adding component-level benchmarks for `base-batcher-encoder` and documenting the step-path budget breakdown.

### Component Benchmark Results (4096 blocks)

| Phase | Median | Per-block |
|---|---|---|
| Composition (`block_to_single_batch`) | 9.52 ms | ~2.34 µs |
| Add-batch (RLP + ShadowCompressor) | 7.35 ms | ~1.80 µs |
| **Combined comp + compression** | **~16.9 ms** | **~4.1 µs** |
| Full step loop (existing bench) | 678.66 ms | — |

### Key Finding

Block composition and per-batch compression together account for **only ~2.5%** of the total encoder step budget. The remaining **~97.5%** is absorbed by:

- `close_current_channel` — flush, output_frame, Arc-frame creation
- Compression finalization during channel close
- Channel management and frame draining

This means future optimization work that targets per-block micro-level (composition, single-batch encoding) has diminishing returns. The next meaningful gain comes from the **channel-close path**: `flush()`, `output_frame()`, and the Arc-ing of frames.

### Files Changed

- `crates/batcher/encoder/benches/step_components.rs` — new component bench (3 Criterion groups)
- `crates/batcher/encoder/Cargo.toml` — wired `[[bench]]` entry
- `crates/batcher/comp/src/channel_out.rs` — cosmetic cleanup (removed unused `vec::Vec` import)
- `docs/autonomy/perf-journal.md` — timestamped entry documenting findings

### Next Steps (for next autopilot cycle)

Focus on `close_current_channel` internals: add a targeted bench measuring flush + output_frame + Arc-frame-creation together, then evaluate batched frame output, buffer recycling, or compressor finalization optimization.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2922,6 +2922,7 @@ dependencies = [
  "base-protocol",
  "base-runtime",
  "base-tx-manager",
+ "criterion",
  "derive_more 2.1.1",
  "eyre",
  "futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2892,6 +2892,7 @@ dependencies = [
  "base-metrics",
  "base-protocol",
  "clap",
+ "criterion",
  "metrics",
  "rand 0.9.4",
  "rstest",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3713,6 +3713,7 @@ dependencies = [
  "base-metrics",
  "base-protocol",
  "bytes",
+ "criterion",
  "derive_more 2.1.1",
  "discv5",
  "futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5415,6 +5415,7 @@ dependencies = [
  "base-common-rpc-types",
  "base-common-rpc-types-engine",
  "brotli",
+ "criterion",
  "derive_more 2.1.1",
  "miniz_oxide 0.9.1",
  "proptest",
--- a/crates/batcher/comp/src/brotli.rs
+++ b/crates/batcher/comp/src/brotli.rs
@@ -56,6 +56,12 @@ pub struct BrotliCompressor {
     /// The lazily-materialised compressed bytes.  Valid (non-dirty) only when
     /// `dirty` is `false`.
     compressed: RefCell<Vec<u8>>,
+    /// Raw byte offset within `compressed` for the next read.
+    ///
+    /// By keeping a cursor instead of calling `Vec::drain()` on every frame
+    /// we turn the per-frame drain step from O(n) into O(1), removing the
+    /// overall O(n²) behaviour when a channel produces many frames.
+    read_pos: Cell<usize>,
     /// The raw accumulated input bytes.
     raw: Vec<u8>,
     /// Marks that the compressor is closed.
@@ -73,6 +79,7 @@ impl BrotliCompressor {
         let level = level.into();
         Self {
             compressed: RefCell::new(Vec::new()),
+            read_pos: Cell::new(0),
             raw: Vec::new(),
             closed: false,
             dirty: Cell::new(false),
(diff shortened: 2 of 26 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Sertug17:
---

*"Excellent breakdown @refcell. The 97.5% / 2.5% split is a clear signal.*

*One angle worth quantifying: *does close_current_channel cost scale with blocks or with channels?**

*With the fixture config (4096 blocks x ~4.2 KB/block raw, target_frame_size: 120_000, target_num_frames: 1, approx_compr_ratio: 0.4), compressed data lands at roughly 6.9 MB, requiring ~57 channel closes. If the full ~662 ms remainder is channel close, each close costs ~11.6 ms and the cost is proportional to the number of full channels, not the block count.*

*This has a concrete implication for the next autopilot cycle: parameterizing the close_current_channel micro-bench over target_frame_size or blocks_per_channel would reveal whether the cost is amortizable. In production, a batcher with fewer, larger channels would hit close_current_channel less often, potentially making the per-block micro-optimizations (the 2.5%) relatively more important again.*

*Also worth splitting the ~662 ms into its three sub-components early:*
1. *flush()  compressor finalization; scales with total unflushed data in the channel*
2. *output_frame()  serialization; ~O(frame_size), fixed per close*
3. *Arc::new(bytes)  allocation + atomic refcount; ~O(frame_size), fixed per close*

*If the dominant term is (3), buffer recycling across frames would be the simplest win. If it's (1), the fix is in compressor configuration  e.g., finalizing incrementally during the step loop rather than all at once at close."*

---

Sertug17:
@refcell nice work on the cursor based read the Zlib fix alone was a correctness bug regardless of perf. A couple of observations:

    1. ensure_compressed() error is swallowed in len()

    Both BrotliCompressor::len() and ZlibCompressor::len() call let _ = self.ensure_compressed();, discarding the Result. If compression fails (OOM, internal brotli error), len() returns saturating_sub(pos) on an empty compressed vec always 0 after read_pos advances past 0. This means ready_bytes() returns 0, and output_frame's drain loop exits early. The channel closes with zero (or fewer) frames silently no error propagated, no metric fired. The operator sees a channel close but no frames produced, indistinguishable from a genuinely empty channel.

    get_compressed() has the same issue after a compression failure it returns [pos..] on an empty vec, which is &[] regardless of what was written.

    2. ShadowCompressor::close() doc comment is stale

    Lines 113–116 still say "read() can drain it directly" with the cursor, read() no longer drains. Minor but helps future readers.

    3. Brotli write() resets read_pos before read() can propagate the error

    In write() (L58-60), read_pos is set to 0 before dirty is set to true. If ensure_compressed() then fails on the subsequent read(), read_pos is already 0 and dirty is still true. A second write() + read() cycle would attempt recompression from scratch (correctly, since dirty=true), but read_pos was already cleared on the first write not a correctness issue but an ordering wart.

PR #3545, opened 2026-06-15
feat(rpc): add deep tracing for eth_call and sendRawTransaction

meyer9:
## Summary
- add Base-owned tracing spans for `eth_call` and `eth_sendRawTransaction`
- break `eth_call` into nested orchestration spans so state load and execution are visible in traces
- split raw tx handling into decode and submission phases
- include the one-pager summary doc for team review

## Validation
- `just check::format`
- `just zepter`
- `just test-affected-ci meyer9/otel-tracing-engine-parenting`
- local affected tests are not fully green yet; this draft PR is intended to let GitHub CI verify the remaining split fallout


Sertug17:
---

*"Good findings. On the span level concern (#1):*

*`debug_span!` for the inner sub spans is the right call. The concrete production benefit: default RUST_LOG configs typically filter out DEBUG spans, so the 11 inner spans become zero-cost (the `enabled!` check short circuits before any metadata allocation). Operators can then opt into the full breakdown per-target:*

```
RUST_LOG=rpc::eth::call=debug,rpc::eth::send_raw_transaction=debug
```

*This gives the detailed trace exactly when needed without paying the cost on every call. The outer `eth_call` at INFO is sufficient for normal ops it signals rate and latency without the per-step overhead.*

*On finding #3 (cancellation error):* *`EthApiError::InternalEthError` masks the root cause. A separate `Canceled` variant (or wrapping the existing error with context) would make traces and error responses actionable.*

---

Sertug17 (commented):
Reviewed the diff for this PR. The bot picked up most of the structural stuff the double permit, span levels, and the nonce dead conditional. A couple of things it didn't flag:

**Re: the double semaphore permit** `acquire_owned_blocking_io()` followed by `spawn_blocking_io_fut()` which also acquires from the same semaphore. The bot mentioned halved concurrency, but there's a sharper edge: if `blocking_io_task_guard_max` is configured to 1 (which some deployments do to strictly limit blocking IO pressure), this deadlocks on the first call. The outer `_permit` never releases until the inner acquire gets a permit, which it can't.

**About the cancellation path** `guard` is dropped after `spawn_blocking_io_fut` awaits, and `cancel.is_cancelled()` is checked inside the closure. This is correct for Tokio's cancellation model (CancelOnDrop uses Arc internally), but the check only fires once, right before `execute_eth_call`. If the caller drops the response handle during EVM execution, the cancellation is essentially a no-op the blocking task keeps running to completion and the result gets thrown away. Not a correctness issue, but worth documenting that this cancellation only covers the setup phase, not the actual transact call.

**The override vs upstream** Since `EthCall::call` is now fully overridden instead of using the default implementation, any upstream reth changes to the `call` path won't flow in automatically. That's fine for now since most of the logic is duplicated inline, but worth keeping in mind for the first rebase after a reth bump.

Other than that, the implementation looks structurally sound. The span tree in the one-pager is well thought out `eth_call` being broken into nested spans is genuinely useful for debugging latency in Jaeger.

Sertug17:
One more thing I noticed that the bot didn't flag:

The `caller_gas_allowance` cap on line ~201 is only applied when `!request_has_gas_limit` i.e., only when the caller didn't supply an explicit gas limit. If a request comes in with `gas: 1_000_000_000` (huge value), it gets capped to `global_gas_cap` but the per-caller balance allowance is never checked. The upstream default `EthCall::call` applies the caller allowance regardless of whether a gas limit was supplied.

Consequence: when a caller supplies a high gas limit that exceeds their balance based allowance, instead of getting a clear early error the call proceeds to execution and would fail inside the EVM with a more opaque error. Not a correctness issue per se (the EVM will still reject it), but the error path is worse and the behavior diverges from upstream.

PR #3435, opened 2026-06-10
chore: Register L3 devnet chain 84534 for nitro proving

jowparks:
## Summary
- Adds `privacy-sepolia` (`ChainConfig` chain ID **84534**, parent L1 **84532**) so nitro host/enclave can resolve `rollup_config!(84534)` and precompute config hashes.
- Adds **Base Sepolia (84532)** to `L1_CONFIGS` as an OP-stack L1 parent config (required by nitro-host `prover_config()` when proving L3 batches derived from Base Sepolia).
- Extends config-hash regression tests for chain 84534.

## Placeholder fields (follow-up required)
Genesis anchors, L1 contract addresses, and azul/beryl timestamps currently use **zero/derived placeholders**. They must be updated after privacy-enclave pins a deploy-once genesis so `PerChainConfig::hash()` matches the live rollup config at runtime.

Privacy devnet params encoded today:
- `block_time=2`, `seq_window=3600`, azul block 20 / beryl block 21 → timestamps 40/42 (assumes `genesis_l2_time=0`)
- `batch_inbox`: `0xff...84534`
- `activation_admin`: devnet Anvil account `0x9965...`

## Test plan
- [x] `cargo test -p base-common-chains --lib`
- [x] `cargo test -p base-proof-tee-nitro-enclave config_hashes_match_chain_configs`
- [x] `cargo test -p base-proof-succinct-client-utils test_config_hash_matches_nitro_enclave`
- [ ] Rebuild base-proofs nitro EIF with this `BASE_COMMIT` and bump privacy-enclave mirror digests (after merge)
- [ ] Pin privacy devnet genesis and update placeholder addresses/hashes in `PRIVACY_SEPOLIA`

Made with [Cursor](https://cursor.com)

PR #3554, opened 2026-06-15
feat: add --fetch-rollup-config flag to nitro-host

jowparks:
## Summary

Adds `--fetch-rollup-config` / `--l2-cl-url` CLI flags to `base-prover-nitro-host` for fetching genesis config from an L2 CL node via `optimism_rollupConfig` RPC. Also fixes the enclave to compute `config_hash` dynamically from the rollup config instead of a hardcoded table, so non-built-in chains (e.g. devnet 84534) can produce valid signed proofs.

## Host changes (`bin/prover/nitro-host/`)

- **`cli.rs`**: `--fetch-rollup-config` bool flag, `--l2-cl-url` CL endpoint, RPC response structs, `fetch_rollup_config_from_rpc()`, async `prover_config()` with override logic, all 3 call sites updated.
- **`Cargo.toml`**: Added `alloy-provider` with `reqwest` feature.
- **`main.rs`**: `use alloy_provider as _;` unused dep guard.

When `--fetch-rollup-config` is set: fetches config from CL, validates chain ID, overrides genesis fields on the resolved `RollupConfig`. Hard error if RPC fails or `--l2-cl-url` missing.

## Enclave change (`crates/proof/tee/nitro-enclave/`)

- **`server.rs`**: Replaced `CONFIG_HASHES` `LazyLock` table (hardcoded from `ChainConfig::all()`) with `config_hash()` that computes `PerChainConfig` from `boot_info.rollup_config`. The rollup config already flows through preimage oracle key 6 — we just weren't using it for `config_hash`. Known chain hashes unchanged (verified by golden-value tests). Matches ZK pipeline's `hash_rollup_config()`.

## Why `--l2-cl-url`?

`optimism_rollupConfig` is a CL method (op-node), not EL (reth/geth). Verified on devnet: EL returns "Method not found", CL works.

## Usage

```bash
base-prover-nitro-host local \
  --l2-eth-url http://base-rpc:8645 \
  --l2-cl-url http://base-builder-cl:7549 \
  --l2-chain-id 84534 \
  --fetch-rollup-config
```

### DIFF
--- a/bin/prover/nitro-host/src/cli.rs
+++ b/bin/prover/nitro-host/src/cli.rs
@@ -602,23 +602,18 @@ struct SystemConfigRpcResponse {
 }
 
 #[cfg(any(target_os = "linux", feature = "local"))]
-async fn fetch_rollup_config_from_rpc(
-    l2_eth_url: &str,
-) -> eyre::Result<RollupConfigRpcResponse> {
+async fn fetch_rollup_config_from_rpc(l2_eth_url: &str) -> eyre::Result<RollupConfigRpcResponse> {
     let provider = ProviderBuilder::new()
         .connect(l2_eth_url)
         .await
         .map_err(|e| eyre!("failed to connect to L2 node at {l2_eth_url}: {e}"))?;
 
-    provider
-        .raw_request("optimism_rollupConfig".into(), ())
-        .await
-        .map_err(|e| {
-            eyre!(
-                "optimism_rollupConfig RPC failed on {l2_eth_url}: {e}. \
+    provider.raw_request("optimism_rollupConfig".into(), ()).await.map_err(|e| {
+        eyre!(
+            "optimism_rollupConfig RPC failed on {l2_eth_url}: {e}. \
                  Ensure the L2 node supports this method"
-            )
-        })
+        )
+    })
 }
 
 #[cfg(all(test, feature = "local"))]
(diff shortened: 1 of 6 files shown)


hughy (approved):

bin/prover/nitro-host/src/cli.rs
```diff
@@ -150,11 +167,49 @@ impl ProverRuntimeArgs {
...
+                "overriding genesis config from L2 node RPC"
+            );
+
+            rollup_config.genesis.l1.hash = rpc.genesis.l1.hash;
+            rollup_config.genesis.l1.number = rpc.genesis.l1.number;
+            rollup_config.genesis.l2.hash = rpc.genesis.l2.hash;
+            rollup_config.genesis.l2.number = rpc.genesis.l2.number;
+            rollup_config.genesis.l2_time = rpc.genesis.l2_time;
+            rollup_config.deposit_contract_address = rpc.deposit_contract_address;
+            rollup_config.l1_system_config_address = rpc.l1_system_config_address;
```
hughy: We might also need to set/override `gas_limit` and `scalar` to make sure that the derived config covers everything

jowparks:
Good catch — `gas_limit` and `scalar` are both included in `PerChainConfig::marshal_binary()` which feeds into `config_hash`. Added them to the RPC deserialization struct and the override logic in 22a2245.

hughy:
After merging in https://github.com/base/base/pull/3435 we might need a change to `crates/proof/proof/src/boot.rs:253` to use the runtime config over the hardcoded config

PR #345, opened 2026-01-10
feat(flashblocks): add UnifiedReceiptBuilder for seamless deposit receipt handling

Himess:
## Summary

Introduces `UnifiedReceiptBuilder` - a wrapper that handles both deposit and non-deposit transaction receipts seamlessly, eliminating the try-catch pattern required by `OpReceiptBuilder`.

### Changes
- Added `receipt_builder.rs` with `UnifiedReceiptBuilder<C>` struct
- Single `build()` method handles both tx types internally
- Simplified `state_builder.rs` - removed ~40 lines of manual deposit handling
- Removed unused `alloy-op-evm` dependency

Closes #309

PR #3584, opened 2026-06-16
feat(PRIV-1984): dev multiproof deploy config template + nitro-host config-hash tool

robriks:
## Summary

- **Deploy config template:** parameterize `multiproofConfigHash`, `teeImageHash`, and `devTeeSigner` so they are driven by env vars from `devnet-l3-env`; forward those vars into the `setup-l2` container (compose + `setup-l2.sh`)
- **Fix `teeProposer`** from `${DEPLOYER_ADDR}` to `${PROPOSER_ADDR}` — the proposer sends `createWithInitData` txs and must match `gameCreator()` in journal verification
- **`extract-artifacts.sh`:** emit `TEEProverRegistryProxy`, `AggregateVerifier`, `TEEVerifier` into `.devnet/l2/configs/l1-addresses.json` so downstream services can read their addresses
- **New `nitro-host config-hash` subcommand:** computes `config_hash` from a live op-node rollup config using the enclave's own `base_proof_primitives::PerChainConfig` path — the single source of truth for the hash
- **Bump `CONTRACTS_COMMIT`** to `65dd69c6` (current `l3-changes` tip = PRIV-1997 / contracts #345)
- Create gitignored `devnet-tee.env` for the dev TEE signer private key

## Context

Part of standing up a fully functional L3 appchain devnet in dev multiproof mode ([PRIV-1984](https://linear.app/priv/issue/PRIV-1984)). After this PR, `just devnet up-l3` deploys the full multiproof contract stack (`DevTEEProverRegistry` with the dev signer registered, `TEEVerifier`, `AggregateVerifier` as game type 621 on `DisputeGameFactory`, zero finality delays), and the repo gains the tool that computes the chain's `config_hash`.

### `config_hash` — placeholder here, runtime registration in a follow-up

`MULTIPROOF_CONFIG_HASH` is a non-zero placeholder (`0x00…01`) that simply enables the multiproof path at deploy time. The correct value **cannot be hardcoded**: it is `keccak256(PerChainConfig::marshal_binary())` — a 212-byte encoding that includes genesis block hashes and deployed contract addresses, which are non-deterministic across ephemeral devnet runs (anvil stamps L1 genesis with wall-clock time).

The real hash is computed from the live chain and set on the `AggregateVerifier` by a **runtime-registration** step in a follow-up PR (PRIV-1985). Nothing consumes the placeholder-hash verifier in this PR (no proposer/games yet).

### `nitro-host config-hash`

Resolves a `RollupConfig` via `--l2-cl-url <op-node>` and prints `PerChainConfig::from_rollup_config(..).force_defaults().hash()` as `0x…`. Pure host logic (no NSM), so it runs without the enclave — in CI, in smoke tests, and in the follow-up's registration one-shot. This guarantees the on-chain `CONFIG_HASH` is byte-identical to what the enclave signs, by construction (same code path, same op-node input).

## Test plan

- [x] Template renders correctly via `envsubst` with the new variables
- [x] `devnet-tee.env` is gitignored (matched by `*.env` in `.gitignore`)
- [x] `extract-artifacts.sh` includes the three new multiproof addresses
- [x] `nitro-host config-hash` compiles + clippy + fmt clean across default/`local`/`worker`/`local,worker`, and prints a valid `0x…64` hash against a live op-node
- [x] `CONTRACTS_COMMIT` resolves to `65dd69c6` (PRIV-1997, includes PRIV-1975 + PRIV-1982 + the register entrypoint)


etc/scripts/devnet/templates/deploy-config.json.template
```diff
@@ -3,6 +3,7 @@
   "baseFeeVaultRecipient": "${DEPLOYER_ADDR}",
   "baseFeeVaultWithdrawalNetwork": 0,
   "batchSenderAddress": "${BATCHER_ADDR}",
+  "devTeeSigner": "${DEV_TEE_SIGNER}",
```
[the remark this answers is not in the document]
robriks: [ef8d9cc](https://github.com/base/base/pull/3584/commits/ef8d9ccb758bbcf148b28b0568e52f79c55ca8ad)

etc/docker/devnet-l3-env
```diff
@@ -29,3 +29,14 @@ BATCHER_MAX_PENDING_TRANSACTIONS=4
...
+# Multiproof / TEE configuration
+# Non-zero placeholder — the real config hash is deployment-specific (depends on genesis
+# block hashes and contract addresses). A two-phase approach in setup-l2.sh will compute
+# the correct hash post-deploy and update the on-chain AggregateVerifier CONFIG_HASH.
+MULTIPROOF_CONFIG_HASH=0x0000000000000000000000000000000000000000000000000000000000000001
+# Dev TEE signer address — registered in DevTEEProverRegistry during SystemDeploy.
+# The corresponding private key lives in devnet-tee.env (gitignored).
+DEV_TEE_SIGNER=0x6cebCF805c5191BCf26602E43DECa89AEe092b5d
+# Zero image hash — matches nitro-local's B256::ZERO (no real PCR0 measurement in dev).
+TEE_IMAGE_HASH=0x0000000000000000000000000000000000000000000000000000000000000000
```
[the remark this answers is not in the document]
robriks: [ef8d9cc](https://github.com/base/base/pull/3584/commits/ef8d9ccb758bbcf148b28b0568e52f79c55ca8ad)

etc/docker/devnet-l3-env
```diff
@@ -29,3 +29,14 @@ BATCHER_MAX_PENDING_TRANSACTIONS=4
 # No L1_BEACON_ADDRESS: leaving it unset makes the ZK prover build a disabled blob provider, but
 # the prover (like the consensus node) requires calldata-only mode to be opted into explicitly.
 L1_CALLDATA_ONLY=true
+
+# Multiproof / TEE configuration
+# Non-zero placeholder — the real config hash is deployment-specific (depends on genesis
+# block hashes and contract addresses). A two-phase approach in setup-l2.sh will compute
```
hughy: Is the two-phase approach still to come later?

Is it possible to update the onchain AggregateVerifier CONFIG_HASH after the contract is deployed?

robriks: two phase approach is indeed planned, but I'm seeing now that `CONFIG_HASH` can only be updated via a full upgrade (which is not how we should do it imo, regardless what claude thinks!)

will investigate

robriks: I think I've come up with a better solution than the two-phase upgrade approach, will push more commits shortly

Sertug17: Nice work @robriks. The multiproof env var plumbing looks clean end to end.

    One bootstrap gap I noticed: devnet-tee.env is gitignored (via *.env) but there's no fallback or creation step if it doesn't exist. Someone cloning the repo and running devnet up-l3 for the first time won't know they need to create this file, or what key to put in it.

    The DEV_TEE_SIGNER address is hardcoded in devnet-l3-env (line 42), but the corresponding private key lives in the gitignored file with no documented generation step. If devnet-tee.env is missing, the setup fails without a clear message.

    Could add a bootstrap step in setup-l2.sh that creates devnet-tee.env with a fresh key if the file doesn't exist (e.g. cast wallet simple), or ship a devnet-tee.env.example with a doc comment explaining how to generate one.

robriks: turned out the two-phase approach is necessary, merged in [base/contracts #345](https://github.com/base/contracts/pull/345)

this PR lays the foundation for the [follow-up PR which wires everything together](https://github.com/base/base/pull/3599)

Sertug17: @robriks the #3599 pipeline looks solid for the config hash issue compute config-hash then register aggregate verifier with the right sequencing.

    Two things I'd look at before merging this:

    1. The PR description still talks about the old two-phase approach. Might be worth updating it to match what actually happens now (deploy with placeholder, compute real hash, register post-genesis with MULTIPROOF_DEFER_REGISTRATION).

    2. devnet-tee.env is still a gap. Someone running devnet up l3 for the first time won't have this file and there's no fallback. By the time the follow-up lands with the proposer/challenger, it'll be a hard requirement. Even just adding a [ -f devnet-tee.env ] || cast wallet simple > devnet-tee.env bootstrap in setup-l2.sh would save people a lot of head-scratching.

hughy (approved):

PR #3586, opened 2026-06-16
feat(eip8130): intrinsic gas accounting + fee/balance validation

chunter-cb:
## Summary

Adds `base-execution-eip8130-gas`, the **gas** step of the EIP-8130 validation pipeline shared by mempool admission and block inclusion. Given the signed transaction plus a couple of state-derived hints, it computes the EIP-8130 intrinsic gas and decides whether the gas payer can afford the transaction.

### Intrinsic gas

`IntrinsicGas::compute(signed, encoded, input)` returns the per-component breakdown of the spec formula. `encoded` is the caller-supplied EIP-2718 serialization (the bytes already held on the admission/build paths), used for the `payload` cost so the hot path performs no extra RLP allocation:

```
intrinsic_gas = AA_BASE_COST + tx_payload_cost + nonce_key_cost + bytecode_cost
              + account_changes_cost + auto_delegation_cost
              + sender_auth_cost + payer_auth_cost
```

| Component | Source |
|---|---|
| `base` | `AA_BASE_COST` (15,000) |
| `payload` | EIP-2028 (16/non-zero, 4/zero byte) over the EIP-2718-serialized **signed** tx (so `sender_auth`/`payer_auth` are charged) |
| `nonce_key` | nonce-free `14,000`; otherwise first-use `22,100` / existing `5,000` |
| `bytecode` | per create entry: `32,000 + 200·code_len` (code only; the entry's initial-actor slot writes are in `account_changes`) |
| `account_changes` | per create entry: one fresh `actor_config` slot write per initial actor (`22,100` each; unrestricted owner, no policy slots); per config-change entry: its `auth` cost + each mutated actor slot (`actor_config`, plus the two policy slots when the authorize carries a policy); per delegation entry: the `4,600` deposit |
| `auto_delegation` | `4,600` when a code-less `sender` is auto-delegated to `DEFAULT_ACCOUNT` |
| `sender_auth` / `payer_auth` | authenticator execution gas + one cold `actor_config` SLOAD (charged only when an authenticator is actually resolved); delegate recurses into its nested authenticator (depth-1); `payer_auth` is `0` for self-pay |

`sender_intrinsic()` excludes `payer_auth` (payer auth is metered on top of `gas_limit`), exposing `execution_gas_available(gas_limit) = gas_limit − sender_intrinsic`.

### Authenticator execution gas (chain policy)

[...]

PR #3589, opened 2026-06-17
feat(eip8130): transaction authorization orchestrator

chunter-cb:
## Summary

Adds the **authorization orchestrator** of the EIP-8130 validation pipeline, shared by mempool admission and block inclusion. It composes the per-stage primitives into a single verdict over a signed transaction: the **final sender/payer signatures** plus the transaction's **ordered set of account-configuration changes**.

Lands as the **`transaction` module** of the consolidated `base-execution-eip8130` crate (after #3595 folded the per-stage `eip8130-{state,authorize,tx,nonce}` crates into one). Reads state but performs **no mutations**.

### `TransactionAuthorizer::authorize`

1. **Final transaction signatures** — resolves and scope-gates the sender (`SCOPE_SENDER`) and, when sponsored, the payer (`SCOPE_PAYER`) via `ActorTxVerifier`. The sender signature commits to the whole body, so it authorizes the envelope (calls, metadata, and the presence of every account change), and resolves the sender account (recovering it for the EOA path) that the config changes apply against.
2. **Account-configuration changes** — walks `account_changes` in order, authorizing each `ConfigChange` with its own `SignedActorChanges` signature (`SCOPE_CONFIG`), the account lock, and the chain binding. `Create`/`Delegation` entries carry no independent signature at this layer (authorized by the sender signature; `Create` is additionally bound by its deterministic deploy address at execution).

Returns `AuthorizedTransaction { actors, config_changes }`, or the first `TxAuthError` (sender → payer → config changes in order).

### Sequence advancement

A config-change channel (multichain `chain_id == 0`, or the local chain) advances by one per applied entry. `ConfigChangeAuthorizer::authorize` alone only validates against the channel's *current* on-chain sequence, so a transaction carrying several same-channel entries would fail on the second. The orchestrator reads each channel's base sequence once and validates entries against the running value (`base`, `base + 1`, …) in tx order. Channel selection is explicit (`chain_id == 0` → multichain, `== local` → local, anything else → `ConfigChainId`), so a foreign chain id can never advance a local counter. A running sequence that would exceed `u64::MAX` is rejected (`ConfigSequenceOverflow`) rather than wrapped/saturated.

To enable this, `ConfigChangeAuthorizer` gains **`authorize_at_sequence`** (explicit expected sequence); the existing `authorize` is now a thin wrapper that reads the sequence from state and delegates.

### Scope (what this is not)

Authorization only. Nonce validation, intrinsic-gas accounting and fee/balance checks (#3586) are separate stages; composing all of them into one end-to-end transaction verdict is the next layer. Structural/expiry validation is upstream in consensus. No state is mutated (no `actor_config` writes, sequence/nonce advancement, or balance debits) — that belongs to the execution/apply layer.

## Test plan

- `cargo test -p base-execution-eip8130` — 72 tests total; the 7 orchestrator tests cover sequential same-channel advancement, stale second same-channel entry rejected, independent multichain/local channels, foreign-chain config change rejected without advancing a channel, channel-sequence overflow rejected (not wrapped), `Create`/`Delegation` pass-through (no sequence consumed), and full configured sender + sponsored payer + config-change composition.
- `cargo clippy -p base-execution-eip8130 --all-targets`
- `cargo +nightly fmt -p base-execution-eip8130 -- --check`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p base-execution-eip8130 --no-deps`

### DIFF
--- a/crates/common/consensus/src/transaction/eip8130/addresses.rs
+++ b/crates/common/consensus/src/transaction/eip8130/addresses.rs
@@ -82,8 +82,8 @@ impl Eip8130Contracts {
     // ─────────────────────────────────────────────────────────────────────────
 
     // Note: secp256k1 has no contract entry. It is the protocol-reserved native
-    // ecrecover sentinel
-    // [`Eip8130Constants::ECRECOVER_AUTHENTICATOR`](super::Eip8130Constants::ECRECOVER_AUTHENTICATOR)
+    // k1 sentinel
+    // [`Eip8130Constants::K1_AUTHENTICATOR`](super::Eip8130Constants::K1_AUTHENTICATOR)
     // (`address(1)`), handled directly by the protocol, not a deployed contract.
 
     /// secp256r1 / P-256 (raw) authenticator contract.
--- a/crates/common/consensus/src/transaction/eip8130/constants.rs
+++ b/crates/common/consensus/src/transaction/eip8130/constants.rs
@@ -87,16 +87,26 @@ impl Eip8130Constants {
     /// `actor_change` operation byte: revoke an existing actor.
     pub const ACTOR_CHANGE_REVOKE: u8 = 0x02;
 
-    /// Lower-bound authenticator address. The `ECRECOVER_AUTHENTICATOR` native is fixed at
-    /// `address(1)`; authenticator addresses smaller than this are reserved.
-    pub const ECRECOVER_AUTHENTICATOR: Address =
-        address!("0x0000000000000000000000000000000000000001");
+    /// The single canonical secp256k1 ("k1") authenticator, fixed at
+    /// `address(1)`. Native `ecrecover`: the protocol recovers from the `data`
+    /// blob (`r || s || v`) rather than `STATICCALL`-ing a contract. The same
+    /// identity serves both the implicit default EOA and any explicitly
+    /// registered k1 actor; the `actor_config` slot alone distinguishes a
+    /// full-owner EOA from a scoped key.
+    ///
+    /// `address(0)` is reserved as the empty / "no actor configured" sentinel and
+    /// is never a valid authenticator selector; addresses below this are reserved.
+    pub const K1_AUTHENTICATOR: Address = address!("0x0000000000000000000000000000000000000001");
 
-    /// Sentinel authenticator address indicating an actor slot has been revoked
-    /// (`type(uint160).max`). Submitting authentication data prefixed with this
-    /// authenticator MUST be rejected.
-    pub const REVOKED_AUTHENTICATOR: Address =
-        address!("0xffffffffffffffffffffffffffffffffffffffff");
+    /// `AccountState.flags` bit that disables the implicit default-EOA path.
+    ///
+    /// The implicit default EOA is a [`Self::K1_AUTHENTICATOR`] signature whose
+    /// recovered signer equals the account; with no explicit `actor_config` it
+    /// resolves to a full owner, gated solely on this flag. Set by
+    /// `createAccount`/`importAccount` (disabled by default), and by authorizing
+    /// or revoking the self-actor; once set it is never cleared (monotonic), so
+    /// an explicit self-actor entry always implies the flag is set.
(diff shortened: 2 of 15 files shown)


Sertug17:
Nice architecture @chunter-cb. The `authorize_at_sequence` extraction and the running counter pattern are clean. Read through the full authorize.rs one observation and one minor documentation note:

### Observation: `else` branch conflates all non-0 chain_ids into a single `local_base` counter

Lines 82–84 in authorize.rs:
```rust
let (base, applied) = if cc.chain_id == 0 {
    (multichain_base, &mut multichain_applied)
} else {
    (local_base, &mut local_applied)  // all non-0 chain_ids share this
};
```

If EIP-8130 is extended to carry config changes for multiple distinct non-zero chains in the same transaction, this shared counter would under count the per channel advancement (two entries for different chains would both pull from and increment the same `local_applied`).

For the current design where all non-zero changes target the local chain, this is correct and `authorize_at_sequence` rejects foreign chain_ids before the sequence check, so no misrouted entry silently advances the counter. Worth noting for future readers though, since the mapping isn't obvious at a glance.

### Minor doc: `expected` calculation could note the one-phase vs two-phase commitment

The comment on lines ~68–73 explains *why* the orchestrator uses running sequences ("re-reading the still-unwritten state"), which is helpful. One extra sentence clarifying that this is safe because the execution/apply layer will re-validate against the post-commit state before advancing would connect the validation and execution phases for readers unfamiliar with the pipeline layering.

---

Solid work. The test coverage (6 tests in validate, 19 in tx including the refactored `authorize_at_sequence`) gives good confidence on the sequence advancement logic.

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

xenoliss (approved):

PR #3596, opened 2026-06-17
fix(zlib): drain consumed bytes in read() to prevent infinite frame loop

### LINKED ISSUE #3594: fix(zlib): read() does not drain consumed bytes causes infinite frame loop on channel close
ZlibCompressor::read() has two bugs, both fixed in the Brotli sibling during PR #3541 but never applied to Zlib.

**Bug 1: bytes never drained**

```rust
// Zlib (buggy)
let compressed = self.compressed.borrow();
let len = compressed.len().min(buf.len());
buf[..len].copy_from_slice(&compressed[..len]);
// NO DRAIN — same bytes returned every time

// Brotli (fixed)
let mut compressed = self.compressed.borrow_mut();
buf[..len].copy_from_slice(&compressed[..len]);
compressed.drain(..len);  // bytes consumed
```

Since `read()` returns the same bytes every time, the `close_current_channel` drain loop becomes infinite — `ready_bytes()` never decreases, frames grow unbounded, process OOMs.

**Bug 2: ensure_compressed() error swallowed**

Zlib calls `self.ensure_compressed()` without `?`, silently swallowing the error. Brotli propagates it with `?`.

**Fix:** copy Brotli's read() implementation (the same two-line change from PR #3541 commit a0ddbc2 that was only applied to one side):

```rust
fn read(&mut self, buf: &mut [u8]) -> CompressorResult<usize> {
    self.ensure_compressed()?;
    let mut compressed = self.compressed.borrow_mut();
    let len = compressed.len().min(buf.len());
    buf[..len].copy_from_slice(&compressed[..len]);
    compressed.drain(..len);
    Ok(len)
}
```

Triggers on every channel close when `compression_algo = "zlib"` is configured (the no_std fallback or explicit config).

@refcell

Sertug17:
## Fix

Applies the same fix that BrotliCompressor got in PR #3541 but was missed for Zlib.

**Two changes in `ZlibCompressor::read()`:**

1. `let compressed` → `let mut compressed` — need `&mut` for `drain()`
2. `compressed.drain(..len)` — consume read bytes so subsequent calls don't return the same data

**Why this matters:**

`ZlibCompressor::read()` was not draining consumed bytes from the internal compressed buffer. When `ChannelOut::output_frame()` calls `compressor.read()` inside the `close_current_channel` drain loop:

```rust
while open.out.ready_bytes() > 0 {     // never decreases with Zlib
    match open.out.output_frame(...) {   // reads same data every time
        Ok(frame) => frames.push(Arc::new(frame)),  // grows forever
    }
}
```

Each iteration creates another frame with identical data. The `frames` vec grows unboundedly until the process OOMs.

**Trigger:** Any batcher configured with `compression_algo = "zlib"` hits this on every channel close.

**Related:**
- Fixes #3594
- Ref: #3541 (same fix for Brotli, commit a0ddbc2)

### DIFF
--- a/crates/batcher/comp/src/zlib.rs
+++ b/crates/batcher/comp/src/zlib.rs
@@ -90,9 +90,10 @@ impl CompressorWriter for ZlibCompressor {
 
     fn read(&mut self, buf: &mut [u8]) -> CompressorResult<usize> {
         self.ensure_compressed();
-        let compressed = self.compressed.borrow();
+        let mut compressed = self.compressed.borrow_mut();
         let len = compressed.len().min(buf.len());
         buf[..len].copy_from_slice(&compressed[..len]);
+        compressed.drain(..len);
         Ok(len)
     }
 }


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Sertug17:
@refcell @jackchuma @eric-ships This is the missing half of the Brotli fix from #3541. Same pattern read() wasn't draining consumed bytes from the compressed buffer, causing an infinite frame loop (OOM) on every channel close when Zlib compression is active.

    The len() method's ensure_compressed() swallow is harmless here since Zlib's compression is infallible (miniz_oxide), but read() was a real OOM-triggering bug.

    Same three-line change that #3541 applied to Brotli, just for Zlib.

PR #3603, opened 2026-06-17
fix(flashblocks): `BLOCKHASH` divergence bw pending and canonical EVM execution, harden sequence linkage

haardikk21:
This PR fixes the divergence between the pending flashblocks EVM executor and the canonical executor that was causing divergent transaction receipts to exist between pending state and canonical state for transactions that directly or indirectly used the `BLOCKHASH` opcode during EVM execution.

Due to reconstructed Flashblock blocks being slightly different than canonical blocks (lack of state root, for example), the reconstructed block hash would differ from the canonical block hash. 

When building pending state across block boundaries, if a transaction that reads `BLOCKHASH` for the parent block comes in before the Flashblock executor has received the canonical block, it would return an incorrect block hash to the EVM runtime. This could cause divergent code paths to be executed - for e.g. for games built on top of Chainlink VRF, that would poison the state and potentially also have cascading affects downstream for other future transactions that relied on the poisoned state even without using `BLOCKHASH` themselves.

The added regression test passes, but will fail if we comment out the additions in `processor.rs` from this PR. 

---

Separately, this PR also updates the block builder to include a `prev_flashblock_id` and updates the Flashblock sequence validation logic in the client to use that (if available) to harden sequence linkage. Once this change is deployed across both the builder and the clients, we will be able to detect cases of the last flashblock in a block being skipped over the stream - which previously we were unable to.

### DIFF
--- a/crates/builder/core/src/flashblocks/payload.rs
+++ b/crates/builder/core/src/flashblocks/payload.rs
@@ -1,7 +1,7 @@
 use core::time::Duration;
 use std::{
     ops::{Div, Rem},
-    sync::Arc,
+    sync::{Arc, Mutex},
     time::Instant,
 };
 
@@ -17,7 +17,8 @@ use base_bundles::RejectedTransaction;
 use base_common_chains::Upgrades;
 use base_common_consensus::{BaseReceipt, BaseTransactionSigned};
 use base_common_flashblocks::{
-    ExecutionPayloadBaseV1, ExecutionPayloadFlashblockDeltaV1, FlashblocksPayloadV1,
+    ExecutionPayloadBaseV1, ExecutionPayloadFlashblockDeltaV1, FlashblockId, FlashblocksPayloadV1,
+    Metadata,
 };
 use base_execution_consensus::{calculate_receipt_root_no_memo, isthmus};
 use base_execution_evm::{BaseEvmConfig, BaseNextBlockEnvAttributes};
@@ -89,11 +90,13 @@ pub(super) struct BasePayloadBuilder<Pool, Client> {
     pub config: BuilderConfig,
     /// Sender for forwarding per-block batches of rejected transactions to the audit-archiver.
     pub rejected_tx_sender: Option<mpsc::Sender<Vec<RejectedTransaction>>>,
+    /// Last flashblock emitted by this builder instance.
+    pub last_emitted_flashblock_id: Arc<Mutex<Option<FlashblockId>>>,
 }
 
 impl<Pool, Client> BasePayloadBuilder<Pool, Client> {
     /// `BasePayloadBuilder` constructor.
-    pub(super) const fn new(
+    pub(super) fn new(
         evm_config: BaseEvmConfig,
         pool: Pool,
         client: Client,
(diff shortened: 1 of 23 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>3/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/flashblocks/src/metadata.rs
```diff
@@ -1,10 +1,107 @@
...
-use serde::{Deserialize, Serialize};
+use std::{fmt, str::FromStr};
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
+
+/// Identifies a flashblock within a canonical block build.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct FlashblockId {
+    /// Canonical block number this flashblock belongs to.
+    pub block_number: u64,
```
BrianBland, line 12: Ideally this should use a block hash to harden against reorgs, in which case index would also be replaced by a chunk hash.

This may be overkill if we are willing to trust that canonical blocks and flashblocks are not reorged.

haardikk21: A block hash is not stable for a block that is actively still being constructed though, the reliable marker is the block number. Block hash would only work in a base flashblock where it's previous flashblock is FB 10 of last block - but FBs 1..10 will not be able to use that

meyer9 (dismissed):

crates/common/flashblocks/src/metadata.rs
```diff
@@ -1,10 +1,147 @@
...
+        /// Index component.
+        index: String,
+        /// Parse error for the index component.
+        source: ParseIntError,
+    },
+}
+
+impl fmt::Display for FlashblockId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}-{}", self.block_number, self.index)
```
mw2000, line 50: non-blocking: should be fine for `FlashblocksId` since block_number >> index, but for similar structs should ideally keep the named fields on the `Display` impl

mw2000 (dismissed):

mw2000 (approved):

mw2000 (approved):

meyer9 (approved):

crates/builder/core/src/flashblocks/payload.rs
```diff
@@ -70,6 +74,26 @@ type NextBestFlashblocksTxs<Pool> = BestFlashblocksTxs<
...
+            block_number: self.block_number.load(Ordering::Relaxed),
+            index: self.index.load(Ordering::Relaxed),
+        }
+    }
+
+    fn store(&self, flashblock_id: FlashblockId) {
+        self.block_number.store(flashblock_id.block_number, Ordering::Relaxed);
+        self.index.store(flashblock_id.index, Ordering::Relaxed);
+    }
+}
```
0x00101010, line 95: worth removing the AtomicU64 and use a Mutex instead here for atomicity

haardikk21: I had a mutex earlier but atomics are a bit more performant. Given they always update together at the same time plus this being single threaded code I dont think there's much risk here tbh

0x00101010 (approved):

PR #3620, opened 2026-06-18
fix(common): Upgrade Typing

refcell:
Unifies the duplicated upgrade enums into a single canonical BaseUpgrade defined in base-common-genesis, removing the parallel ContractUpgrade type that existed only to work around the dependency graph. Everything downstream now depends on the one enum and dispatches through exhaustive typed matches instead of matching upgrades by string.

BaseUpgrade is generated via alloy_hardforks::hardfork! and spans two domains: the execution fork ladder (EXECUTION_VARIANTS) that maps onto the reth/revm hardfork schedule, and the contract-backed upgrade set (CONTRACT_VARIANTS) keyed by the L1 upgrade-signal contract ids and genesis UpgradeConfig timestamps. Bedrock is execution-only, while Delta and PectraBlobSchedule are contract-only and excluded from the execution ladder. snake_case contract ids come from contract_id(), kept separate from the PascalCase hardfork name().

Execution-layer behavior that cannot live in genesis (revm specs, ChainConfig schedules) is added through a BaseUpgradeExt trait in base-common-chains. ChainUpgrades now indexes a fixed-size array by execution_idx() rather than relying on enum discriminant ordering, so adding a future non-execution variant cannot silently shift the ladder.

Validated with tests and clippy across genesis, chains, chainspec, upgrade-signal, evm, and precompiles, plus a check build across dependent consensus, protocol, execution, builder, and cli crates.

### DIFF
--- a/crates/common/genesis/src/chain/upgrade.rs
+++ b/crates/common/genesis/src/chain/upgrade.rs
@@ -2,12 +2,11 @@
 
 use alloc::{
     collections::BTreeMap,
-    format,
     string::{String, ToString},
 };
-use core::{fmt::Display, str::FromStr};
+use core::fmt::Display;
 
-use alloy_hardforks::{EthereumHardfork, ParseHardforkError};
+use alloy_hardforks::{EthereumHardfork, hardfork};
 use spin::{Once, RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 /// Upgrade configuration for Base-specific upgrades.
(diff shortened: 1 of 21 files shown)


crates/common/genesis/src/chain/upgrade.rs
```diff
@@ -86,9 +134,45 @@ impl ContractUpgrade {
...
+    pub const fn is_execution(self) -> bool {
+        self.execution_idx().is_some()
+    }
+
+    /// Returns true if this upgrade is contract-backed (i.e. signaled by the L1 upgrade-signal
+    /// contract and stored in [`UpgradeConfig`]). False only for [`Bedrock`](Self::Bedrock).
+    pub const fn is_contract_backed(self) -> bool {
+        !matches!(self, Self::Bedrock)
+    }
+
```
PelleKrab, line 147: Why is this upgrade not contract backed?

refcell: This is because bedrock was block activated. `ContractUpgrade` enum also chose to ignore bedrock. It shouldn't make a difference since this is a very old upgrade anyways :)

PelleKrab (approved):

PR #3673, opened 2026-06-20
Decode TEEProverRegistry registration reverts

leopoldjoy:
## Summary
- Adds local `TEEProverRegistry` custom-error selectors to the contract binding.
- Exposes a small helper to decode known registry revert data to names like `AttestationTooOld`, `AttestationVerificationFailed`, `InvalidPublicKey`, and `PCR0NotFound`.
- Uses the decoded name in registrar registration errors when `TxManagerError::ExecutionReverted` carries registry revert bytes.

## Tangible value
This prevents on-call and operators from seeing only a generic `execution reverted` or raw selector bytes when the contract already returned a precise reason. Different errors imply different actions: `AttestationTooOld` means fetch/generate fresh proof material, `AttestationVerificationFailed` points at verifier/proof output, `InvalidPublicKey` or `PCR0NotFound` points at malformed attestation journal data, and `ImageHashReadFailed` points at registry/factory configuration.

## Tests
- `cargo test -p base-proof-contracts`
- `RISC0_SKIP_BUILD_KERNELS=1 cargo test -p base-proof-tee-registrar signer_manager`

### DIFF
--- a/crates/proof/contracts/src/tee_prover_registry.rs
+++ b/crates/proof/contracts/src/tee_prover_registry.rs
@@ -5,7 +5,7 @@
 
 use alloy_primitives::Address;
 use alloy_provider::RootProvider;
-use alloy_sol_types::sol;
+use alloy_sol_types::{SolError, sol};
 use async_trait::async_trait;
 
 use crate::ContractError;
@@ -16,6 +16,27 @@ sol! {
     /// `TEEProverRegistry` contract interface.
     #[sol(rpc)]
     interface ITEEProverRegistry {
+        /// Thrown when the attestation document is too old.
+        error AttestationTooOld();
+
+        /// Thrown when the ZK attestation verification fails.
+        error AttestationVerificationFailed();
+
+        /// Thrown when the attestation's public key is malformed.
+        error InvalidPublicKey();
+
+        /// Thrown when PCR0 is not found in the attestation's PCR list.
+        error PCR0NotFound();
+
+        /// Thrown when the dispute game factory is not configured.
+        error DisputeGameFactoryNotSet();
+
+        /// Thrown when reading TEE_IMAGE_HASH from the AggregateVerifier fails.
+        error ImageHashReadFailed();
+
+        /// Thrown when the selected game type has no TEE_IMAGE_HASH.
+        error InvalidGameType();
+
         /// Registers a signer using a ZK-proven AWS Nitro attestation.
         function registerSigner(bytes calldata output, bytes calldata proofBytes) external;
 
@@ -32,6 +53,26 @@ sol! {
 
         /// Returns all currently registered signer addresses.
         function getRegisteredSigners() external view returns (address[]);
+    }
+}
+
+/// Returns a human-readable name for known `TEEProverRegistry` custom-error
+/// revert data.
+pub fn decode_tee_prover_registry_revert(data: &[u8]) -> Option<&'static str> {
+    let selector = data.get(..4)?;
+    match selector {
+        s if s == ITEEProverRegistry::AttestationTooOld::SELECTOR => Some("AttestationTooOld"),
+        s if s == ITEEProverRegistry::AttestationVerificationFailed::SELECTOR => {
+            Some("AttestationVerificationFailed")
+        }
+        s if s == ITEEProverRegistry::InvalidPublicKey::SELECTOR => Some("InvalidPublicKey"),
+        s if s == ITEEProverRegistry::PCR0NotFound::SELECTOR => Some("PCR0NotFound"),
+        s if s == ITEEProverRegistry::DisputeGameFactoryNotSet::SELECTOR => {
+            Some("DisputeGameFactoryNotSet")
+        }
+        s if s == ITEEProverRegistry::ImageHashReadFailed::SELECTOR => Some("ImageHashReadFailed"),
+        s if s == ITEEProverRegistry::InvalidGameType::SELECTOR => Some("InvalidGameType"),
+        _ => None,
     }
 }
 
@@ -89,7 +130,7 @@ impl TEEProverRegistryClient for TEEProverRegistryContractClient {
 #[cfg(test)]
 mod tests {
     use alloy_primitives::{Address, Bytes};
-    use alloy_sol_types::SolCall;
+    use alloy_sol_types::{SolCall, SolError};
 
     use super::*;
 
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/contracts/src/tee_prover_registry.rs
```diff
@@ -35,6 +56,26 @@ sol! {
     }
 }
 
+/// Returns a human-readable name for known `TEEProverRegistry` custom-error
+/// revert data.
+pub fn decode_tee_prover_registry_revert(data: &[u8]) -> Option<&'static str> {
```
jackchuma: Could we shrink this by leaning on the error enum generated by `sol!`? Alloy should already generate `ITEEProverRegistryErrors::name_by_selector`, so this helper can avoid manually matching every selector:

```rust
  data.get(..4)
      .and_then(|selector| selector.try_into().ok())
      .and_then(ITEEProverRegistry::ITEEProverRegistryErrors::name_by_selector)
```

That keeps future registry errors from needing updates in two places

crates/proof/tee/registrar/src/error.rs
```diff
@@ -44,6 +44,16 @@ pub enum RegistrarError {
         tx_hash: B256,
     },
 
+    /// Registration reverted with a known `TEEProverRegistry` custom error.
```
jackchuma: Do we need a new public `RegistrarError` variant here? It looks like this only re-labels an existing `TxManagerError::ExecutionReverted` with a decoded reason that we already log at the failure site. Unless something downstream needs to branch on `RegistryRevert`, I’d keep the existing `Transaction(ExecutionReverted)` path and put the decoded registry error only in structured logs

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -297,12 +299,26 @@ where
...
+                                if let Some(data) = data
+                                    && let Some(reason) =
+                                        decode_tee_prover_registry_revert(data.as_ref())
+                                {
+                                    warn!(
+                                        signer = %signer_address,
+                                        registry_error = reason,
+                                        "registration reverted with known TEEProverRegistry error"
+                                    );
+                                    return Err(RegistrarError::RegistryRevert {
```
jackchuma: This wrapper seems unnecessary if the goal is observability. We already emit `registry_error = reason` just above, so returning `RegistrarError::from(e)` preserves the original tx-manager error without adding another registrar-level error shape. I’d only keep the wrapper if there’s a caller that needs typed handling for known registry reverts

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -858,6 +875,33 @@ mod tests {
         }
     }
 
+    #[tokio::test(start_paused = true)]
```
jackchuma: If `RegistryRevert` goes away, this test can go too. The useful coverage is the decoder unit test in `tee_prover_registry.rs`; the existing tx outcome test already covers non-retryable execution reverts blocking recovery

crates/proof/contracts/src/tee_prover_registry.rs
```diff
@@ -35,6 +56,14 @@ sol! {
     }
 }
 
+/// Returns a human-readable name for known `TEEProverRegistry` custom-error
```
jackchuma: This helper feels like extra public API for a single logging call. Alloy already generated the selector-to-name lookup here (`ITEEProverRegistry::ITEEProverRegistryErrors::name_by_selector`), so `decode_tee_prover_registry_revert` mostly wraps an existing API and then needs its own `lib.rs` export plus tests.

  I’d keep this local to the caller for now: import/use the generated error enum directly in `signer_manager.rs`, do the `data.get(..4).and_then(...)` there, and drop the exported helper plus decode tests. If a second caller needs this, that’s the point where a shared helper pays for itself

leopoldjoy: Done: https://github.com/base/base/pull/3673/commits/41bdfa0c7489e789a03e77866cdb94da54617324

crates/proof/contracts/src/tee_prover_registry.rs
```diff
@@ -16,6 +16,27 @@ sol! {
...
+        error PCR0NotFound();
+
+        /// Thrown when the dispute game factory is not configured.
+        error DisputeGameFactoryNotSet();
+
+        /// Thrown when reading `TEE_IMAGE_HASH` from the `AggregateVerifier` fails.
+        error ImageHashReadFailed();
+
+        /// Thrown when the selected game type has no `TEE_IMAGE_HASH`.
+        error InvalidGameType();
```
jackchuma, line 41: Add PCR0Mismatch to registry error decoding. For registrations that reach the contract with a valid proof but the enclave image/PCR0 does not match the expected `TEE_IMAGE_HASH`, the TEE attestation docs say `TEEProverRegistry` reverts with `PCR0Mismatch`. Because the new error enum does not include that custom error, this selector will still decode as `None`, so the warning added in `register_signer` hides the main registry failure this change is meant to diagnose. Please include `PCR0Mismatch` with the exact Solidity signature used by the registry

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -492,7 +501,7 @@ mod tests {
         time::Duration,
     };
 
-    use alloy_primitives::{Address, B256};
+    use alloy_primitives::{Address, B256, Bytes};
```
jackchuma: Seems like we could probably remove this `Bytes` import?

leopoldjoy: Good catch: https://github.com/base/base/pull/3673/changes/13bdcd70438c79f2bdfb4bdba3f8223cb81942ec

crates/proof/contracts/src/tee_prover_registry.rs
leopoldjoy: True, done: https://github.com/base/base/pull/3673/changes/13bdcd70438c79f2bdfb4bdba3f8223cb81942ec

jackchuma (approved):

PR #3674, opened 2026-06-20
Validate registrar attestation nonce binding

leopoldjoy:
## Summary
- Derives deterministic attestation nonces from registrar domain, registry address, and signer address.
- Requests signer-specific attestations with those deterministic nonces, so recovered Boundless proofs can be validated after registrar restart without persisted nonce state.
- Decodes the generated proof journal before `registerSigner` submission and verifies the proven nonce matches the deterministic nonce.
- Blocks proof recovery for the signer when the proof journal is undecodable or carries the wrong nonce, so the next attempt can produce valid attestation material.

## Tangible value
This adds nonce binding on top of the freshness checks while preserving stateless Boundless recovery. It prevents accepting attestations that do not carry the registrar's expected domain/signer/registry nonce, without rejecting recovered proofs solely because they were generated during an earlier registrar cycle.

## Tests
- `RISC0_SKIP_BUILD_KERNELS=1 cargo test -p base-proof-tee-registrar`
- `just check::clippy-affected-ci "origin/main"`


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -235,6 +237,11 @@ where
             }
             Err(e) => return Err(e.into()),
         };
+
+        if let Err(e) = Self::ensure_proof_nonce(signer_address, &proof, requested_nonce) {
+            self.proof_provider.block_recovery_for_signer(signer_address);
+            return Err(e);
```
jackchuma: Preserve the original nonce when recovering proofs. When Boundless recovers a proof submitted by a previous registration attempt, that proof's journal contains the nonce from the earlier attestation, while `requested_nonce` here is regenerated on every `resolve_instance` cycle. In a restart/ cancellation during the proof window, `generate_proof_for_signer` can wait for and return that recovered proof, but this check then rejects it, blocks recovery, and forces a second proof on a later cycle. Either carry the original nonce through recovery or avoid waiting on recovered requests that cannot match the current nonce

crates/proof/tee/registrar/src/error.rs
```diff
@@ -44,6 +44,26 @@ pub enum RegistrarError {
         tx_hash: B256,
     },
 
+    /// Generated proof journal could not be decoded before submission.
```
jackchuma: `InvalidProofJournal` and `AttestationNonceMismatch` are very specific public error variants for one private validation step. Nothing else matches on the structured fields except tests, so the structure is not buying much. A single `InvalidProofJournal(String)` or `InvalidAttestationProof(String)` would preserve the useful operator-facing message without adding enum surface area

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -353,6 +360,29 @@ where
         Ok(())
     }
 
+    fn ensure_proof_nonce(
```
jackchuma: `ensure_proof_nonce` is a tiny one-caller helper that takes the whole `AttestationProof` just to decode `proof.output`. The abstraction does not hide repeated logic or clarify a reused concept. Inline `let journal = VerifierJournal::decode(&proof.output)?;` beside the proof generation path and compare `journal.nonce` there

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -353,6 +360,29 @@ where
...
+        requested_nonce: &[u8],
+    ) -> Result<()> {
+        let journal = VerifierJournal::decode(&proof.output)
+            .map_err(|e| RegistrarError::InvalidProofJournal { reason: e.to_string() })?;
+
+        if journal.nonce.as_ref() != requested_nonce {
+            return Err(RegistrarError::AttestationNonceMismatch {
+                signer,
+                expected_hex: format!("0x{}", hex::encode(requested_nonce)),
+                actual_hex: if journal.nonce.is_empty() {
```
jackchuma: the `<missing>` branch treats empty nonce as a special display case. But `VerifierJournal` stores nonce as `bytes`, so “missing” arrives as empty bytes anyway. Hex-encoding empty bytes as `0x` is enough, and avoids a tiny custom convention only this error uses

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -877,6 +947,70 @@ mod tests {
...
+
+        assert!(
+            matches!(result, Err(RegistrarError::AttestationNonceMismatch { signer, .. }) if signer == SIGNER_A),
+            "mismatched nonce should fail before submission: {result:?}"
+        );
+        assert_eq!(manager.tx_manager.send_count(), 0, "nonce mismatch must not submit a tx");
+        assert_eq!(*proof_provider.blocked_signers.lock().unwrap(), vec![SIGNER_A]);
+    }
+
+    #[tokio::test]
```
jackchuma: `register_signer_rejects_missing_nonce_before_tx` mostly exists to test that `<missing>` string. If the special string goes away, the existing mismatch test already covers the behavior that matters: nonce mismatch blocks tx submission and blocks proof recovery for the signer

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -632,9 +671,29 @@ mod tests {
     struct RecordingProofProvider {
         cancel_then_error: bool,
         blocked_signers: Arc<Mutex<Vec<Address>>>,
+        proof_output: Option<Bytes>,
```
jackchuma, line 720: `RecordingProofProvider` now has both `proof_output` and `proof_nonce`, two ways to customize the same returned proof. `proof_nonce` is just a convenience for one mismatch test. Use `proof_output: Some(proof_output_with_nonce(b"old-nonce"))` there and delete the extra field/path

leopoldjoy:
Updated #3674 to use deterministic signer/registry nonces instead of per-cycle random nonce storage. This preserves stateless Boundless recovery while still validating that the proof journal carries the registrar's expected nonce. Also applied the cleanup comments by simplifying the error shape, inlining the validation, removing the `<missing>` convention/test, and simplifying the proof-provider test override path.

Verification:
- `RISC0_SKIP_BUILD_KERNELS=1 cargo test -p base-proof-tee-registrar`
- `just check::clippy-affected-ci "origin/main"`

crates/proof/tee/registrar/src/driver.rs
```diff
@@ -238,40 +237,52 @@ where
...
-        );
-        let all_attestations = match self
-            .signer_client
-            .signer_attestation(&instance.endpoint, Some(nonce.to_vec()))
-            .await
-        {
-            Ok(attestations) => attestations,
-            Err(e) => {
+        let mut all_attestations = Vec::with_capacity(addresses.len());
+        for (index, signer) in addresses.iter().copied().enumerate() {
```
jackchuma, line 254: If I understand correctly, we are now requesting attestations per-signer instead of per-instance, and each request returns N attestations where N is the number of signers for that instance. We are doing N^2 work here. A follow up PR should be to improve the API shape to either request an attestation scoped to a specific signer, or accept a vec of nonces

jackchuma (approved):

jackchuma (approved):

PR #3675, opened 2026-06-21
fix(registrar): back off tx submission retries

leopoldjoy:
## Summary
- use exponential backoff for retryable registrar registration tx submission errors
- treat the existing tx retry delay config as the initial backoff delay
- include the selected retry delay in retry warning logs
- clarify CLI/config docs that the retry delay is now the initial delay

## Test plan
- `cargo +nightly fmt --package base-proof-tee-registrar --package base-proof-tee-registrar-bin`
- `RISC0_SKIP_BUILD_KERNELS=1 cargo test -p base-proof-tee-registrar signer_manager::tests::register_signer`
- `RISC0_SKIP_BUILD_KERNELS=1 cargo test -p base-proof-tee-registrar`
- `RISC0_SKIP_BUILD_KERNELS=1 cargo clippy -p base-proof-tee-registrar --all-targets -- -D warnings`

### DIFF
--- a/crates/proof/tee/registrar/src/signer_manager.rs
+++ b/crates/proof/tee/registrar/src/signer_manager.rs
@@ -30,6 +30,9 @@ pub const DEFAULT_MAX_TX_RETRIES: u32 = 3;
 /// Default initial delay between transaction submission retries in seconds.
 pub const DEFAULT_TX_RETRY_DELAY_SECS: u64 = 5;
 
+/// Maximum exponential backoff delay between transaction submission retries.
+const MAX_TX_RETRY_BACKOFF_DELAY: Duration = Duration::from_secs(60);
+
 /// State for a proof-generation task currently in-flight.
 ///
 /// One entry per signer address. The pending map is keyed by [`Address`] so
@@ -78,12 +81,6 @@ impl<P, R, T> SignerManager<P, R, T> {
             max_tx_retries,
             tx_retry_delay,
         }
-    }
-
-    /// Returns the exponential backoff delay before the next retry attempt.
-    fn tx_retry_backoff_delay(&self, retry: u32) -> Duration {
-        let multiplier = 1_u32.checked_shl(retry.saturating_sub(1)).unwrap_or(u32::MAX);
-        self.tx_retry_delay.saturating_mul(multiplier)
     }
 }
 
@@ -318,7 +315,11 @@ where
                         }
 
                         let retry = retry + 1;
-                        let retry_delay = self.tx_retry_backoff_delay(retry);
+                        let max_retry_delay = MAX_TX_RETRY_BACKOFF_DELAY.max(self.tx_retry_delay);
+                        let retry_delay = self
+                            .tx_retry_delay
+                            .saturating_mul(2_u32.saturating_pow(retry - 1))
+                            .min(max_retry_delay);
                         warn!(
                             error = %e,
                             signer = %signer_address,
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -79,6 +79,12 @@ impl<P, R, T> SignerManager<P, R, T> {
             tx_retry_delay,
         }
     }
+
+    /// Returns the exponential backoff delay before the next retry attempt.
+    fn tx_retry_backoff_delay(&self, retry: u32) -> Duration {
+        let multiplier = 1_u32.checked_shl(retry.saturating_sub(1)).unwrap_or(u32::MAX);
+        self.tx_retry_delay.saturating_mul(multiplier)
```
jackchuma: Cap exponential retry delays. When operators configure a larger `MAX_TX_RETRIES`, this uncapped multiplier makes the registrar effectively stop retrying: with the default 5s base delay, retry 20 sleeps for about 30 days before the next attempt, and retry 32 sleeps for centuries. Since the CLI still accepts any `u32` retry count, a configuration that previously meant "keep retrying every 5s" now leaves the signer registration pending until cancellation. Please cap the backoff delay or validate `max_tx_retries` to a safe range

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -79,6 +79,12 @@ impl<P, R, T> SignerManager<P, R, T> {
             tx_retry_delay,
         }
     }
+
+    /// Returns the exponential backoff delay before the next retry attempt.
```
jackchuma: This helper only has one caller, and most of its defensive shape is covering `retry == 0`, which the caller already rules out by incrementing before calculating the delay. We can delete the helper and use the stdlib saturating power directly at the call site:

```rust
let retry_delay = self.tx_retry_delay.saturating_mul(2_u32.saturating_pow(retry - 1));
```

Same behavior for the real call path, fewer lines, one less private API to keep around

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -858,6 +866,30 @@ mod tests {
         }
     }
 
+    #[tokio::test(start_paused = true)]
```
jackchuma, line 870: This test duplicates the surrounding `register_signer_tx_outcome` setup and retry path; the only new assertion is elapsed retry sleep time. We can fold it into that table with an `expected_elapsed` column and assert `start.elapsed()` per case. That keeps the backoff coverage, but avoids another full manager setup and separate test body

jackchuma (commented):
The current code already has bounded retry for retryable tx errors. This changes the retry schedule from fixed delay to exponential delay, which is reasonable if we have evidence that quick retries are causing repeated provider/nonce/mempool failures. Without that evidence, this feels like speculative retry tuning: more latency, slightly more code, and no clearly demonstrated behavior gap

leopoldjoy:
Good catch on the uncapped delay; pushed a 60s cap.

The motivation for exponential backoff is the tx-manager caller contract for retryable `Rpc(_)` errors, see [here](https://github.com/base/base/blob/main/crates/utilities/tx-manager/src/error.rs#L179-L186) and [here](https://github.com/base/base/blob/main/crates/utilities/tx-manager/src/error.rs#L197-L202). It's quite explicit that:

> Callers **must** enforce bounded retry counts and exponential backoff ...

crates/proof/tee/registrar/src/signer_manager.rs
```diff
@@ -858,6 +867,55 @@ mod tests {
...
```
jackchuma, line 870: these two backoff tests are almost identical setup with only retry count, base delay, error count, and expected elapsed changing. A single table-driven test would cover both the normal exponential path and the capped path with less test code and less fixture duplication

leopoldjoy: Thanks! Addressed here: https://github.com/base/base/pull/3675/commits/1e6879adba7c1631505401db63b0497fabf41a1a

jackchuma (approved):

PR #3691, opened 2026-06-22
fix(prover-service, challenger): close metric gaps

0xth4nh:
## Summary
- Record terminal prover-service success and duration metrics when worker `submit_proof` newly completes a job.
- Split idempotent `submit_proof` replays into `AlreadyCompleted` so repeated worker submissions return the stored result without double-counting terminal metrics.
- Reuse the same terminal metric helper for expired-claim reaper failures and stuck-request permanent failures so `proof_requests_completed` and `proof_request_duration_ms` stay aligned on normal terminal paths.
- Route fraudulent ZK challenge checkpoint validation through the existing output validator method, now named `validate_claimed_root_at_block`, so Path 2 emits the existing validation latency, error, and invalid-root metrics without adding wrapper indirection.

## Why
The existing metric definitions were present, but some terminal paths did not emit them. Successful worker completions skipped `prover_service.proof_requests_completed` and `prover_service.proof_request_duration_ms`, and the fraudulent-ZK-challenge path computed one checkpoint directly instead of using the shared instrumented validator. That made dashboards undercount successful proof outcomes and hide Path 2 validation work.

The README metric sections added earlier were removed after review so the source constants remain the single source of truth for metric names.

## How to test
- `RUSTFMT_BOOTSTRAP=1 cargo fmt --check` (passes; rustfmt prints the workspace stable-channel warnings for unstable config keys)
- `git diff --check`
- `cargo clippy -p base-challenger --features metrics --all-targets -- -D warnings`
- `cargo test -p base-prover-service terminal_proof_job_records_outcome_and_duration`
- `cargo test -p base-challenger --features metrics validate_claimed_root_at_block_emits_shared_validation_metrics`
- `cargo test -p base-prover-service`
- `cargo test -p base-prover-service-db`
- `cargo test -p base-challenger --features metrics`

Notes: Postgres-backed integration tests in `base-prover-service` and `base-prover-service-db` remain ignored unless `DATABASE_URL` is set, which matches the existing test setup.

### DIFF
--- a/crates/proof/challenge/src/driver.rs
+++ b/crates/proof/challenge/src/driver.rs
@@ -420,8 +420,12 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
         //   starting_block + interval * (i + 1)
         let checkpoint_block = candidate.checkpoint_start_block(challenged_index + 1)?;
 
-        let expected_root = match self.validator.compute_output_root(checkpoint_block).await {
-            Ok(root) => root,
+        let validation = match self
+            .validator
+            .validate_claimed_root_at_block(game_address, checkpoint_block, on_chain_root)
+            .await
+        {
+            Ok(result) => result,
             Err(ValidatorError::BlockNotAvailable { .. }) => {
                 debug!(
                     game = %game_address,
@@ -443,12 +447,12 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
 
         // If the onchain root at the challenged index does not match the
         // locally computed root, the ZK challenge was legitimate — skip.
-        if on_chain_root != expected_root {
+        if !validation.is_valid {
             debug!(
                 game = %game_address,
                 challenged_index = challenged_index,
                 on_chain = %on_chain_root,
-                expected = %expected_root,
+                expected = %validation.expected_root,
                 "ZK challenge is legitimate (challenged root was wrong), skipping"
             );
             return Ok(());
@@ -463,8 +467,13 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
 
         ChallengerMetrics::fraudulent_zk_challenge_detected_total().increment(1);
 
-        self.initiate_zk_proof(candidate, challenged_index, on_chain_root, DisputeIntent::Nullify)
-            .await
+        self.initiate_zk_proof(
+            candidate,
+            challenged_index,
+            validation.expected_root,
+            DisputeIntent::Nullify,
+        )
+        .await
     }
 
     /// Attempts TEE-first proof sourcing with ZK fallback.
--- a/crates/proof/challenge/src/validator.rs
+++ b/crates/proof/challenge/src/validator.rs
@@ -265,18 +265,18 @@ impl<L2: L2Provider> OutputValidator<L2> {
         Ok(None)
     }
 
-    /// Validates the final output root of a candidate dispute game.
+    /// Validates one claimed output root at a specific L2 block.
     ///
     /// Fetches the L2 header and `L2ToL1MessagePasser` storage proof at the
-    /// game's L2 block number, computes the expected output root, and compares
-    /// it against the game's `rootClaim`.
-    pub async fn validate_final_root(
+    /// provided L2 block number, computes the expected output root, and compares
+    /// it against the claimed root.
+    pub async fn validate_claimed_root_at_block(
         &self,
         game_address: Address,
         l2_block_number: u64,
         claimed_root: B256,
     ) -> Result<ValidationResult, ValidatorError> {
-        info!(game = %game_address, block = l2_block_number, "validating final output root");
+        info!(game = %game_address, block = l2_block_number, "validating claimed output root");
 
         let checkpoint = Checkpoint { block: l2_block_number, claimed_root };
         let mismatch = self.validate_output_roots(game_address, &[checkpoint]).await?;
@@ -401,6 +401,11 @@ mod tests {
     use alloy_consensus::Header as ConsensusHeader;
     use alloy_primitives::{Address, B256};
     use alloy_rpc_types_eth::Header as RpcHeader;
+    #[cfg(feature = "metrics")]
+    use metrics_util::{
+        MetricKind,
+        debugging::{DebugValue, DebuggingRecorder},
+    };
     use rstest::rstest;
 
     use super::*;
@@ -431,7 +436,7 @@ mod tests {
     #[case::valid(None, true)]
     #[case::invalid(Some(B256::repeat_byte(0xFF)), false)]
     #[tokio::test]
-    async fn test_validate_final_root(
+    async fn test_validate_claimed_root_at_block(
         #[case] wrong_root: Option<B256>,
         #[case] expect_valid: bool,
     ) {
@@ -441,7 +446,10 @@ mod tests {
         let game_address = Address::repeat_byte(0x01);
         let claimed_root = wrong_root.unwrap_or(expected_root);
 
-        let result = validator.validate_final_root(game_address, 100, claimed_root).await.unwrap();
+        let result = validator
+            .validate_claimed_root_at_block(game_address, 100, claimed_root)
+            .await
+            .unwrap();
 
         assert_eq!(result.is_valid, expect_valid);
         assert_eq!(result.expected_root, expected_root);
@@ -519,7 +527,7 @@ mod tests {
 
         // Final root validation succeeds
         let final_result =
-            validator.validate_final_root(game_address, 100, roots[1]).await.unwrap();
+            validator.validate_claimed_root_at_block(game_address, 100, roots[1]).await.unwrap();
         assert!(final_result.is_valid);
 
         // Intermediate root validation fails (corrupt index 0)
@@ -554,7 +562,7 @@ mod tests {
         let validator = OutputValidator::new(Arc::new(provider));
         let game_address = Address::repeat_byte(0x06);
 
-        let result = validator.validate_final_root(game_address, 100, B256::ZERO).await;
+        let result = validator.validate_claimed_root_at_block(game_address, 100, B256::ZERO).await;
 
         assert!(result.is_err());
         let err = result.unwrap_err();
@@ -772,7 +780,7 @@ mod tests {
         let validator = OutputValidator::new(Arc::new(provider));
         let game_address = Address::repeat_byte(0x0B);
 
-        let result = validator.validate_final_root(game_address, 100, B256::ZERO).await;
+        let result = validator.validate_claimed_root_at_block(game_address, 100, B256::ZERO).await;
 
         assert!(result.is_err());
         let err = result.unwrap_err();
@@ -812,7 +820,7 @@ mod tests {
         let validator = OutputValidator::new(Arc::new(provider));
         let game_address = Address::repeat_byte(0x0F);
 
-        let result = validator.validate_final_root(game_address, 100, B256::ZERO).await;
+        let result = validator.validate_claimed_root_at_block(game_address, 100, B256::ZERO).await;
 
         assert!(result.is_err());
         let err = result.unwrap_err();
@@ -839,7 +847,8 @@ mod tests {
         let validator = OutputValidator::new(Arc::new(provider));
         let game_address = Address::repeat_byte(0x10);
 
-        let result = validator.validate_final_root(game_address, 100, substituted_root).await;
+        let result =
+            validator.validate_claimed_root_at_block(game_address, 100, substituted_root).await;
 
         let err = result.expect_err("substituted account proof should be rejected");
         match err {
@@ -852,5 +861,54 @@ mod tests {
             }
             other => panic!("expected AccountProofFailed, got: {other:?}"),
         }
+    }
+
+    #[cfg(feature = "metrics")]
+    #[test]
+    fn validate_claimed_root_at_block_emits_shared_validation_metrics() {
+        let recorder = DebuggingRecorder::new();
+        let snapshotter = recorder.snapshotter();
+        let (provider, roots) = mock_with_blocks(&[100]);
+        let validator = OutputValidator::new(Arc::new(provider));
+        let game_address = Address::repeat_byte(0x11);
+        let wrong_root = B256::repeat_byte(0xFF);
+
+        metrics::with_local_recorder(&recorder, || {
+            let rt = tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
+            rt.block_on(async {
+                let result = validator
+                    .validate_claimed_root_at_block(game_address, 100, wrong_root)
+                    .await
+                    .expect("validation should complete");
+
+                assert!(!result.is_valid);
+                assert_eq!(result.expected_root, roots[0]);
+                assert_eq!(result.invalid_intermediate_index, None);
+            });
+        });
+
+        let snapshot = snapshotter.snapshot().into_vec();
+        assert_eq!(
+            snapshot.iter().find_map(|(ck, _, _, value)| {
+                if ck.kind() != MetricKind::Counter
+                    || ck.key().name() != "base_challenger.games_invalid_total"
+                {
+                    return None;
+                }
+                match value {
+                    DebugValue::Counter(value) => Some(*value),
+                    _ => None,
+                }
+            }),
+            Some(1),
+        );
+        assert!(
+            snapshot.iter().any(|(ck, _, _, value)| {
+                ck.kind() == MetricKind::Histogram
+                    && ck.key().name() == "base_challenger.validation_latency_seconds"
+                    && matches!(value, DebugValue::Histogram(_))
+            }),
+            "checkpoint validation should record latency",
+        );
     }
 }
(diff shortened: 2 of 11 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

0xth4nh:
Addressed the GitHub Actions review finding by rebasing onto latest `main` (which includes the prover-service namespace rename) and updating the prover-service README metric references from `zk_prover_service.*` to `prover_service.*`. Re-ran `git diff --check`, `RUSTFMT_BOOTSTRAP=1 cargo fmt --check`, and the two targeted metrics tests.

0xth4nh:
Addressed the two non-blocking review suggestions as well:

- `validate_claimed_root_at_block` now delegates through `validate_final_root`, preserving the shared validator metrics without duplicating the single-root validation body.
- `RetryOutcome::PermanentlyFailed` now carries the updated terminal `ProofJob` from `UPDATE ... RETURNING`, so the status poller records terminal failure metrics without an extra DB read or fallback branches.

Re-ran:
- `git diff --check`
- `RUSTFMT_BOOTSTRAP=1 cargo fmt --check`
- `cargo test -p base-prover-service`
- `cargo test -p base-prover-service-db`
- `cargo test -p base-challenger --features metrics`

0xth4nh:
Addressed the latest metrics nit by replacing the signed `num_milliseconds() as f64` conversion with a nonnegative `to_std()` duration conversion and an upper clamp before casting to `f64`.

Re-ran:
- `git diff --check`
- `RUSTFMT_BOOTSTRAP=1 cargo fmt --check`
- `cargo test -p base-prover-service terminal_proof_job_records_outcome_and_duration`
- `cargo test -p base-prover-service`

0xth4nh:
Pushed follow-up fixes on top of latest `origin/main` (head `bf1abd937`).

Changes since the last review pass:
- Added a structured warning when terminal proof-job duration is negative, while still clamping the emitted duration sample to `0.0`.
- Added regression coverage for the negative-duration terminal metric path.
- Removed the duplicate `validate_claimed_root_at_block` info log so the delegated validation emits one log line.
- Boxed `RetryOutcome::PermanentlyFailed` to address the CI Clippy `large_enum_variant` failure while preserving the no-extra-DB-read terminal metrics path.
- Rebased onto the latest `origin/main`.

Local verification:
- `git diff --check`
- `RUSTFMT_BOOTSTRAP=1 cargo fmt --check` (passes with existing rustfmt unstable-config warnings)
- `cargo test -p base-prover-service-db`
- `cargo test -p base-prover-service`
- `cargo test -p base-challenger --features metrics validate_claimed_root_at_block_emits_shared_validation_metrics`
- `cargo clippy -p base-prover-service-db --all-targets -- -D warnings`
- `cargo clippy -p base-prover-service --all-targets -- -D warnings`
- `cargo clippy -p base-challenger --features metrics --all-targets -- -D warnings`

0xth4nh:
Addressed the two non-blocking observations from the latest review as well:

- Documented that `validate_claimed_root_at_block` is intentionally a semantic alias over `validate_final_root` for single-checkpoint validation.
- Used the already computed request proof-type label for `stuck_requests` in the permanent-failure branch, while `record_terminal_proof_job` continues to derive terminal outcome labels from the returned terminal job.

Additional local verification after these changes:
- `git diff --check`
- `RUSTFMT_BOOTSTRAP=1 cargo fmt --check`
- `cargo test -p base-prover-service`
- `cargo test -p base-challenger --features metrics validate_claimed_root_at_block_emits_shared_validation_metrics`
- `cargo clippy -p base-prover-service --all-targets -- -D warnings`
- `cargo clippy -p base-challenger --features metrics --all-targets -- -D warnings`

0xth4nh:
Clarifying the latest GitHub Actions review summary:

- The pure delegation in `validate_claimed_root_at_block` is intentional. It keeps the driver call site semantically named for the fraudulent-ZK single-checkpoint path, while routing through the same instrumented validation path as final-root validation. The method doc now calls this out explicitly.
- The duplicate `info!` log item appears stale. Current head has no `info!` in `validate_claimed_root_at_block`; the only matching validation log is in `validate_final_root`, so this path emits one log line, not two.
- The prior proof-type label consistency nit in `status_poller` was also addressed: `stuck_requests` now uses the precomputed request proof-type label, and terminal outcome/duration metrics continue to use `record_terminal_proof_job` from the returned terminal job.

crates/proof/challenge/src/driver.rs
jackchuma, line 1: Nit: this refactoring has nothing to do with metrics. Please undo this

crates/proof/challenge/src/validator.rs
```diff
@@ -293,6 +293,22 @@ impl<L2: L2Provider> OutputValidator<L2> {
...
     }
 
+    /// Validates one claimed output root at a specific L2 block.
+    ///
+    /// This is used by flows that only need to check one intermediate
+    /// checkpoint, while preserving the same latency, error, and invalid-root
+    /// metrics emitted by full intermediate validation. It intentionally
+    /// delegates to final-root validation because the driver only needs the
+    /// validity and expected-root result for this single checkpoint.
+    pub async fn validate_claimed_root_at_block(
```
jackchuma: This wrapper function adds no value. Please remove

crates/proof/challenge/README.md
```diff
@@ -21,6 +21,28 @@ ZK-proof dispute game challenger.
 - **CLI**: CLI argument definitions (`BASE_CHALLENGER_` env-var prefix).
 - **Metrics**: Prometheus metric definitions.
 
+## Metrics Notes
```
jackchuma: Why do we need metrics notes in our readme?

crates/proof/prover-service/db/src/models.rs
jackchuma, line 1: Not sure I agree with these changes to the models

crates/proof/prover-service/db/tests/postgres_integration.rs
jackchuma, line 1: Why are these test changes needed?

crates/proof/prover-service/README.md
```diff
@@ -14,3 +14,26 @@ integration.
 
 Enable `rpc-server` to generate the server trait. Use
 `base-prover-service-client` for requester and worker JSON-RPC clients.
+
+## Metrics
```
jackchuma: Why do we need metrics notes in this readme?

crates/proof/prover-service/src/server/worker_api.rs
```diff
@@ -199,13 +202,22 @@ impl ProverServiceServer {
...
             SubmitProofOutcome::Completed(job) => {
+                metrics::record_terminal_proof_job(metrics::PROOF_STATUS_SUCCEEDED, &job);
                 info!(
                     worker_id = %request.worker_id,
                     session_id = %request.session_id,
                     "worker submitted proof result"
                 );
                 Ok(WorkerSubmitProofResponse { job: into_protocol_job(job)? })
             }
+            SubmitProofOutcome::AlreadyCompleted(job) => {
```
jackchuma, line 213: Don't think this is needed

crates/proof/prover-service/src/worker/status_poller.rs
```diff
@@ -110,14 +110,14 @@ impl StatusPoller {
...
-                    Ok(RetryOutcome::PermanentlyFailed) => {
+                    Ok(RetryOutcome::PermanentlyFailed(job)) => {
                         error!(
                             proof_request_id = %request.id,
                             retry_count = request.retry_count,
                             "Permanently failing stuck request — max retries exceeded"
                         );
                         metrics::inc_stuck_requests(proof_type_label);
-                        metrics::inc_proof_requests_completed("failed", proof_type_label);
+                        metrics::record_terminal_proof_job(metrics::PROOF_STATUS_FAILED, &job);
```
jackchuma, line 120: What does `record_terminal_proof_job` get us that `inc_proof_requests_completed` did not already do?

crates/proof/challenge/src/driver.rs
0xth4nh: So the intention here was not to refactor / adding new function name, but to route this single checkpoint comparison through the validator’s instrumented path. 

The current path `compute_output_root` computed the expected root but skipped the shared validation instrumentation around latency, validation errors, and invalid-root counting.

I guess a better way to do this is to call the validate_final_root directly. ( think the original intention on the wrapper is to provide a better logging for the case, not just "validating final output root", but I do think it is a bit excessive

crates/proof/challenge/src/validator.rs
0xth4nh: The original intuition was to make the driver call site describe what this branch is actually doing: validating the single challenged checkpoint root, not validating a dispute game’s final root.

This is linked to the driver change above. Previously the fraudulent-ZK path called `compute_output_root(checkpoint_block)` directly, which gave us the expected root but bypassed the validator’s shared validation pipeline.

I do think it is not efficient just to add a wrapper on this, maybe an internal logic within validate_final_root, but maybe not in this PR

crates/proof/challenge/README.md
0xth4nh: Deleting now!

crates/proof/prover-service/README.md
0xth4nh: deleting now

crates/proof/prover-service/db/src/models.rs
0xth4nh: The model changes are supporting two metric semantics that were hard to express with the previous enums:

  1. `SubmitProofOutcome::AlreadyCompleted` distinguishes a first terminal transition from an idempotent replay. Without that split, the worker API cannot tell whether it should emit terminal success metrics or simply return the existing completed job. This prevents double-counting `proof_requests_completed` / duration on retrying the same `submit_proof`.

  2. `RetryOutcome::PermanentlyFailed(Box<ProofJob>)` lets the stuck-request path return the terminal row produced by `UPDATE ... RETURNING`. That gives the caller access to `completed_at`, `created_at`, and `api_proof_type`, which are needed by `record_terminal_proof_job` to emit both terminal outcome and duration metrics. The `Box` is only to avoid Clippy’s large-enum-variant warning; semantically this is just
returning the updated terminal job.

 The alternative would be to keep the enum unchanged and either skip duration metrics for this failure path or issue a second DB query after the update. I chose `RETURNING` because it keeps the transition atomic and avoids the extra read.

crates/proof/prover-service/db/tests/postgres_integration.rs
0xth4nh: These test changes follow from the enum semantics above.

 ---> For `RetryOutcome`, the permanent-failure case now carries the terminal `ProofJob`, so exact `assert_eq!(RetryOutcome::PermanentlyFailed)` no longer applies. The tests now match the variant and assert the useful
 new behavior: the returned job has the expected id, failed status, and populated `completed_at`.

---> For `SubmitProofOutcome`, an idempotent replay is no longer reported as `Completed`; it is `AlreadyCompleted`, so the test expectation changed to reflect that a replay is not a fresh terminal transition. That distinction is what prevents terminal success metrics from being emitted twice.

crates/proof/prover-service/src/server/worker_api.rs
0xth4nh: This is needed if we want terminal success metrics from the worker completion path.

--> `complete_claimed_proof_job` is where a worker-submitted proof transitions to terminal success. Recording `record_terminal_proof_job` only in the `Completed` branch means we emit
 --> `proof_requests_completed{status="succeeded"}` and `proof_request_duration_ms` exactly once for the first successful completion.

The new `AlreadyCompleted` branch intentionally does not record metrics, because it represents an idempotent replay of a result that was already completed.

crates/proof/prover-service/db/src/models.rs
0xth4nh: if this does not make sense to add, i can revert on the other 4 related comments too

crates/proof/prover-service/src/worker/status_poller.rs
0xth4nh: `inc_proof_requests_completed` only increments the terminal outcome counter.

`record_terminal_proof_job` increments that same counter and also records `proof_request_duration_ms` from `created_at` to `completed_at`, with handling for missing or invalid timestamps. That is the metric gap this path was closing: stuck permanent failures and expired-claim reaper failures were counted as terminal failures, but did not emit the end-to-end terminal duration.

So the helper centralizes the terminal metric pair and keeps success/failure paths consistent.

crates/proof/challenge/src/validator.rs
jackchuma: If anything, we can just rename `validate_final_root` to `validate_claimed_root_at_block`. I understand that name makes more sense. We should not add a new function that just routes to an existing function though. That kind of indirection will compound into significant code smell over time

crates/proof/prover-service/src/metrics.rs
```diff
@@ -129,6 +134,35 @@ pub fn inc_worker_jobs_failed(reason: &str, proof_type: &str) {
     .increment(1);
 }
 
+/// Record terminal outcome and duration metrics for a proof job.
+pub fn record_terminal_proof_job(status: &str, job: &ProofJob) {
+    let proof_type = api_proof_type_label(job.api_proof_type);
+    inc_proof_requests_completed(status, proof_type);
+
+    if let Some(completed_at) = job.completed_at {
```
jackchuma, line 142: This duration conversion feels heavier than it needs to be for a metrics helper. The `MAX_EXACT_F64_INTEGER` clamp and `to_std()` branch add machinery for values that should never matter operationally here. I’d shrink this to:

```rust
  let duration_ms = (completed_at - job.created_at).num_milliseconds().max(0) as f64;
```

crates/proof/prover-service/src/metrics.rs
```diff
@@ -149,3 +183,155 @@ pub const fn api_proof_type_label(proof_type: ApiProofType) -> &'static str {
...
+
+    use super::*;
+
+    #[test]
+    fn terminal_proof_job_records_outcome_and_duration() {
+        let recorder = DebuggingRecorder::new();
+        let snapshotter = recorder.snapshotter();
+        let created_at = Utc::now();
+        let completed_at = created_at + Duration::milliseconds(2500);
+        let job = ProofJob {
```
jackchuma: These two tests duplicate nearly the full `ProofJob` literal just to vary timestamps. A small local `proof_job(created_at, completed_at)` helper would keep the test focused on the metric behavior and avoid maintaining the same 20-field fixture twice

crates/proof/prover-service/src/metrics.rs
```diff
@@ -149,3 +183,155 @@ pub const fn api_proof_type_label(proof_type: ApiProofType) -> &'static str {
...
+            find_histogram(
+                &snapshot,
+                PROOF_REQUEST_DURATION_MS,
+                &[("status", PROOF_STATUS_SUCCEEDED), ("proof_type", "tee")],
+            ),
+            Some(vec![0.0]),
+        );
+    }
+
+    fn find_counter(
```
jackchuma: The `find_counter` / `find_histogram` / `has_labels` helpers create a mini assertion API for only a couple of checks. I’d inline the `snapshot.iter().find_map(...)` or `any(...)` at the assertion sites; less code, and the test remains easy to read

crates/proof/challenge/src/validator.rs
```diff
@@ -853,4 +862,60 @@ mod tests {
...
+        });
+
+        let snapshot = snapshotter.snapshot().into_vec();
+        assert_eq!(find_counter(&snapshot, "base_challenger.games_invalid_total"), Some(1));
+        assert!(
+            find_histogram(&snapshot, "base_challenger.validation_latency_seconds").is_some(),
+            "checkpoint validation should record latency",
+        );
+
+        type SnapEntry =
```
jackchuma: Same pattern here: `SnapEntry`, `find_counter`, and `find_histogram` are one-off helpers inside a single test. Inline the two snapshot checks so this test doesn’t grow extra local abstraction just to inspect two metrics

crates/proof/challenge/src/driver.rs
```diff
@@ -443,17 +447,19 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
                 challenged_index = challenged_index,
                 on_chain = %on_chain_root,
-                expected = %expected_root,
+                expected = %validation.expected_root,
                 "ZK challenge is legitimate (challenged root was wrong), skipping"
             );
             return Ok(());
         }
 
+        let expected_root = validation.expected_root;
```
jackchuma: `expected_root` is only used once immediately below. This can pass `validation.expected_root` directly into `initiate_zk_proof(...)` and drop the temporary

jackchuma (approved):

PR #3693, opened 2026-06-22
fix(precompile-storage): clear stale tail chunks in store_bytes_like on shrink

eric-ships:
[BOP-339](https://linear.app/coinbase/issue/BOP-339)

## Motivation

`store_bytes_like` (used by `String::store` and `Bytes::store`) did not clear data slots belonging to a previous longer value when overwriting with a shorter one. The `delete_bytes_like` function already handles this correctly: it loads the prior encoded length, then zeros every chunk slot before clearing the base slot. `store_bytes_like` skipped that read entirely, so any shrinking write left ghost bytes in raw storage.

The affected fields with a reachable post-initialization mutation path are `name`, `symbol`, and `contract_uri` on `B20CoreStorage`, all updatable via the `Configurable` trait by accounts with the `Metadata` role. Typed reads (`name()`, `symbol()`, `contract_uri()`) are always length-bounded and return correct values; the stale bytes are only visible via raw slot inspection (`eth_getStorageAt`). The practical impact is missed EIP-2200 gas refunds and ghost data in raw storage explorers.

This was partially addressed in https://github.com/base/base/pull/3384 for `Vec<T>` (via explicit `VecHandler::truncate` / `VecHandler::clear` helpers), but `bytes_like.rs` was not updated at the same time.

## What changed

- `store_bytes_like`: before writing the new value, load the existing base slot. If it encodes a long string, compute `old_chunks` vs `new_chunks` and zero the range `new_chunks..old_chunks`. This mirrors `delete_bytes_like` and adds one extra `SLOAD` only when the previous value was a long string.

## What was considered

Reusing `delete_bytes_like` to fully clear the slot before every store was rejected: it always zeros the base slot and every chunk, so a same-or-growing write would pay redundant `SSTORE`s. Bounding the clear to the retired tail range (`new_chunks..old_chunks`) keeps the common non-shrinking path at zero extra writes.

## Tests

Four regression tests added to `bytes_like.rs`, each verifying the raw slot contents after a shrinking write:

- `test_long_to_short_clears_stale_chunks`: 64-byte `String` overwritten with a short (<=31 byte) value; both old data chunks must be zero.
- `test_long_to_shorter_long_clears_tail_chunk`: 96-byte `String` overwritten with a 32-byte value (1 chunk); tail chunks 1 and 2 must be zero, chunk 0 preserved.
- `test_long_to_empty_clears_all_chunks`: 64-byte `String` overwritten with `""`.
- `test_bytes_long_to_short_clears_stale_chunks`: same scenario for `Bytes`.

### DIFF
--- a/crates/common/precompile-storage/src/types/bytes_like.rs
+++ b/crates/common/precompile-storage/src/types/bytes_like.rs
@@ -211,19 +211,40 @@ where
 
 #[inline]
 fn store_bytes_like<S: StorageOps>(bytes: &[u8], storage: &mut S, base_slot: U256) -> Result<()> {
-    let length = bytes.len();
-    if length <= 31 {
+    let new_length = bytes.len();
+
+    // If the previous value was a long string, zero any tail chunks that the new
+    // (shorter) value will not overwrite. Without this, shrinking writes leave
+    // stale data in raw storage even though typed reads are length-bounded.
+    let old_base = storage.load(base_slot)?;
+    if is_long_string(old_base) {
+        let old_chunks = calc_chunks(calc_string_length(old_base, true)?);
+        let new_chunks = if new_length > 31 { calc_chunks(new_length) } else { 0 };
+        if new_chunks < old_chunks {
+            let slot_start = calc_data_slot(base_slot);
+            for i in new_chunks..old_chunks {
+                storage.store(
+                    slot_start
+                        .checked_add(U256::from(i))
+                        .ok_or(BasePrecompileError::SlotOverflow)?,
+                    U256::ZERO,
+                )?;
+            }
+        }
+    }
+
+    if new_length <= 31 {
         storage.store(base_slot, encode_short_string(bytes))
     } else {
-        storage.store(base_slot, encode_long_string_length(length))?;
+        storage.store(base_slot, encode_long_string_length(new_length))?;
         let slot_start = calc_data_slot(base_slot);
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #3712, opened 2026-06-23
fix: seed AnchorStateRegistry with real L2 genesis output root (PRIV-2006)

robriks:
## Summary

The dev-multiproof L3 settles to an **anchor** — the L2 genesis output root — that is only known *after* L2 genesis. The main deploy runs before genesis, so it seeded the `AnchorStateRegistry` with a placeholder root (`0x00..01`). The TEE prover then rejects every proof built from that anchor (`Output root does not match L2 head`), the proposer creates **zero dispute games**, and L3→L1 withdrawals never finalize.

This is the devnet-scripts half of the fix; the contract-side deferral is in base/contracts#349.

Linear: [PRIV-2006](https://linear.app/coinbase/issue/PRIV-2006)

## Changes

- **`templates/deploy-config.json.template`**: make `multiproofGenesisOutputRoot` substitutable (`${MULTIPROOF_GENESIS_OUTPUT_ROOT}`) instead of a hard-coded `0x..01`.
- **`setup-l2.sh`**: export a non-zero placeholder default for `MULTIPROOF_GENESIS_OUTPUT_ROOT`. The main deploy still needs a non-zero `startingAnchorRoot` to pass `SystemDeploy`'s validation, and the real root is unknowable pre-genesis.
- **`register-aggregate-verifier.sh`** (post-genesis one-shot): compute the **real** L2 genesis output root from the L2 EL and inject it into the regenerated deploy-config, so `SystemDeploy.registerAggregateVerifier(bytes32)` initializes the `AnchorStateRegistry` with the true anchor before deploying the verifier.

## Output-root derivation

op-node is being deprecated, so the root is computed directly from the L2 EL using the standard OP formula at block 0:

```
keccak256(version(0x0) ++ stateRoot ++ L2ToL1MessagePasser.storageHash ++ blockHash)
```

via `cast block 0`, `cast proof 0x42..16 --block 0`, and `cast keccak`. The script waits for the L2 RPC, validates each field is `0x` + 64 hex, and fails loudly otherwise.

## Companion PRs

- base/contracts#349 — defers ASR `initialize()` in `SystemDeploy` to `registerAggregateVerifier`.
- privacy-enclave — vendored `setup-l2.sh` mirror + `services.yml` (pass `L2_RPC_URL` to the one-shot and depend on `base-rpc`).

### DIFF
--- a/etc/scripts/devnet/register-aggregate-verifier.sh
+++ b/etc/scripts/devnet/register-aggregate-verifier.sh
@@ -13,7 +13,7 @@ set -euo pipefail
 # real hash and points the DisputeGameFactory's game type at it.
 
 L1_RPC_URL="${L1_RPC_URL:-http://l1-el:4545}"
-L2_RPC_URL="${L2_RPC_URL:-http://base-rpc:8545}"
+L2_RPC_URL="${L2_RPC_URL:-http://base-rpc:8645}"
 L1_CHAIN_ID="${L1_CHAIN_ID:-1337}"
 OUTPUT_DIR="${OUTPUT_DIR:-/devnet/l2/configs}"
 TEMPLATE_DIR="${TEMPLATE_DIR:-/templates}"
--- a/etc/scripts/devnet/register-aggregate-verifier.sh
+++ b/etc/scripts/devnet/register-aggregate-verifier.sh
@@ -13,12 +13,15 @@ set -euo pipefail
 # real hash and points the DisputeGameFactory's game type at it.
 
 L1_RPC_URL="${L1_RPC_URL:-http://l1-el:4545}"
+L2_RPC_URL="${L2_RPC_URL:-http://base-rpc:8545}"
 L1_CHAIN_ID="${L1_CHAIN_ID:-1337}"
 OUTPUT_DIR="${OUTPUT_DIR:-/devnet/l2/configs}"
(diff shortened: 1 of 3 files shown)


etc/scripts/devnet/register-aggregate-verifier.sh
```diff
@@ -13,12 +13,15 @@ set -euo pipefail
 # real hash and points the DisputeGameFactory's game type at it.
 
 L1_RPC_URL="${L1_RPC_URL:-http://l1-el:4545}"
+L2_RPC_URL="${L2_RPC_URL:-http://base-rpc:8545}"
```
hughy: Is this the right port for the default? I think the default listener port for the Base RPC migh be 8645
```suggestion
L2_RPC_URL="${L2_RPC_URL:-http://base-rpc:8645}"
```

hughy (commented):
minor improvement we could make would be to update the docker-compose so that `register-aggregate-verifier` has an explicit `depends_on` for `base-rpc`. We're probably ok without it because of the retry loop in the script

hughy (approved):
minor improvement we could make would be to update the docker-compose so that `register-aggregate-verifier` has an explicit `depends_on` for `base-rpc`. We're probably ok without it because of the retry loop in the script

PR #3721, opened 2026-06-23
refactor(proposer): clean up proof collector

jackchuma:
### What changed? Why?

Refactors the proposer proof collector into explicit collector state, target retry state, and a `ProofCollectorOrchestrator` that owns prover-service polling, dispatch retries, and sequential submit handling.

Keeps the proving pipeline responsible for running dispatcher and collector loops while moving collector-specific retry/session behavior out of the pipeline.

### Notes to reviewers

The main behavior to review is retry/session preservation around failed proofs, missing prover-service sessions, discard retries, and submit-timeout restarts.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -1,1980 +1,648 @@
...
+                    Ok(Some(proof)) => proof,
+                    Ok(None) => return false,
+                    Err(error) => {
+                        warn!(
+                            target_block,
+                            session_id = %session_id,
+                            error = %error,
+                            "Deleting proof request after proof collection failure"
+                        );
+                        return self.delete_proof_request(&session_id, target_block).await;
```
leopoldjoy: Deleting a terminal failed proof and restarting here bypasses any per-target retry accounting, so a deterministic prover failure can be re-dispatched forever instead of exhausting max_retries; is that intentional despite max_retries being documented as covering proof failures?

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -1,1980 +1,648 @@
...
+                warn!(target_block, error = %error, "Submission failed");
+            }
+            Err(SubmitAction::Discard(error)) => {
+                Metrics::errors_total(error.metric_label()).increment(1);
+                warn!(
+                    target_block,
+                    error = %error,
+                    "Submission discarded, deleting proof request for re-prove"
+                );
+                return if self.delete_proof_request(session_id, target_block).await {
```
leopoldjoy, line 352: Discarded proofs are retried by deleting and reusing the root-derived session with no discard retry counter or retry-specific session id, so permanent discard causes like invalid signer or stale parent can loop unbounded, again, is that likely acceptable behaviour given that games are deterministic so in this case the system is essentially borked?

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -1,1980 +1,648 @@
...
-                });
-                TargetPoll::Failed {
-                    session_id,
-                    claimed_l2_output_root,
-                    error: ProposerError::Prover(message),
-                }
+            Err(SubmitAction::RootMismatch) => {
+                Metrics::root_mismatch_total().increment(1);
+                warn!(target_block, "Output root mismatch at submit time");
+                return if self.delete_proof_request(session_id, target_block).await {
```
leopoldjoy: A root mismatch should force the pipeline to restart even if deleting the stale proof request fails, otherwise the dispatcher can remain advanced past this target while the collector starts looking for a new root-derived session that was never dispatched.

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -1,1980 +1,648 @@
...
-                        }
-                    };
                 debug!(
                     target_block,
                     session_id = %session_id,
-                    "Retry session missing from prover service, dispatch needed",
+                    "Proof request not found yet"
                 );
-                return TargetPoll::NotFound { session_id, claimed_l2_output_root };
+                return Ok(None);
```
leopoldjoy: Treating prover-service `NotFound` exactly like a pending proof gives the pipeline no way to distinguish a missing session from normal progress, which can stall collection after the dispatcher has advanced.

crates/proof/proposer/src/proof_collector.rs
jackchuma: Yes that's intended. If we run into a deterministic prover failure, it will require manual intervention to unblock

crates/proof/proposer/src/proof_collector.rs
jackchuma: Yes I think that is likely acceptable behavior. If we run into permanent discard causes, we will need manual intervention to unblock

crates/proof/proposer/src/proof_collector.rs
```diff
@@ -1,1980 +1,738 @@
...
+                Ok(Some(proof)) => proof,
+                Ok(None) => return false,
+                Err(error) => {
+                    warn!(
+                        target_block,
+                        session_id = %session_id,
+                        error = %error,
+                        "Deleting proof request after proof collection failure"
+                    );
+                    return self.delete_proof_request(&session_id, target_block).await;
```
leopoldjoy, line 134: Since terminal collector-side proof failures intentionally restart here without consuming `max_retries`, can we update the `DriverConfig::max_retries` docs to clarify that the retry budget only bounds dispatcher-side proof request failures?

crates/proof/proposer/src/pipeline.rs
```diff
@@ -133,10 +161,21 @@ where
...
-                    self.proof_collector_orchestrator
-                        .tick(&mut state, &mut cache, recovered, safe_head, cancel)
+                    if cursor_source != Some(recovered) || cursor.is_none() {
+                        cursor_source = Some(recovered);
+                        cursor = Some(recovered);
+                    }
+
+                    let current =
+                        cursor.as_mut().expect("collector cursor initialized from recovered state");
+                    self.proof_collector
```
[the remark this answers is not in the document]
jackchuma: I'll do this in a follow-up

leopoldjoy (approved):

PR #3722, opened 2026-06-23
fix(eip8130): keep enshrined system accounts non-empty under EIP-161 via Cobalt code stub

chunter-cb:
## Summary

The enshrined EIP-8130 execution path writes persistent state (the 2D-nonce channels in the native `NonceManager`, account configuration, etc.) through the journal's `sstore`, which only mutates the **storage trie** — it never sets nonce/balance/code. A code-less system account therefore stays EIP-161-"empty" (`code == ∅ && nonce == 0 && balance == 0`) and is **reaped together with its storage** by end-of-block state clearing, silently discarding the enshrined state.

The native `NonceManager` (`0x8130…aa01`) is the account that hits this: it holds persistent storage but carries **no code on any chain**, so it is empty-with-storage even on mainnet.

## Fix

Make the code-less enshrined system accounts EIP-161-**non-empty by giving them code**, via an irregular state transition at the Cobalt (EIP-8130) activation — mirroring the existing Canyon create2-deployer force-deploy. `ensure_eip8130_system_accounts` plants a one-byte `0xEF` stub on the `NonceManager` account on the first Cobalt block:

- **`0xEF`** is the EIP-3541-reserved prefix: it can never come from a real `CREATE`/`CREATE2`, so it is an unambiguous protocol-system-account sentinel. The address is still serviced by the native precompile, not by executing the byte.
- **Idempotent & non-destructive** — gated on `is_cobalt_active` + empty code hash, so it fires once and never overwrites a real deployment.
- **Protects every write path** — once the account has code, both the gas-free journal provider and `EvmPrecompileStorageProvider` are safe, with no per-`sstore` guard.

This supersedes the earlier per-`sstore` nonce-bump guard, which is removed (along with its four tests). The unrelated `journal.rs` `with_account_info` clone-removal nit is retained.

### Why nothing else needs a stub
- The transaction-context precompile (`0x8130…aa02`) uses **transient** storage only — cleared every transaction, never trie-resident.
- `AccountConfiguration` is a genuinely **deployed** contract (it carries code) on every chain where EIP-8130 is enabled; it is expected to be deployed before Cobalt is ready, so it needs no special handling.

### Note
Once Cobalt is live this makes `EXTCODESIZE(0x8130…aa01) == 1` and `EXTCODEHASH == keccak(0xEF)`, where the `NonceManager` previously reported as code-less. That is the intended trade-off of the code-stub approach.

## Test plan

* `cargo test -p base-common-evm cobalt` — 3 tests: stub planted when Cobalt active; no-op when inactive; existing deployed code never overwritten.
* `cargo test -p base-precompile-storage --lib` — journal/storage suite green after guard removal.
* `cargo test -p base-common-evm --lib executor` — `eip8130_transaction_is_included_with_typed_receipt` still passes with the new pre-execution hook.
* `cargo clippy -p base-common-evm -p base-precompile-storage --all-targets`

### DIFF
--- a/crates/common/evm/src/executor/block_executor.rs
+++ b/crates/common/evm/src/executor/block_executor.rs
@@ -144,6 +144,17 @@ where
         )
         .map_err(BlockExecutionError::other)?;
 
+        // At the Cobalt (EIP-8130) transition, plant a code stub on the code-less
+        // enshrined system accounts (the 2D nonce manager) so the persistent state
+        // the enshrined path writes to them is not reaped by EIP-161 end-of-block
+        // state clearing.
+        crate::cobalt::ensure_eip8130_system_accounts(
+            &self.spec,
+            self.evm.block().timestamp().saturating_to(),
+            self.evm.db_mut(),
+        )
+        .map_err(BlockExecutionError::other)?;
+
         Ok(())
     }
 
--- a/crates/common/evm/src/cobalt.rs
+++ b/crates/common/evm/src/cobalt.rs
@@ -1,0 +1,145 @@
+use alloy_evm::Database;
+use alloy_primitives::{Address, Bytes};
+use base_common_chains::Upgrades;
+use base_common_precompiles::NonceManagerStorage;
+use revm::{DatabaseCommit, primitives::HashMap, state::Bytecode};
+
+/// Single-byte code stub planted on otherwise code-less EIP-8130 system accounts.
+///
+/// `0xEF` is the EIP-3541 reserved prefix: it can never be produced by a normal
+/// `CREATE`/`CREATE2` deployment, so it is an unambiguous "this is a protocol
+/// system account" sentinel. Its only job is to give the account a non-empty
+/// code hash; the address is still serviced by the native precompile, not by
+/// executing this byte.
+const SYSTEM_ACCOUNT_STUB: [u8; 1] = [0xEF];
+
+/// Code-less EIP-8130 system accounts that hold persistent storage but carry no
+/// code on any chain, and therefore must be made non-empty so EIP-161
+/// end-of-block state clearing does not reap them together with their storage.
+///
+/// Only the [`NonceManager`](NonceManagerStorage) qualifies: it persists the 2D
+/// nonce channels in the state trie while never being a deployed contract. The
+/// transaction-context precompile (`0x8130…aa02`) uses transient storage only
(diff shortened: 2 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/evm/src/executor/block_executor.rs
```diff
@@ -144,6 +144,17 @@ where
         )
         .map_err(BlockExecutionError::other)?;
 
+        // At the Cobalt (EIP-8130) transition, plant a code stub on the code-less
+        // enshrined system accounts (the 2D nonce manager) so the persistent state
+        // the enshrined path writes to them is not reaped by EIP-161 end-of-block
+        // state clearing.
+        crate::cobalt::ensure_eip8130_system_accounts(
```
xenoliss, line 151: Flashblocks has its own `apply_pre_execution_changes` (`flashblocks/src/state_builder.rs:192`) mirroring this one and already replicates `ensure_create2_deployer`, but not this new call. So post-Cobalt the sequencer won't plant the `NonceManager` stub and diverges from the validator. Add the same call there, after create2.

xenoliss (approved):

PR #3726, opened 2026-06-23
feat(upgrade-signal): wire upgrade-signal support into consensus nodes

PelleKrab:
## Summary

  - wires L1 upgrade-signal schedule reads into standalone and integrated consensus startup
  - adds consensus runtime admin refresh support for applying upgrade-signal schedules
  - applies startup upgrade-signal schedules to both EL and CL runtime views in integrated nodes
  - defaults integrated upgrade-signal L1 RPC config from the consensus L1 RPC when no explicit upgrade-signal RPC is provided
  - adds separate opt-in EL and CL upgrade-signal metrics with `layer` and `upgrade` labels
  - routes startup apply and runtime admin refresh through shared upgrade-signal schedule validation/application helpers
  - renames public upgrade-signal surfaces from hardfork terminology to upgrade terminology
  - narrows `base-upgrade-signal` exports

  ## Testing

  - `cargo fmt --check`
  - `git diff --check origin/main..HEAD`
  - `cargo check -p base-upgrade-signal -p base-consensus-node -p base-execution-cli -p base-node-core -p base`
  - `cargo test -p base-upgrade-signal --lib`
  - `cargo test -p base-consensus-cli -p base-execution-cli -p base-node-core --lib upgrade_signal`
  - `cargo test -p base --bin base upgrade_signal`
  - tested on devnet


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

bin/base/src/commands/integrated_upgrade_signal.rs
refcell, line 1: I'd prefer if we placed this logic behind types (following `AGENTS.md` standards) in the `upgrade-signal` utility crate. This way we keep the `base` binary crate lightweight and allow the logic to be re-used easier later.

crates/consensus/service/src/actors/l1_watcher/query_processor.rs
```diff
@@ -80,7 +80,7 @@ where
         query_started_at: Instant,
         sender: oneshot::Sender<RollupConfig>,
     ) {
-        if let Err(error) = sender.send((*self.rollup_config).clone()) {
+        if let Err(error) = sender.send(self.rollup_config.with_runtime_upgrade_overrides()) {
```
refcell, line 83: Is `.with_runtime_upgrade_overrides` expensive here? We don't want to do any expensive operations here ideally

crates/consensus/service/src/actors/upgrade_signal.rs
```diff
@@ -0,0 +1,76 @@
...
+        loop {
+            tokio::select! {
+                _ = cancellation.cancelled() => return Ok(()),
+                _ = interval.tick() => {
+                    tokio::select! {
+                        _ = cancellation.cancelled() => return Ok(()),
+                        _ = self.poll_l1_signal() => {}
+                    }
+                }
+            }
```
refcell, line 126: This is ok, but we could flatten the nesting here to make it a bit more legible. e.g.

```rust
async fn start(mut self, _ctx: ()) -> Result<(), Self::Error> {
      let cancellation = self.cancellation.clone();
      let mut interval = tokio::time::interval(UpgradeSignalDefaults::POLL_INTERVAL);
      interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);

      loop {
          // Wait until the interval is reached before polling
          tokio::select! {
              _ = cancellation.cancelled() => return Ok(()),
              _ = interval.tick() => {}
          }

          // Interval reached, poll l1 signal
          tokio::select! {
              _ = cancellation.cancelled() => return Ok(()),
              _ = self.poll_l1_signal() => {},
          }
      }
  }
```

crates/consensus/service/src/service/builder.rs
```diff
@@ -83,6 +84,12 @@ pub struct RollupNodeBuilder {
     /// When set, enables persistent safe head tracking via redb and serves
     /// `optimism_safeHeadAtL1Block` RPC requests from the database.
     pub safedb_path: Option<PathBuf>,
+    /// Optional L1 upgrade signal metrics observer configuration.
+    pub upgrade_signal_metrics_config: Option<UpgradeSignalConfig>,
+    /// Optional L1 RPC endpoint override for upgrade signal reads.
+    pub upgrade_signal_l1_rpc: Option<Url>,
+    /// Optional runtime upgrade signal validation context.
+    pub upgrade_signal_runtime_validation: Option<UpgradeSignalRuntimeValidation>,
```
refcell, line 92: Maybe this could just be collapsed into one config type containing all 3 of these? Not worth addressing in this PR, but maybe open a KTLO ticket on linear to track for later.

crates/consensus/service/src/service/node.rs
```diff
@@ -532,9 +544,38 @@ impl RollupNode {
...
+                UpgradeSignalRefresher::new(
+                    config,
+                    upgrade_signal_l1_provider.clone(),
+                    self.config.l2_chain_id.id(),
+                    runtime_validation,
+                    UpgradeSignalMetricLayer::Consensus,
+                )
+            })
+        });
+        let node_mode = self.mode();
```
refcell, line 542: I think this can be hoisted / abstracted out of the node startup logic. The start method is already too bloated and this module is getting too complex. We should just have one upgrade_signal_config that we can construct the upgrade signal metrics actor from and the refresher from.

Ideally something like:

```rust
let upgrade_signal_refresher = self.upgrade_signal_config.refresher();
let metrics_actor = self.upgrade_signal_config.metrics_actor();
```

crates/utilities/upgrade-signal/src/config/args.rs
```diff
@@ -53,128 +53,168 @@ pub struct UpgradeSignalArgs {
...
     )]
     pub l1_block_tag: UpgradeSignalBlockTag,
+
+    /// Enable execution-layer live upgrade-signal metrics.
+    #[arg(long = "upgrade-signal.el-metrics", env = "BASE_NODE_UPGRADE_SIGNAL_EL_METRICS")]
+    pub el_metrics_enabled: bool,
+
+    /// Enable consensus-layer live upgrade-signal metrics.
+    #[arg(long = "upgrade-signal.cl-metrics", env = "BASE_NODE_UPGRADE_SIGNAL_CL_METRICS")]
+    pub cl_metrics_enabled: bool,
```
refcell: Let's remove this. We're trying to make our cli increasingly opinionated so there's less config required for our nodes. Since metrics are very lightweight, we should always enable them.

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -17,34 +18,72 @@ pub struct UpgradeSignalConfig {
...
     /// Local schedule mutation mode.
     pub mode: UpgradeSignalMode,
     /// L1 block tag used to read the contract.
     pub l1_block_tag: BlockNumberOrTag,
     /// Node protocol version supported by this binary.
     pub node_protocol_version: U256,
+    /// Whether execution-layer metric recording is enabled.
+    pub el_metrics_enabled: bool,
+    /// Whether consensus-layer metric recording is enabled.
+    pub cl_metrics_enabled: bool,
```
refcell: rm

refcell (changes requested):
Looking very good. Left a few comments that I think are worth addressing before we merge.

cb-heimdall:
Review Error for refcell @ 2026-06-24 11:38:25 UTC

bin/base/src/commands/rpc.rs
```diff
@@ -48,42 +50,74 @@ impl RpcCommand {
...
+                    &execution.standard.rollup_args.upgrade_signal_l1_rpc,
+                )?;
+                integrated_upgrade_signal::apply_startup_signal(
+                    &mut execution_chain,
+                    &mut rollup_config,
+                    &signal_config,
+                    l1_rpc,
+                    "integrated startup",
+                )
+                .await?;
```
refcell: If this fails, it will prevent the node from starting entirely. This is ok for now since we won't be running nodes with apply at startup mode right?

bin/base/src/commands/sequencer.rs
```diff
@@ -47,49 +52,70 @@ impl SequencerCommand {
...
+                )?;
+                integrated_upgrade_signal::apply_startup_signal(
+                    &mut execution_chain,
+                    &mut rollup_config,
+                    &signal_config,
+                    l1_rpc,
+                    "integrated sequencer startup",
+                )
+                .await?;
+            }
```
refcell: All this startup logic is duplicated between the rpc and sequencer commands. Maybe this could be refactored?

crates/utilities/upgrade-signal/src/config/args.rs
```diff
@@ -14,27 +14,27 @@ pub struct UpgradeSignalArgs {
...
     /// permits schedule mutation.
     #[arg(
-        long = "upgrade-signal.apply-hardfork-id",
-        env = "BASE_NODE_UPGRADE_SIGNAL_APPLY_HARDFORK_ID",
+        long = "upgrade-signal.apply-upgrade-id",
+        env = "BASE_NODE_UPGRADE_SIGNAL_APPLY_UPGRADE_ID",
         value_delimiter = ','
     )]
-    pub apply_hardfork_ids: Vec<String>,
+    pub apply_upgrade_ids: Vec<String>,
```
refcell: I think there is a subtle runtime issue here when `upgrade-signal.apply-upgrade-id` is configured to a subset of upgrades. Let me walk through what I'm seeing happens.

The runtime refresh path reads the L1 schedule for `upgrade_ids`, then filters it down to `apply_upgrade_ids` via `application_schedule()`. The filtered schedule is then passed into `UpgradeSignalRuntimeApplier::apply_schedule()`, which creates a fresh `RuntimeRegistrySink`.

When `apply_schedule_to_sink()` finishes, it calls `staged_sink.finalize()`, and `RuntimeRegistrySink::finalize()` commits by calling `RuntimeUpgradeRegistry::replace_overrides(chain_id, updates)` or `clear_chain()` if the filtered updates are empty. So the filtered application schedule is treated as the complete replacement set for the whole chain.

This means a runtime refresh can unintentionally drop existing runtime overrides for upgrades that were not included in `apply_upgrade_ids`. For example, if the registry currently has `{Azul: 100, Cobalt: 300}`, and the node is configured to read `[Azul, Beryl, Cobalt]` but apply only `[Beryl]`, then the L1 read may correctly return all three signals, but `application_schedule()` reduces it to only `{Beryl: 200}`. `finalize()` then replaces the entire chain registry with just `{Beryl: 200}`, removing Azul and Cobalt even though L1 did not clear them and the user only scoped which upgrade should be applied.

I'm not sure we should allow this behavior.

One option is to just remove `apply_upgrade_ids` and use `upgrade_ids`, requiring the applied schedule to be complete. Alternatively, we could support a subset by making the runtime path merge touched IDs instead of replacing the whole registry, e.g. update only the IDs present in the filtered schedule and avoid calling `clear_chain()` for an empty scoped schedule. If both full replacement and scoped updates are desired, the code probably needs to carry that distinction explicitly so `finalize()` knows whether it is committing a complete schedule or only a selected subset.

bin/base/src/commands/rpc.rs
```diff
@@ -48,42 +50,74 @@ impl RpcCommand {
...
+            }
+
+            if metrics_enabled {
+                CliMetrics::init_rollup_config(&rollup_config);
+            }
             let _upgrade_countdown_metrics = metrics_enabled
                 .then(|| CliMetrics::spawn_upgrade_countdown_recorder(rollup_config.clone()));
```
refcell, line 93: So we have these metrics which basically the number of seconds until an upgrade activates. I think this needs to be updated with the dynamic upgrade activation timestamps in mind. Otherwise, we'll have incorrect metrics for when upgrades will activate if the timestamp is changed on L1 and the node isn't restarted.

refcell (changes requested):
Took another look and I think there's a couple issues with how we configure applying upgrades as well as blocking node startup on failure.

bin/base/src/commands/rpc.rs
PelleKrab: Correct.

crates/consensus/service/src/actors/l1_watcher/query_processor.rs
PelleKrab: The most expensive part is a clone, we clone `RollupConfig` and then we do a pass over to update any timestamp changes. So as it stands its pretty cheap, but we could definitely get this back to a single clone in future. I think it would be good to do this in a separate Dynamic upgrade activation optimization PR, since there are some other changes I would like to make.

crates/utilities/upgrade-signal/src/config/args.rs
PelleKrab: I went for the full schedule apply to keep it simple for now, as we don't gain too much from individual upgrade reads. We may want to add a read full schedule to the smart contract to further simplify it.

crates/consensus/service/src/actors/l1_watcher/query_processor.rs
refcell: Fine with that!

bin/base/src/commands/rpc.rs
```diff
@@ -48,41 +50,72 @@ impl RpcCommand {
...
+            .apply_default_from(&consensus_config.l1_rpc_args.l1_eth_rpc);
+        consensus_config.upgrade_signal = execution.standard.rollup_args.upgrade_signal.clone();
+        let consensus_args = ConsensusNodeArgs::new(consensus_chain, consensus_config);
+        let mut rollup_config = consensus_args.load_rollup_config()?;
 
         CliRunner::try_default_runtime()?.run_command_until_exit(|ctx| async move {
+            let upgrade_signal_runtime_validation =
+                UpgradeSignalRuntimeValidation::with_activation_admin_address(
+                    execution_chain.activation_admin_address,
+                );
```
refcell, line 74: I think we should be able to remove this now. In cobalt, the activation admin address will be in L2 state so it won't be a `ChainConfig` owned address. As such, I think we should be able to simplify this. Can be done in a follow on, maybe just track it with a linear ticket please so it doesn't get lost

refcell (approved):
Happy to land this as is and clean up in follow-ons

PR #3731, opened 2026-06-23
fix(flashblocks): guard against panic on empty flashblocks vec after reorg filter

### LINKED ISSUE #3728: bug(flashblocks): HandleReorg and DepthLimitExceeded can panic with empty flashblocks vec going into build_pending_state
Found this while tracing the flashblocks reconciliation paths. Both `HandleReorg` and `DepthLimitExceeded` filter the flashblocks vec with `.retain()` before passing it to `build_pending_state`, but neither checks whether the vec is empty afterward.

`crates/execution/flashblocks/src/processor.rs`:

- Line 289-291: `HandleReorg` retains only flashblocks > block.number, then calls `build_pending_state(None, &flashblocks)`
- Line 299-301: `DepthLimitExceeded` does the same
- Line 701: `build_pending_state` does `flashblocks_per_block.keys().min().unwrap()` on the BTreeMap built from that vec

If the retain filter removes everything (all pending flashblocks are at or below the reorg block), the vec is empty, the BTreeMap is empty, `min()` returns `None`, and the node panics.

Trigger conditions: a reorg or depth-limit reset where all pending flashblocks have `block_number <= block.number`. This isn't an edge case during a reorg, it's likely that the pending flashblocks are from around the same height as the reorg point. If none of them are strictly ahead, the node crashes instead of handling the reorg gracefully.

Quick fix:

```rust
flashblocks.retain(|flashblock| flashblock.metadata.block_number > block.number);
if flashblocks.is_empty() {
    return self.clear_live_state().map(|()| None);
}
```

Same guard on both the HandleReorg and DepthLimitExceeded paths.

Sertug17:
HandleReorg and DepthLimitExceeded filter pending flashblocks with `.retain()` before passing them to `build_pending_state`, which panics via `.keys().min().unwrap()` on an empty BTreeMap. If no flashblock has a block_number strictly greater than the reorg block, the retain produces an empty vec and the node crashes.

Adds an early return with `clear_live_state()` when the filtered vec is empty, matching the existing pattern in the `NoPendingState` and `CanonicalCaughtUp` paths.

Fixes #3728


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

Sertug17:
@danyalprout @refcell @eric-ships @roethke  this is the fix for #3728. The HandleReorg and DepthLimitExceeded paths filter flashblocks without checking if the result is empty, and build_pending_state panics on an empty map.

Would appreciate a review when either of you gets a chance. If the approach looks right, happy to make any changes.

PR #3735, opened 2026-06-23
feat(challenger): Route challenger ZK proofs through requester

mw2000:
Wire the challenger through base-proof-zk-requester so Groth16 proof submission and polling use the requester shim instead of building raw prover-service SNARK requests. The challenger still owns dispute state, retries, and submission intent, while the requester crate owns the range-plus-aggregation proof flow.

### DIFF
--- a/crates/proof/challenge/src/driver.rs
+++ b/crates/proof/challenge/src/driver.rs
@@ -24,8 +24,9 @@ use alloy_primitives::{Address, B256};
 use base_proof_contracts::{AggregateVerifierClient, GameStatus};
 use base_proof_primitives::ProofRequest as TeeProofRequest;
 use base_proof_rpc::L2Provider;
+use base_proof_zk_requester::{Groth16ProofState, Groth16RangeProofRequest, ZkProofRequester};
 use base_prover_service_client::ProofRequesterProvider;
-use base_prover_service_protocol::{SnarkGroth16ProofRequest, TeeKind, ZkProofRequest, ZkVm};
+use base_prover_service_protocol::{TeeKind, ZkProofRequest, ZkVm};
 use base_runtime::{Clock, TokioRuntime};
 use base_tx_manager::TxManager;
 use tokio::select;
@@ -610,16 +611,20 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
         })
     }
 
-    /// Builds a [`SnarkGroth16ProofRequest`] for the given candidate and invalid index.
+    /// Builds a [`Groth16RangeProofRequest`] for the given candidate and invalid index.
     fn build_zk_request(
         &self,
         candidate: &CandidateGame,
         invalid_index: u64,
-    ) -> eyre::Result<SnarkGroth16ProofRequest> {
+    ) -> eyre::Result<Groth16RangeProofRequest> {
         let start_block_number = candidate.checkpoint_start_block(invalid_index)?;
 
-        Ok(SnarkGroth16ProofRequest {
-            proof: ZkProofRequest {
+        Ok(Groth16RangeProofRequest::new(
+            crate::ChallengerProofAdapter::snark_groth16_session_id(
+                candidate.factory.proxy,
+                invalid_index,
+            ),
+            ZkProofRequest {
                 start_block_number,
                 number_of_blocks_to_prove: candidate.intermediate_block_interval,
                 sequence_window: None,
@@ -627,8 +632,8 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
                 intermediate_root_interval: Some(candidate.intermediate_block_interval),
                 zk_vm: ZkVm::Sp1,
             },
-            prover_address: self.submitter.sender_address(),
-        })
+            self.submitter.sender_address(),
+        ))
     }
 
     /// Requests a ZK proof, stores the session, and polls for the result.
@@ -646,25 +651,23 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
         // needs to cover the single interval that contains the invalid
         // checkpoint: [prior_checkpoint .. invalid_checkpoint].
         let proof_request = self.build_zk_request(&candidate, invalid_index)?;
-        let request = crate::ChallengerProofAdapter::snark_groth16_prove_block_range_request(
-            game_address,
-            invalid_index,
-            proof_request.clone(),
-        );
 
-        let prove_response = self.proof_requester.prove_block_range(request).await?;
+        let prove_response = ZkProofRequester::new(&*self.proof_requester)
+            .request_groth16_proof(&proof_request)
+            .await?;
 
         info!(
             game = %game_address,
-            session_id = %prove_response.session_id,
+            range_session_id = %prove_response.range.session_id,
             "proof job initiated"
         );
 
-        let pending = PendingProof::awaiting(
-            prove_response.session_id,
+        let pending = PendingProof::awaiting_zk(
+            proof_request.parent_session_id.clone(),
             invalid_index,
             expected_root,
             proof_request,
+            prove_response.state,
             intent,
         );
         self.pending_proofs.insert(game_address, pending);
@@ -840,8 +843,6 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
         };
 
         let retry_count = pending.retry_count;
-        let invalid_index = pending.invalid_index;
-
         if retry_count > Self::MAX_PROOF_RETRIES {
             warn!(
                 game = %game_address,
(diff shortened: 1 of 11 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/challenge/src/pending.rs
```diff
@@ -269,6 +268,69 @@ impl PendingProofs {
             return Ok(Some(ProofUpdate::NeedsRetry));
         }
 
+        let zk_prove_request = match &pending.kind {
+            ProofKind::Zk { prove_request } => Some(prove_request.clone()),
```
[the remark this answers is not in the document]
mw2000: Minor finding, I think its worth fixing in a larger refactor but don't want it to block this PR

crates/proof/challenge/src/pending.rs
```diff
@@ -269,6 +268,69 @@ impl PendingProofs {
...
+            ProofKind::Tee { .. } => None,
+        };
+
+        if let Some(prove_request) = zk_prove_request {
+            let status =
+                ZkProofRequester::new(proof_requester).groth16_status(&prove_request).await?;
```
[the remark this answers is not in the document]
mw2000: This is acknowledged and should be fine

0xth4nh (dismissed):

crates/proof/zk/requester/src/request.rs
```diff
@@ -24,21 +24,21 @@ pub struct Groth16RangeProofRequest {
...
     /// Return the child session id used for the range proof stage.
     pub fn range_session_id(&self) -> String {
-        format!("{}:{RANGE_SESSION_LABEL}", self.session_id)
+        format!("{}:{RANGE_SESSION_LABEL}", self.parent_session_id)
     }
 
     /// Return the child session id used for the Groth16 aggregation proof stage.
     pub fn aggregation_session_id(&self) -> String {
-        format!("{}:{AGGREGATION_SESSION_LABEL}", self.session_id)
+        format!("{}:{AGGREGATION_SESSION_LABEL}", self.parent_session_id)
```
leopoldjoy, line 41: This aggregation child is still a standalone `SnarkGroth16` request that does not reference the submitted `:range` session, so the new flow adds redundant range work and lets an unrelated range result gate a valid Groth16 challenger proof.

crates/proof/zk/requester/src/requester.rs
```diff
@@ -93,6 +114,28 @@ where
...
+    ///
+    /// This stateless poll rechecks the range stage before checking the
+    /// aggregation stage. Callers should stop polling after [`Groth16ProofStatus::Succeeded`].
+    pub async fn groth16_status(
+        &self,
+        request: &Groth16RangeProofRequest,
+    ) -> Result<Groth16ProofStatus, ZkProofRequesterError> {
+        match self.groth16_result(request).await {
+            Ok(Some(result)) => Ok(Groth16ProofStatus::Succeeded(result)),
+            Ok(None) => Ok(Groth16ProofStatus::Pending),
```
leopoldjoy: Because both child sessions are submitted up front, returning `Pending` as soon as the range child is queued or running hides a failed or malformed aggregation child until the range completes or the challenger times out.

crates/proof/challenge/src/driver.rs
```diff
@@ -924,4 +920,43 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
+                    range_session_id, ..
+                } = &error
+                {
+                    warn!(
+                        game = %game_address,
+                        range_session_id = %range_session_id,
+                        error = %error,
+                        "aggregation request failed after range request was accepted"
+                    );
+                    if let Err(delete_error) = self
```
leopoldjoy: This cleanup assumes `deleteProofRequest` can cancel an accepted range request, but the real prover-service API only deletes terminal rows, so an aggregation-submit failure still leaves in-flight range work orphaned despite the mock test passing.

crates/proof/challenge/src/pending.rs
jackchuma: Do not gate Groth16 proofs on an unused range request. When the challenger takes the ZK path, this now polls `groth16_status`, which first requires the separately submitted `:range Compressed` request to succeed before it will use the `:aggregation` Groth16 result. In the prover service, a `SnarkGroth16` request already runs its own STARK stage and does not consume this separate compressed request, so if the extra range request is still running, fails, or returns a malformed result while the Groth16 request succeeds, the challenger will keep waiting or retry instead of submitting the valid dispute proof. This adds an unnecessary failure/delay dependency to every ZK challenge

crates/proof/zk/requester/src/requester.rs
```diff
@@ -17,6 +17,27 @@ pub struct Groth16ProofRequestResponse {
...
+        session_id: String,
+        /// Failure message returned by prover-service.
+        message: String,
+    },
+    /// A proof stage returned a success response that could not be used.
+    Malformed {
+        /// Prover-service session identifier for the malformed stage.
+        session_id: String,
+    },
+}
```
jackchuma: This `Groth16ProofStatus` enum looks like an extra abstraction over `groth16_result()` rather than a separate domain concept. It has one production caller, and each variant mostly repackages either `Ok(Some(_))`, `Ok(None)`, or specific `ZkProofRequesterError` variants. I’d cut the enum plus `groth16_status()`, and have `PendingProofs::poll` match directly on `groth16_result()` / `ZkProofRequesterError`. That should also let us drop the re-export and the status-only tests

crates/proof/challenge/src/driver.rs
```diff
@@ -924,4 +920,43 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
+                            error = %delete_error,
+                            "accepted range proof cleanup failed"
+                        );
+                    }
+                }
+
+                Err(error)
+            }
+        }
+    }
```
jackchuma: This driver helper is mostly cleanup plumbing around `ZkProofRequester::request_groth16_proof()`. Since the accepted-range cleanup is part of the Groth16 two-stage request lifecycle, I think it belongs in `base-proof-zk-requester` instead of every caller needing to remember it. Moving the `delete_proof_request` call into `ZkProofRequester::request_groth16_proof()` would let the driver call the requester directly and delete this wrapper

crates/proof/challenge/tests/driver.rs
```diff
@@ -678,6 +696,71 @@ async fn test_step_proof_retry_reuses_deterministic_session_id() {
...
+
+    let state = zk.state.lock().unwrap();
+    assert_eq!(
+        state.delete_proof_request_log.len(),
+        1,
+        "accepted retry range session should be deleted when aggregation submission fails"
+    );
+    assert_eq!(state.delete_proof_request_log[0].session_id, expected_range_session_id);
+}
+
```
jackchuma: These two driver integration tests are asserting cleanup behavior that would be better owned by the Groth16 requester itself. Once the accepted-range cleanup moves into `base-proof-zk-requester`, these can shrink to one requester unit test that simulates aggregation submission failure and asserts the accepted range session is deleted. The driver tests don’t need to know that internal cleanup detail

crates/proof/prover-service/client/src/requester.rs
```diff
@@ -227,6 +227,40 @@ impl ProofRequesterProvider for ProofRequesterClient {
...
+        (**self).delete_proof_request(request).await
+    }
+
+    async fn list_proofs(
+        &self,
+        request: ListProofsRequest,
+    ) -> Result<ListProofsResponse, ProverServiceClientError> {
+        (**self).list_proofs(request).await
+    }
+}
```
jackchuma: This blanket `ProofRequesterProvider for &Client impl` appears to exist only so `ZkProofRequester::new(&*self.proof_requester)` type-checks. That’s a lot of public forwarding surface for one borrow convenience. I’d avoid exporting the blanket impl and instead make `ZkProofRequester` work with a borrowed provider locally, so the adapter stays where the need exists

PR #3737, opened 2026-06-23
feat(devnet): wire prover-service and zk-host to devnet

mw2000:
## Summary
Replace the legacy local zk-prover service with the prover service coordinator and zk host worker in devnet and standalone compose flows. Update local just helpers to submit prover-service JSON-RPC requests against the new endpoint.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

etc/just/zk-prover.just
```diff
@@ -1,67 +1,229 @@
 set positional-arguments := true
 set working-directory := '../..'
 
-_script := "etc/scripts/zk-prover/grpc.sh"
 _ca_compose := "etc/scripts/zk-prover/ca-compose.sh"
```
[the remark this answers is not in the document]
mw2000: Will delete this in a follow up

etc/docker/Justfile
```diff
@@ -188,7 +188,7 @@ _up-devnet compose_profile zk:
         just --justfile {{ source_file() }} down
         just --justfile {{ source_file() }} _build-setup-image
         just --justfile {{ source_file() }} _build-rust-images "devnet" "dev"
-        SP1_PROVER=dry-run docker compose {{ _env }} "${compose_files[@]}" up -d --no-build
+        ZK_BACKEND=dry_run docker compose {{ _env }} "${compose_files[@]}" up -d --no-build
```
leopoldjoy: This devnet startup now exposes prover-service JSON-RPC, but `etc/systems/benches/common/zk_proof.rs` still uses the legacy gRPC `ZkProofClient` and expects `execution_stats`, so the advertised B20 ZK bench path is broken.

etc/just/zk-prover.just
```diff
@@ -1,67 +1,229 @@
...
-# Describe the ZK prover gRPC service
-describe:
-    bash "{{ _script }}" describe "{{ _endpoint }}"
+# Alias for listing JSON-RPC methods
+list:
+    @just --justfile {{ source_file() }} methods
 
 # Request a proof or dry-run for a block range: proof_type=compressed|snark
-prove start_block number_of_blocks='1' proof_type='compressed':
+prove start_block number_of_blocks='1' proof_type='compressed' session_id='' prover_address='' l1_head='':
```
leopoldjoy: This helper still advertises `proof_type=snark`, but both compose flows start only a compressed zk-host worker, so snark requests will stay queued unless a snark worker is added or snark is rejected locally.

etc/just/zk-prover.just
```diff
@@ -1,67 +1,229 @@
...
 _ca_combined_file := "/tmp/zk-prover-ca-certificates.crt"
 _ca_extra_file := "/etc/ssl/certs/zk-prover-extra-ca.pem"
-_ca_service := "base-prover-zk"
-_ca_service_command := "./base-prover-zk"
+_ca_service := "base-prover-zk-host"
+_ca_service_command := "./base-prover-zk-host"
 _ca_system_file := "/etc/ssl/certs/ca-certificates.crt"
 _compose := "-f etc/docker/docker-compose.zk-prover.yml"
-_endpoint := "${ZK_PROVER_ENDPOINT:-localhost:${ZK_PROVER_GRPC_PORT:-9000}}"
+_endpoint := "${PROVER_SERVICE_ENDPOINT:-http://localhost:${PROVER_SERVICE_RPC_PORT:-9000}}"
```
leopoldjoy: Using `PROVER_SERVICE_ENDPOINT` for requester helpers is ambiguous because the zk-host uses the same variable for the worker API on port 9001, so a requester-specific endpoint would avoid silently targeting the wrong RPC surface.

etc/docker/docker-compose.yml
```diff
@@ -624,27 +624,59 @@ services:
...
       <<: *rust-service-build
-      target: zk-prover
-    container_name: base-prover-zk
+      target: prover-service
+    container_name: base-prover-service
+    depends_on:
+      prover-service-postgres:
+        condition: service_healthy
+    ports:
+      - "${PROVER_SERVICE_RPC_PORT}:9000"
```
jackchuma, line 645: Preserve a compatible endpoint for the ZK benchmark. When the devnet is started and `just --justfile etc/docker/Justfile bench b20-zk-proving` is run, port 9000 now points at the JSON-RPC prover-service, but `etc/systems/benches/common/zk_proof.rs` still constructs a `base_zk_client::ZkProofClient` and calls the old gRPC `ProveBlock`/`GetProof` API. That benchmark will fail against this endpoint, and it also expects dry-run `execution_stats` that the new JSON-RPC response path does not provide, so either the benchmark needs to be migrated or a compatible gRPC prover endpoint needs to remain exposed

etc/docker/docker-compose.yml
```diff
@@ -624,27 +624,59 @@ services:
...
-    ports:
-      - "${ZK_PROVER_GRPC_PORT}:9000"
     volumes:
       - ../../.devnet/l1/configs/el/chain-config.json:/app/configs/L1/${L1_CHAIN_ID}.json:ro
     environment:
-      - SP1_PROVER=${SP1_PROVER:-dry-run}
+      - PROVER_SERVICE_ENDPOINT=http://base-prover-service:9001
+      - PROVER_WORKER_ID=devnet-zk-host-compressed
+      - PROOF_TYPE=compressed
+      - ZK_BACKEND=${ZK_BACKEND:-dry_run}
```
jackchuma, line 685: Start a worker for SNARK proof jobs. If callers use the new helper against devnet with `proof_type=snark`, the requester API accepts a `snark_groth16` request, but the only ZK host in this stack is configured with `PROOF_TYPE=compressed`. The worker discovery code filters claims by that exact proof type, so those SNARK requests stay queued forever unless a SNARK worker is also started or the helper rejects SNARK for this stack

etc/docker/docker-compose.yml
```diff
@@ -604,18 +604,18 @@ services:
         ipv4_address: ${L2_BASE_RPC_ADVERTISE_IP}
     restart: unless-stopped
 
-  zk-prover-postgres:
+  prover-service-postgres:
```
jackchuma, line 615: The prover-service, Postgres, and zk-host service definitions are now duplicated across the devnet compose file and the standalone zk-prover compose file. That gives us two places to keep ports, env vars, images, healthchecks, migrations, and service names in sync. Can we keep the common prover stack in one compose file and use an override file only for the devnet-specific parts?

etc/just/zk-prover.just
```diff
@@ -1,67 +1,229 @@
...
 logs *containers:
     #!/usr/bin/env bash
     set -euo pipefail
-    docker compose -p "{{ _project }}" {{ _compose }} logs -f {{ if containers == "" { "base-prover-zk zk-prover-postgres" } else { containers } }}
+    docker compose -p "{{ _project }}" {{ _compose }} logs -f {{ if containers == "" { "base-prover-service base-prover-zk-host prover-service-postgres" } else { containers } }}
 
-# List services exposed by the active ZK prover endpoint
-list:
-    bash "{{ _script }}" list "{{ _endpoint }}"
+# List JSON-RPC methods exposed by the active prover-service endpoint
```
jackchuma: The JSON-RPC recipes all repeat the same curl setup: content-type header, endpoint, JSON-RPC envelope, `id`, and trailing newline. A small private helper for “send method + params to endpoint” would make the public recipes shorter and reduce the chance that future request helpers drift from each other

etc/just/zk-prover.just
```diff
@@ -1,67 +1,229 @@
...
+    proof_type={{ quote(proof_type) }}
     proof_type="${proof_type#proof_type=}"
-    bash "{{ _script }}" prove "{{ start_block }}" "{{ number_of_blocks }}" "{{ _endpoint }}" "$proof_type"
+    proof_type="$(printf '%s' "$proof_type" | tr '[:upper:]-' '[:lower:]_')"
+    session_id={{ quote(session_id) }}
+    session_id="${session_id#session_id=}"
+    prover_address={{ quote(prover_address) }}
+    prover_address="${prover_address#prover_address=}"
+    l1_head={{ quote(l1_head) }}
+    l1_head="${l1_head#l1_head=}"
```
jackchuma: These lines normalize every argument by stripping `name=` from the value, even though Just already gives us named parameters. Supporting both invocation styles adds a lot of shell noise to every recipe. I’d use the Just parameters directly unless we have a real caller that needs the `start_block=...` style inside the shell script

etc/just/zk-prover.just
```diff
@@ -1,67 +1,229 @@
...
+      session_id="$(uuidgen | tr '[:upper:]' '[:lower:]')"
+    fi
+    if [ -z "$prover_address" ]; then
+      prover_address="0x0000000000000000000000000000000000000000"
+    fi
 
-# Get proof status and, for dry-run mode, executionStats
-get session_id receipt_type='':
-    bash "{{ _script }}" get "{{ session_id }}" "{{ _endpoint }}" "{{ receipt_type }}"
+    case "$proof_type" in
```
jackchuma, line 60: The compressed and snark branches duplicate most of the same request construction: session id, block range, zk VM, optional `l1_head`, method name, params shape, and JSON-RPC envelope. The only real difference is the proof type and the snark wrapper fields. Build the shared block payload once, then branch only for the snark-specific wrapping

etc/just/zk-prover.just
```diff
@@ -1,205 +1,177 @@
...
-    proof_type="${proof_type#proof_type=}"
-    bash "{{ _script }}" prove "{{ start_block }}" "{{ number_of_blocks }}" "{{ _endpoint }}" "$proof_type"
-
-# Get proof status and, for dry-run mode, executionStats
-get session_id receipt_type='':
-    bash "{{ _script }}" get "{{ session_id }}" "{{ _endpoint }}" "{{ receipt_type }}"
+    command -v jq >/dev/null 2>&1 || { echo "jq is required" >&2; exit 1; }
+    start_block={{ quote(start_block) }}
+    number_of_blocks={{ quote(number_of_blocks) }}
+    proof_type={{ quote(proof_type) }}
```
leopoldjoy, line 25: Please strip the optional `proof_type=` prefix before normalisation, because `just zk-prover prove 1 1 proof_type=snark` now becomes `proof_type=snark` and is rejected before it reaches prover-service.

etc/just/zk-prover.just
```diff
@@ -1,205 +1,177 @@
...
+    http_status="$(curl -sS \
+      --output "$response_file" \
+      --write-out '%{http_code}' \
+      --header 'content-type: application/json' \
+      --data "$payload" \
+      "{{ _requester_endpoint }}")"
+    if [ -s "$response_file" ]; then
+      jq . "$response_file" || cat "$response_file"
+    fi
+    if [ "$http_status" -ge 400 ]; then
```
leopoldjoy, line 176: Please fail when the JSON-RPC body contains an `error` field, because jsonrpsee application errors are returned with HTTP 200 and these helpers currently exit successfully after printing the failure.

etc/just/zk-prover.just
```diff
@@ -1,205 +1,177 @@
...
-_ca_service_command := "./base-prover-zk"
-_ca_system_file := "/etc/ssl/certs/ca-certificates.crt"
-_compose := "-f etc/docker/docker-compose.zk-prover.yml"
-_endpoint := "${ZK_PROVER_ENDPOINT:-localhost:${ZK_PROVER_GRPC_PORT:-9000}}"
-_project := "zk-prover"
-_yaml_to_env := "etc/scripts/yaml-to-compose-env.py"
+_requester_endpoint := "${PROVER_SERVICE_REQUESTER_ENDPOINT:-http://localhost:${PROVER_SERVICE_RPC_PORT:-9000}}"
 
-# Show ZK prover request commands
+# Show prover-service request commands
```
leopoldjoy, line 6: The standalone `up/down/logs` recipes and compose file were removed rather than replaced, so the PR-described standalone prover-service compose workflow can no longer be started via `just zk-prover`.

etc/docker/Justfile
```diff
@@ -65,12 +65,12 @@ sequencer-logs:
...
 # Stream logs from the devnet ZK prover services
 zk-logs *containers:
-    docker compose {{ _env }} {{ _ha }} logs -f {{ if containers == "" { "base-prover-zk zk-prover-postgres" } else { containers } }}
+    docker compose {{ _env }} {{ _ha }} logs -f {{ if containers == "" { "base-prover-service base-prover-zk-host prover-service-postgres" } else { containers } }}
 
 # Stop only the devnet ZK prover services, leaving the rest of devnet running
 zk-down:
-    -docker compose {{ _env }} {{ _ha }} stop base-prover-zk zk-prover-postgres
-    -docker compose {{ _env }} {{ _ha }} rm -f base-prover-zk zk-prover-postgres
+    -docker compose {{ _env }} {{ _ha }} stop base-prover-zk-host base-prover-service prover-service-postgres
```
leopoldjoy, line 72: Since these service names changed, can't we also remove the legacy `base-prover-zk`/`zk-prover-postgres` containers? Or use `--remove-orphans` so branch switches do not leave port-conflicting orphans.

etc/docker/docker-bake.hcl
```diff
@@ -101,16 +98,26 @@ target "batcher" {
...
+  ]
+}
+
+target "zk-host" {
   inherits = ["_rust-service-common"]
-  target = "zk-prover"
+  target = "zk-host"
   args = {
-    PROFILE                   = "${ZK_PROVER_PROFILE}"
+    PROFILE                   = "${PROFILE}"
```
leopoldjoy: Should `zk-host` keep a separate release build profile like the old `zk-prover` target did? Since devnet now builds it with `PROFILE=dev`, local dry-run witness / SP1 execution and the B20 benchmark may be significantly slower if this path is still performance-sensitive.

etc/just/zk-prover.just
mw2000: This is okay since we knowingly removed the standalone prover service. However, we still want to be able to send json rpc requests to the live endpoints and hence keep some of the `just zk-prover` commands

etc/just/zk-prover.just
mw2000: Fixed, we now strip the optional `proof_type=` prefix before normalization.

etc/just/zk-prover.just
mw2000: Fixed, `_rpc` now exits non-zero when the response body contains a JSON-RPC error

etc/docker/Justfile
mw2000: Fixed, devnet `compose up` now passes `--remove-orphans`

etc/docker/docker-compose.yml
```diff
@@ -625,27 +625,59 @@ services:
...
-      target: zk-prover
-    container_name: base-prover-zk
+      target: prover-service
+    container_name: base-prover-service
+    depends_on:
+      prover-service-postgres:
+        condition: service_healthy
+    ports:
+      - "${PROVER_SERVICE_RPC_PORT}:9000"
+      - "${PROVER_SERVICE_WORKER_RPC_PORT}:9001"
```
leopoldjoy: Worker RPC is internal to Compose, so couldn't we not publish it unless external workers are intended?

mw2000: Yup dropped the host mapping for 9001

etc/just/zk-prover.just
```diff
@@ -1,205 +1,181 @@
...
+        zk_vm: "sp1"
+      } + if $l1_head == "" then {} else {l1_head: $l1_head} end')"
+
+    case "$proof_type" in
+      compressed)
+        request="$(jq -n -c \
+          --argjson proof "$block_payload" \
+          '{proof_type: "compressed", payload: $proof}')"
+        ;;
+      snark|groth16|snark_groth16)
```
leopoldjoy, line 66: SNARK is still accepted here, but the configured ZK backends reject it; please reject it locally or wire a backend that can complete it.

etc/docker/docker-compose.yml
```diff
@@ -625,27 +625,57 @@ services:
...
+      interval: 250ms
+      timeout: 1s
+      retries: 240
+      start_period: 250ms
+    restart: unless-stopped
+
+  base-prover-zk-host:
+    image: base-prover-zk-host:local
+    build:
+      <<: *rust-service-build
```
leopoldjoy: Direct compose builds still inherit `CARGO_PROFILE:-dev` for zk-host; please override `PROFILE=release` here too or document compose builds as unsupported.

etc/just/zk-prover.just
mw2000: This should be resolved in #3869 so should be okay to have it not be rejected for now

leopoldjoy (approved):

PR #3738, opened 2026-06-24
feat(proof): execute zk dry runs locally

mw2000:
## Summary
Make the dry-run ZK backend generate a witness and execute the SP1 range program with LightProver. Return local execution stats in ZK proof results and keep dry-run compressed-only.

### DIFF
--- a/crates/proof/prover-service/src/server/get_proof.rs
+++ b/crates/proof/prover-service/src/server/get_proof.rs
@@ -3,11 +3,11 @@ use base_prover_service_db::{
     canonical_session_id,
 };
 use base_prover_service_protocol::{
-    GetProofRequest, GetProofResponse, PROOF_REQUEST_NOT_FOUND_MESSAGE, ProofResult, ProofStatus,
-    ZkProofResult, ZkVm,
+    ExecutionStats, GetProofRequest, GetProofResponse, PROOF_REQUEST_NOT_FOUND_MESSAGE,
+    ProofResult, ProofStatus, ZkProofResult, ZkVm,
 };
 use jsonrpsee::core::RpcResult;
-use tracing::info;
+use tracing::{info, warn};
 use uuid::Uuid;
 
 use crate::{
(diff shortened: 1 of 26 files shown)


crates/proof/zk/backend/src/succinct/dry_run.rs
```diff
@@ -1,46 +1,260 @@
...
+}
 
-/// [`ZkProver`] that completes instantly with an empty proof payload.
-#[derive(Debug, Clone, Copy, Default)]
-pub struct DryRunZkProver;
+impl std::fmt::Debug for DryRunZkProver {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DryRunZkProver").field("config", &self.config).finish_non_exhaustive()
+    }
+}
```
jackchuma: `DryRunZkProverConfig` looks like extra public API for a type that only has one internal caller. Since the config is just three fields copied into `DryRunZkProver`, this can be simpler as fields directly on `DryRunZkProver`, with `new(provider, base_consensus_url, l1_node_url, default_sequence_window)` or construction inline in `build_until_cancelled`. That also lets us drop the `DryRunZkProverConfig` re-exports

crates/proof/zk/backend/src/succinct/dry_run.rs
```diff
@@ -1,46 +1,260 @@
...
     pub fn backend_session_id(request: &ZkProofRequestKind, request_session_id: &str) -> String {
-        let session_type = if request.is_snark_groth16() { "snark" } else { "stark" };
-        format!("dry-run-{session_type}-{request_session_id}")
+        match request {
+            ZkProofRequestKind::Compressed(_) => format!("{DRY_RUN_PREFIX}{request_session_id}"),
+            ZkProofRequestKind::SnarkGroth16(_) => {
+                format!("dry-run-unsupported-snark-{request_session_id}")
+            }
+        }
+    }
```
jackchuma: `backend_session_id` appears to be test-only API now, and the Snark branch returns a synthetic `"dry-run-unsupported-snark-*"` id that production never uses because `submit` rejects Snark before creating a backend session. I’d delete this helper and the two tests around it, and keep the single `format!("{DRY_RUN_PREFIX}{request_session_id}")` at the actual call site in `submit`

crates/proof/zk/backend/src/succinct/dry_run.rs
```diff
@@ -1,102 +1,256 @@
...
+        let (execution_result, execution_ms) = tokio::task::spawn_blocking(move || {
+            info!("starting local SP1 zkVM execution");
+
+            let execution_start = std::time::Instant::now();
+            let prover = LightProver::new();
+            let result = prover
+                .execute(Elf::Static(get_range_elf_embedded()), stdin)
+                .calculate_gas(true)
+                .deferred_proof_verification(false)
+                .run();
```
jackchuma, line 118: Honor cycle limits during dry-run execution. When `ZK_BACKEND=dry-run` is used, this local SP1 execution runs without the configured `RANGE_CYCLE_LIMIT`, unlike the cluster/network paths that pass the limit to the backend. For requests whose range exceeds the intended limit, or if execution gets unexpectedly expensive, this can tie up a blocking thread indefinitely; the comment above also notes cancellation cannot stop an already-dispatched execution. Please thread the dry-run cycle limit through the config and apply it before run()

crates/proof/prover-service/src/server/get_proof.rs
```diff
@@ -3,26 +3,59 @@ use base_prover_service_db::{
...
 
-fn is_dry_run_metadata(metadata: &serde_json::Value) -> bool {
+#[cfg(test)]
+fn execution_stats_from_metadata(metadata: &serde_json::Value) -> Option<ExecutionStats> {
+    is_dry_run_result_metadata(metadata).then_some(())?;
+    let stats = metadata.get(OP_SUCCINCT_EXECUTION_STATS_METADATA_KEY)?;
+    execution_stats_from_value(stats)
+}
+
+fn execution_stats_from_value(stats: &serde_json::Value) -> Option<ExecutionStats> {
```
jackchuma, line 25: This metadata parsing path looks heavier than it needs to be. We already have `ExecutionStats` as a serde type, so the manual field-by-field parsing, custom millisecond coercion, `(bool, Option<ExecutionStats>)` return, and warning path for malformed stats all feel like extra compatibility surface.

  Could this be reduced to checking the dry-run marker and then using `serde_json::from_value::<ExecutionStats>(stats.clone()).ok()` with `find_map` over completed sessions? If the stats are malformed, treating it as absent seems enough for this endpoint

crates/proof/zk/backend/src/succinct/dry_run.rs
```diff
@@ -1,102 +1,268 @@
...
+impl std::fmt::Debug for DryRunZkProver {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DryRunZkProver")
+            .field("base_consensus_url", &self.base_consensus_url)
+            .field("l1_node_url", &self.l1_node_url)
+            .field("default_sequence_window", &self.default_sequence_window)
+            .field("range_cycle_limit", &self.range_cycle_limit)
+            .finish_non_exhaustive()
+    }
+}
```
jackchuma, line 54: The manual `Debug` impl mostly repeats fields that are already debug-safe, and `OpSuccinctWitnessProvider` already has a non-exhaustive `Debug` impl. This can probably just be `#[derive(Clone, Debug)]` on `DryRunZkProver`

jackchuma (approved):

PR #3740, opened 2026-06-24
chore: update SP1 dependencies

mw2000:
## Summary
Bump the SP1 host and guest crates to 6.3.0 and update the SP1 cluster git dependencies to v2.5.1. Regenerate both the workspace and SP1 program lockfiles from the updated manifests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1071,7 +1071,7 @@ dependencies = [
  "alloy-rlp",
  "alloy-serde 1.8.3",
  "alloy-sol-types",
- "itertools 0.14.0",
+ "itertools 0.13.0",
  "serde",
  "serde_json",
  "serde_with",
@@ -1093,7 +1093,7 @@ dependencies = [
  "alloy-serde 2.0.5",
  "alloy-sol-types",
  "arbitrary",
- "itertools 0.14.0",
+ "itertools 0.13.0",
  "serde",
  "serde_json",
  "serde_with",
@@ -1375,7 +1375,7 @@ checksum = "e8597d36d546e1dab822345ad563243ec3920e199322cb554ce56c8ef1a1e2e7"
 dependencies = [
  "alloy-json-rpc 1.8.3",
  "alloy-transport 1.8.3",
- "itertools 0.14.0",
+ "itertools 0.13.0",
  "reqwest 0.13.4",
  "serde_json",
  "tower 0.5.3",
@@ -1396,7 +1396,7 @@ dependencies = [
  "hyper 1.10.1",
  "hyper-tls",
  "hyper-util",
- "itertools 0.14.0",
+ "itertools 0.13.0",
  "jsonwebtoken",
  "opentelemetry 0.31.0",
  "opentelemetry-http",
@@ -6805,7 +6805,7 @@ dependencies = [
  "bitflags 2.13.0",
  "cexpr",
  "clang-sys",
- "itertools 0.11.0",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -6825,7 +6825,7 @@ dependencies = [
  "bitflags 2.13.0",
  "cexpr",
  "clang-sys",
- "itertools 0.11.0",
+ "itertools 0.13.0",
  "proc-macro2",
  "quote",
  "regex",
@@ -11031,7 +11031,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.59.0",
+ "windows-core 0.62.2",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/succinct/programs/Cargo.toml
```diff
@@ -42,8 +42,8 @@ base-proof-succinct-client-utils = { path = "../utils/client" }
 base-proof-succinct-ethereum-client-utils = { path = "../utils/ethereum/client" }
 
 # SP1 (must match the cargo-prove Docker tag)
-sp1-lib = { version = "=6.2.3", features = ["verify"] }
-sp1-zkvm = { version = "=6.2.3", features = ["verify"] }
+sp1-lib = { version = "=6.3.0", features = ["verify"] }
```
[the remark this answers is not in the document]
mw2000: No 6.3.0 patch tags exist, 6.2.0 tags should be okay

refcell (dismissed):

etc/just/succinct.just
```diff
@@ -1,5 +1,5 @@
 _cargo_prove := "~/.sp1/bin/cargo-prove"
-_sp1_tag := "v6.2.3"
+_sp1_tag := "v6.3.0"
```
jackchuma, line 2: Invalidate cached ELFs when bumping SP1. When a checkout already has `crates/proof/succinct/elf/{range-elf-embedded,aggregation-elf}` from the previous v6.2.3 build, changing this tag does not make `just succinct ensure-elfs` rebuild them because it only checks that the files are non-empty. That means warm local/CI builds can embed and prove the old v6.2.3 guest ELFs while the host crates and documented Docker tag are v6.3.0, effectively shipping stale circuits; please key the cache on the SP1 version or force a rebuild when the tag changes

Cargo.lock
```diff
@@ -21719,9 +21722,9 @@ dependencies = [
 
 [[package]]
 name = "sp1-primitives"
-version = "6.2.4"
+version = "6.3.0"
```
leopoldjoy, line 21725: The root lockfile still leaves the neighboring `sp1-lib` package on `6.2.4` while the rest of this SP1 stack moves to `6.3.0`, so the lockfile was not fully regenerated to the version this PR intends.

leopoldjoy:
Posting here since it's an unmodified file:

`.github/actions/setup/action.yml:22` - The shared CI setup action still defaults `sp1-version` to `v6.2.3` even though this PR moves the manifests and `succinct.just` tag to `v6.3.0`, so CI jobs that enable SP1 without overriding the input can run stale SP1 tooling.

etc/just/succinct.just
mw2000: Yeah this is a good catch, this is likely since we moved away from the `manifest.toml` check that would have done the invalidation.

mw2000: ~~Addressed by making `just succinct ensure-elfs` call `just succinct build-elfs` before validating the outputs. That removes the warm-cache path where non-empty old ELFs could be reused after an SP1 bump, without reintroducing the maintained ELF hash manifest.~~

Actually I think there's a better solution for the caching. I will add this in a follow up PR since it was actually broken since before this. The current state is that it's rebuilding from scratch everytime since the checkout doesn't have the ELF binaries in it

Cargo.lock
mw2000: Addressed. Regenerated the root lockfile so `sp1-lib` now resolves to `6.3.0` as well (`cargo tree -i sp1-lib@6.2.4 --workspace --target all` no longer matches any package).

mw2000:
Addressed Leopold’s setup-action comment too: `.github/actions/setup/action.yml` now defaults `sp1-version` to `v6.3.0`, matching `etc/just/succinct.just`.

leopoldjoy (approved):
Given that the caching issue will be addressed via a follow-up, approving.

PR #3745, opened 2026-06-24
fix(challenger): classify known dispute reverts

leopoldjoy:
## Gap
Challenger `submit_dispute()` collapsed named `AggregateVerifier` custom errors into generic tx-manager failures. The driver then left ready proofs queued for retry, even when the revert already told us the proof was stale, already attached, or the parent game was no longer actionable.

## Value-add
- Reuses `base-proof-submission::KnownRevert` decoding for challenger dispute submissions.
- Drops already-proven and invalid-parent cases instead of retrying the same proof every tick.
- Re-requests a fresh proof for `L1OriginTooOld` and keeps the existing TEE-to-ZK fallback with clearer signer/stale-origin diagnostics.

## Testing
- `cargo test -p base-challenger submit_dispute_classifies_known_revert_error --lib`
- `cargo test -p base-challenger proof_already_verified_revert_drops_pending_proof --lib`
- `cargo test -p base-challenger stale_l1_origin_revert_requests_fresh_proof --lib`
- `cargo test -p base-challenger --lib`

### DIFF
--- a/crates/proof/challenge/src/driver.rs
+++ b/crates/proof/challenge/src/driver.rs
@@ -24,6 +24,7 @@ use alloy_primitives::{Address, B256};
 use base_proof_contracts::{AggregateVerifierClient, GameStatus};
 use base_proof_primitives::ProofRequest as TeeProofRequest;
 use base_proof_rpc::L2Provider;
+use base_proof_submission::KnownRevert;
 use base_prover_service_client::ProofRequesterProvider;
 use base_prover_service_protocol::{SnarkGroth16ProofRequest, TeeKind, ZkProofRequest, ZkVm};
 use base_runtime::{Clock, TokioRuntime};
@@ -806,17 +807,78 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
                 }
             }
             Err(e) => {
-                if targets_tee && let Some(p) = self.pending_proofs.get_mut(&game_address) {
+                if matches!(e.known_revert(), Some(KnownRevert::ProofAlreadyVerified)) {
+                    info!(
+                        error = %e,
+                        game = %game_address,
+                        "dispute proof already verified onchain, dropping pending proof"
+                    );
+                    self.pending_proofs.remove(&game_address);
+                    return Ok(());
+                }
+
+                if matches!(e.known_revert(), Some(KnownRevert::InvalidParentGame)) {
                     warn!(
                         error = %e,
                         game = %game_address,
-                        "TEE dispute tx failed, falling back to ZK"
+                        "dispute game parent is invalid onchain, dropping pending proof"
                     );
+                    self.pending_proofs.remove(&game_address);
+                    return Ok(());
+                }
+
+                if targets_tee && let Some(p) = self.pending_proofs.get_mut(&game_address) {
+                    match e.known_revert() {
+                        Some(KnownRevert::InvalidSigner) => {
+                            warn!(
+                                error = %e,
+                                game = %game_address,
+                                "TEE proof signer is invalid onchain, falling back to ZK"
+                            );
+                        }
+                        Some(KnownRevert::L1OriginTooOld) => {
+                            warn!(
+                                error = %e,
+                                game = %game_address,
+                                "TEE proof L1 origin is too old, falling back to ZK"
+                            );
+                        }
+                        _ => {
+                            warn!(
+                                error = %e,
+                                game = %game_address,
+                                "TEE dispute tx failed, falling back to ZK"
+                            );
+                        }
+                    }
                     // Don't retry the failed TEE submission — switch to the ZK
                     // fallback so the next retry uses the pre-built ZK request.
                     p.phase = ProofPhase::NeedsRetry;
                     return self.handle_proof_retry(game_address).await;
                 }
+
+                if matches!(e.known_revert(), Some(KnownRevert::L1OriginTooOld)) {
+                    warn!(
+                        error = %e,
+                        game = %game_address,
+                        "dispute proof L1 origin is too old, requesting a fresh proof"
+                    );
+                    if let Some(p) = self.pending_proofs.get_mut(&game_address) {
+                        p.phase = ProofPhase::NeedsRetry;
+                    }
+                    return self.handle_proof_retry(game_address).await;
+                }
+
+                if matches!(e.known_revert(), Some(KnownRevert::InvalidSigner)) {
+                    warn!(
+                        error = %e,
+                        game = %game_address,
+                        "dispute proof signer is invalid onchain, dropping pending proof"
+                    );
+                    self.pending_proofs.remove(&game_address);
+                    return Ok(());
+                }
+
                 warn!(
                     error = %e,
                     game = %game_address,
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/challenge/src/error.rs
```diff
@@ -13,7 +14,30 @@ pub enum ChallengeSubmitError {
...
+}
+
+impl ChallengeSubmitError {
+    /// Returns the decoded known revert, if this error wraps one.
+    pub const fn known_revert(&self) -> Option<KnownRevert> {
+        match self {
+            Self::KnownRevert(revert) => Some(*revert),
+            Self::TxReverted { .. } | Self::TxManager(_) => None,
+        }
+    }
```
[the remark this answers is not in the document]
leopoldjoy: Agreed this is a real limitation of the mined-receipt path, but it needs broader receipt/replay support and is out of scope for this narrow classifier PR.

crates/proof/challenge/src/driver.rs
```diff
@@ -806,17 +807,78 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
+                if matches!(e.known_revert(), Some(KnownRevert::L1OriginTooOld)) {
+                    warn!(
+                        error = %e,
+                        game = %game_address,
+                        "dispute proof L1 origin is too old, requesting a fresh proof"
+                    );
+                    if let Some(p) = self.pending_proofs.get_mut(&game_address) {
+                        p.phase = ProofPhase::NeedsRetry;
+                    }
+                    return self.handle_proof_retry(game_address).await;
```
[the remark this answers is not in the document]
leopoldjoy: Fixed in 5a132e309: pure ZK L1OriginTooOld now drops the stale pending proof instead of regenerating with the same onchain l1_head; TEE can still fall back to ZK.

crates/proof/challenge/src/driver.rs
```diff
@@ -806,17 +807,76 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
+
+                if matches!(e.known_revert(), Some(KnownRevert::InvalidSigner)) {
+                    warn!(
+                        error = %e,
+                        game = %game_address,
+                        "dispute proof signer is invalid onchain, dropping pending proof"
+                    );
+                    self.pending_proofs.remove(&game_address);
+                    return Ok(());
+                }
```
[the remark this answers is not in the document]
leopoldjoy: Thanks, leaving this as-is since the current branch ordering is intentional, the repeated known_revert() calls are cheap enum matches, and this is readability-only rather than a behavior issue.

crates/proof/challenge/src/driver.rs
```diff
@@ -806,17 +807,76 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
+                                game = %game_address,
+                                "TEE proof signer is invalid onchain, falling back to ZK"
+                            );
+                        }
+                        Some(KnownRevert::L1OriginTooOld) => {
+                            warn!(
+                                error = %e,
+                                game = %game_address,
+                                "TEE proof L1 origin is too old, falling back to ZK"
+                            );
```
jackchuma: Do not fall back to ZK after a stale TEE revert. When a TEE `nullify()` fails with `L1OriginTooOld`, falling back to ZK cannot make this same game submit successfully: both `build_tee_request` and the prebuilt ZK fallback use the game's fixed `candidate.l1_head`, and once that L1 origin is outside the history window it only becomes older. This path starts an expensive ZK proof job that will hit the same non-retryable revert instead of treating the game as terminal/ignored

crates/proof/challenge/src/driver.rs
```diff
@@ -806,17 +807,76 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
                 }
+
+                if matches!(e.known_revert(), Some(KnownRevert::L1OriginTooOld)) {
+                    warn!(
+                        error = %e,
+                        game = %game_address,
+                        "dispute proof L1 origin is too old, dropping pending proof"
+                    );
+                    self.pending_proofs.remove(&game_address);
+                    return Ok(());
```
jackchuma: Avoid re-proving after stale L1-origin reverts. For a ZK dispute tx that reverts with `L1OriginTooOld`, this removes the pending proof but does not mark the game terminal; because the reverted tx leaves the game `InProgress` with the same prover slots and fixed `l1_head`, the next `step()` scan will select the same candidate and call `prove_block_range` again. That turns a non-retryable stale-origin failure into an expensive proof-generation loop, so the game needs an ignored/terminal state rather than just deleting the pending entry

crates/proof/challenge/src/driver.rs
```diff
@@ -806,17 +807,76 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
                 }
             }
             Err(e) => {
-                if targets_tee && let Some(p) = self.pending_proofs.get_mut(&game_address) {
+                if matches!(e.known_revert(), Some(KnownRevert::ProofAlreadyVerified)) {
```
jackchuma: This repeats `e.known_revert()` across five branches, and most branches do the same thing: log, remove the pending proof, and return `Ok(())`.

  Could match the known revert once and group the terminal cases:

  - `ProofAlreadyVerified`
  - `InvalidParentGame`
  - non-TEE `L1OriginTooOld`
  - non-TEE `InvalidSigner`

  For the TEE fallback path, one generic warning should be enough because `error = %e` already includes the specific revert. That keeps the behavior but cuts the branch/log duplication

crates/proof/challenge/src/error.rs
```diff
@@ -13,7 +14,30 @@ pub enum ChallengeSubmitError {
...
+    /// A known proof-contract custom error was decoded from the revert.
+    #[error(transparent)]
+    KnownRevert(#[from] KnownRevert),
     /// Transaction manager error (nonce, fees, RPC, signing, etc.).
     #[error(transparent)]
-    TxManager(#[from] TxManagerError),
+    TxManager(TxManagerError),
+}
+
+impl ChallengeSubmitError {
```
jackchuma: `ChallengeSubmitError::known_revert()` is a public helper with one caller, and that caller is already doing driver-specific behavior branching.

  I’d inline the match at the call site for now:

  `match &e { ChallengeSubmitError::KnownRevert(r) => Some(*r), _ => None }`

  That avoids adding API surface until there’s a second caller that actually benefits from the helper

crates/proof/challenge/src/driver.rs
```diff
@@ -925,3 +985,115 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
+                Bytes::from(vec![0x01, 0xAA]),
+                0,
+                B256::repeat_byte(0x22),
+                proof_request(),
+                DisputeIntent::Challenge,
+            ),
+        );
+    }
+
+    #[tokio::test]
```
jackchuma, line 1120: These two tests share the same arrange/act/assert shape and only vary the revert encoding plus the “proveBlockRange was not called” assertion.

  A small table-style test like `terminal_known_reverts_drop_pending_proof` can cover both `AlreadyProven` and `L1OriginTooOld` without duplicating the driver setup path. Keep the extra `prove_block_range_log.is_empty()` assertion for the stale L1 origin case if that behavior matters

crates/proof/challenge/src/driver.rs
leopoldjoy: Done in 9b18d70f5: stale L1-origin games are added to an ignored set so later scan ticks skip the same game instead of re-proving.

crates/proof/challenge/src/driver.rs
leopoldjoy: Done in 9b18d70f5: the branch now matches the known revert once and groups terminal cases before the TEE fallback path.

crates/proof/challenge/src/driver.rs
leopoldjoy: Leaving these as separate tests for now because the terminal ignored-game behaviour added here gives each case distinct assertions and keeps the regression intent explicit. Please let me know if you disagree with that.

crates/proof/challenge/src/error.rs
leopoldjoy: Done in 9b18d70f5: removed the public known_revert() helper and inlined the KnownRevert extraction at the driver call site.

crates/proof/challenge/src/driver.rs
leopoldjoy: Done in 9b18d70f5: L1OriginTooOld is now terminal before the TEE fallback path, so we do not start a ZK fallback for the same stale game.

crates/proof/challenge/src/driver.rs
```diff
@@ -806,29 +817,83 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
+                        return Ok(());
+                    }
+                    _ if targets_tee => {
+                        warn!(
+                            error = %e,
+                            game = %game_address,
+                            "TEE dispute tx failed, falling back to ZK"
+                        );
+                    }
+                    _ => {
```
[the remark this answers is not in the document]
leopoldjoy: Done in 0acd7f5a1: GameAlreadyExists now has an explicit terminal drop arm before the TEE fallback/default retry paths.

crates/proof/challenge/src/driver.rs
```diff
@@ -112,6 +115,8 @@ where
     pub verifier_client: Arc<dyn AggregateVerifierClient>,
     /// In-flight proof sessions keyed by game address.
     pub pending_proofs: PendingProofs,
+    /// Games that hit terminal contract reverts and should not be rediscovered.
+    ignored_games: HashSet<Address>,
```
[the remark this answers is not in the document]
leopoldjoy: Partially addressed in 0acd7f5a1: added an ignored_games gauge for visibility; leaving pruning/cleanup for a follow-up since it needs an explicit retention or extra state-read policy.

crates/proof/challenge/src/driver.rs
```diff
@@ -806,29 +818,93 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
+                        return Ok(());
+                    }
+                }
+
+                if let Some(p) = self.pending_proofs.get_mut(&game_address) {
                     // Don't retry the failed TEE submission — switch to the ZK
                     // fallback so the next retry uses the pre-built ZK request.
                     p.phase = ProofPhase::NeedsRetry;
                     return self.handle_proof_retry(game_address).await;
                 }
```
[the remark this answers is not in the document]
leopoldjoy: Leaving TEE fallback behavior unchanged here because the broad fallback predates this PR; tightening infra-error vs proof-error fallback deserves a focused follow-up with targeted tests.

crates/proof/challenge/src/driver.rs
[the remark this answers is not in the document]
leopoldjoy: Leaving pruning out of this PR: the new ignored_games gauge gives visibility, while cleanup needs a separate retention/state-read policy to avoid reintroducing stale proof loops.

jackchuma (approved):

PR #3696, opened 2026-06-22
feat(eip8130): phased call execution + policy gate + full fee settlement

chunter-cb:
## Summary

Drives an EIP-8130 transaction's `calls` (`Vec<Vec<Call>>`) as real EVM call frames after the pre-call pipeline, adds the actor policy gate, and reworks gas/fee settlement to route the full fee.

- **Phased call execution** — phases share a single gas pool (`gas_limit - sender_intrinsic`) and commit independently in sequence. Calls within a phase are atomic; if any call reverts (or is policy-gated), that phase's state is discarded and every later phase is skipped, but the tx is still included (nonce consumed, fee paid). Each call dispatches from `sender` to `call.to` with `msg.value == 0` and `tx.origin == sender`.
- **Transaction context** — publishes sender / payer / `sender_actor_id` to `TxContextStorage` before dispatching calls.
- **Policy gate** — resolves the sender actor's policy manager once; gates every `call.to`, reverting the phase with `ActorPolicyViolation(bytes32,address)` on mismatch.
- **Full fee settlement** — pre-charges the worst case (gas + L1 + operator), caps the refund per EIP-3529, then routes base fee, priority tip, L1 cost, and operator fee to their vaults and refunds the surplus to the payer.
- **Status** — overall transaction outcome reported via `ExecutionResult::Success` / `Revert`.

EIP-8037 (Amsterdam) maps to `ForkCondition::Never` on Base, so the state-gas reservoir is always zero and is not threaded through call frames.

## Test plan

- [x] `eoa_self_pay_transaction_executes_and_charges_sender`
- [x] `underfunded_payer_is_rejected`
- [x] `single_phase_call_executes_against_contract_code`
- [x] `reverting_call_includes_tx_with_revert_status`
- [x] `later_phase_skipped_after_earlier_phase_reverts`
- [x] `policy_gate_blocks_call_to_unauthorized_target`
- [x] `policy_gate_allows_call_to_authorized_target`
- [x] `actor_policy_violation_data_is_abi_encoded`
- [x] `cargo clippy --features std --all-targets` clean
- [x] `cargo check --no-default-features` (no_std) compiles

PR #3748, opened 2026-06-24
fix(eip8130): surface mid-call database errors as fatal, not reverts

chunter-cb:
## Summary

Follow-up to #3696. Three fixes to the EIP-8130 phased-call executor, which runs `calls` directly and bypasses the mainnet single-frame handler — and with it several of the handler's steps.

### 1. Mid-call DB errors must be fatal, not reverts (consensus-critical)

EIP-8130 calls are dispatched as depth-0 top-level frames via `run_call`. revm only converts a frame's recorded DB error into an `Err` when a *child* frame's outcome is folded into its parent (`EthFrame::return_result` → `take_error`); the **root** frame is returned directly by `run_exec_loop` without that check. So a DB failure during a call (e.g. an `SLOAD` against a missing trie node) halted the interpreter with `FatalExternalError` and came back as a non-ok `InterpreterResult`, which `execute_calls` folded into the `phase_reverted` path.

This is consensus-unsafe: a node hitting the DB failure would include the tx as reverted (nonce consumed, fee paid), while a healthy node would execute it normally — a fork. Every other tx type is protected via `Handler::execution_result`'s `take_error` guard; the 8130 path bypasses it.

**Fix:** call `take_error` after `run_exec_loop` in `run_call`.

### 2. Warm coinbase + precompiles + set journal spec before calls (EVM equivalence)

The 8130 path skips the handler's `load_accounts`, so dispatched calls ran against the journal's default spec id with a **cold coinbase** and an unwarmed precompile set — charging cold access (2600) where a call in a normal tx is charged warm (100). Deterministic, but a gas/EVM-equivalence bug.

**Fix:** `warm_pre_call_accounts` mirrors `pre_execution::load_accounts` (minus access-list handling, which 8130 has no analogue for): sets the journal EVM spec id, warms precompiles, and warms the coinbase (Shanghai+), invoked right before `execute_calls`.

### 3. Centralized post-error teardown (defensive)

The `execute_calls` / `settle_fees` error paths only reverted the checkpoint and cleared the L1 cost. A DB error inside a *nested subcall* surfaces while the parent frame is still on the stack, so stale frame/local state could leak into the next tx reusing the same `BaseEvm`.

**Fix:** `teardown_after_error` mirrors the mainnet `catch_error` cleanup (checkpoint revert + clear L1 cost + `local_mut().clear()` + `frame_stack().clear()`), used on both error paths.

## Tests
- `db_failure_during_call_propagates_as_error_not_revert` — a call whose `SLOAD` is refused by a wrapping DB must fail with `EVMError::Database`, not be included as a revert. (Verified it fails without fix #1.)
- `call_warms_coinbase_per_eip3651` — `BALANCE(coinbase)` must be strictly cheaper than a byte-identical `BALANCE` of an untouched address, which only holds when the coinbase is pre-warmed. (Verified it fails without fix #2.)

## Test plan
- [x] `cargo test -p base-common-evm --features std eip8130` (13 passed)
- [x] `cargo clippy -p base-common-evm --features std --all-targets` clean
- [x] `cargo check -p base-common-evm --no-default-features` (no_std) compiles

### DIFF
--- a/crates/common/evm/src/eip8130.rs
+++ b/crates/common/evm/src/eip8130.rs
@@ -731,8 +731,8 @@ impl Eip8130Executor {
     where
         DB: AlloyDatabase,
         P: PrecompileProvider<BaseContext<DB>, Output = InterpreterResult>,
-        BaseContext<DB>: BaseContextTr
-            + ContextTr<Db = DB, Tx = BaseTransaction<TxEnv>, Block = BlockEnv>,
+        BaseContext<DB>:
+            BaseContextTr + ContextTr<Db = DB, Tx = BaseTransaction<TxEnv>, Block = BlockEnv>,
     {
         let (ctx, precompiles) = evm.ctx_precompiles();
 
@@ -764,8 +764,8 @@ impl Eip8130Executor {
     where
         DB: AlloyDatabase,
         P: PrecompileProvider<BaseContext<DB>, Output = InterpreterResult>,
-        BaseContext<DB>: BaseContextTr
-            + ContextTr<Db = DB, Tx = BaseTransaction<TxEnv>, Block = BlockEnv>,
+        BaseContext<DB>:
+            BaseContextTr + ContextTr<Db = DB, Tx = BaseTransaction<TxEnv>, Block = BlockEnv>,
     {
         let ctx = evm.ctx_mut();
         ctx.journal_mut().checkpoint_revert(checkpoint);
@@ -1142,10 +1142,7 @@ mod tests {
         let mut db = InMemoryDB::default();
         db.insert_account_info(
             sender,
-            AccountInfo {
-                balance: U256::from(10u64).pow(U256::from(18u64)),
-                ..Default::default()
-            },
+            AccountInfo { balance: U256::from(10u64).pow(U256::from(18u64)), ..Default::default() },
         );
         db.insert_account_info(
             target,
--- a/crates/common/evm/src/eip8130.rs
+++ b/crates/common/evm/src/eip8130.rs
@@ -56,6 +56,7 @@ use revm::{
     context::{BlockEnv, LocalContextTr, TxEnv, journaled_state::account::JournaledAccountTr},
     context_interface::{
         Block, Cfg, ContextTr, JournalTr,
+        context::take_error,
         result::{EVMError, ExecutionResult, Output, ResultGas, SuccessReason},
     },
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/evm/src/eip8130.rs
```diff
@@ -686,7 +703,75 @@ impl Eip8130Executor {
...
+    /// Mirrors the mainnet handler's `load_accounts` pre-execution step (minus
+    /// the access-list handling, for which an EIP-8130 transaction has no
+    /// analogue): sets the journal's EVM spec id, warms the precompile address
+    /// set, and warms the coinbase per EIP-3651 (Shanghai+). The 8130 path runs
+    /// `calls` directly without the single-frame handler, so without this the
+    /// dispatched calls would execute against the journal's default spec id and
+    /// a cold coinbase / precompile set — charging cold-access gas (2600) where a
+    /// call in a normal transaction is charged warm (100) and otherwise drifting
+    /// from EVM equivalence on the same chain.
+    fn warm_pre_call_accounts<DB, I, P>(evm: &mut BaseEvm<DB, I, P>)
```
chunter-cb, line 730: We may also add the account config contract here.

xenoliss (approved):

PR #3751, opened 2026-06-24
refactor(proposer): clean up proof dispatcher

jackchuma:
### What changed? Why?

Cleaned up the proposer proof dispatcher by simplifying dispatcher state, metrics wiring, and the pipeline/service call sites around proof targets.

### Notes to reviewers

The main review surface is `crates/proof/proposer/src/proof_dispatcher.rs`; the surrounding changes keep the driver, pipeline, service, metrics, and test utilities aligned with the simpler dispatcher shape.

### DIFF
--- a/crates/proof/proposer/src/pipeline.rs
+++ b/crates/proof/proposer/src/pipeline.rs
@@ -130,6 +130,9 @@ where
                     Metrics::safe_head().set(safe_head as f64);
                     Metrics::last_proposed_block().set(recovered.l2_block_number as f64);
 
+                    // Dispatch failures retry from the in-memory cursor. A fresh recovery walk is
+                    // only needed when onchain state changes; try_recover_and_plan returns that as
+                    // a different recovered state, which resets the cursor here.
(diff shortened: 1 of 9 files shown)


crates/proof/proposer/src/pipeline.rs
```diff
@@ -137,26 +130,17 @@ where
...
                     }
 
-                    if drop_recovery_cache {
-                        cache = None;
-                        dispatched_through.store(0, Ordering::Relaxed);
-                    }
+                    let current = cursor
+                        .as_mut()
+                        .expect("dispatcher cursor initialized from recovered state");
+                    self.proof_dispatcher.tick(current, safe_head).await;
```
leopoldjoy, line 144: This removes the retry-exhaustion path that previously dropped the recovery cache after repeated dispatch failures; if indefinitely retrying from the cached dispatcher cursor is intended, can we document why a fresh recovery walk is no longer needed here?

leopoldjoy (approved):

PR #3752, opened 2026-06-24
fix(txpool): admit EIP-8130 against pending state and retry on lag

chunter-cb:
## Summary

EIP-8130 mempool admission authorizes against **live** precompile state (account configuration, nonce channels) and the sender's nonce/code. It previously read \`client().latest()\`, the last fully committed block, which can trail the block currently being built by one block. A snapshot that lags the in-progress block transiently rejects a *structurally valid* transaction whose authorizing state was written by the **preceding** committed block — e.g. an owner change immediately followed by a transaction signed by the new owner, which fails as \`actor … is not bound\` even though the binding is already canonical.

## Changes

- **Read \`client().pending()\`** for the admission state snapshot. \`pending()\` is the state of the block currently being built (not yet committed), one block ahead of \`latest()\`. On the builder this reflects precompile state already mutated by the in-progress block; on non-builder nodes it may be unavailable, so this falls back to \`latest()\`.
- **Reclassify \`BaseTxPoolError::Eip8130Validation\` as non-bad (retryable).** EIP-8130 admission is state-relative, so a lagging snapshot can yield a transient false rejection; caching the tx as permanently bad or penalizing the relaying peer would drop traffic that becomes valid moments later. The deterministic, state-independent DA-footprint failure stays penalizing. The dominant structural invariants are already surfaced as \`InvalidTransactionError\` variants and classified by the inner validator.

## Test plan

- \`cargo test -p base-execution-txpool\`
- Devnet: rebuild, run the EOA + smart-account 8130 suite, expect 6/6 with no inter-tx delay (previously 3/6).

### DIFF
--- a/crates/execution/txpool/src/validator.rs
+++ b/crates/execution/txpool/src/validator.rs
@@ -440,7 +440,24 @@ impl PrecompileStorageProvider for OverlayPrecompileStorage<'_> {
 
 impl PoolTransactionError for BaseTxPoolError {
     fn is_bad_transaction(&self) -> bool {
-        true
+        match self {
+            // Deterministic and state-independent: the DA footprint exceeds the
+            // block gas limit under every possible state, so the transaction can
+            // never be included. Penalize the peer and cache it as bad.
+            Self::DaFootprintExceedsBlockGasLimit { .. } => true,
+            // EIP-8130 admission is state-relative: it authorizes against the
+            // live account-configuration and nonce-channel precompile state and
+            // the sender's nonce/code. A pool state snapshot that trails the
+            // canonical tip can transiently reject a structurally valid
+            // transaction whose authorizing state was written by the preceding
+            // block, so these failures are retryable and must not be cached as
+            // permanently bad or penalize the relaying peer. The dominant
+            // structural invariants are surfaced separately as
+            // `InvalidTransactionError` variants (classified by the inner
+            // validator); the residual deterministic cases routed here are still
+            // rejected from the pool, they merely do not penalize the peer.
+            Self::Eip8130Validation { .. } => false,
+        }
     }
 
     fn as_any(&self) -> &dyn Any {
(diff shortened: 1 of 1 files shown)


crates/execution/txpool/src/validator.rs
```diff
@@ -433,7 +450,26 @@ where
     ) -> Result<Eip8130ValidationState, InvalidPoolTransactionError> {
         let local_chain_id = self.inner.chain_spec().chain().id();
         let now = self.block_timestamp();
-        let state = self.client().latest().map_err(|error| Self::provider_unavailable(error))?;
+        // Admit against the freshest available state. `pending()` reflects the
+        // in-memory canonical tip (including the block the sequencer just
+        // produced), whereas `latest()` can trail it by a block (~one block
+        // time). EIP-8130 admission authorizes against live precompile state
```
haardikk21: this description isnt entirely accurate, as `pending` basically refers to the block not yet fully built whereas `latest` is the last fully built block. for mempool nodes i dont think they'll be able to do much here just by looking at pending state, we're probably gonna have to order txns and determine if a specific ordering as a whole is valid or not, but this will help on the builder who does have a real notion of a "pending" state as its building blocks so we should do it anyway

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/txpool/src/validator.rs
chunter-cb: Good catch — fixed. The comment and commit message now describe `pending()` correctly as the state of the block currently being built (not yet committed), one block ahead of `latest()` (the last fully committed block), and note that the benefit is primarily on the builder (which has a real in-progress block) while non-builder nodes fall back to `latest()`.

crates/execution/txpool/src/validator.rs
```diff
@@ -440,7 +440,24 @@ impl PrecompileStorageProvider for OverlayPrecompileStorage<'_> {
...
+            // canonical tip can transiently reject a structurally valid
+            // transaction whose authorizing state was written by the preceding
+            // block, so these failures are retryable and must not be cached as
+            // permanently bad or penalize the relaying peer. The dominant
+            // structural invariants are surfaced separately as
+            // `InvalidTransactionError` variants (classified by the inner
+            // validator); the residual deterministic cases routed here are still
+            // rejected from the pool, they merely do not penalize the peer.
+            Self::Eip8130Validation { .. } => false,
+        }
```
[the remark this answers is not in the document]
xenoliss: Imo that's a blocker cc @chunter-cb

PR #3753, opened 2026-06-24
feat(eip8130): account-abstraction receipt

chunter-cb:
## Summary

Adds the EIP-8130 receipt surface so AA transactions
report their extra execution data through `eth_getTransactionReceipt`

- **Consensus receipt:** a `BaseReceipt::Eip8130` variant wrapping the standard
  receipt with the per-phase `phaseStatuses` array, plus its node-local Compact
  (de)serialization for storage.
- **Executor handoff:** the EIP-8130 executor computes a per-phase status array
  while running a transaction's `calls` and publishes it to the receipt builder
  via a single-threaded thread-local slot (`Eip8130PhaseStatuses`); the receipt
  builder takes it immediately after execution. The slot is cleared at the start
  of every `execute` so a value leaked by an earlier transaction (e.g. a panic
  caught between its `set` and the builder's `take`) can never be misattributed.
  Rationale is documented on the type (the receipt builder is generic over the
  EVM and has no typed channel for per-tx metadata).
- **RPC fields:** `BaseTransactionReceipt` gains the AA fields (`payer`,
  `status`, `phaseStatuses`, `metadata`), derived in the receipt response
  builder. Empty `metadata` is omitted rather than serialized as `"0x"`,
  matching how empty `phaseStatuses` is skipped.

## Test plan

- [x] `cargo test` for `base-common-consensus`, `base-common-evm`,
      `base-common-rpc-types`, `base-execution-eip8130-rpc-node` (incl. the new
      end-to-end `tests/receipt.rs`, covering self-pay, phase statuses, and the
      sponsored-payer precedence) — all pass.
- [x] `cargo check --workspace --all-features`
- [x] `cargo clippy --all-features --all-targets` on touched crates — clean.
- [x] `cargo +nightly fmt --all -- --check` — clean.


xenoliss (approved):
nit: the phase-status handoff relies on reth executing each tx as `execute` -> `build_receipt` sequentially on one thread, but no test covers >1 EIP-8130 tx per block. We might want to add a 2-tx case (one with a reverting phase) to lock the per-tx attribution against future executor changes.

xenoliss (approved):

PR #3765, opened 2026-06-24
refactor(proposer): clean up proof recovery

jackchuma:
### What changed? Why?

Simplifies proof recovery by making ProofRecovery hold a dyn RollupProvider instead of being generic, consolidating safe/finalized head lookup into the recovery tick, reusing ProofTarget next-block math, and collecting ordered intermediate roots directly during the forward walk.

This keeps recovery cache handling smaller while preserving the existing proposer recovery behavior.

### Notes to reviewers

The main behavior change is in crates/proof/proposer/src/proof_recovery.rs. The driver, pipeline, and service changes are type glue for the non-generic recovery client.

The colocated recovery tests were reduced to direct tokio tests while still covering chain walking, cache reuse, cache invalidation, safe-head stopping, games lookup failures, and intermediate roots.

### How has it been tested?

cargo test -p base-proposer proof_recovery

### DIFF
--- a/crates/proof/proposer/src/driver.rs
+++ b/crates/proof/proposer/src/driver.rs
@@ -311,7 +311,7 @@ mod tests {
                 anchor_state_registry_address: config.anchor_state_registry_address,
                 scan_concurrency: config.recovery_scan_concurrency,
             },
-            rollup.clone(),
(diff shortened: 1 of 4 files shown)


refcell (approved):
Nice diff

PR #3766, opened 2026-06-24
fix(eip8130): authorize counterfactual smart-account creates in execution

chunter-cb:
## Problem

A type `0x7b` smart-account counterfactual CREATE transaction was accepted by the txpool but rejected at every flashblock with `rejection_reasons=["other"]`, and the account was never mined. All subsequent transactions from that account then failed with `EIP-8130 validation failed: actor is not bound`.

## Root cause

A pool-vs-execution asymmetry in how EIP-8130 sender authorization handled counterfactual creates.

- **Txpool** (`validator.rs`): detected `AccountChange::Create` as the first entry and routed to a _stateless_ create-sender path that dispatched the auth blob and checked the recovered `(actor_id, authenticator)` was in `initial_actors`. The account did not need to exist.
- **Block execution** (`transaction.rs` + `eip8130.rs`): `TransactionAuthorizer::authorize` always called `ActorTxVerifier::verify` → `resolve_bound` on the counterfactual address. The account was empty, so `resolve_bound` returned `AuthorizeError::NotBound`. `apply_create` (which installs `initial_actors`) ran _after_ the authorize step and was never reached.

The resulting `BaseTransactionError::Eip8130(...)` took the non-`record_rejection` path in the flashblock builder, surfacing as `pool_drained` + `rejection_reasons=["other"]` at INFO rather than as a named EIP-8130 error.

## Fix

Rather than patch the stateless create path, this collapses the asymmetric pool and execution paths into a single orchestrator, `TransactionAuthorizer::authorize_and_apply`, which **interleaves authorization and application** against an evolving state — mirroring `AccountConfiguration.sol`:

1. Apply each `account_changes` entry in order against the live (mutable) state. A `Create` installs `initial_actors` via `initialize_actors` (scope=0, expiry=0, i.e. an unrestricted owner) before any later entry or the sender/payer check runs.
2. Each `ConfigChange` authorizes and applies against the post-apply state, so a `Create + ConfigChange` config authorizes against the freshly-created account.
3. Sender and payer are authenticated against the **final** post-apply state via `ActorTxVerifier::verify`. The create sender is therefore authenticated against its installed unrestricted owner — there is no separately "synthesised" actor.
4. EOA sender `ecrecover` runs exactly once: the recovered token is threaded into `verify_with_recovered_sender` rather than recovered again.

Both layers now run the **same** routine:

- **Execution** (`eip8130.rs`) runs `authorize_and_apply` against the EVM journal. Apply/code writes happen before the nonce check; any `Err` is reverted wholesale by the surrounding journal checkpoint in `execute` (documented inline).
- **Txpool** (`validator.rs`) runs `authorize_and_apply` against a new writable `OverlayPrecompileStorage` layered over the read-only pool state snapshot, so admission validates against the same evolving state the builder sees at inclusion. The overlay buffers `sload`/`sstore`; it is discarded wholesale on error and never mutates canonical state.

The old asymmetric `validate_create_sender_auth` / `validate_eip8130_create_sender` stateless re-implementations are deleted.

## Behavior change: interim per-tx account-changes cap

This adds `Eip8130Constants::MAX_ACCOUNT_CHANGES_PER_TX = 3`, a pool-only **total** cap on `account_changes` entries (any mix of `Create` / `ConfigChange` / `Delegation`).

- **Before:** there was no total cap. Only per-type caps applied (≤1 `Create`, ≤1 `Delegation`, ≤`MAX_CONFIG_CHANGES_PER_TX` = 10 `ConfigChange`), so a single tx could carry up to ~12 entries.
- **After:** at most 3 entries total (e.g. 1 create + 2 config changes), so a previously-admissible `create + many configs + delegation` now bounces at admission.

This is an interim guardrail that keeps per-transaction admission work (and the overlay it applies against) small and bounded while the interleaved flow beds in. It can be raised toward `MAX_CONFIG_CHANGES_PER_TX` later. The cap sits next to `MAX_CONFIG_CHANGES_PER_TX` in `Eip8130Constants` so the relationship is discoverable.

## Other hardening

- `OverlayPrecompileStorage::checkpoint_revert` now `debug_assert!`s: the admission flow must abort wholesale, never revert internally (a nested revert would silently leak partial writes).
- `TxAuthError::Apply(ApplyError::*)` maps to specific named pool rejection reasons (e.g. "create address does not match the sender", "actor authenticator is not canonical") instead of a single generic string.

## Test plan

- [x] `cargo test -p base-execution-eip8130` — 120 pass
- [x] `cargo test -p base-execution-txpool` — 108 pass (incl. new overlay admission test)
- [x] `cargo test -p base-common-consensus` — 93 pass
- [x] `cargo clippy` on affected crates — clean
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] Devnet: re-run the SA counterfactual create scenario — account is mined and follow-up txs succeed

## New tests

- `counterfactual_create_authorizes_on_empty_account` — the regression case: create tx succeeds on an empty counterfactual account.
- `create_then_config_change_authorizes_against_freshly_created_account` — config change authorizes against the post-create actor set (`(multichain, local)` sequences == `(1, 1)` after apply).
- `counterfactual_create_wrong_signer_is_rejected` / `counterfactual_create_without_explicit_sender_is_rejected` — rejection paths.
- `admits_eip8130_create_then_config_change_via_overlay` (txpool) — pool-side coverage that the `OverlayPrecompileStorage` buffers the create so the config change authorizes against the freshly-created account during admission.
- `rejects_eip8130_too_many_account_changes` — the new total cap.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

xenoliss (approved):

PR #3821, opened 2026-06-27
fix(consensus/engine): recover pruned nodes from 4444 "pruned history unavailable" on sync start

renfeide112:
## Summary

On a **pruned** mainnet node, `base-consensus` enters a permanent crash-restart loop the moment the execution layer finishes syncing. The safe/finalized head never advances past `0` and derivation is stuck in `AwaitingELSyncCompletion`. The execution layer (`base-reth-node`) is fully healthy — only the consensus layer is broken.

This patch routes reth's `4444 "pruned history unavailable"` hard error through the existing pruned-tip recovery so the node anchors within its retained range and resumes derivation instead of crashing.

## Root cause

After the EL finishes syncing, the initial engine reset calls
`find_starting_forkchoice → L2ForkchoiceState::current_with_checkpoint_reader`, which reads reth's persisted `finalized` / `safe` forkchoice pointers. After restoring from a pruned snapshot these pointers can reference blocks **below the history floor** (observed: pointer `< earliest`). reth **hard-errors** such reads with RPC `code: 4444 "pruned history unavailable"` — it cannot serve even a header:

```
$ eth_getBlockByNumber("finalized", false)  ->  {"error":{"code":4444,"message":"pruned history unavailable"}}
```

The existing pruned-tip recovery (`find_earliest_unpruned_block`, the checkpoint reader) only handles the **soft** case where reth returns a body-less block and `L2BlockInfo::from_block_and_genesis` raises `MissingL1InfoDeposit`. The **hard** 4444 is an `RpcError` that propagates before any of that recovery runs, so `current_with_checkpoint_reader` fails on its very first request:

```
ERROR engine: Failed to drain engine tasks
      err=EngineReset(SyncStart(RpcError(ErrorResp(
          ErrorPayload { code: 4444, message: "pruned history unavailable", data: None }))))
Location: crates/consensus/cli/src/node.rs:341
```

Genesis (block 0) is also below the floor, so the existing `None`-branch fallback that fetches the genesis block hits 4444 as well — making genesis an unusable anchor on a fully pruned node.

## Fix

`crates/consensus/engine/src/sync/forkchoice.rs` — treat the 4444 response as an **absent block** consistently throughout sync start, reusing the recovery machinery that already exists for the soft case:

1. **`get_block_compat`** maps `4444 "pruned history unavailable"` to `Ok(None)` (new `is_pruned_history_error` helper), alongside the existing `"block not found"` / `"Unknown block"` handling.
2. The **finalized `None` fallback** no longer fetches the (pruned) genesis block. When the labeled block and genesis are both unavailable it recovers the **earliest unpruned block** via `find_earliest_unpruned_block`.
3. **`find_earliest_unpruned_block`'s binary search** treats a hard 4444 the same as a pruned body (`MissingL1InfoDeposit`) — search higher — instead of propagating it.

This anchors finalized/safe at the earliest block reth still retains, and the existing sequencing-window walk then places the safe head within the retained range, so derivation resumes.

## Behavior after the fix (logs)

```
INFO  engine:     Finished execution layer sync.
INFO  engine:     Performing initial engine reset
WARN  sync_start: finalized label and genesis both pruned; recovering to earliest unpruned block
WARN  sync_start: binary searching for earliest unpruned block  lo=0 hi=47891868
WARN  sync_start: found earliest unpruned block  block_number=46500000
INFO  sync_start: Loaded current L2 EL forkchoice state  unsafe=47891868 safe=46500000 finalized=46500000
```

`RestartCount` stays `0`, no further `4444`, and the safe head climbs from the anchor to chain tip; `finalized` advances from the anchor as L1 finality progresses.

## Testing

- **Unit test** added for `is_pruned_history_error` (the error-string detector — the part most likely to silently break if reth changes the message), covering the 4444 message and several unrelated RPC errors.
- **Live**: verified end-to-end on a Base **mainnet** pruned node (`v1.1.1`, `--prune.*.distance=1339200`, storage v1). Before: consensus crash-looped every ~10–70s, `safe`/`finalized` = `null`. After: `RestartCount=0`, `eth_getBlockByNumber("safe"/"finalized")` return real blocks and advance to tip.

> Note: the full sync-start recovery path isn't covered by an integration test because `MockEngineClient::get_l2_block` currently can only return `Ok(Some/None)` (no per-block error injection). Happy to extend the mock with error injection and add that test if you'd prefer it in this PR.

## Related

- Same class as the `AwaitingELSyncCompletion` / engine-reset reports on pruned nodes (e.g. #1041), but the failure here surfaces as a hard `4444` rather than `MissingL1InfoDeposit`, which the current recovery doesn't catch.

### DIFF
--- a/crates/consensus/engine/src/sync/forkchoice.rs
+++ b/crates/consensus/engine/src/sync/forkchoice.rs
@@ -5,7 +5,7 @@ use std::fmt::Display;
 use alloy_eips::{BlockId, BlockNumberOrTag};
 use alloy_provider::Network;
 use alloy_rpc_types_eth::Block as RpcBlock;
-use alloy_transport::TransportResult;
+use alloy_transport::{RpcError, TransportErrorKind, TransportResult};
 use base_common_genesis::RollupConfig;
 use base_common_network::Base;
 use base_common_rpc_types::Transaction;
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #3855, opened 2026-07-06
feat(eip8130-rpc): price estimateGas from raw sender/payer auth blobs

chunter-cb:
## Summary

Fixing `eth_estimateGas` for EIP-8130 to price the estimate from **raw `senderAuth` / `payerAuth` blobs** formatted exactly as on the wire (`authenticator(20) || data`, or a bare secp256k1 signature for the default EOA), replacing the `senderAuthScheme` / `payerAuthScheme` enum (plus `senderAuthSize` / `payerAuthSize`).

The estimator prices each blob **verbatim** — its EIP-2028 calldata cost plus the intrinsic-schedule execution gas selected by the leading 20-byte authenticator address. No signature is required or verified; only the blob's *shape* is priced, so a caller can estimate any key type by passing a correctly-shaped stub without signing first.

### Account

The sender account is named by the EIP-8130 **`sender`** field or the standard **`from`** — the two are interchangeable, and:

- if both are present they **must agree** (mismatch → `INVALID_PARAMS`);
- if **neither** is present → `INVALID_PARAMS` (no silent fallback to the zero address).

`sender` matches the on-wire `TxEip8130.sender` identity; `from` keeps existing `eth_estimateGas` / `eth_call` callers working unchanged.

### Path

The `senderAuth` blob's *form* selects the authentication path, mirroring the wire:

| `senderAuth` | Path | Priced as |
|---|---|---|
| bare signature | default-EOA (`tx.sender` unset) | verbatim |
| `authenticator(20) ‖ data` (k1 / P-256 / WebAuthn / delegate) | configured-account (`tx.sender` set) | verbatim |
| absent, `from`-only | default-EOA | bare k1 stub (65B) |
| absent, `sender` declared | configured-account | k1-prefixed stub (20+65B) |

The absent-blob default keys off intent: a declared `sender` (configured-account request) defaults to a **configured** k1 stub so it isn't under-estimated as a bare EOA; a `from`-only request defaults to the **bare** EOA form. An unrecognized 20-byte prefix on `senderAuth` is treated as a bare signature (EOA path, priced verbatim by length) — it cannot under-price, as no authenticator execution gas applies to the bare path.

A declared `payer` adds payer authentication from `payerAuth` (defaulting to a representative secp256k1 authorization). Unlike `senderAuth`, a supplied `payerAuth` is always the prefixed form and must carry a recognized enshrined authenticator selector (unrecognized → `INVALID_PARAMS`). A per-blob cap bounds the authentication *data* at `MAX_AUTH_SIZE` (excluding the 20-byte selector on the configured path).

**Recognized-selector fix:** prefix recognition (`is_prefixed_auth`) now checks the protocol's actual canonical authenticator set (`Eip8130Contracts::is_canonical_authenticator` + the native k1 sentinel) rather than the RPC estimator's own `Eip8130AuthScheme` enum. The enum only lists flat, schedule-priceable *leaf* schemes (k1/P-256/WebAuthn) and was missing `DELEGATE_AUTHENTICATOR` — a recognized selector, but a structured 3-segment blob (`delegate(20) ‖ delegate_account(20) ‖ nested_authenticator(20) ‖ nested_data`), not a flat leaf, so it can't be a variant. Without this, a delegate-authenticated `senderAuth` was silently misclassified as a bare EOA and flat-priced at k1 (under-pricing), and a delegate-authenticated `payerAuth` was wrongly rejected as `INVALID_PARAMS`. The already-correct, already-delegate-aware intrinsic-gas schedule (shared with the verifying `execute` path) now prices it correctly once the prefix is recognized — no change needed there.

### Why

`estimateGas` should speak the same authentication format the transaction actually carries (`authenticator || data`) and identify the account by the same `sender` the wire uses, rather than a closed, hand-maintained enum. This:
- mirrors the on-wire `sender`, `sender_auth` / `payer_auth` encoding (`Eip8130Signed`), so the client builds one representation for both estimate and submit;
- drops the `*AuthScheme` + `*AuthSize` surface — length and authenticator are implicit in the blob;
- extends to new authenticators without an RPC/enum change (the intrinsic schedule already keys execution gas off the 20-byte prefix);
- keeps both the default-EOA and configured-account estimate paths.

## Testing

- `cargo test -p base-common-rpc-types --features reth reth::` — 22/22 pass (account resolution incl. from/sender mismatch + neither, EOA vs configured path by blob form, absent-blob defaults by intent, unrecognized-prefix → bare EOA, prefixed vs bare data caps, payer defaults + unrecognized rejection, delegate-prefixed sender/payer auth accepted on the configured path).
- `cargo test -p base-execution-eip8130-rpc-node --test rpc` — 13/13 pass against the node harness, including `estimate_gas_prices_the_supplied_authentication_blob`, `estimate_gas_accepts_explicit_sender`, `estimate_gas_rejects_mismatched_from_and_sender`, and sponsored-payer pricing.
- `cargo clippy` + `cargo +nightly fmt --check` clean on the affected crates.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

haardikk21 (approved):

PR #3863, opened 2026-07-07
feat(zk-host): orchestrate multi-stage backend sessions in host

mw2000:
## Summary

- Move Groth16 range-to-aggregation orchestration into `proof_generator.rs`: resume SNARK sessions first, record STARK completion, call `submit_next`, and continue polling until the final proof is ready.
- Extend `ZkProver` in `prover.rs` with `submit_next` (default `Unimplemented`), plus session-aware `poll`/`download` and `initial_session_type` / `session_type` helpers on `ZkProofRequestKind`.
- Add minimal backend trait compliance (`session_type` params on cluster/network/dry-run) and mock `submit_next` support so the orchestration path is testable without cluster aggregation wiring.

This is PR 1 of a stacked Groth16 orchestration split. Backend aggregation implementation (cluster witness/submit, provider, CLI limits) will follow in a separate PR on top of this one.

### DIFF
--- a/crates/proof/zk/backend/src/succinct/dry_run.rs
+++ b/crates/proof/zk/backend/src/succinct/dry_run.rs
@@ -9,7 +9,7 @@ use async_trait::async_trait;
 use base_proof_succinct_client_utils::client::DEFAULT_INTERMEDIATE_ROOT_INTERVAL;
 use base_proof_succinct_proof_utils::get_range_elf_embedded;
 use base_proof_zk_host::{ZkProofRequestKind, ZkProver, ZkProverError, ZkSessionState};
-use base_prover_service_protocol::{ExecutionStats, ProofResult, ZkProofResult, ZkVm};
+use base_prover_service_protocol::{ExecutionStats, ProofResult, SessionType, ZkProofResult, ZkVm};
 use sp1_sdk::{
     Elf, SP1Stdin,
     blocking::{LightProver, Prover},
@@ -245,7 +245,11 @@ impl ZkProver for DryRunZkProver {
         if contains_result { Ok(ZkSessionState::Completed) } else { Ok(ZkSessionState::NotFound) }
     }
 
-    async fn download(&self, backend_session_id: &str) -> Result<ProofResult, ZkProverError> {
+    async fn download(
+        &self,
+        _session_type: SessionType,
+        backend_session_id: &str,
+    ) -> Result<ProofResult, ZkProverError> {
         self.completed_results
             .lock()
             .map_err(|e| backend_error!("dry-run result store lock poisoned: {e}"))?
--- a/crates/proof/zk/backend/src/succinct/mock.rs
+++ b/crates/proof/zk/backend/src/succinct/mock.rs
@@ -2,7 +2,10 @@
 
 use async_trait::async_trait;
 use base_proof_zk_host::{ZkProofRequestKind, ZkProver, ZkProverError, ZkSessionState};
-use base_prover_service_protocol::{ProofResult, SnarkGroth16ProofResult, ZkProofResult, ZkVm};
+use base_prover_service_protocol::{
+    ProofResult, SessionType, SnarkGroth16ProofRequest, SnarkGroth16ProofResult, ZkProofResult,
+    ZkVm,
+};
 
 /// Placeholder proof bytes returned by the mock prover.
 pub const MOCK_PROOF_BYTES: &[u8] = b"mock-zk-proof";
@@ -14,39 +17,45 @@ pub const MOCK_SNARK_PREFIX: &str = "mock-snark-";
 #[derive(Debug, Clone, Copy, Default)]
 pub struct MockZkProver;
 
-impl MockZkProver {
-    /// Derive the deterministic backend session id for a request.
-    pub fn backend_session_id(request: &ZkProofRequestKind, request_session_id: &str) -> String {
-        let session_type = if request.is_snark_groth16() { "snark" } else { "stark" };
-        format!("mock-{session_type}-{request_session_id}")
-    }
-}
-
 #[async_trait]
 impl ZkProver for MockZkProver {
     async fn submit(
         &self,
-        request: &ZkProofRequestKind,
+        _request: &ZkProofRequestKind,
         request_session_id: &str,
     ) -> Result<String, ZkProverError> {
-        Ok(Self::backend_session_id(request, request_session_id))
+        Ok(format!("mock-stark-{request_session_id}"))
+    }
+
+    async fn submit_next(
+        &self,
+        _request: &SnarkGroth16ProofRequest,
+        request_session_id: &str,
+        _completed_backend_session_id: &str,
+    ) -> Result<String, ZkProverError> {
+        Ok(format!("{MOCK_SNARK_PREFIX}{request_session_id}"))
     }
 
     async fn poll(&self, _backend_session_id: &str) -> Result<ZkSessionState, ZkProverError> {
         Ok(ZkSessionState::Completed)
     }
 
-    async fn download(&self, backend_session_id: &str) -> Result<ProofResult, ZkProverError> {
+    async fn download(
+        &self,
+        session_type: SessionType,
+        _backend_session_id: &str,
+    ) -> Result<ProofResult, ZkProverError> {
         let zk_proof = ZkProofResult {
             zk_vm: ZkVm::Sp1,
             proof: MOCK_PROOF_BYTES.to_vec().into(),
             execution_stats: None,
         };
 
-        if backend_session_id.starts_with(MOCK_SNARK_PREFIX) {
-            Ok(ProofResult::SnarkGroth16(SnarkGroth16ProofResult { proof: zk_proof }))
-        } else {
-            Ok(ProofResult::Compressed(zk_proof))
+        match session_type {
+            SessionType::Snark => {
+                Ok(ProofResult::SnarkGroth16(SnarkGroth16ProofResult { proof: zk_proof }))
+            }
+            SessionType::Stark => Ok(ProofResult::Compressed(zk_proof)),
         }
     }
 }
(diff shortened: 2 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/zk/host/src/proof_generator.rs
```diff
@@ -314,7 +350,56 @@ where
...
+                    {
+                        let next_session_type = SessionType::Snark;
+                        let next_backend_session_id = self
+                            .prover
+                            .submit_next(
+                                proof_request,
+                                &request.claim.session_id,
+                                &backend_session_id,
+                            )
+                            .await?;
```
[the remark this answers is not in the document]
mw2000: We document this and this will be idempotent for the cluster and network implementations

crates/proof/zk/host/src/proof_generator.rs
```diff
@@ -220,92 +215,108 @@ where
...
-            self.wait_for_backend_session(
+        // Every request first proves a range (STARK) proof.
+        let range_session_id = self
+            .drive_stage(
                 request,
                 &handle,
-                session_type,
-                &resolved.backend_session_id,
+                SessionType::Stark,
+                self.prover.submit(&request.request, &request.claim.session_id),
```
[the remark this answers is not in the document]
mw2000: The stacked aggregation PRs update submit to pull the inner range proof out of a SnarkGroth16 request

crates/proof/zk/backend/src/succinct/dry_run.rs
```diff
@@ -245,7 +245,11 @@ impl ZkProver for DryRunZkProver {
         if contains_result { Ok(ZkSessionState::Completed) } else { Ok(ZkSessionState::NotFound) }
     }
 
-    async fn download(&self, backend_session_id: &str) -> Result<ProofResult, ZkProverError> {
+    async fn download(
+        &self,
+        _session_type: SessionType,
+        backend_session_id: &str,
```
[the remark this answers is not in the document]
mw2000: Dry-run doesn't derive anything from session_type. It returns the already-typed ProofResult that was stored (store.remove(id)). The stored value is the source of truth

crates/proof/zk/host/src/proof_generator.rs
```diff
@@ -220,136 +215,160 @@ where
...
                     sleep(self.normalized_poll_interval()).await;
                 }
-                ZkSessionState::Completed => return Ok(()),
+                ZkSessionState::Completed => {
+                    handle
+                        .record(
+                            session_type,
+                            backend_session_id.clone(),
+                            BackendSessionState::Completed,
+                        )
```
jackchuma: Stop recording terminal session states via worker API. When any backend poll returns `Completed`, this tries to upsert `BackendSessionState::Completed`. The prover-service worker API rejects terminal session states before writing (`record_worker_proof_session` returns `TerminalSessionStatus` for `Completed`/`Failed`), so a newly finished compressed or STARK stage now errors as `ZkProverError::Session` and never reaches download/submission. Keep the worker-recorded state non-terminal or add a server path that accepts completion before using this

crates/proof/zk/host/src/proof_generator.rs
```diff
@@ -220,136 +215,160 @@ where
...
+        // Every request first proves a range (STARK) proof.
+        let range_session_id = self
+            .drive_stage(
                 request,
                 &handle,
-                session_type,
-                &resolved.backend_session_id,
+                SessionType::Stark,
+                self.prover.submit(&request.request, &request.claim.session_id),
             )
```
jackchuma, line 240: Submit a compressed range request for SNARK first stage. For a `SnarkGroth16` job with no existing STARK session, this passes the SNARK request itself into the range stage. The concrete non-mock backends' `submit` methods only accept `ZkProofRequestKind::Compressed` and return an error for `SnarkGroth16`, so the new two-stage flow never reaches `submit_next`; build the first-stage request from `proof_request.proof` instead

crates/proof/zk/backend/src/succinct/mock.rs
```diff
@@ -15,38 +18,51 @@ pub const MOCK_SNARK_PREFIX: &str = "mock-snark-";
 pub struct MockZkProver;
 
 impl MockZkProver {
```
jackchuma: `backend_session_id` is a public helper, but it only wraps one `format!` call and only has one production caller. I’d inline it in `submit`; the tests can assert the returned string directly.

```rust
  Ok(format!("mock-stark-{request_session_id}"))
```

crates/proof/zk/host/src/prover.rs
```diff
@@ -97,11 +84,37 @@ pub trait ZkProver: Send + Sync + std::fmt::Debug {
         request_session_id: &str,
     ) -> Result<String, ZkProverError>;
 
+    /// Submit the Groth16 aggregation backend session for a completed range proof.
```
jackchuma: This doc block is carrying more design detail than the trait needs. The key contract is idempotency; the crash-window and backend-provider explanation makes the public trait harder to scan.

```rust
  /// Implementations must be idempotent for `(request_session_id, completed_backend_session_id)`.
```

crates/proof/zk/host/src/prover.rs
```diff
@@ -97,11 +84,37 @@ pub trait ZkProver: Send + Sync + std::fmt::Debug {
...
+        _request_session_id: &str,
+        _completed_backend_session_id: &str,
+    ) -> Result<String, ZkProverError> {
+        Err(ZkProverError::Unimplemented)
+    }
+
     /// Poll the backend session, returning its current state.
     async fn poll(&self, backend_session_id: &str) -> Result<ZkSessionState, ZkProverError>;
 
     /// Download the completed proof for a backend session.
```
jackchuma, line 104: The `download` doc repeats information already visible in the signature and enum names. I’d keep the summary line and let `SessionType`/`ProofResult` carry the rest.

```rust
  /// Download the completed proof for a backend session.
```

crates/proof/zk/host/src/proof_generator.rs
```diff
@@ -220,136 +215,160 @@ where
...
             .await?;
-        }
 
-        let proof = self.prover.download(&resolved.backend_session_id).await?;
+        // Groth16 requests then aggregate the completed range proof into a SNARK.
+        let (final_session_type, final_session_id) = match &request.request {
```
jackchuma: `final_session_type` and `final_session_id` only exist to feed the next call. Returning the download from each `match` arm keeps the two paths local and avoids the temporary tuple.

```rust
  match &request.request {
      ZkProofRequestKind::SnarkGroth16(proof_request) => {
          let snark_session_id = self
              .drive_stage(...)
              .await?;
          self.prover.download(SessionType::Snark, &snark_session_id).await
      }
      ZkProofRequestKind::Compressed(_) => {
          self.prover.download(SessionType::Stark, &range_session_id).await
      }
  }
```

crates/proof/zk/host/src/proof_generator.rs
mw2000: Confirmed the full chain — `BackendSessionState::Completed` → `SessionStatus::Completed` → `is_terminal()` → `record_worker_proof_session` returns `TerminalSessionStatus`, which `record_proof_session_inner` maps to an ownership rejection, surfacing as `ZkProverError::Session`. So the STARK stage errored the instant it completed. `poll_to_completion` no longer records the terminal state on completion; it just returns the backend session id (with a debug log). `record_backend_resubmission` already only writes the non-terminal `Submitting` state.

crates/proof/zk/host/src/proof_generator.rs
mw2000: Fixed — `prove_to_completion` now builds a `ZkProofRequestKind::Compressed` for the first stage (from `snark.proof` for Groth16 jobs), so backends always receive a compressed range request in `submit` and the two-stage flow reaches `submit_next`.

crates/proof/zk/backend/src/succinct/mock.rs
mw2000: Done — inlined the `format!` into `submit` and removed the helper; the tests already assert the returned string directly.

crates/proof/zk/host/src/proof_generator.rs
mw2000: Done — each match arm now returns its own `download(...)`, dropping the `final_session_type`/`final_session_id` temporaries.

crates/proof/zk/host/src/prover.rs
mw2000: Trimmed the block down to the idempotency contract as suggested. One deviation from the exact wording: I kept "should" rather than "must" (plus one clause noting the exception) because the network backend can't be idempotent.

crates/proof/zk/host/src/proof_generator.rs
```diff
@@ -220,136 +215,169 @@ where
...
+                backend_session_id
             }
-        }
+        };
+
+        self.poll_to_completion(request, handle, session_type, backend_session_id).await
     }
 
-    async fn wait_for_backend_session(
+    /// Poll a running backend session until it completes, recording the terminal state.
```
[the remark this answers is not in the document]
mw2000: Good catch — reworded to "Poll a running backend session until it reaches a terminal state." to match the new behavior where completion is not persisted.

crates/proof/zk/host/src/proof_generator.rs
```diff
@@ -220,136 +215,169 @@ where
...
-            self.wait_for_backend_session(
+        // Every request begins with a range (STARK) proof. A Groth16 job proves the range over the
+        // compressed request nested in its SNARK request, since backends only submit compressed
+        // proofs in the first stage.
+        let range_request = match &request.request {
+            ZkProofRequestKind::Compressed(proof) => ZkProofRequestKind::Compressed(proof.clone()),
+            ZkProofRequestKind::SnarkGroth16(snark) => {
+                ZkProofRequestKind::Compressed(snark.proof.clone())
+            }
+        };
```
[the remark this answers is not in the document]
mw2000: Kept the clone here. `ZkProofRequest` has no heap-allocated fields — `start_block_number`/`number_of_blocks_to_prove` are `u64`, `sequence_window`/`intermediate_root_interval` are `Option<u64>`, `l1_head` is `Option<B256>` (a Copy 32-byte array), and `zk_vm` is a small enum — so the clone is a trivial stack copy with no allocation (`sequence_window` isn't Vec-like). The deferred-init borrow pattern would trade the readable `match` for marginal-to-no benefit, so I left it as-is.

crates/proof/zk/host/src/proof_generator.rs
```diff
@@ -220,136 +215,169 @@ where
...
-            .map_err(|error| ZkProverError::Session(Box::new(error)))?;
-        let resolved =
-            self.resolve_backend_session(request, &handle, session_type, existing).await?;
-
-        if resolved.needs_polling {
-            self.wait_for_backend_session(
+        // Every request begins with a range (STARK) proof. A Groth16 job proves the range over the
+        // compressed request nested in its SNARK request, since backends only submit compressed
+        // proofs in the first stage.
+        let range_request = match &request.request {
```
jackchuma, line 228: Now that Groth16 jobs explicitly run the range proof first, `submit` is only called for compressed range requests. This re-wraps the inner proof into `ZkProofRequestKind::Compressed` just to pass through a trait method that still accepts the broader enum.

  Can `ZkProver::submit` take `&ZkProofRequest` instead? Then cluster/network/dry-run no longer need their `SnarkGroth16` rejection branches, and this caller can pass the range request directly

jackchuma (approved):

PR #3892, opened 2026-07-09
feat(eip8130-rpc): accept senderActorId hint for estimateGas simulation

chunter-cb:
## Summary
- Estimation never recovers a signature, so `simulate` always published the account's **self-actor** to the `TxContext` precompile. Policy-gated session-key calls then looked up the wrong policy (`NoActivePolicy`) and the estimate threw.
- Add optional `senderActorId` on the EIP-8130 `eth_estimateGas` request (orthogonal to the existing `senderAuth` / verifier shape hint for auth-gas pricing).
- Thread it onto `Eip8130TransactionParts.simulation_sender_actor_id` (RPC-only; consensus path stays `None`).
- In `simulate_resolve`, apply `accountChanges` first, then resolve the hinted actor's policy via `get_policy` so same-tx authorizations are visible. Absent hint → self-actor (backward compatible).

## Test plan
- [x] `cargo test -p base-common-rpc-types --features reth --lib reth::tests` (24 passed, including new `sender_actor_id_*` tests)
- [x] `cargo test -p base-common-evm --lib eip8130::tests::simulate` (4 passed)
- [x] `just check::format`

### DIFF
--- a/crates/common/evm/src/eip8130.rs
+++ b/crates/common/evm/src/eip8130.rs
@@ -625,9 +625,7 @@ impl Eip8130Executor {
     ///
     /// `acting_actor_hint` is the optional `senderActorId` from the estimate
     /// request. Without it, simulation publishes the self-actor (backward
-    /// compatible). With it, policy-gated session-key estimates can name the
-    /// intended actor — including one authorized in the same request's
-    /// `accountChanges`, which is why policy is resolved *after* apply.
+    /// compatible).
     fn simulate_resolve<DB>(
         ctx: &mut BaseContext<DB>,
         signed: &base_common_consensus::Eip8130Signed,
@@ -686,8 +684,8 @@ impl Eip8130Executor {
             //    RPC hint names the intended actor (e.g. a session key); absent
             //    that, fall back to the account's self-actor. Policy is read from
             //    the post-apply journal so same-tx authorizations are visible.
-            //    Revocation/expiry are not enforced here because estimation
-            //    prices the happy-path gas.
+            //    Expiry is not enforced (estimation prices the happy path);
+            //    `get_policy` still treats a revoked default-EOA self as ungated.
             let acc = AccountConfigurationStorage::new(sctx);
             let sender_actor_id = acting_actor_hint
                 .unwrap_or_else(|| AccountConfigurationStorage::self_actor_id(sender));
@@ -1451,9 +1449,10 @@ impl Eip8130Executor {
 mod tests {
     use alloy_evm::{Evm, FromTxWithEncoded, precompiles::PrecompilesMap};
     use alloy_primitives::{Address, B256, Bytes, U256, address, bytes, keccak256};
+    use alloy_sol_types::{SolValue, sol};
     use base_common_consensus::{
-        AccountChange, BaseTxEnvelope, Call, CreateEntry, Eip8130Signed, InitialActor, Predeploys,
-        TxEip8130,
+        AccountChange, ActorChange, ActorChangeType, BaseTxEnvelope, Call, ConfigChange,
+        CreateEntry, Eip8130Signed, InitialActor, Predeploys, TxEip8130,
     };
     use base_execution_eip8130::AccountChangeApplier;
     use k256::ecdsa::SigningKey;
@@ -1468,7 +1467,10 @@ mod tests {
     };
 
     use super::*;
-    use crate::{BaseEvm, BaseSpecId, BaseTransaction, BaseUpgrade, Builder, DefaultBase};
+    use crate::{
+        BaseEvm, BaseSpecId, BaseTransaction, BaseUpgrade, Builder, DefaultBase,
+        Eip8130ExecutionMode,
+    };
 
     const CHAIN_ID: u64 = 8453;
     const NOW: u64 = 1_000;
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/evm/src/eip8130.rs
```diff
@@ -679,11 +676,24 @@ impl Eip8130Executor {
...
+            //    priced. Must precede actor/policy resolution so an actor
+            //    authorized in this same estimate request is visible.
             Self::apply_account_changes(signed, sctx, sender)?;
 
+            // 3. Resolve the acting actor. No signature recovery: the optional
+            //    RPC hint names the intended actor (e.g. a session key); absent
+            //    that, fall back to the account's self-actor. Policy is read from
+            //    the post-apply journal so same-tx authorizations are visible.
+            //    Revocation/expiry are not enforced here because estimation
+            //    prices the happy-path gas.
```
xenoliss: Comment now inaccurate. `get_policy` does enforce default-EOA revocation (returns `(0, ZERO, ZERO)` `via default_eoa_revoked()`), unlike the removed hand-rolled path.

crates/common/evm/src/eip8130.rs
```diff
@@ -610,20 +613,28 @@ impl Eip8130Executor {
...
+    /// authentication gas for the sender's (and any payer's) declared
+    /// authenticator is priced from the synthesized auth-blob shape via
+    /// [`IntrinsicGas`]. Storage writes land on the journal; the caller's
+    /// checkpoint reverts them.
+    ///
+    /// `acting_actor_hint` is the optional `senderActorId` from the estimate
+    /// request. Without it, simulation publishes the self-actor (backward
+    /// compatible). With it, policy-gated session-key estimates can name the
+    /// intended actor — including one authorized in the same request's
+    /// `accountChanges`, which is why policy is resolved *after* apply.
```
lukasrosario: nit: second part of this comment seems kind of redundant / verbose

crates/common/evm/src/eip8130.rs
```diff
@@ -679,11 +676,24 @@ impl Eip8130Executor {
...
+            //    RPC hint names the intended actor (e.g. a session key); absent
+            //    that, fall back to the account's self-actor. Policy is read from
+            //    the post-apply journal so same-tx authorizations are visible.
+            //    Revocation/expiry are not enforced here because estimation
+            //    prices the happy-path gas.
+            let acc = AccountConfigurationStorage::new(sctx);
+            let sender_actor_id = acting_actor_hint
+                .unwrap_or_else(|| AccountConfigurationStorage::self_actor_id(sender));
+            let (policy_type, policy_target, _) =
+                acc.get_policy(sender, sender_actor_id).map_err(BaseTransactionError::eip8130)?;
```
xenoliss, line 693: We should probably add a test for this resolution path. The new tests only check that the hint is threaded through rpc-types, not that it actually changes the resolved policy here. Something like a simulate test with a session-key actor authorized in `accountChanges` and passed via the hint, checking that `policy_type`/`policy_target` resolve to that actor and the estimate no longer reverts with `ActorPolicyViolation`, would cover the actual fix.

lukasrosario (dismissed):
one nit, lgtm otherwise

crates/common/evm/src/eip8130.rs
chunter-cb: Good catch — updated the comment to note that expiry is not enforced, while `get_policy` still treats a revoked default-EOA self as ungated. Also added `simulate_sender_actor_id_hint_resolves_policy_after_account_changes` covering the resolution path.

crates/common/evm/src/eip8130.rs
chunter-cb: Trimmed — dropped the redundant second half.

crates/common/evm/src/eip8130.rs
chunter-cb: Added in `50b6decda`: `simulate_sender_actor_id_hint_resolves_policy_after_account_changes` gates the self-actor to a wrong target, authorizes a session actor in the same estimate's `accountChanges`, and asserts the call reverts without the hint / succeeds with it.

crates/common/rpc-types/src/transaction/request.rs
```diff
@@ -182,6 +191,7 @@ impl Eip8130RequestFields {
             || self.sender_auth.is_some()
             || self.payer.is_some()
             || self.payer_auth.is_some()
+            || self.sender_actor_id.is_some()
```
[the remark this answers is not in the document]
chunter-cb: Excluded `sender_actor_id` from `is_some()` — it's simulation-only metadata about an 8130 request, not a defining field. A lone `senderActorId` is now ignored rather than routing onto the AA path. Added `sender_actor_id_alone_does_not_mark_request_as_eip8130`.

xenoliss (approved):

PR #3894, opened 2026-07-09
fix(consensus): preserve L2 finalization progress

yannickhilber:
We observed this issue across several of our RPC nodes. The investigation was initially triggered by unbounded memory growth and repeated OOM kills. We traced the root cause to lost finalization progress, applied this patch to our nodes, and confirmed it resolves the problem. Details in the commit message below:

The L2Finalizer loses finalization progress in two independent ways:

try_finalize_next() silently drops a finalized L1 signal when the awaiting_finalization queue is empty. The L1 watcher does not re-send signals, so the EL's finalized head can stay stuck indefinitely; nodes restarted during sync report finalized=0 for days. Since reth's changeset cache eviction keys off the finalized block, the wedge also causes unbounded memory growth and eventual OOM.

Every pipeline reset clears the whole queue. An entry only becomes finalizable once the finalized L1 overtakes its L1 origin (~2 L1 epochs, ~13 min). When resets occur more frequently than that (e.g. transient BlockNotFound resets every few minutes), entries never survive long enough, and finalization only advances in multi-hour batches.

Fix:

- Remember the highest finalized L1 block number observed and preserve it across clear(). After each enqueue, retry finalization against the stored signal via try_finalize_pending().

- On Signal::Reset, drop only the queue entries above the safe head instead of clearing the whole queue. Blocks at or below the safe head survive the reset unchanged, so their entries remain valid.

Neither change relaxes the finality rule: an L2 block is only ever finalized when its derived_from L1 block is itself finalized. The changes only stop already-satisfied finalizations from being lost.

### DIFF
--- a/crates/consensus/service/src/actors/derivation/actor.rs
+++ b/crates/consensus/service/src/actors/derivation/actor.rs
@@ -107,10 +107,11 @@ where
 
     /// Handles a [`Signal`] received over the derivation signal receiver channel.
     async fn signal(&mut self, signal: Signal) {
-        if let Signal::Reset(ResetSignal { l2_safe_head: _reset_safe_head }) = signal {
-            Metrics::derivation_l1_origin().absolute(_reset_safe_head.l1_origin.number);
-            // Clear the finalization queue on reset.
-            self.finalizer.clear();
+        if let Signal::Reset(ResetSignal { l2_safe_head: reset_safe_head }) = signal {
+            Metrics::derivation_l1_origin().absolute(reset_safe_head.l1_origin.number);
+            // Prune finalization queue entries above the reset anchor; entries at or
+            // below the safe head remain canonical across a reset.
+            self.finalizer.prune_above(reset_safe_head.block_info.number);
             // Discard any in-flight derived_from so that a stale pre-reset L1 inclusion
             // block is never recorded for a post-reset safe head confirmation.
             self.pending_derived_from = None;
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #3896, opened 2026-07-09
refactor(challenger): extract anchor updater

jackchuma:
### What changed? Why?

Extracts anchor advancement out of `BondManager` into a dedicated `AnchorUpdater` owned by the challenger driver. Each poll reads the current `AnchorStateRegistry` snapshot, computes the next dispute-game UUID from the current anchor, L2 output roots, and AggregateVerifier intervals, then submits `setAnchorState()` only after that next game is a finalized defender win and passes registry preflight checks.

Challenger startup now reads the respected game type from `AnchorStateRegistry`, resolves its AggregateVerifier implementation from `DisputeGameFactory`, and validates the onchain block intervals before constructing the updater. The old `--anchor-update-retention` CLI/config path, retained-game metrics, and per-game anchor update caches are removed so `BondManager` is back to bond lifecycle only: discovery, resolve, unlock, and claim.

Shared `DisputeGameFactory` game lookup helpers now build the checkpoint blocks, `rootClaim`, and `extraData` used by challenger anchor updates and proposer submit/recovery paths. This keeps proposer and challenger UUID lookups consistent for game chains with intermediate roots, while terminally ignored challenger games are bounded to avoid unbounded memory growth.

### Notes to reviewers

`AnchorUpdater` follows only the next game after the current anchor instead of collecting candidates from successful challenges or bond discovery. It caches that lookup while the anchor is unchanged, waits on transient readiness checks, and relies on external anchor advancement to move past stale or permanently ineligible next games.

The respected game type and AggregateVerifier intervals are read once during challenger startup. Startup fails fast if the respected game type has no registered implementation, either interval is zero, or the block interval is not divisible by the intermediate interval.

Successful challenges still register games for bond tracking when bond claiming is enabled, but they no longer register anchor update work with the bond manager.

### How has it been tested?

- `cargo fmt --package base-challenger --check`
- `cargo test -p base-challenger anchor`
- `cargo test -p base-proof-contracts game_lookup`
- `cargo test -p base-proposer recovery`

### DIFF
--- a/crates/proof/challenge/src/anchor.rs
+++ b/crates/proof/challenge/src/anchor.rs
@@ -99,6 +99,7 @@ impl AnchorUpdater {
                     status = %status,
                     "next game cannot update anchor"
                 );
+                self.cached_next_game = None;
                 return;
             }
             Err(e) => {
@@ -467,11 +468,14 @@ mod tests {
             mock_state(GameStatus::ChallengerWins, Address::ZERO, 100),
         )]));
         let submitter = MockBondTransactionSubmitter::with_responses(vec![]);
-        let mut updater = updater(factory, anchor_registry, Arc::new(l2));
+        let mut updater = updater(Arc::clone(&factory), anchor_registry, Arc::new(l2));
 
+        updater.poll(&verifier, &submitter).await;
+        factory.uuid_games.lock().unwrap().clear();
         updater.poll(&verifier, &submitter).await;
 
         assert!(submitter.recorded_calls().is_empty());
+        assert_eq!(verifier.status_read_count(challenger_win), 1);
     }
 
     #[tokio::test]
--- a/crates/proof/challenge/src/anchor.rs
+++ b/crates/proof/challenge/src/anchor.rs
@@ -1,0 +1,486 @@
+//! Anchor root update management.
+
+use std::sync::Arc;
+
+use alloy_primitives::Address;
+use base_proof_contracts::{
+    AggregateVerifierClient, AnchorRoot, AnchorSnapshot, AnchorStateRegistryClient,
+    DisputeGameFactoryClient, GameStatus, encode_set_anchor_state_calldata, game_lookup_blocks,
+    game_lookup_key,
+};
+use base_proof_rpc::L2Provider;
+use futures::stream::{self, StreamExt};
+use tracing::{debug, info, warn};
+
+use crate::{BondTransactionSubmitter, ChallengerMetrics, OutputValidator};
+
+/// Best-effort updater for the `AnchorStateRegistry`.
+pub struct AnchorUpdater {
+    factory_client: Arc<dyn DisputeGameFactoryClient>,
+    anchor_registry_client: Arc<dyn AnchorStateRegistryClient>,
+    output_validator: OutputValidator<dyn L2Provider>,
+    cached_next_game: Option<(AnchorSnapshot, Address)>,
+    anchor_state_registry_address: Address,
+    game_type: u32,
+    block_interval: u64,
+    intermediate_block_interval: u64,
+}
+
+impl std::fmt::Debug for AnchorUpdater {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AnchorUpdater")
+            .field("anchor_state_registry_address", &self.anchor_state_registry_address)
+            .field("game_type", &self.game_type)
+            .field("block_interval", &self.block_interval)
+            .field("intermediate_block_interval", &self.intermediate_block_interval)
+            .finish_non_exhaustive()
+    }
+}
+
+impl AnchorUpdater {
+    /// Creates an anchor updater.
+    pub fn new(
+        factory_client: Arc<dyn DisputeGameFactoryClient>,
+        anchor_registry_client: Arc<dyn AnchorStateRegistryClient>,
+        l2_provider: Arc<dyn L2Provider>,
+        anchor_state_registry_address: Address,
+        game_type: u32,
+        block_interval: u64,
+        intermediate_block_interval: u64,
+    ) -> Self {
+        Self {
+            factory_client,
+            anchor_registry_client,
+            output_validator: OutputValidator::new(l2_provider),
+            cached_next_game: None,
+            anchor_state_registry_address,
+            game_type,
+            block_interval,
+            intermediate_block_interval,
+        }
+    }
+
+    /// Finds the next game after the anchor game and advances the anchor root if it is ready.
+    pub async fn poll(
+        &mut self,
+        verifier_client: &dyn AggregateVerifierClient,
+        submitter: &dyn BondTransactionSubmitter,
+    ) {
+        let anchor = match self.anchor_registry_client.anchor_snapshot().await {
+            Ok(anchor) => anchor,
+            Err(e) => {
+                warn!(error = %e, "failed to read anchor snapshot for anchor update");
+                return;
+            }
+        };
+
+        let game_address = match self.cached_next_game {
+            Some((cached_anchor, address)) if cached_anchor == anchor => address,
+            _ => {
+                self.cached_next_game = None;
+                let Some(address) = self.next_game(anchor.anchor_root, anchor.anchor_game).await
+                else {
+                    return;
+                };
+                self.cached_next_game = Some((anchor, address));
+                address
+            }
+        };
+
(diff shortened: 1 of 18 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/challenge/src/anchor.rs
```diff
@@ -0,0 +1,528 @@
...
+                    return;
+                };
+                self.cached_next_game = Some((anchor, address));
+                address
+            }
+        };
+
+        match verifier_client.status(game_address).await {
+            Ok(GameStatus::DefenderWins) => {}
+            Ok(GameStatus::InProgress) => {
```
leopoldjoy, line 92: `AnchorUpdater` still waits on `InProgress` instead of resolving ready games, so it doesn't own the full flow yet.

crates/proof/challenge/src/bond.rs
```diff
@@ -568,7 +522,7 @@ impl<C: Clock> BondManager<C> {
         )
         .await;
 
-        let mut discovered = 0u64;
+        let mut discovered = 0;
```
leopoldjoy, line 525: Discovery still only updates internal bond tracking, so the driver can't seed these games into `AnchorUpdater`.

crates/proof/challenge/src/driver.rs
```diff
@@ -1236,6 +1270,18 @@ mod tests {
         assert!(!driver.pending_proofs.contains_key(&addr(0)));
     }
 
+    #[tokio::test]
+    async fn challenge_success_does_not_track_anchor_update() {
```
leopoldjoy, line 1274: This locks in missing `AnchorUpdater` registration after successful challenges.

crates/proof/challenge/src/driver.rs
```diff
@@ -197,11 +210,13 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
...
     /// First polls any in-flight proof sessions that are not in the current
     /// scan batch, then discovers claimable bonds, advances bond lifecycle
-    /// claims, and finally scans for new candidates and processes them.
+    /// claims, polls anchor updates, and finally scans for new candidates and
+    /// processes them.
     pub async fn step(&mut self) -> eyre::Result<()> {
         self.poll_pending_proofs().await;
         self.discover_claimable_bonds().await;
         self.poll_bond_claims().await;
+        self.anchor_updater.poll(&*self.verifier_client, &self.submitter).await;
```
leopoldjoy, line 219: Polling the updater here is not the promised scanner recovery pass for defender wins/read failures.

crates/proof/challenge/src/anchor.rs
```diff
@@ -0,0 +1,528 @@
...
+
+        let preflight = match verifier_client.anchor_preflight(asr_address, game_address).await {
+            Ok(preflight) => preflight,
+            Err(e) => {
+                warn!(game = %game_address, asr = %asr_address, error = %e, "failed to read anchor preflight");
+                return false;
+            }
+        };
+
+        if preflight.permanently_ineligible() {
```
leopoldjoy, line 224: Permanent/stale skips return `false`, so the cached game is retried and `skipped` increments every poll.

crates/proof/challenge/src/cli.rs
```diff
@@ -60,6 +60,10 @@ pub struct ChallengerArgs {
     #[arg(long = "anchor-state-registry-addr", env = cli_env!("ANCHOR_STATE_REGISTRY_ADDR"))]
     pub anchor_state_registry_addr: Address,
 
+    /// Game type ID for `AggregateVerifier` dispute games.
+    #[arg(long = "game-type", env = cli_env!("GAME_TYPE"))]
```
leopoldjoy: Requiring `--game-type` is avoidable and can silently target the wrong UUID when ASR already exposes it.

crates/proof/challenge/src/anchor.rs
jackchuma: Not sure I understand the point you're making here

crates/proof/challenge/src/bond.rs
jackchuma: Not sure I understand the point you're making here either .. `AnchorUpdater` purposely does not accept seeded games

crates/proof/challenge/src/driver.rs
jackchuma: Is that a bad thing?

crates/proof/challenge/src/driver.rs
jackchuma: Is that a bad thing?

crates/proof/challenge/src/anchor.rs
jackchuma: I don't follow your point

crates/proof/challenge/src/cli.rs
leopoldjoy: To clarify, I mean this value already exists on-chain as ASR’s respected game type, so a separate required config can drift unless we read it from ASR or validate the CLI value against ASR.

crates/proof/challenge/src/anchor.rs
leopoldjoy: `try_update()` returns false for both retryable and permanent skips, and `poll()` only clears the cache on true, so blacklisted / retired / stale games can be reprocessed and counted as skipped every poll.

crates/proof/challenge/src/driver.rs
leopoldjoy: Not necessarily bad if the UUID lookup is the intended replacement, but the PR notes specifically mention a scanner-based recovery pass for post-anchor `DefenderWins` games / status-read failures, and I don’t see that path.

crates/proof/challenge/src/driver.rs
leopoldjoy: Only if the PR description is intended literally: it lists successful challenges as an anchor-eligibility source, but the code / test assert successful challenges do not feed `AnchorUpdater`.

crates/proof/challenge/src/bond.rs
leopoldjoy: That’s the mismatch I was pointing at: the PR notes say `BondManager` returns recovered / discovered games for the driver to track, but this implementation has `AnchorUpdater` discover via UUID lookup instead.

crates/proof/challenge/src/anchor.rs
leopoldjoy: I meant this as a PR-description mismatch: if `AnchorUpdater` owns resolve plus `setAnchorState()`, it needs to resolve `InProgress` games whose dispute period elapsed; otherwise the description should say it only handles already-resolved `DefenderWins` games.

0xth4nh (approved):

crates/proof/challenge/src/service.rs
```diff
@@ -109,6 +110,36 @@ impl ChallengerService {
         );
 
         let verifier_client = AggregateVerifierContractClient::new(l1_rpc_url.clone())?;
+        let impl_address = factory_client.game_impls(config.game_type).await?;
```
leopoldjoy, line 113: The PR says startup reads ASR's respected game type, but this still uses configured `--game-type`; please read or validate ASR's value before resolving the impl.

PR #3918, opened 2026-07-10
feat(prover-service): route zk jobs by requested backend

mw2000:
## Summary
- add a protocol-level ZK backend selector with cluster-compatible defaults
- persist backend selection and filter claims by worker-advertised capabilities
- preserve rolling-deploy behavior for legacy requests and nullable database rows

## Stack
1. **Backend-aware prover queue** (this PR)
2. Host backend routing
3. Simultaneous backend configuration

### DIFF
--- a/crates/proof/challenge/src/driver.rs
+++ b/crates/proof/challenge/src/driver.rs
@@ -27,7 +27,9 @@ use base_proof_primitives::ProofRequest as TeeProofRequest;
 use base_proof_rpc::L2Provider;
 use base_proof_submission::KnownRevert;
 use base_prover_service_client::ProofRequesterProvider;
-use base_prover_service_protocol::{SnarkGroth16ProofRequest, TeeKind, ZkProofRequest, ZkVm};
+use base_prover_service_protocol::{
+    SnarkGroth16ProofRequest, TeeKind, ZkBackend, ZkProofRequest, ZkVm,
+};
 use base_runtime::{Clock, TokioRuntime};
 use base_tx_manager::{TxManager, TxManagerError};
 use tokio::select;
@@ -662,6 +664,7 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
                 l1_head: Some(candidate.l1_head),
                 intermediate_root_interval: Some(candidate.intermediate_block_interval),
                 zk_vm: ZkVm::Sp1,
+                zk_backend: ZkBackend::Cluster,
             },
             prover_address: self.submitter.sender_address(),
         })
@@ -1098,6 +1101,7 @@ mod tests {
                 l1_head: Some(B256::repeat_byte(0xAA)),
                 intermediate_root_interval: Some(10),
                 zk_vm: ZkVm::Sp1,
+                zk_backend: ZkBackend::Cluster,
             },
             prover_address: Address::repeat_byte(0xCC),
         }
--- a/crates/proof/prover-service/db/migrations/014_add_zk_backend.sql
+++ b/crates/proof/prover-service/db/migrations/014_add_zk_backend.sql
@@ -1,0 +1,62 @@
+-- Migration 014: Persist the ZK proving backend selected by each request.
+--
+-- Callers choose mock / dry_run / cluster / network per request. Workers claim
+-- jobs whose zk_backend is in their advertised capability list.
+BEGIN;
+
+ALTER TABLE proof_requests ADD COLUMN IF NOT EXISTS zk_backend VARCHAR(32);
+
+-- Preserve historical updated_at values while backfilling.
+ALTER TABLE proof_requests DISABLE TRIGGER update_proof_requests_updated_at;
+
+-- Existing ZK rows were always executed by cluster-configured hosts.
+UPDATE proof_requests
+SET zk_backend = COALESCE(zk_backend, 'cluster')
+WHERE api_proof_type IN ('compressed', 'snark_groth16')
+   OR (api_proof_type IS NULL AND proof_type IS NOT NULL);
+
+-- Backfill request_payload JSON so legacy rows deserialize with zk_backend.
+UPDATE proof_requests
+SET request_payload = CASE
+    WHEN request_payload -> 'request' ->> 'proof_type' = 'snark_groth16' THEN
+        jsonb_set(
+            request_payload,
+            '{request,payload,proof,zk_backend}',
+            '"cluster"'::jsonb,
+            true
+        )
+    WHEN request_payload -> 'request' ->> 'proof_type' = 'compressed' THEN
+        jsonb_set(
+            request_payload,
+            '{request,payload,zk_backend}',
+            '"cluster"'::jsonb,
+            true
+        )
+    ELSE request_payload
+END
+WHERE request_payload IS NOT NULL
+  AND (
+      (request_payload -> 'request' ->> 'proof_type' = 'compressed'
+          AND request_payload -> 'request' -> 'payload' ->> 'zk_backend' IS NULL)
+      OR (request_payload -> 'request' ->> 'proof_type' = 'snark_groth16'
+          AND request_payload -> 'request' -> 'payload' -> 'proof' ->> 'zk_backend' IS NULL)
+  );
+
+ALTER TABLE proof_requests ENABLE TRIGGER update_proof_requests_updated_at;
+
+-- Keep NULL equivalent to the compatibility default while legacy service
+-- binaries may still insert rows during a rolling deployment.
+CREATE INDEX IF NOT EXISTS idx_proof_requests_zk_job_claim
+ON proof_requests(
+    job_status,
+    api_proof_type,
+    zk_vm,
+    (COALESCE(zk_backend, 'cluster')),
+    start_block_number,
+    created_at
+)
+WHERE api_proof_type IN ('compressed', 'snark_groth16');
+
+COMMENT ON COLUMN proof_requests.zk_backend IS 'Protocol ZK proving backend: mock, dry_run, cluster, network.';
+
+COMMIT;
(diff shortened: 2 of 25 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/prover-service/db/src/repo.rs
```diff
@@ -574,17 +579,38 @@ impl ProofRequestRepo {
...
+                sqlx::query(&sql)
+                    .bind(&req.worker_id)
+                    .bind(lock_id)
+                    .bind(i64::from(req.lock_duration_seconds))
+                    .bind(req.api_proof_type.as_str())
+                    .bind(&zk_vms)
+                    .bind(&zk_backends)
+                    .bind(i64::from(req.max_attempts))
+                    .fetch_optional(&self.pool)
+                    .await?
```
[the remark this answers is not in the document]
mw2000: Fixed in 408f51774. claim_next_proof_job now applies the Cluster compatibility default when a direct DB caller supplies an empty backend list, and the existing claim integration test exercises that path.

crates/proof/zk/host/src/host.rs
```diff
@@ -112,15 +113,19 @@ impl ZkHostConfig {
...
+        )
+        .with_poll_interval(self.job_discovery_poll_interval)
+        .with_lock_duration_seconds(self.job_discovery_lock_duration_seconds)
+        .with_max_concurrent_jobs(self.job_discovery_max_concurrent_jobs)
     }
 
     /// Converts this config into the shared worker discovery config.
     pub fn into_job_discovery_config(self) -> JobDiscoveryConfig {
-        JobDiscoveryConfig::zk(self.worker_id, self.zk_vms)
+        JobDiscoveryConfig::zk(self.worker_id, self.zk_vms, vec![ZkBackend::Cluster])
```
jackchuma: Advertise the selected ZK backend. When `bin/prover/zk-host` is started with `--backend dry-run`, `mock`, or `network`, this hardcoded `Cluster` capability makes that worker claim only cluster jobs and never claim requests routed to its actual backend; worse, a dry-run/mock worker can now claim cluster requests and submit the wrong kind of result. The discovery config needs to carry the backend selected by the host/CLI instead of always advertising `Cluster`

crates/proof/prover-service/db/src/repo.rs
```diff
@@ -2321,6 +2362,12 @@ impl CreateRequestParams<'_> {
...
             return Some("tee_kind");
         }
+        let stored_zk_backend = row.get::<Option<&str>, _>("zk_backend").or_else(|| {
+            matches!(self.api_proof_type, "compressed" | "snark_groth16").then_some("cluster")
+        });
+        if stored_zk_backend != self.zk_backend {
+            return Some("zk_backend");
+        }
         if let Some(request_payload) = row.get::<Option<serde_json::Value>, _>("request_payload")
             && !request_payload_matches(&request_payload, self.request_payload, mode)
```
jackchuma, line 2372: Normalize legacy payloads before idempotency checks. During the rolling-deploy case described by the migration, an old writer can insert a ZK row after migration with `zk_backend = NULL` and a `request_payload` that omits `zk_backend`; this fallback treats the column as `cluster`, but the next raw JSON comparison still compares that old payload to a new serialized payload containing `"zk_backend":"cluster"`, causing an identical retry to fail with an id collision on `request_payload`. The compatibility default needs to be applied to the stored/incoming payload before this comparison as well

crates/proof/prover-service/db/migrations/014_add_zk_backend.sql
```diff
@@ -0,0 +1,62 @@
...
+        )
+    ELSE request_payload
+END
+WHERE request_payload IS NOT NULL
+  AND (
+      (request_payload -> 'request' ->> 'proof_type' = 'compressed'
+          AND request_payload -> 'request' -> 'payload' ->> 'zk_backend' IS NULL)
+      OR (request_payload -> 'request' ->> 'proof_type' = 'snark_groth16'
+          AND request_payload -> 'request' -> 'payload' -> 'proof' ->> 'zk_backend' IS NULL)
+  );
```
jackchuma: This JSON backfill looks like extra migration surface without buying much. `ZkProofRequest.zk_backend` already has `#[serde(default)]`, so old `request_payload` rows deserialize as `cluster` without rewriting stored JSON. I’d keep the column backfill/index, but delete this `request_payload` update block

crates/proof/prover-service/src/server/worker_api.rs
```diff
@@ -77,11 +77,17 @@ impl ProverServiceServer {
...
+            vec![ZkBackend::Cluster]
+        } else {
+            request.zk_backends
+        };
         let claim = ClaimProofJob {
             worker_id: request.worker_id,
             api_proof_type: request.proof_type.into(),
             tee_kinds: request.tee_kinds.into_iter().map(Into::into).collect(),
             zk_vms: request.zk_vms.into_iter().map(Into::into).collect(),
+            zk_backends,
```
jackchuma: The empty `zk_backends => cluster` compatibility default is implemented again in `ProofRequestRepo::claim_next_proof_job`. Keeping it in the repo is enough and covers direct repo callers too. I’d pass `request.zk_backends` through unchanged here and drop the `ZkBackend` import

crates/proof/worker/src/job_discovery.rs
```diff
@@ -64,8 +68,8 @@ impl JobClaimFilter {
     }
 
     /// Creates a ZK claim filter.
-    pub fn zk(zk_vms: impl Into<Vec<ZkVm>>) -> Self {
-        Self::Zk { zk_vms: zk_vms.into() }
+    pub fn zk(zk_vms: impl Into<Vec<ZkVm>>, zk_backends: impl Into<Vec<ZkBackend>>) -> Self {
+        Self::Zk { zk_vms: zk_vms.into(), zk_backends: zk_backends.into() }
```
jackchuma: `JobClaimFilter::Zk` takes a list of backends, but a worker host is configured with one prover backend. This makes every caller/thread of config carry a vector for flexibility we don’t use. I’d store a single `ZkBackend` in the worker config and only wrap it in `vec![backend]` when building the RPC request

crates/proof/prover-service/protocol/src/types.rs
```diff
@@ -49,6 +49,55 @@ pub enum ZkVm {
     Sp1,
 }
 
+/// ZK proving backend that executes a proof request.
+#[derive(
+    Debug, Clone, Copy, Default, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize,
+)]
```
jackchuma: `PartialOrd`, `Ord`, and `Hash` on `ZkBackend` look unused. Unless something actually sorts or hashes backends, these derives are just extra API promise. I’d keep only the derives callers need today

crates/proof/challenge/src/driver.rs
```diff
@@ -662,6 +664,7 @@ impl<L2: L2Provider, P: ProofRequesterProvider, T: TxManager, C: Clock> Driver<L
                 l1_head: Some(candidate.l1_head),
                 intermediate_root_interval: Some(candidate.intermediate_block_interval),
                 zk_vm: ZkVm::Sp1,
+                zk_backend: ZkBackend::Cluster,
```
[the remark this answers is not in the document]
mw2000: This should be okay for now. We don't need to make this configurable. We will do so in a future PR if we think that is necessary

crates/proof/prover-service/protocol/src/types.rs
```diff
@@ -49,6 +49,53 @@ pub enum ZkVm {
...
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        match value {
+            "mock" => Ok(Self::Mock),
+            "dry_run" => Ok(Self::DryRun),
+            "cluster" => Ok(Self::Cluster),
+            "network" => Ok(Self::Network),
+            other => Err(format!("Unknown ZK backend: {other}")),
+        }
+    }
```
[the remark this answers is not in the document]
mw2000: Since it mirrors existing convention we will leave it for now

crates/proof/prover-service/db/src/repo.rs
mw2000: Fixed in bcdf29f8e. Stored and incoming ZK payloads now normalize a missing zk_backend to cluster before comparison, with coverage for compressed and Groth16 legacy payloads.

crates/proof/prover-service/db/src/repo.rs
```diff
@@ -2341,17 +2388,26 @@ fn request_payload_matches(
...
+            Some("snark_groth16") => Some("/request/payload/proof"),
+            _ => None,
+        };
+    if let Some(map) = zk_backend_path.and_then(|path| value.pointer_mut(path)?.as_object_mut()) {
+        map.entry("zk_backend").or_insert_with(|| serde_json::Value::String("cluster".to_owned()));
+    }
+    if mode == RequestMismatchMode::AllowL1HeadReplacement {
+        remove_l1_head_fields(&mut value);
+    }
     value
```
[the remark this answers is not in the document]
mw2000: Fixed in b92551e82. Identical payloads now return before normalization, avoiding both JSON clones on the common idempotent retry path while preserving legacy normalization for differing payloads

jackchuma (approved):

PR #3930, opened 2026-07-12
fix(engine): do not poison blocks when parent state is unavailable

### LINKED ISSUE #3701: Isthmus withdrawals-root validation races against canonical-state visibility, permanently poisoning valid blocks
## Summary

On a `base-reth-node`-equivalent execution client running pinned `v1.1.1`, we
repeatedly see valid blocks rejected as "invalid" with:

```
failed to load parent state for Isthmus withdrawals root validation: no state found for block 0x...
```

The rejected block's parent is one that was canonicalized only milliseconds earlier — the validator loses a race against the provider becoming able to serve that parent's state. Once a block is marked invalid this way, reth's in-memory invalid-block cache poisons it permanently for the life of the process: every subsequent payload that (transitively) builds on it is rejected immediately via `"Bad block with existing invalid ancestor"`, without re-validation. The node never recovers on its own — only a full process restart clears the cache and lets it resync.

This happens repeatedly (every ~13–14 minutes in our deployment) and is, as far as we can tell, **not** specific to our setup — it reproduces on the officially pinned `v1.1.1` commit and is still present, unchanged, on `main` as of this writing.

## Environment

- **base/base version:** `v1.1.1`, commit `01e732cdbae0c624d652da9e608d7d3fe0f9c74b` (same commit pinned in the official `base/node` reference repo's `versions.env`)
- **Network:** Base mainnet
- **Consensus client:** `base-consensus node` (NOT `follow` mode — `BASE_NODE_SOURCE_L2_RPC` is unset)
- **Execution client:** a custom binary (`reth-indexer`) that links `base-execution-evm`, `base-common-consensus`, `base-node-core`, etc. as libraries and uses the official `base_node_core::BaseNode` launcher via the standard `NodeBuilder` extension API (`.install_exex(...)`, `.extend_rpc_modules(...)`, `.on_node_started(...)`) — we do **not** reimplement any node/engine wiring. This is functionally equivalent to running the stock `base-reth-node` binary with an ExEx attached; the `BaseEngineValidator` construction path is identical.
- **reth version:** `2.3.0-dev` (`9384bc53d8c0c77e59cac83fdaaf3b372c6d2216`, `paradigmxyz/reth` tag `v2.3.0`)
- **Build profile:** Initially built with plain `cargo build --release`. We rebuilt with a `[profile.maxperf]` (`inherits = "release"`, `lto = "fat"`, `codegen-units = 1`) matching what `base/node`'s own Dockerfile uses for `base-reth-node`/`base-consensus`/`basectl`, on the theory that a less-optimized build might be losing a narrow timing race more often. **This did not change the failure cadence** (see "What we ruled out" below).
- **Flashblocks:** enabled, `--websocket-url=wss://mainnet.flashblocks.base.org/ws`. We have not yet conclusively confirmed or ruled out Flashblocks as a contributing factor (see "Open questions").
- **OS / infra:** Ubuntu 24.04 Docker host, `execution` and `node` (consensus) as separate `docker-compose` services, communicating over the engine API (`ws://execution:8551`). `consensus-entrypoint` is byte-identical to the one in `base/node`.

## Root cause

`crates/execution/node/src/engine.rs`, `BaseEngineValidator`'s
`PayloadValidator::validate_block_post_execution_with_hashed_state`:

```rust
fn validate_block_post_execution_with_hashed_state(
    &self,
    state_updates: &HashedPostState,
    block: &RecoveredBlock<Self::Block>,
) -> Result<(), ConsensusError> {
    if !self.chain_spec().is_isthmus_active_at_timestamp(block.timestamp()) {
        return Ok(());
    }

    let parent_state =
        self.provider.state_by_block_hash(block.parent_hash()).map_err(|err| {
            ConsensusError::Other(Arc::from(Box::<dyn core::error::Error + Send + Sync>::from(
                format!(
                    "failed to load parent state for Isthmus withdrawals root validation: {err}"
                ),
            )))
        })?;

    self.validate_block_post_execution_with_state(state_updates, parent_state, block.header())
}
```

This calls `self.provider.state_by_block_hash(parent_hash)` directly on the plain `StateProviderFactory` — it does **not** consult any in-memory overlay for a parent that was just canonicalized but hasn't fully propagated to wherever this lookup resolves state from. Contrast with the doc comment on the sibling method a few lines above it:

```rust
/// Verifies hardfork-gated post-execution rules against the supplied parent state.
/// ...
/// Callers supply `parent_state` explicitly so engine pipelines can pass
/// in-memory-aware overlay providers when the parent block isn't canonical yet.
pub fn validate_block_post_execution_with_state<DB, H>(...)
```

So the overlay-aware design *exists* and is clearly intentional for that call path — it's just not wired into the `PayloadValidator` trait hook above, which is the one actually invoked during normal new-payload handling.

We believe this is a regression from #2573 (`fix(engine): validate Isthmus withdrawals with overlay state`, merged
2026-05-11), which made this check fail-closed (error out the block) without giving *this specific* call site the overlay-aware state it needs to safely be strict.

We also looked at #3480 / #3486 (`fix engine-tree proof worker overlay for in-memory parents` / `fix(engine-tree): use in-memory overlay for proof workers`, merged 2026-06-11/12) — these fix the *same class* of race (state for a freshly-canonicalized-but-unpersisted parent not being visible) but for a different call path: the trie/proof-worker overlay in `crates/execution/engine-tree/src/validator.rs`. That fix does not touch `crates/execution/node/src/engine.rs`, so this Isthmus-withdrawals-root call site is still exposed. We confirmed this is unchanged on current `main` (`e635b2dea...` at time of writing).

## Evidence

Caught the exact race in our execution-client logs:

```
15:31:57.546908Z  INFO  Block added to canonical chain  number=47676478 hash=0x30286a6b3159cb7e81673d2469d1ec7848b565946e8bf54366e83e9dd4bbb115
15:31:57.613794Z  WARN  Encountered invalid block        number=47676479  hash=0xbfb1068ffa13206b1d8a6d40864259f79c4828359e3cd07ef9e499f30e24977c
                         error: failed to load parent state for Isthmus withdrawals root validation: no state found for block 0x30286a6b3159cb7e81673d2469d1ec7848b565946e8bf54366e83e9dd4bbb115
```

Block `47676479`'s validation asked for its parent's state **67ms** after that parent (`47676478`) was logged as added to the canonical chain, and lost the race.

Once that happens, every subsequent payload that builds on the poisoned hash
is rejected immediately and indefinitely (until process restart):

```
15:16:59.852042Z  WARN  Bad block with existing invalid ancestor
                         hash=0x731623ce6ddfa19e1e0ce25433737a33cc428991a193d605e720da2ce82fa661
                         invalid_ancestor=BlockWithParent { parent: 0x024a8bd1...4cebc, block: NumHash { number: 47675998, hash: 0x3cdd7191...eaab29e } }
```

This repeats every time the consensus client retries delivering the next
payload (sub-second intervals in our logs), spamming the log and leaving the
chain stuck until we restart the `execution` process.

In our deployment, restarts happen automatically: we run an ExEx alongside
the execution client, and a liveness watchdog restarts the process if the
ExEx goes ~600s without a new canonical-block notification (which it will,
once a block gets permanently poisoned and the chain stops advancing). So we
see this recur repeatedly over time, but the **interval between occurrences
in our deployment reflects that watchdog/restart cycle, not the natural
frequency of the underlying race** — we can't use our own uptime data to
infer how often this would happen on a long-running node without such a
watchdog.

## What we ruled out

- **Not a config/flag divergence:** our `consensus-entrypoint` is
  byte-identical to `base/node`'s, our pinned commit matches `base/node`'s
  `versions.env` exactly, and the execution-side flags we pass differ from
  `base/node`'s `execution-entrypoint` only in cosmetic ways (log verbosity,
  RPC namespace selection, discovery v5 port) — nothing engine/consensus
  related.
- **Not our own node-construction code:** our binary uses the official
  `BaseNode` launcher via the supported ExEx extension API, not a
  reimplementation.
- **Not (solely) a build-optimization/CPU-scheduling timing issue:** we
  rebuilt with `[profile.maxperf]` (`lto = "fat"`, `codegen-units = 1` —
  matching `base/node`'s own Dockerfile) instead of plain `--release`, and
  the issue still recurred.
- **Not follow mode:** we run `base-consensus node`, not `follow`
  (`BASE_NODE_SOURCE_L2_RPC` unset), so this is unrelated to the
  follow-mode Isthmus issues we found in #3197.

## Open questions (for maintainers' awareness, not confirmed root causes)

- We have Flashblocks enabled (`--websocket-url=...`); its
  `crates/execution/flashblocks/src/processor.rs` handles a
  `"processing canonical block"` event on every new canonical block and
  itself reads from the same provider (`state_by_block_number_or_tag`,
  `header_by_number`). We have not confirmed it causes or worsens this race
  (no exclusive lock found that would directly explain a delay), but it is
  a concurrent consumer of the same provider at the same trigger point, so
  we flag it as a plausible aggravating factor worth checking from your
  side.

## Current mitigation

We're aware of and have adopted
[dalechyn/base@e272257](https://github.com/dalechyn/base/commit/e272257a5b0655c9f65964c1c9e905f83c0baef9)
as a stopgap, which we've ported onto our own fork on top of `v1.1.1` at
[spilnotaxyz/base@fbb4c06](https://github.com/spilnotaxyz/base/commit/fbb4c0616158a6638702a0afef64312a0ddce7ab):

```rust
// NOTE: pre-#2573 lenient behavior — skip the Isthmus withdrawals-root
// check when the parent state can't be loaded (node still syncing, or parent only
// in the in-memory tree), instead of failing closed. Failing closed marks valid
// canonical blocks "bad" and wedges the node permanently. See base #2573.
let Ok(parent_state) = self.provider.state_by_block_hash(block.parent_hash()) else {
    return Ok(());
};
```

This trades away the Isthmus withdrawals-root check whenever parent state
happens to be unavailable (sync, or this race) in exchange for never wedging
the node. We understand this weakens defense-in-depth (a buggy/malicious L2
withdrawals root could slip through unnoticed if it happens to arrive during
an unavailability window) and would much rather run the strict, fail-closed
check with correct overlay-aware state — i.e. the same fix shape as
#3480/#3486, applied to this call site.

## Ask

- Could this call site (`validate_block_post_execution_with_hashed_state` in
  `crates/execution/node/src/engine.rs`) be wired to use overlay-aware state
  for the parent, the way `validate_block_post_execution_with_state`'s own
  doc comment says callers should, and the way #3480/#3486 did for the
  trie/proof-worker path?
- Any guidance on whether the ~13–14 minute periodicity rings a bell on your
  side (a known periodic internal operation), or whether Flashblocks is
  expected to interact with this code path, would help us narrow this down
  further from our end too.

Happy to share full raw logs (Loki export) around any of the timestamps
above if useful.

danilaverbena:
## Summary

Fixes #3701.

The legacy `PayloadValidator::validate_block_post_execution_with_hashed_state` path (used by reth `BasicEngineValidator`) loaded parent state via `state_by_block_hash` and returned `ConsensusError` on failure. That permanently marks the child invalid in reth's invalid-block cache, so every subsequent payload that builds on it is rejected until process restart.

Parent state can be transiently invisible right after canonicalization even for valid blocks. This change skips Isthmus withdrawals-root validation **only** when the parent cannot be loaded; when state is available, verification still fails closed on a bad withdrawals root.

The engine-tree path continues to use overlay-aware parent state via `BasePostExecutionValidator`.

## Changes
- `crates/execution/node/src/engine.rs`: fail-open on missing parent state in the legacy hook
- `crates/execution/engine-tree/tests/isthmus.rs`: regression test updated for #3701

## Test plan
- [ ] `cargo test -p base-engine-tree --test isthmus`
- [ ] `cargo check -p base-node-core`

Closes #3701

---
Supersedes #3929 (auto-closed while the fork was behind; this branch is rebased on current `main`).

### DIFF
--- a/crates/execution/engine-tree/tests/isthmus.rs
+++ b/crates/execution/engine-tree/tests/isthmus.rs
@@ -26,7 +26,9 @@ use reth_trie::{
 use revm::database::BundleState;
 
 #[test]
-fn legacy_isthmus_validation_errors_when_parent_is_only_in_memory() {
+fn legacy_isthmus_validation_skips_when_parent_is_only_in_memory() {
+    // Regression for #3701: the legacy PayloadValidator hook must not fail-closed (and
+    // permanently poison the block) when parent state is not yet visible via StateProviderFactory.
     let (chain_spec, provider, _genesis_hash, parent_executed_block, _parent_withdrawals_root) =
         isthmus_parent_overlay_fixture();
     let parent_hash = parent_executed_block.recovered_block().hash();
@@ -36,6 +38,11 @@ fn legacy_isthmus_validation_errors_when_parent_is_only_in_memory() {
         KeccakKeyHasher,
     >(Arc::clone(&chain_spec), provider.clone());
 
+    assert!(
+        provider.state_by_block_hash(parent_hash).is_err(),
+        "test setup requires the parent to exist only in memory"
+    );
+
     let legacy_result =
         PayloadValidator::<BaseEngineTypes>::validate_block_post_execution_with_hashed_state(
             &validator,
@@ -43,12 +50,8 @@ fn legacy_isthmus_validation_errors_when_parent_is_only_in_memory() {
             &child_block,
         );
     assert!(
-        legacy_result.is_err(),
-        "legacy validation should fail closed when the parent state is not canonical"
-    );
-    assert!(
-        provider.state_by_block_hash(parent_hash).is_err(),
-        "test setup requires the parent to exist only in memory"
+        legacy_result.is_ok(),
+        "legacy validation must skip (not poison) when parent state is unavailable: {legacy_result:?}"
     );
 }
 
--- a/crates/execution/node/src/engine.rs
+++ b/crates/execution/node/src/engine.rs
@@ -215,14 +215,28 @@ where
             return Ok(());
         }
 
-        let parent_state =
-            self.provider.state_by_block_hash(block.parent_hash()).map_err(|err| {
-                ConsensusError::Other(Arc::from(Box::<dyn core::error::Error + Send + Sync>::from(
-                    format!(
-                        "failed to load parent state for Isthmus withdrawals root validation: {err}"
-                    ),
-                )))
-            })?;
+        // This trait hook is used by reth's BasicEngineValidator (no tree overlay). Prefer the
+        // overlay-aware path via [`BasePostExecutionValidator`] / base-engine-tree when available.
+        //
+        // #3701: after a parent is just canonicalized, `state_by_block_hash` can transiently fail
+        // even though the parent is valid. Returning `ConsensusError` here marks the child
+        // permanently invalid in reth's invalid-block cache, wedges the node until restart, and
+        // rejects every subsequent payload that builds on the poisoned ancestor. Skip the check
+        // only when parent state cannot be loaded; when state *is* available, still fail closed
+        // on a bad withdrawals root via `validate_block_post_execution_with_state`.
+        let parent_state = match self.provider.state_by_block_hash(block.parent_hash()) {
+            Ok(state) => state,
+            Err(err) => {
+                tracing::warn!(
+                    target: "base::engine",
+                    parent_hash = %block.parent_hash(),
+                    block_number = block.number(),
+                    error = %err,
+                    "parent state unavailable for Isthmus withdrawals-root validation;                      skipping check to avoid permanent invalid-block poison (see #3701)"
+                );
+                return Ok(());
+            }
+        };
 
         self.validate_block_post_execution_with_state(state_updates, parent_state, block.header())
     }


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

danilaverbena:
Auto-closed as unvouched (expected per [CONTRIBUTING](https://github.com/base/base/blob/main/CONTRIBUTING.md)).

Requested **vouch** + assignment on https://github.com/base/base/issues/3701. Branch is rebased on current `main` and ready once vouched.

PR #3941, opened 2026-07-13
feat(precompiles): add foo versioned reference precompile (copy-fork, minimal)

stephancill:
type=routine
risk=low
impact=sev5

## Summary

Copy-fork counterpart to #3933, using the **most minimal abstraction that still keeps each version self-contained**. Same `foo` precompile, same append-only ABI/storage/dispatch, but:

- Each version is a **self-contained** type (`FooV1`, `FooV2`) — no `previous` composition.
- The **`FooVersion` enum and the `FooVersion::logic()` gate are gone**. A single seam maps fork → implementation:

```rust
impl FooVersions {
    pub fn resolve(upgrade: BaseUpgrade) -> Option<&'static dyn FooLogic> {
        if upgrade >= BaseUpgrade::Cobalt { Some(&FooV2) }
        else if upgrade >= BaseUpgrade::Beryl { Some(&FooV1) }
        else { None }
    }
}
```

## Abstractions kept vs cut

- **Kept:** the `FooLogic` trait (the one interface that lets the dispatcher stay version-agnostic and keeps each version a whole unit) and self-contained per-version structs.
- **Cut:** the `FooVersion` enum and the separate `logic()` method — `resolve` returns the `&'static dyn FooLogic` directly. Versions are zero-sized, so that reference is a static vtable pointer, not an allocation.
- `FooLogic` gains a `Sync` supertrait so the resolved `&'static dyn FooLogic` can be captured by the (thread-safe) precompile closure; the ZST versions satisfy it trivially.

## Trade-off vs #3933

- Versions are fully independent (no runtime link between them); each file fully describes its version.
- One consequence of dropping the enum: version *identity* is no longer a comparable value, so tests assert behavior (`resolve(Cobalt).hello_world() == "…"`) rather than `== Some(V2)`.

## Activation

Installed from Beryl (V1) and Cobalt (V2), self-gated in `Foo::install` via `FooVersions::resolve`. No-op before Beryl.

## Test plan

- \`cargo test -p base-common-precompiles foo\` — 9 passed
- \`cargo clippy -p base-common-precompiles --all-targets\` — clean

PR #3942, opened 2026-07-13
feat(precompiles): add foo V3 — demonstrate adding a new version (copy-fork, minimal)

stephancill:
type=routine
risk=low
impact=sev5

## Summary

Stacked on #3941. Copy-fork counterpart to #3939, in the minimal-abstraction style: adds a **new self-contained version** with no `previous` link, no `FooVersion` enum, and no per-version dispatch match.

## The entire diff to add a version

- **`logic/v3.rs`** (new) — `FooV3` (plain unit struct implementing `FooLogic`).
- **`versions.rs`** — **one `resolve()` branch** (`if upgrade >= Zombie { Some(&FooV3) }`) plus the import. That's the whole routing change.
- **`logic/mod.rs` / `foo/mod.rs` / `lib.rs`** — re-exports.

**`logic/v1.rs` and `logic/v2.rs` have zero changes** — activated versions stay frozen.

Because the enum and `logic()` gate were removed in #3941, adding a version here does **not** touch a version enum, a `logic()` static table, or any per-method match. Contrast #3939 (composition), where the equivalent change also adds an enum variant and a `logic()` arm.

## Copy-fork vs composition (#3939)

- `hello_world` is unchanged since V2. Composition delegates (`self.previous.hello_world()`); here V3 **copies the literal**, so it has no runtime dependency on V2.
- Goal 1 (change existing logic): `greet` returns `"Hey <name>, welcome to Base!"` — same in both.

## Activation

V3 is staged behind the permanently-off **`Zombie`** gate: testable (`resolve(Zombie)` returns V3's impl), never selected on a live chain until repointed to a real fork.

## Test plan

- \`cargo test -p base-common-precompiles foo\` — 12 passed
- \`cargo clippy -p base-common-precompiles --all-targets\` — clean

### DIFF
--- a/crates/common/precompiles/src/foo/dispatch.rs
+++ b/crates/common/precompiles/src/foo/dispatch.rs
@@ -65,7 +65,7 @@ mod tests {
     use alloy_sol_types::SolCall;
     use base_precompile_storage::{HashMapStorageProvider, StorageCtx};
 
-    use crate::{FooLogic, FooStorage, FooV1, FooV2, IFoo};
+    use crate::{FooLogic, FooStorage, FooV1, FooV2, FooV3, IFoo};
 
     const CALLER: Address = address!("0x1111111111111111111111111111111111111111");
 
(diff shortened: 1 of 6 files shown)


stephancill:
Superseded by the self-contained-versions variant #3946. Closing to keep one live copy-fork proposal.

PR #3933, opened 2026-07-13
feat(precompiles): add foo versioned reference precompile

rayyan224:
## Summary

Adds `foo`, a self-contained **reference precompile** that demonstrates the hardfork-versioned execution-consensus design (per the *Execution Consensus* TDD) against the real Base precompile machinery. It exists to give the team a concrete template for versioning precompile logic by hardfork.

Business logic is split into immutable, fork-activated versions selected by a central version manager and routed by an append-only dispatcher.

## Design

- **`versions.rs`** — `FooVersions::resolve(upgrade)` maps the active hardfork to a version: `None` before Beryl, `V1` from Beryl, `V2` from Cobalt. Resolved **once at install time** (matching how all existing fork gating works in `provider.rs`), then threaded into dispatch via the existing `#[precompile(args(...))]` macro. Routing is a single `match` — **no dynamic dispatch**, satisfying the "minimal execution-path overhead" constraint.
- **`logic/{v1,v2}.rs`** — frozen per-version implementations behind the `FooLogic` trait. `FooV2 { previous: FooV1 }` uses composition (Rust's stand-in for inheritance).
- **`storage.rs`** — append-only state (`FooStorage`) with a shared `record_greeting` primitive reused across versions.

## What V1 → V2 demonstrates

- **Goal 1 (change existing logic):** `helloWorld` returns `"Hello, World!"` in V1, `"Hello, World! Welcome to Base."` in V2.
- **Goal 3 (append-only new method):** `greet(name)` is new in V2; before its activation fork the trait's default impl reverts `UnsupportedBeforeActivation`, preserving pre-activation behavior for the frozen version.

Installed for Beryl+ via `BasePrecompiles::install` at `0xf00f…0000`.

## Module layout

```
foo/
  mod.rs  abi.rs  storage.rs  versions.rs  dispatch.rs  precompile.rs
  logic/  mod.rs  v1.rs  v2.rs
```

## Test plan

- `cargo build -p base-common-precompiles` — clean
- `cargo test -p base-common-precompiles foo` — 8 passed (version resolution, cross-version `helloWorld`, `greet` unsupported pre-V2, `greet` from V2, unknown selector, install)
- `cargo clippy -p base-common-precompiles --all-targets` — clean

Generated with Claude Code

PR #3946, opened 2026-07-13
feat(precompiles): add foo versioned reference precompile (self-contained versions)

stephancill:
type=routine
risk=low
impact=sev5

## Summary

A third structural option for the execution-consensus versioning design, alongside composition (#3933) and minimal copy (#3941). This is the **pragmatic split** for a B20-scale precompile: keep the genuine cross-version invariants **shared**, make each version **self-contained** for the parts that legitimately vary.

`foo` stands in for a B20-scale token precompile; the structure is the point.

## The split

- **Shared** (cross-version invariants — copying these is a hazard):
  - `abi.rs` — the append-only `IFoo` interface. A selector must always decode the same way across versions.
  - `storage.rs` — `FooStorage`. Every version reads/writes the **same** state slots; the layout cannot actually be made per-version (state is shared on-chain), so one definition is the safety mechanism.
  - `dispatch.rs` — a tiny shared entry that charges calldata gas and hands off. No selector routing.
- **Self-contained per version** (what actually varies):
  - Each version implements the whole `FooVersion::call(storage, ctx, calldata)` seam — it **decodes and routes its own selectors** and holds its own logic (`logic/v1.rs`, `logic/v2.rs`).

The entire shared interface between the entry and a version is one method:

```rust
pub trait FooVersion: Sync {
    fn call(&self, storage: &mut FooStorage<'_>, ctx: StorageCtx<'_>, calldata: &[u8]) -> Result<Bytes>;
}

pub fn resolve(upgrade: BaseUpgrade) -> Option<&'static dyn FooVersion> {
    if upgrade >= Cobalt { Some(&FooV2) } else if upgrade >= Beryl { Some(&FooV1) } else { None }
}
```

## How this differs from #3941 (minimal copy)

- No shared `FooLogic` per-method trait and no shared selector-routing. A version owns its selector table: V1 routes `helloWorld` and lets everything else fall through to `UnsupportedBeforeActivation`; V2 routes both.
- "Which selectors exist and what they do" is fully readable in one file per version.

## Honest tradeoffs

- **Win:** each version is a self-contained routing+logic unit; changing existing logic or adding a new version touches nothing shared. Storage/ABI invariants stay single-source.
- **Cost:** per-version selector routing is duplicated (each version re-runs `decode_precompile_call!`). And because the repo forbids `#[allow(unreachable_patterns)]`, a version that handles every current selector (V2) uses an exhaustive match with no catch-all — so adding a *brand-new selector* to the ABI later is compiler-enforced to touch each existing version (add an `unsupported` arm). That is a deliberate safety prompt, but it is not "zero edits to frozen files" for the new-selector case. Changing logic / adding versions remains edit-free on frozen code.
- Same foot-gun as every option: a version that omits an introduced selector reverts unsupported — covered by per-version golden tests.

## Test plan

- \`cargo test -p base-common-precompiles foo\` — 9 passed
- \`cargo clippy -p base-common-precompiles --all-targets\` — clean

### DIFF
--- a/crates/common/precompiles/src/foo/abi.rs
+++ b/crates/common/precompiles/src/foo/abi.rs
@@ -1,0 +1,34 @@
+//! ABI definitions for the `foo` reference precompile.
+
+use alloy_sol_types::sol;
+
+sol! {
+    /// `foo` reference precompile ABI.
+    ///
+    /// The interface is append-only across versions: methods and errors are
+    /// added, never removed or repurposed, so historical calldata keeps
+    /// decoding the same way.
+    interface IFoo {
+        /// Precompile cannot be executed via delegatecall or callcode.
+        error DelegateCallNotAllowed();
+
+        /// The called method is not supported by the version active at this block.
+        ///
+        /// Returned for methods that had not yet been introduced at the target
+        /// hardfork, preserving their original pre-activation revert behavior.
+        error UnsupportedBeforeActivation();
+
+        /// Emitted when `greet` records a greeting for `caller`.
+        event Greeted(address indexed caller, string greeting);
+
+        /// Returns a fixed greeting. The value is version-specific and frozen
+        /// once its version is activated.
+        function helloWorld() external view returns (string);
+
+        /// Records and returns a personalized greeting for the caller.
+        ///
+        /// Introduced in a later version; reverts with
+        /// `UnsupportedBeforeActivation` before its activation fork.
+        function greet(string name) external returns (string);
+    }
+}
--- a/crates/common/precompiles/src/foo/dispatch.rs
+++ b/crates/common/precompiles/src/foo/dispatch.rs
@@ -1,0 +1,106 @@
+//! Shared precompile entry for the `foo` reference precompile.
+//!
+//! This is the *only* shared execution glue: it charges the flat calldata cost
+//! and hands off to the active version's own [`FooVersion::call`]. Selector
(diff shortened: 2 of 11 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #3949, opened 2026-07-13
feat(precompiles): b20_asset hardfork versioning — fork-routed lookup + V1 seam (Phase 0+1)

stephancill:
## Summary

Applies the P/PS "Option 3 / hybrid" versioning pattern to the B20 asset
precompile: shared ABI + storage + dispatch **entry** (append-only), with
per-version **self-contained** logic + selector routing, selected by a single
fork→version resolver. Storage and the ABI are **not** duplicated per version.

This is delivered as **Phase 0 + Phase 1** (two commits), wrapping today's
behavior as **V1 with zero behavior change**. The existing `b20` test suite
passes **unedited** — that is the proof of "no behavior change."

## Phase 0 — thread the current hardfork (`BaseUpgrade`) into the B20 lookup

B20 tokens are resolved per token-address by `BerylLookup`, a path that
previously carried **no fork signal**. Phase 0 threads it (pure plumbing,
byte-identical behavior):

- `lookup.rs`: `BerylLookupWithObserver` now stores `upgrade`;
  `install`/`install_with_observer`/`lookup`/`lookup_with_observer` accept a
  `BaseUpgrade` and pass it through to the `create_precompile*` calls.
- `b20_asset` / `b20_stablecoin` `create_precompile*` signatures accept
  `upgrade`. In Phase 0 both **accept and ignore** it (`_upgrade`).
- `provider.rs`: installs `BerylLookup` with `self.spec.upgrade()`. The
  precompiles map is already built per-spec (`BasePrecompiles::new_with_spec`
  matches on `spec.upgrade()`), so the active fork is known at install time.

## Phase 1 — version seam on `b20_asset`, wrap today's logic as V1 (code motion)

- **SHARED (unchanged):** `abi.rs`, `storage.rs`, `accounting.rs`, and the
  dispatch **entry** in `dispatch.rs` (non-payable check, calldata-gas
  deduction, init check, observer/recorder plumbing).
- **NEW `versions.rs`:** the `B20AssetVersion` seam trait + `B20AssetVersions::resolve(upgrade)`
  returning `Some(B20AssetVersionId::V1)` from `BaseUpgrade::Beryl` onward, else `None`.
- **NEW `logic/{mod,v1}.rs`:** the selector routing (`inner*`, `handle_b20_call`,
  `handle_asset_call`, `announce`) is **relocated verbatim** (code motion, no
  logic change). `B20AssetV1` (a ZST) implements the seam by delegating to the
  moved `inner_with_observer` routing.
- **`dispatch.rs`:** the shared entry now resolves to a version and delegates:
  after the init check it calls `version.run(self, ctx, calldata, observer)`.
  `dispatch_with_observer` is preserved as a V1 convenience for existing callers
  and tests; the new fork-routed entry is `dispatch_with_version`.
- **`precompile.rs`:** `create_precompile*` resolve via `B20AssetVersions::resolve(upgrade)`
  and return `Option<DynPrecompile>`; `None` (pre-Beryl) does not register.
  `lookup.rs` threads that `None` through.

Phase 1 is `b20_asset` **only**. The stablecoin variant received only the
Phase-0 signature change (accepts and ignores `upgrade`); its logic is not
restructured. Storage/ABI are not duplicated; B20 business logic is not rewritten.

## No behavior change — existing tests pass unedited

The `#[cfg(test)] mod tests` block in `b20_asset/dispatch.rs` is **byte-identical**
to `main`, and `token.rs`'s tests are untouched. No test bodies were edited.
Phase 1 added only new resolver unit tests in `versions.rs` (the `+3`).

## Design decision / compromise on the generic seam

The plan's ideal seam is `B20AssetVersions::resolve(upgrade) -> Option<&'static dyn B20AssetVersion>`.
That is **intractable here**, and this PR takes the documented compromise:

- B20 logic lives on the generic `B20AssetToken<S: AssetAccounting, P: Policy>`
  and dispatch threads a generic `O: PrecompileCallObserver`, so the seam's
  `run<S, P, O>(..)` method is **generic** → the trait is **object-unsafe**
  (no `&dyn`).
- Even a generic-**parameter** trait (`dyn B20AssetVersion<S, P, O>`) can't be
  `&'static`, because in the real path `S = B20AssetStorage<'ctx>` borrows the
  call's `StorageCtx` and is **never `'static`** (unlike `foo`, whose storage is
  concrete and `'static`, which is why `foo` can use `&'static dyn FooVersion: Sync`).

**Chosen structure:** keep the `B20AssetVersion` trait as the extensible per-version
seam (implemented by the `B20AssetV1` ZST), and have `B20AssetVersions::resolve`
return a small `'static`, `Copy` enum `B20AssetVersionId` that is dispatched
**statically** (`match self { V1 => B20AssetV1.run(..) }`). This compiles cleanly,
preserves behavior, centralizes fork→version resolution in one resolver, and keeps
V1's routing in its own file (`logic/v1.rs`). A side benefit: static dispatch means
the seam needs no `Sync` bound (foo required `Sync` only for the `&dyn` closure capture).

## Open risks for reviewers

- The relocated routing methods (`inner*`, `handle_*`, `announce`) remain
  **inherent methods on `B20AssetToken<S,P>`** (just moved into `logic/v1.rs`),
  because `announce` self-recurses via `inner_with_privilege` and existing tests
  call `token.inner(..)` directly. V1's trait impl is the thin seam adapter over
  them. This is deliberate: it keeps the move a true no-op and tests unedited.
- The shared token **business** methods (transfer/mint/`batch_mint`/multiplier/…)
  stay in `token.rs` (shared, generic), invoked by V1's routing. V1 owns "which
  selectors exist and what they do"; it does not fork the arithmetic. Moving those
  buys no isolation (storage is shared) and would only add churn.
- `create_precompile*` now returns `Option<DynPrecompile>`. In the installed
  provider the lookup only exists for `>= Beryl`, where `resolve` is always
  `Some`, so this is byte-identical in practice; the `None` arm is a defensive
  pre-Beryl guard (address simply not registered).

## Test plan

- `cargo test -p base-common-precompiles b20` → **97 passed, 0 failed** (94 pre-existing + 3 new resolver tests).
- `cargo test -p base-common-precompiles --lib` → **429 passed, 0 failed** (full crate).
- `cargo clippy -p base-common-precompiles --all-targets` → **clean** (no `#[allow(...)]` used).

### DIFF
--- a/crates/common/precompiles/src/b20_asset/dispatch.rs
+++ b/crates/common/precompiles/src/b20_asset/dispatch.rs
@@ -1,37 +1,59 @@
-//! ABI dispatch for the asset B-20 variant.
+//! Shared precompile entry for the asset B-20 variant.
 //!
-//! Asset-specific selectors are tried first via `IB20Asset::IB20AssetCalls`.
-//! The `IB20` match block handles inherited selectors.
+//! This is the version-agnostic execution glue: the non-payable check,
+//! calldata-gas deduction, initialization check, and observer/recorder
+//! plumbing. After the init check it resolves to the active
+//! [`B20AssetVersionId`] and hands off to that version's self-contained
+//! selector routing + business logic (see [`crate::b20_asset::logic`]).
 
-use alloc::string::ToString;
-
-use alloy_primitives::{Bytes, U256};
-use alloy_sol_types::{SolCall, SolEvent, SolInterface, SolValue};
+use alloy_primitives::Bytes;
 use base_precompile_storage::{BasePrecompileError, StorageCtx};
 use revm::precompile::PrecompileResult;
 
 use crate::{
-    AssetAccounting, B20AssetStorage, B20AssetToken, B20TokenRole, BerylAuxiliaryMetrics,
-    BerylCallRecorder, BerylMetricLabels, BerylSelector, Burnable, Configurable,
-    IB20::{self, IB20Calls as C},
-    IB20Asset::{self, IB20AssetCalls as SC},
-    Mintable, NoopPrecompileCallObserver, Pausable, PermitArgs, Permittable, Policy,
-    PrecompileCallObserver, RoleManaged, Token, Transferable,
-    macros::decode_precompile_call,
+    AssetAccounting, B20AssetToken, B20AssetVersionId, BerylCallRecorder, BerylMetricLabels,
+    IB20, NoopPrecompileCallObserver, Policy, PrecompileCallObserver, Token,
 };
 
 impl<S: AssetAccounting, P: Policy> B20AssetToken<S, P> {
-    /// ABI-dispatches `calldata` to the appropriate `IB20Asset` handler.
+    /// ABI-dispatches `calldata` using the current (V1) execution logic.
     pub fn dispatch(&mut self, ctx: StorageCtx<'_>, calldata: &[u8]) -> PrecompileResult {
         self.dispatch_with_observer(ctx, calldata, NoopPrecompileCallObserver)
     }
 
-    /// ABI-dispatches `calldata` and observes the decoded asset B-20 operation.
+    /// ABI-dispatches `calldata` and observes the decoded asset B-20 operation,
+    /// using the current (V1) execution logic.
+    ///
+    /// Convenience entry for callers and tests that exercise today's behavior
+    /// directly. The fork-routed entry point is [`dispatch_with_version`], which
+    /// the precompile install path uses to select the version active at the
+    /// current hardfork.
+    ///
+    /// [`dispatch_with_version`]: Self::dispatch_with_version
     pub fn dispatch_with_observer<O>(
         &mut self,
         ctx: StorageCtx<'_>,
         calldata: &[u8],
         observer: O,
+    ) -> PrecompileResult
+    where
+        O: PrecompileCallObserver,
+    {
+        self.dispatch_with_version(ctx, calldata, observer, B20AssetVersionId::V1)
+    }
+
+    /// Shared precompile entry: performs the non-payable check, calldata-gas
+    /// deduction, initialization check, and observer/recorder plumbing, then
+    /// delegates to `version`'s self-contained selector routing + business logic.
+    ///
+    /// Everything up to and including the init check is shared across versions;
+    /// only the final `version.run` hand-off is version-specific.
+    pub fn dispatch_with_version<O>(
+        &mut self,
+        ctx: StorageCtx<'_>,
+        calldata: &[u8],
+        observer: O,
+        version: B20AssetVersionId,
     ) -> PrecompileResult
     where
         O: PrecompileCallObserver,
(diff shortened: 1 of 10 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/b20_asset/logic/v1.rs
```diff
@@ -0,0 +1,464 @@
...
+        }
+
+        // Fall through to inherited IB20 selectors.
+        let call = decode_precompile_call!(calldata, IB20::IB20Calls);
+        let label = call.as_label();
+
+        observer.observe(label, || self.handle_b20_call(ctx, call, privileged))
+    }
+
+    fn handle_b20_call(
```
rayyan224, line 119: Why would the dispatching logic be here ?

crates/common/precompiles/src/b20_asset/versions.rs
```diff
@@ -0,0 +1,117 @@
...
+/// invariants); routing and behavior do not.
+pub trait B20AssetVersion {
+    /// Routes `calldata` to this version's selectors and executes the matching
+    /// business logic against `token`, returning the ABI-encoded output (or an
+    /// error/revert).
+    ///
+    /// Generic over storage (`S`), policy (`P`), and observer (`O`) because B20
+    /// logic lives on the generic [`B20AssetToken`] and dispatch threads a
+    /// generic observer.
+    fn run<S, P, O>(
```
rayyan224, line 39: I don't think it should have a run function here this should return an instance of implementation

crates/common/precompiles/src/b20_asset/dispatch.rs
```diff
@@ -1,38 +1,60 @@
...
+    /// ABI-dispatches `calldata` and observes the decoded asset B-20 operation,
+    /// using the current (V1) execution logic.
+    ///
+    /// Convenience entry for callers and tests that exercise today's behavior
+    /// directly. The fork-routed entry point is [`dispatch_with_version`], which
+    /// the precompile install path uses to select the version active at the
+    /// current hardfork.
+    ///
+    /// [`dispatch_with_version`]: Self::dispatch_with_version
     pub fn dispatch_with_observer<O>(
```
rayyan224, line 33: We shouldn't have this function it's a place where a bug can enter, you should only call with a version. I also don't think an external party should know about this internal versioning system. It should only pass in the fork. Versioning is an implementation detail.

PR #3960, opened 2026-07-14
refactor(precompiles): IB20Stablecoin → SOLID versioned structure (BOP-419)

rayyan224:
## Summary

Restructures the `b20_stablecoin` precompile into the hardfork-versioned **SOLID** shape (dispatcher / business-logic interface / ABI / version manager / per-version implementation), mirroring the `foo` reference precompile. Implements [BOP-419](https://linear.app/coinbase/issue/BOP-419/ib20stablecoin-to-solid-structure).

## Structure

| Component | Responsibility | File |
|---|---|---|
| **Version Manager** | `resolve(upgrade)` → active `StablecoinVersion`; `logic()` → frozen impl | `b20_stablecoin/versions.rs` |
| **Business-Logic Interface** | append-only trait the dispatcher consumes / each version implements | `b20_stablecoin/logic/interface.rs` |
| **Contract Implementation** | `StablecoinV1` — self-contained logic + the `B20StablecoinToken` holder | `b20_stablecoin/logic/v1.rs`, `logic/mod.rs` |
| **Precompile Dispatcher** | decode selector, gate by hardfork, route to the active version | `b20_stablecoin/dispatch.rs` |
| **ABI / Storage** | unchanged | `abi.rs`, `storage.rs` |

## Key points

- **Self-contained v1**: `StablecoinV1` reimplements every operation's orchestration directly against the storage/policy port, reusing only *stateless* shared primitives (`B20Guards`, `PermitArgs` EIP-712 crypto, `IB20` ABI/enum types). It calls **no** capability-trait methods — a change to the shared traits can never alter a frozen version (open/closed).
- **`token.rs` deleted**: `B20StablecoinToken` is reduced to a minimal `(accounting, policy)` holder implementing `Token`, relocated into `logic/mod.rs`.
- **Hardfork threading (runtime path only)**: the dispatcher receives a raw `BaseUpgrade` and resolves the version via the Version Manager. `BaseUpgrade` threads from `provider.rs` through the `BerylLookup` runtime path.
- **`B20Factory` untouched**: factory-initiated setup calls run at the token's introduction version (`StablecoinVersion::V1`) via `inner_with_privilege`, and the token retains a `grant_role_unchecked` bootstrap primitive for the initial-admin grant. (When a second version is introduced, thread the creation-fork version through the factory instead of pinning.)
- **Behavior-preserving**: v1 is a faithful copy of the shared ops, so the existing suite (429 tests) passes unchanged; new `versions.rs` resolve tests added.

## Verification

[...]

PR #3971, opened 2026-07-15
test(precompiles): add golden tests for stablecoin V1 (BOP-425)

stephancill:
## Summary

Adds a golden/snapshot test suite pinning current **Stablecoin V1** behavior of the B-20 precompile (BOP-425), before/while it is frozen behind the versioning seam from #3960 (BOP-419).

Every op is driven through the **version-resolver-gated** path (`inner(..., BaseUpgrade::Beryl)` → `StablecoinVersions::from_base_upgrade` → V1) against the real EVM-backed `B20StablecoinStorage` over `HashMapStorageProvider`, with `InMemoryPolicy` for deterministic allow/block. Each mutation case asserts:

1. exact returned ABI bytes (or the typed revert),
2. resulting state (balances / supply / roles / allowances / storage),
3. emitted event signature(s), and
4. a pinned keccak **storage hash** over the sorted storage triples (`hash_state`) — a reference baseline for the CI frozen-manifest check (BOP-421).

Plus a dedicated `golden_gas_footprints` test pinning each mutating op's **storage-access footprint** — the `(SLOAD, SSTORE, KECCAK256)` op counts, the deterministic gas-schedule-independent signal that drives real gas — so an extra storage access in V1 is caught even when bytes/state/events match.

**Op coverage:** every op the dispatcher routes — ERC-20 (transfer / transferFrom / approve + memo variants), mint / burn / burnBlocked, pause / unpause, updateSupplyCap|Name|Symbol|ContractURI, grant/revoke/renounce/renounceLastAdmin/setRoleAdmin, updatePolicy, permit; computed reads (isPaused, pausedFeatures, policyId, DOMAIN_SEPARATOR, eip712Domain, currency); all direct + role/policy-constant reads. Privileged vs unprivileged via `inner_with_privilege` vs `inner`; the guard envelope (nonpayable / uninitialized / pre-Beryl) via `dispatch_with_observer`; plus the factory `grant_role_unchecked` bootstrap.

**Edge cases:** zero-address (sender / receiver / approver), insufficient balance/allowance, supply-cap, paused, policy-blocked sender + executor policy, empty-feature-set, last-admin protection (revoke/renounce), expired permit, and every **unprivileged authorization revert** (caller lacking the required role) for the admin/config/role ops.

**Meta coverage guard:** a compile-time `v1_op_coverage_checklist` — an exhaustive `match` over the generated `IB20Calls` / `IB20StablecoinCalls` enums whose arms reference the covering golden `#[test]` fns. Adding an ABI op fails the build (non-exhaustive match) until a golden is added; renaming/removing a golden fails the build (missing fn reference). This intentionally couples to the shared ABI surface: when a future version extends the ABI, this suite is expected to be updated to assert the new op reverts for V1.

No source/logic changes — frozen V1 is untouched; only a new integration test file and one `[[test]]` stanza are added.

## Coverage

Region/line coverage of the frozen logic from this suite alone (via `cargo llvm-cov`, this test target only):

- `b20_stablecoin/logic/v1.rs`: **100% line, 100% function, 90.7% region**
- `b20_stablecoin/dispatch.rs`: **97.7% line, 92.4% region**

Remaining uncovered regions are defensive `checked_*().ok_or_else(under_overflow)` closures and `&&`/`||` short-circuit halves — unreachable without contriving arithmetic overflow.

## Testing

- `cargo test -p base-common-precompiles --features test-utils --test b20_stablecoin_v1_golden` → **80 passed**
- full crate (`cargo test -p base-common-precompiles --features test-utils`) green; clippy (`--all-targets`) + fmt clean

State hashes are pinned; regenerate with `BLESS_GOLDEN=1 cargo test ... -- --nocapture` (documented in the file header).

### DIFF
--- a/crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
+++ b/crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
@@ -27,6 +27,7 @@ use base_common_precompiles::{
     B20TokenRole, IB20, IB20Stablecoin, InMemoryPolicy, NoopPrecompileCallObserver, PermitArgs,
     Stablecoin, StablecoinV1, StablecoinVersion, StablecoinVersions, TokenAccounting,
 };
+use base_common_precompiles::{IB20::IB20Calls as C, IB20Stablecoin::IB20StablecoinCalls as SC};
 use base_precompile_storage::{BasePrecompileError, HashMapStorageProvider, StorageCtx};
 use k256::ecdsa::SigningKey;
 
@@ -2072,9 +2073,6 @@ fn golden_gas_footprints() {
 /// reviewed.
 #[allow(dead_code)]
 fn v1_op_coverage_checklist(call: IB20::IB20Calls, ext: IB20Stablecoin::IB20StablecoinCalls) {
-    use IB20::IB20Calls as C;
-    use IB20Stablecoin::IB20StablecoinCalls as SC;
-
     // No-op: forces each arm to name real golden `#[test]` fns by path.
     fn covered(_goldens: &[fn()]) {}
 
--- a/crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
+++ b/crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
@@ -1818,6 +1818,242 @@ fn golden_grant_role_unchecked_bootstraps_first_admin() {
 }
 
 // ============================================================================
+// gas: storage-access footprint per op
+// ============================================================================
+//
+// `gas_deducted` is 0 under the test gas schedule, so we pin the deterministic,
+// schedule-independent signal instead: the SLOAD / SSTORE / KECCAK256 op counts a
+// call performs. These are the storage-access footprint that drives real gas, so a
+// change here (e.g. an extra SLOAD in V1) is caught even when bytes/state/events match.
+
+/// Runs `calldata` privileged after `setup`, returning `(sload, sstore, keccak256)` counts.
+fn gas(
+    setup: impl FnOnce(&mut B20StablecoinStorage<'_>),
+    caller: Address,
+    policy: InMemoryPolicy,
+    calldata: Vec<u8>,
+) -> (u64, u64, u64) {
+    let mut s = fresh();
+    seed(&mut s, setup);
+    s.set_caller(caller);
+    s.reset_counters();
+    StorageCtx::enter(&mut s, |ctx| {
+        B20StablecoinToken::with_storage_and_policy(
+            B20StablecoinStorage::from_address(TOKEN, ctx),
+            policy,
+        )
+        .inner_with_privilege(ctx, &calldata, true)
+    })
+    .expect("gas-footprint op must succeed");
+    (s.counter_sload(), s.counter_sstore(), s.counter_keccak256())
+}
+
+/// An `InMemoryPolicy` authorizing `who` under the default (0) scope.
+fn allow0(who: Address) -> InMemoryPolicy {
+    let mut p = InMemoryPolicy::new();
+    p.allow(0, who);
+    p
+}
+
+#[test]
+fn golden_gas_footprints() {
+    let actual: Vec<(&str, (u64, u64, u64))> = vec![
+        (
+            "transfer",
+            gas(
+                |t| fund(t, ALICE, u(100)),
+                ALICE,
+                InMemoryPolicy::new(),
+                IB20::transferCall { to: BOB, amount: u(30) }.abi_encode(),
+            ),
+        ),
+        (
+            "transfer_from",
+            gas(
+                |t| {
+                    fund(t, ALICE, u(100));
+                    t.set_allowance(ALICE, BOB, u(40)).unwrap();
+                },
+                BOB,
+                InMemoryPolicy::new(),
+                IB20::transferFromCall { from: ALICE, to: BOB, amount: u(30) }.abi_encode(),
+            ),
+        ),
+        (
+            "approve",
+            gas(
+                |_t| {},
+                ALICE,
+                InMemoryPolicy::new(),
+                IB20::approveCall { spender: BOB, amount: u(50) }.abi_encode(),
+            ),
+        ),
+        (
+            "mint",
+            gas(
+                |_t| {},
+                ADMIN,
+                allow0(BOB),
+                IB20::mintCall { to: BOB, amount: u(100) }.abi_encode(),
+            ),
+        ),
+        (
+            "burn",
+            gas(
+                |t| {
+                    fund(t, ALICE, u(100));
+                    give_role(t, B20TokenRole::Burn.id(), ALICE);
+                },
+                ALICE,
+                InMemoryPolicy::new(),
+                IB20::burnCall { amount: u(40) }.abi_encode(),
+            ),
+        ),
+        (
+            "burn_blocked",
+            gas(
+                |t| {
+                    fund(t, ALICE, u(100));
+                    t.set_policy_id(B20PolicyType::TransferSender.id(), POLICY_ID).unwrap();
+                },
+                ADMIN,
+                InMemoryPolicy::new(),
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
```diff
@@ -0,0 +1,1396 @@
...
+/// A concrete (non-sentinel) policy id. Unconfigured scopes default to the
+/// `ALWAYS_ALLOW_ID` (0) EVM zero-slot, so blocking/executor guards must be
+/// exercised against an explicitly configured policy id like this one.
+const POLICY_ID: u64 = 7;
+
+// Anvil/Hardhat account 0 — well-known test key, never used in production.
+const PRIVATE_KEY: [u8; 32] =
+    alloy_primitives::hex!("ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80");
+
+// --- pinned state-roots (bless with BLESS_GOLDEN=1; see module docs) --------
```
meyer9: nice method of generating these!

crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
```diff
@@ -0,0 +1,1396 @@
...
+    })
+}
+
+/// Topic-0 (signature hash) of the last event emitted by the token.
+fn last_topic0(storage: &HashMapStorageProvider) -> B256 {
+    storage.get_events(TOKEN).last().expect("an emitted event").topics()[0]
+}
+
+/// Deterministic keccak state-root over the sorted `(address, slot, value)` storage triples.
+fn state_root(storage: HashMapStorageProvider) -> B256 {
```
meyer9: Maybe we can call call this `hash_state` since `state_root` generally refers to an MPT state root? There are also ways of generating an MPT state root from KV pairs, but concatting all of them seems fine for this use-case.

meyer9 (commented):
looks good in general! Only one small suggestion around "state root" naming.

crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
stephancill: nice job claude :D

crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
stephancill: Good call — renamed to `hash_state` (with a doc note that it's a plain content hash of KV pairs, not an MPT root) in 192fbe31b. Hashes are unchanged.

rayyan224:
Question, i think we should also measure the gas used by the transaction right ?

rayyan224:
Also can we structure this someway to be more explicit on the methods being tested ?

crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
```diff
@@ -1396,3 +1396,46 @@ fn dispatch_reverts_when_uninitialized() {
...
+// meta: op coverage guard
+// ============================================================================
+
+/// Extracts op identifiers from `C::<op>(` / `SC::<op>(` dispatch match arms.
+///
+/// The dispatcher's `match` over `IB20Calls` / `IB20StablecoinCalls` is a
+/// compiler-exhaustive list of every routable op, so this is the authoritative
+/// method surface.
+fn dispatch_ops(src: &str) -> std::collections::BTreeSet<String> {
+    let mut ops = std::collections::BTreeSet::new();
```
rayyan224: I’m not sure if this is the right abstraction, but I’m thinking of something along these lines:

```rust
struct TestExecutor;

impl TestExecutor {
    fn run_all(&self) {
        // Runs all registered test suites
    }

    fn run_for_method(&self, method: Method) {
        match method {
            Method::Transfer => transfer_test_suite().run(),
        }
    }
}

trait TestSuite {
    fn name(&self) -> &'static str;
    fn method(&self) -> Method;
    fn run(&self);
}
```

Each method would have its own `TestSuite` containing all relevant tests. The `TestExecutor` would be responsible for running every suite.

The nice part about using a Rust `match` here is that it must be exhaustive for the code to compile. If a new method is added but its test suite is not registered, the compiler will catch it. This gives us a compile-time guarantee that every supported method has an associated test suite.

crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
```diff
@@ -0,0 +1,2238 @@
...
+    storage.get_events(TOKEN).last().expect("an emitted event").topics()[0]
+}
+
+/// Deterministic keccak hash of the per-case snapshot: the token's emitted events
+/// (topics + data) followed by its sorted `(address, slot, value)` storage triples.
+///
+/// A plain content hash (not an MPT state root). Events are included so a regression in
+/// an event's payload — indexed args or data not otherwise reflected in storage, e.g. a
+/// `Memo`'s bytes — is pinned here even though logs are not storage.
+fn hash_state(storage: HashMapStorageProvider) -> B256 {
```
rayyan224, line 205: Nice, can we hash gas used also ?

rayyan224: Thoughts ?

crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
```diff
@@ -0,0 +1,2238 @@
...
+        buf.extend_from_slice(addr.as_slice());
+        buf.extend_from_slice(&slot.to_be_bytes::<32>());
+        buf.extend_from_slice(&value.to_be_bytes::<32>());
+    }
+    keccak256(&buf)
+}
+
+/// Asserts the storage hash, or prints it under `BLESS_GOLDEN` for (re)pinning.
+#[track_caller]
+fn assert_root(label: &str, storage: HashMapStorageProvider, expected: B256) {
```
rayyan224, line 226: I think a lot of these are utils, we should use throught golden tests ?

crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
stephancill: yeah we count opcodes since we don't have gas costs in this context

crates/common/precompiles/tests/b20_stablecoin_v1_golden.rs
stephancill: yeah can maybe refactor once we have another set of golden tests?

rayyan224 (approved):

cb-heimdall:
Review Error for rayyan224 @ 2026-07-16 22:08:27 UTC

PR #3979, opened 2026-07-15
feat(genesis): add zombie_block_timestamp and 200ms timing constants

meyer9:
## Summary

Adds deterministic timestamp computation to `RollupConfig`. Because the pre-Zombie chain advances by exactly `block_time` seconds per block, every block's millisecond-precise timestamp is derivable from block number and config alone — no extra wire data needed.

The function is fork-agnostic: it returns whole-second timestamps for pre-Zombie blocks and sub-second (200ms cadence) timestamps for post-Zombie blocks.

**Depends on #3978.**

## Changes (`crates/common/genesis/src/rollup.rs` only)

- `RollupConfig::timestamp_from_block_number(block_number) → Option<(u64, u16)>`: returns `(seconds, millis_part)` for any block; `None` only on integer overflow
  - Pre-Zombie: `(genesis_time + block_time * n, 0)` — no millis
  - Post-Zombie: anchored at activation, increments by 200ms per block
  - Debug-asserts that the Zombie activation timestamp is grid-aligned (multiple of 200ms from genesis)
- `MILLIS_PER_SECOND = 1000u16` and `ZOMBIE_BLOCK_TIME_MILLIS = 200u16` constants
- `RollupConfig::is_first_zombie_block(block_number)`: true for the activation block only
- Tests: `test_timestamp_from_block_number`, `test_timestamp_from_block_number_unscheduled`, `test_timestamp_from_block_number_non_zero_genesis`

## Stack
- PR1 (#3978): Zombie hardfork scaffolding
- **→ This PR (#3979)**: `timestamp_from_block_number` timing math
- PR3 (#3980): `SequencerTimestampPlanner`
- PR4 (#3981): Wire configurable block interval into build loop + CLI flag

### DIFF
--- a/crates/common/genesis/src/rollup.rs
+++ b/crates/common/genesis/src/rollup.rs
@@ -195,14 +195,6 @@ macro_rules! rollup_fork_methods {
 impl RollupConfig {
     /// Returns this rollup config's runtime-aware activation for a contract upgrade ID.
     pub fn contract_upgrade_activation(&self, upgrade_id: BaseUpgrade) -> UpgradeActivation {
-        // Zombie is a permanently-off production gate: runtime overrides cannot activate it.
-        if matches!(upgrade_id, BaseUpgrade::Zombie) {
-            #[cfg(any(test, feature = "test-utils"))]
-            return RuntimeUpgradeRegistry::activation(self.l2_chain_id.id(), upgrade_id)
-                .unwrap_or(UpgradeActivation::Never);
-            #[cfg(not(any(test, feature = "test-utils")))]
-            return UpgradeActivation::Never;
-        }
         RuntimeUpgradeRegistry::activation(self.l2_chain_id.id(), upgrade_id)
             .unwrap_or_else(|| self.upgrades.activation(upgrade_id))
     }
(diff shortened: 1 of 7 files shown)


crates/common/genesis/src/rollup.rs
```diff
@@ -374,6 +366,38 @@ impl RollupConfig {
...
+    /// activation_block  = (zombie_timestamp - genesis_l2_time) / block_time
+    /// anchor_seconds    = genesis_l2_time + block_time * activation_block
+    /// full_millis       = anchor_seconds * 1000 + (block_number - activation_block) * 200
+    /// ```
+    ///
+    /// Returns `None` when the Zombie upgrade is not scheduled or when `block_number` precedes the
+    /// activation block. Otherwise returns `(seconds, millis_part)` where `seconds` is the
+    /// whole-second timestamp and `millis_part` is the sub-second remainder (0-999, a multiple of
+    /// 200).
+    pub fn zombie_block_timestamp(&self, block_number: u64) -> Option<(u64, u16)> {
```
meyer9: in the future, this won't be zombie, so let's call this just `timestamp_from_block_number` or something. Can we also extend it to work before zombie? I don't want this to be zombie-specific. Every block in the past has had the exact timestamp it should (genesis_time + block_time * block_number).

crates/common/genesis/src/rollup.rs
```diff
@@ -374,6 +366,38 @@ impl RollupConfig {
...
+        let activation = self.contract_upgrade_activation_timestamp(BaseUpgrade::Zombie)?;
+        let activation_block = self.block_number_lower_bound_from_timestamp(activation);
+        let slots = block_number.checked_sub(activation_block)?;
+        let anchor_seconds =
+            self.genesis.l2_time.checked_add(self.block_time.checked_mul(activation_block)?)?;
+        let full_millis = anchor_seconds
+            .checked_mul(Self::MILLIS_PER_SECOND)?
```
meyer9: let's use std::time::Duration here

meyer9 (commented):
All addressed — see responses below.

crates/common/genesis/src/rollup.rs
meyer9: Done in 05c3f960. Renamed to `timestamp_from_block_number` and extended to cover all blocks — pre-Zombie blocks return `(genesis_l2_time + block_time * block_number, 0)` so it's the single timestamp oracle for any block number, fork-agnostic. Added a non-zero genesis test too. The caller in build.rs now uses `is_zombie_active(timestamp)` to gate the millis path instead of the old `if let Some` pattern.

crates/common/genesis/src/rollup.rs
meyer9: Done in 05c3f960. Using `core::time::Duration` (not `std` — the crate is no_std). The arithmetic is now:
```rust
let total = Duration::from_secs(anchor_seconds)
    .checked_add(Duration::from_millis(slots * ZOMBIE_BLOCK_TIME_MILLIS))?;
Some((total.as_secs(), total.subsec_millis() as u16))
```

crates/common/genesis/src/rollup.rs
```diff
@@ -374,6 +366,38 @@ impl RollupConfig {
...
+            .checked_add(slots.checked_mul(Self::ZOMBIE_BLOCK_TIME_MILLIS)?)?;
+        Some((
+            full_millis / Self::MILLIS_PER_SECOND,
+            (full_millis % Self::MILLIS_PER_SECOND) as u16,
+        ))
```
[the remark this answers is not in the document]
meyer9: Added a `debug_assert_eq!` in 05c3f960 that fires when the activation timestamp isn't grid-aligned to the block time. Network upgrade timestamps are always configured on block boundaries; this catches misconfiguration in dev/test builds without adding overhead in release. Went with option 1 from your suggestion — no test for the non-aligned case since the `debug_assert!` makes that a precondition violation.

meyer9:
Closing: superseded. `base/main` independently landed a Zombie hardfork identity as a permanently-off gate (contract-backed checks, EL derive plumbing) with different names/shape than this PR's `zombie_block_timestamp`/timing-constant additions. Rather than reconcile two divergent designs, the stack has been rebuilt on top of upstream's version: #3978 now makes upstream's Zombie gate genesis-configurable directly, and #3981 carries the sequencer-side 200ms cadence wiring. Collapsed the 4-PR stack down to these 2.

PR #3985, opened 2026-07-15
fix(eip8130): encode account_changes entry as a single flat RLP list

chunter-cb:
## Summary

Each `account_changes` entry in an EIP-8130 transaction was encoded as a bare `type_byte || rlp([body_fields...])` written **raw** into the surrounding `account_changes` list. This diverges from the EIP-8130 spec and is not self-framing RLP.

The spec encodes each entry as a single flat RLP list whose first element is the type byte:

```
rlp([0x00, user_salt, code, initial_actors])          // Create
rlp([0x01, chain_id, sequence, actor_changes, auth])  // ConfigChange
rlp([0x02, target])                                   // Delegation
```

This PR makes `AccountChange` encode/decode that flat form, matching both the spec and the sibling `ActorChange` codec (`rlp([change_type, actorId, data])`), which already carries its discriminant as the first list element.

## Why the old encoding was wrong

A bare `0x00` / `0x01` / `0x02` is itself a complete RLP token. So `type_byte || rlp(body)` placed unwrapped inside the `account_changes` list is **two** RLP tokens, not one. A generic RLP decoder walking the list sees **two items per entry**, and the resulting bytes (and therefore the tx hash / `sender_auth` signing domain) differ from a spec-compliant implementation. It only round-tripped because our custom decoder peeked the first byte.

## The three options (context from review)

A reviewer noted that other transaction types put the type byte *outside* the RLP (EIP-2718). That's true, but the detail that makes it work is that a typed transaction is wrapped as an **opaque RLP byte string** wherever it's nested (block bodies, devp2p), i.e. `rlp(type || rlp(payload))` — still one self-framing item. For interior list elements like `account_changes`, there are three designs:

1. **Type inside the list** (spec text, and this PR): `rlp([type, f1, f2, ...])` — one list item.
2. **EIP-2718 style, done correctly**: each entry is a byte string `rlp(type || rlp([f1, f2, ...]))` — the outer list holds opaque strings; read the type from the first byte. Also self-framing.
3. **What the code did**: `type || rlp([f1, f2, ...])` written raw into the list — matches neither (1) nor (2), and is not self-framing (2 tokens/entry).

Options 1 and 2 are both valid; this is genuinely a spec choice. This PR implements **option 1** because:

- The EIP already uses the type-inside-list style for the sibling `actor_change`, and our code implements that one correctly — option 1 keeps the two consistent.
- The "type outside" convention belongs to the transaction **envelope**, which we already do correctly (`0x79 || rlp([...])`, treated as an opaque string on the wire). `account_changes` entries are a different (interior list) context.
- Option 2 adds an extra length prefix + re-encode per entry for no benefit here.

Either way, the encoding must be self-framing and the spec + code must agree — option 3 should not stand. If we'd rather standardize on the envelope-style convention, we can switch this to option 2 (still a code change, plus a spec clarification). The EIP prose should also be updated to explicitly pin whichever option is chosen (it currently shows option 1 but does not rule out the ambiguous reading).

## Changes

- `AccountChange` `Encodable`/`Decodable` now produce/consume a single flat RLP list `rlp([type_byte, ...fields])`.
- Dropped the now-unused list-wrapped `RlpEncodable`/`RlpDecodable` derives from `CreateEntry`, `ConfigChange`, and `Delegation` (only ever reachable through `AccountChange`; a standalone body codec would be misleading).
- Updated module/type docs.

## Tests

- Existing round-trip tests updated to assert the type byte is the first *list element* (not a raw prefix).
- New `account_change_entry_is_single_rlp_item`: an entry decodes as exactly one self-contained RLP list.
- New `account_changes_vec_frames_one_item_per_entry`: the outer `account_changes` list contains exactly one RLP item per entry (regression guard against the 2-token framing).
- Invalid-type test now uses a well-formed list carrying an unknown type byte.

```
cargo test -p base-common-consensus   # 128 passed
cargo test -p base-execution-eip8130   # passed
```

## Note

The Solidity `AccountConfiguration.sol` contract does **not** constrain this: it works in ABI-encoded structs and never parses the `account_changes` RLP. So a contract-parity test cannot catch this framing bug — it must be pinned by RLP/tx-hash vectors (added here). Contract parity remains relevant for the semantic pieces (address-derivation commitment, signed-actor-change digests, `data` ABI).

### DIFF
--- a/crates/common/consensus/src/transaction/eip8130/account_changes.rs
+++ b/crates/common/consensus/src/transaction/eip8130/account_changes.rs
@@ -313,9 +313,7 @@ impl AccountChange {
     /// RLP-encoded, e.g. `0x00` -> `0x80`).
     fn rlp_payload_length(&self) -> usize {
         let fields_len = match self {
-            Self::Create(b) => {
-                b.user_salt.length() + b.code.length() + b.initial_actors.length()
-            }
+            Self::Create(b) => b.user_salt.length() + b.code.length() + b.initial_actors.length(),
             Self::ConfigChange(b) => {
                 b.chain_id.length()
                     + b.sequence.length()
@@ -457,7 +455,10 @@ mod tests {
         });
         let mut buf = Vec::new();
         ac.encode(&mut buf);
-        assert_eq!(first_list_element_type_byte(&buf), Eip8130Constants::ACCOUNT_CHANGE_TYPE_CREATE);
+        assert_eq!(
+            first_list_element_type_byte(&buf),
+            Eip8130Constants::ACCOUNT_CHANGE_TYPE_CREATE
+        );
         assert_eq!(buf.len(), ac.length());
         let decoded = AccountChange::decode(&mut buf.as_slice()).unwrap();
         assert_eq!(ac, decoded);
@@ -479,7 +480,10 @@ mod tests {
         });
         let mut buf = Vec::new();
         ac.encode(&mut buf);
-        assert_eq!(first_list_element_type_byte(&buf), Eip8130Constants::ACCOUNT_CHANGE_TYPE_CONFIG);
+        assert_eq!(
+            first_list_element_type_byte(&buf),
+            Eip8130Constants::ACCOUNT_CHANGE_TYPE_CONFIG
+        );
         assert_eq!(buf.len(), ac.length());
         let decoded = AccountChange::decode(&mut buf.as_slice()).unwrap();
         assert_eq!(ac, decoded);
--- a/crates/common/consensus/src/transaction/eip8130/account_changes.rs
+++ b/crates/common/consensus/src/transaction/eip8130/account_changes.rs
@@ -1,7 +1,8 @@
 //! [EIP-8130] `account_changes` entry types.
 //!
 //! An [`AccountChange`] is a tagged-union entry inside `TxEip8130::account_changes`.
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

haardikk21 (approved):

PR #4004, opened 2026-07-16
feat(upgrade-signal): auto-apply live L1 updates in runtime-admin mode

PelleKrab:
## Summary

Wires live auto-apply into the runtime-admin upgrade signal mode.

Runtime-admin previously applied the L1 signal only at startup and via the manual
`admin_refreshUpgradeSignal` RPC; live polling was metrics-only, so a schedule change after
startup was never applied unless every node was manually refreshed.

This PR makes live polling apply observed changes by:

- returning the read schedule from `UpgradeSignalMonitor::poll` when any signals updated
- splitting `UpgradeSignalRefresher` into `apply` (validate + apply an already-read schedule),
  `read_schedule`, and `refresh` (read + `apply`)
- giving the consensus metrics actor an optional refresher that auto-applies the polled schedule
  on observed changes
- routing the execution observer and the manual admin RPC through the same
  `ExecutionUpgradeSignalRuntimeRefresher::apply` path, validated against the startup chain spec

The polled schedule is applied directly rather than re-read, so the apply always matches what the
monitor observed. Apply failures are logged but not retried (validation is deterministic); failed
reads don't advance the baseline, so those are re-detected on the next poll. Metrics-only and
startup-apply modes still never auto-apply, though the first successful poll now logs an observed
update where it previously stayed silent.

## Tests

```
git diff --check
cargo +nightly fmt --check
cargo test -p base-upgrade-signal -p base-consensus-node -p base-execution-cli
cargo clippy -p base-upgrade-signal -p base-consensus-node -p base-execution-cli -- -D warnings

### DIFF
--- a/crates/utilities/upgrade-signal/src/runtime/refresher.rs
+++ b/crates/utilities/upgrade-signal/src/runtime/refresher.rs
@@ -37,23 +37,17 @@ impl UpgradeSignalRefresher {
         Self { config, reader, chain_id, runtime_validation, metrics_layer }
     }
 
-    /// Reads, metrics-records, logs, and validates the current L1 schedule.
-    pub async fn read_validated_schedule(
+    /// Validates and applies an already-read schedule without touching L1.
+    ///
+    /// Callers do not retry failures: validation failures are deterministic, and failed reads
+    /// never advance the monitor baseline, so changed signals are re-detected next poll.
+    pub fn apply(
         &self,
-    ) -> Result<UpgradeSignalSchedule, UpgradeSignalError> {
-        let schedule = self
-            .config
-            .read_validated_schedule(&self.reader, "runtime refresh", &[self.metrics_layer])
-            .await?;
-        self.runtime_validation.validate_schedule(self.chain_id, &schedule)?;
-
-        Ok(schedule)
-    }
-
-    /// Reads, validates, metrics-records, logs, and applies the current L1 schedule.
-    pub async fn refresh(&self) -> Result<UpgradeSignalApplySummary, UpgradeSignalError> {
-        let schedule = self.read_validated_schedule().await?;
-        let summary = UpgradeSignalRuntimeApplier::apply_schedule(self.chain_id, &schedule);
+        schedule: &UpgradeSignalSchedule,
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by cb-heimdall** [comment]
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

File: crates/utilities/upgrade-signal/src/runtime/refresher.rs [hunk 1]
```diff
@@ -37,23 +37,17 @@ impl UpgradeSignalRefresher {
         Self { config, reader, chain_id, runtime_validation, metrics_layer }
     }
 
-    /// Reads, metrics-records, logs, and validates the current L1 schedule.
-    pub async fn read_validated_schedule(
+    /// Validates and applies an already-read schedule without touching L1.
+    ///
+    /// Callers do not retry failures: validation failures are deterministic, and failed reads
```
mw2000, line 42: nit: The PR description says apply failures retry next poll, but this says they don’t. Could we update the description to match?

mw2000 (approved):

PR #4020, opened 2026-07-17
refactor(builder-core): componentize the flashblock build loop into gates + a reporter

BrianBland:
## What

Restructures `BasePayloadBuilderCtx::execute_best_transactions` (the ~575-line flashblock build loop) into a thin walk that composes reusable components.

## Shape

The walk is now: **source raw candidate → emit `Considered` → run gate chain → execute → emit terminal event**.

- **`PoolCandidateSource`** yields raw, fully materialized `Candidate`s. Bundle-window
  validity travels on the candidate as raw facts (`BundleWindow`); the source makes no
  admission decisions.
- **`Gate` trait**, run as an ordered chain (first rejection wins):
  - **`BundleGate`** — bundle-window validity (target block, expiry, not-yet-valid).
  - **`ResourceLimitsGate`** — metering estimation (and its wait/skip), the
    always-enforced hard limits, and the dry-run/enforce metering limit, folded into
    one gate. It produces the estimate once and preserves the hard-before-metering
    precedence. Metering is checked live (`is_enabled()`), so runtime toggling works.
  - **`SequencerGate`** — rejects blob/deposit transactions sourced from the pool.
- **`OutcomeReporter`** is the single layer that emits every builder transaction event
  and records rejection diagnostics, metrics, and the rejected-tx audit trail. Gates
  return a `GateRejection` the reporter renders; nothing else emits.
- Execution-time rejections (nonce-too-low, invalid, max-gas) reuse the same reporter
  via `GateRejection::Limit`; they stay inline pending a future post-execution gate.

## Behavior

Block contents, diagnostics, metrics, and the rejected-tx audit trail are **unchanged**.
The builder transaction **events** change in two intended ways:

1. **`Considered` is emitted once, up front**, when the builder begins evaluating a
   candidate — so its timestamp marks the start of evaluation and the
   `Considered`→terminal interval is a uniform measure of evaluation time.
2. **Decision-specific context** (bundle target/current block, metering wait, dry-run)
   is emitted **only on the terminal `Rejected` event**, no longer duplicated onto
   `Considered` (a small bugfix). The metering execution-time prediction likewise lands
   on the terminal event — it is an evaluation result, not known when the candidate is
   first considered (and the metering cache's LRU `get` can't be probed for
   bundle-gated candidates).

## Commits

1. **Split `is_tx_over_limits`** into hard + metering checks (byte-identical).
2. **Componentize into gates + a reporter** (the restructure above).

## Testing

`cargo clippy -p base-builder-core --lib` (clean, no allow-lints);
`cargo test -p base-builder-core --lib`; `--test data_availability
uncompressed_block_size miner_gas_limit bundles`; and `RUST_TEST_THREADS=1 --test
flashblocks`.

### DIFF
--- a/crates/builder/core/src/flashblocks/candidate_source.rs
+++ b/crates/builder/core/src/flashblocks/candidate_source.rs
@@ -1,0 +1,139 @@
+//! Pool candidate sourcing for the flashblock build loop.
+//!
+//! [`PoolCandidateSource`] wraps the payload transaction iterator and yields raw, fully
+//! materialized [`Candidate`]s. It makes no admission decisions — bundle-window validity travels
+//! on the candidate as a [`BundleWindow`] for the gates to judge — so the build walk owns the gate
+//! sequence and the source owns only iteration and materialization.
+
+use alloy_consensus::Transaction;
+use alloy_eips::Encodable2718;
+use alloy_primitives::{Address, B256, TxHash};
+use base_bundles::MeterBundleResponse;
+use base_common_consensus::BaseTransactionSigned;
+use base_execution_txpool::{
+    BasePooledTx, BundleTransaction, TimestampedTransaction, WatchManifest,
+    estimated_da_size::DataAvailabilitySized,
+};
+use reth_primitives_traits::Recovered;
+use reth_transaction_pool::PoolTransaction;
+
+use crate::{PayloadTxsBounds, TxResources};
+
+/// A bundle transaction's validity window, carried on a [`Candidate`] so the gate can evaluate a
+/// materialized candidate without retaining the pool transaction.
+///
+/// Implements [`BundleTransaction`] purely so it reuses that trait's default `is_bundle_expired` /
+/// `is_bundle_not_yet_valid` predicates — the single source of truth for bundle-window logic.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct BundleWindow {
+    /// The target block number, if set.
+    pub target_block: Option<u64>,
+    /// The minimum validity timestamp in milliseconds, if set.
+    pub min_timestamp_millis: Option<u64>,
+    /// The maximum validity timestamp in milliseconds, if set.
+    pub max_timestamp_millis: Option<u64>,
+}
+
+impl BundleTransaction for BundleWindow {
+    fn target_block_number(&self) -> Option<u64> {
+        self.target_block
+    }
+
+    fn min_timestamp_millis(&self) -> Option<u64> {
+        self.min_timestamp_millis
+    }
+
+    fn max_timestamp_millis(&self) -> Option<u64> {
+        self.max_timestamp_millis
+    }
+}
+
+/// A materialized pool candidate threaded through the admission gates.
+///
+/// The source populates everything except the metering-derived fields: [`resources`] starts with
+/// no execution-time prediction and [`resource_usage`] is `None`. The resource-limits gate enriches
+/// both from metering-service data during evaluation.
+///
+/// [`resources`]: Candidate::resources
+/// [`resource_usage`]: Candidate::resource_usage
+#[derive(Debug)]
+pub struct Candidate {
+    /// Recovered consensus transaction.
+    pub tx: Recovered<BaseTransactionSigned>,
+    /// Transaction hash.
+    pub tx_hash: TxHash,
+    /// Time the transaction was received, in milliseconds since the Unix epoch.
+    pub received_at_ms: u128,
+    /// Bundle validity window (empty for non-bundle transactions).
+    pub bundle: BundleWindow,
+    /// EIP-8130 authorization manifest to revalidate against on-chain state, if the transaction
+    /// carries one.
+    pub watch_manifest: Option<WatchManifest>,
+    /// EIP-8130 replay identifier, if applicable. Nonce-free replay-ID entries are independent, so
+    /// a manifest-precheck drop must not mark the sender's other entries invalid.
+    pub eip8130_replay_id: Option<B256>,
+    /// Effective priority fee (tip per gas), used as the value for rejection metrics.
+    pub priority_fee: f64,
+    /// Estimated resource usage; execution time is filled in by the resource-limits gate.
+    pub resources: TxResources,
+    /// Raw metering response, if any; filled in by the resource-limits gate.
+    pub resource_usage: Option<MeterBundleResponse>,
+}
+
+/// Sources fully materialized candidate transactions from a payload transaction iterator.
+#[derive(Debug)]
+pub struct PoolCandidateSource<'a, T> {
+    best_txs: &'a mut T,
+    base_fee: u64,
+}
+
+impl<'a, T: PayloadTxsBounds> PoolCandidateSource<'a, T> {
+    /// Creates a source over the given iterator, using `base_fee` to compute each candidate's
+    /// priority fee.
+    pub const fn new(best_txs: &'a mut T, base_fee: u64) -> Self {
+        Self { best_txs, base_fee }
+    }
+
+    /// Returns the next fully materialized candidate, or `None` when the pool is drained.
+    pub fn next_candidate(&mut self) -> Option<Candidate> {
+        let tx = self.best_txs.next(())?;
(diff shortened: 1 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/builder/core/src/flashblocks/reporter.rs
```diff
@@ -0,0 +1,312 @@
...
+                builder_mode: "flashblocks",
+                source_queue: "txpool_best",
+            },
+            event_type,
+            tx_hash,
+            data,
+        );
+    }
+
+    /// Reports a candidate admitted by every gate, emitting the considered event.
```
niran: From Grok:

> This docstring says the candidate was "admitted by every gate", but `considered` is emitted *before* the gate chain runs (so the timestamp marks start of evaluation). Worth rewording to match — e.g. that this reports the start of evaluating a candidate, not a successful admission.

crates/builder/core/src/flashblocks/candidate_source.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// The minimum validity timestamp in milliseconds, if set.
+    pub min_timestamp_millis: Option<u64>,
+    /// The maximum validity timestamp in milliseconds, if set.
+    pub max_timestamp_millis: Option<u64>,
+}
+
+impl BundleWindow {
+    /// Returns `true` if the bundle's constraints have expired relative to the given block number
+    /// and block timestamp (in seconds).
+    pub fn is_expired(&self, block_number: u64, block_timestamp_secs: u64) -> bool {
```
niran: From Grok:

> `BundleWindow::is_expired` / `is_not_yet_valid` reimplement the defaults on `BundleTransaction` after `into_consensus()` has dropped the trait. Easy to drift.
>
> Prefer one source of truth: either run the bundle checks via `BundleTransaction` methods before materializing into `Candidate`, or extract a shared helper both the trait defaults and `BundleWindow` call. Field-only `BundleWindow` + trait/helper predicates seems cleaner than two copies of the same boolean logic.

crates/builder/core/src/flashblocks/context.rs
```diff
@@ -674,136 +598,52 @@ impl BasePayloadBuilderCtx {
...
+                resources: tx_resources,
+                resource_usage,
+                priority_fee,
+                ..
+            } = candidate;
+            let tx_da_size = tx_resources.da_size;
+            let tx_uncompressed_size = tx_resources.uncompressed_size;
+            let predicted_execution_time_us = tx_resources.execution_time_us;
 
             let log_txn = |result: Result<TxnOutcome, TxnExecutionError>| {
```
niran: From Grok:

> The nested `reject` → `Option` → `log_txn` pattern is a bit hard to follow, and it splits observability: reporter owns events/metrics/diag, while `log_txn` stays a walk-local closure.
>
> Have you considered folding that debug line into the reporter? Then call sites could be just `reporter.reject(...)` / `reporter.accepted(...)` without the `if let Some(err)` dance.

crates/builder/core/src/flashblocks/context.rs
```diff
@@ -1043,52 +698,25 @@ impl BasePayloadBuilderCtx {
...
-                                    BuilderRejectedEventData::from_error(
-                                        &diag_err,
-                                        info,
-                                        limits,
-                                        Some(&tx_resources),
-                                    )
-                                },
-                            );
-                            log_txn(Err(diag_err));
+                            let rejection = GateRejection::Limit(TxnExecutionError::NonceTooLow);
```
niran: From Grok:

> Constructing `GateRejection::Limit(...)` for nonce-too-low / invalid / max-gas feels like the wrong abstraction. The pre-exec `Gate`s are static checks on the candidate (bundle window, resource limits, sequencer tx type). These post-exec failures are a different question: given the simulation result, is this tx valid to **include in the payload**?
>
> Rather than overloading `Gate` / inventing `GateRejection`s outside gates, a sibling type fits better — e.g. `InclusionCheck` with its own verdict/rejection enum — so the walk is `gates → execute → inclusion checks → reporter`, and the reporter has a dedicated entry point for inclusion failures. Fine as a follow-up if you want to keep this PR focused; I'd just avoid growing more call sites that launder execution outcomes through `GateRejection`.

niran (dismissed):
From Grok:

> Requesting changes on four design/docs points from review: `considered` docstring, `BundleWindow` predicate drift, folding `log_txn` into the reporter, and a sibling inclusion-check abstraction instead of hand-built `GateRejection`s for post-exec failures.

crates/builder/core/src/transaction_events.rs
```diff
@@ -129,48 +129,24 @@ impl From<&TxResources> for BuilderTransactionResources {
...
 pub(crate) struct BuilderConsideredEventData {
     #[serde(flatten)]
     budget: BuilderBudgetFields,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    bundle_target_block: Option<u64>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    tx_age_ms: Option<u128>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    metering_wait_duration_ms: Option<u128>,
 }
```
refcell: Don't think there's a point of this struct. It's just a thin wrapper around `BuilderBudgetFields`. Couldn't we just use that struct instead?

crates/builder/core/src/flashblocks/context.rs
BrianBland: I was planning on splitting gates into pre and post execution but wanted to bound the scope of this single PR.
Agreed that the naming is off here, and we can use something other than Limit for here.

crates/builder/core/src/transaction_events.rs
BrianBland: Agreed, this wrapper is no longer relevant after refactoring the "considered" event.

crates/builder/core/src/flashblocks/context.rs
```diff
@@ -1092,58 +693,38 @@ impl BasePayloadBuilderCtx {
...
                     }
                     // this is an error that we should treat as fatal for this attempt
-                    log_txn(Err(TxnExecutionError::EvmError));
+                    debug!(
+                        target: "payload_builder",
+                        message = "Considering transaction",
+                        tx_hash = ?tx_hash,
+                        tx_da_size = ?tx_da_size,
+                        result = %TxnExecutionError::EvmError,
+                    );
```
niran: Doesn't `execution_rejected()`/`log_outcome()` do this now?

niran (dismissed):

crates/builder/core/src/flashblocks/context.rs
BrianBland: Good catch, deleting this now

niran (approved):
needs a rebase

cb-heimdall:
Review Error for niran @ 2026-07-24 20:25:03 UTC

PR #4025, opened 2026-07-17
feat(rpc): populate Base millisecond timestamps

0x00101010:
Populates Base block and header RPC responses with millisecond timestamps derived from the canonical BaseTime metadata deposit while preserving omission for historical or invalid metadata.

### DIFF
--- a/crates/consensus/protocol/src/base_time.rs
+++ b/crates/consensus/protocol/src/base_time.rs
@@ -92,6 +92,16 @@ impl BaseTimeUpdateTx {
         Ok(base_time)
     }
 
+    /// Extracts the block timestamp in milliseconds from its transactions and header fields.
+    pub fn extract_timestamp_ms<T: BaseTransaction>(
+        transactions: &[T],
+        block_number: u64,
+        timestamp: u64,
+    ) -> Result<u64, BaseTimeMetadataError> {
+        let base_time = Self::extract_from_transactions(transactions, block_number)?;
+        Ok(timestamp.wrapping_mul(1_000).wrapping_add(u64::from(base_time.timestamp_millis_part())))
+    }
+
     /// Validates and decodes a `BaseTime` metadata deposit.
     pub fn validate_deposit(
         deposit: &TxDeposit,
@@ -310,6 +320,16 @@ mod tests {
         let base_time = BaseTimeUpdateTx::extract_from_transactions(&transactions, 9).unwrap();
 
         assert_eq!(base_time.timestamp_millis_part(), 600);
+    }
+
+    #[test]
+    fn extracts_timestamp_ms() {
+        let transactions: Vec<BaseTransactionSigned> = vec![
+            TxDeposit::default().seal_slow().into(),
+            base_time_deposit(9, 600).seal_slow().into(),
+        ];
+
+        assert_eq!(BaseTimeUpdateTx::extract_timestamp_ms(&transactions, 9, 42), Ok(42_600));
     }
 
     #[test]
--- a/crates/execution/rpc/src/eth/block.rs
+++ b/crates/execution/rpc/src/eth/block.rs
@@ -13,22 +13,6 @@ use reth_rpc_eth_api::{
 
 use crate::{BaseEthApi, BaseEthApiError, eth::RpcNodeCore};
 
-impl<N: RpcNodeCore, Rpc: RpcConvert> BaseEthApi<N, Rpc> {
-    fn extract_timestamp_ms<B: BlockBody>(
-        body: &B,
-        block_number: u64,
-        timestamp: u64,
-    ) -> Option<u64>
-    where
-        B::Transaction: BaseTransaction,
-    {
-        let millis = BaseTimeUpdateTx::extract_from_transactions(body.transactions(), block_number)
-            .ok()?
-            .timestamp_millis_part();
-        Some(timestamp.wrapping_mul(1_000).wrapping_add(u64::from(millis)))
-    }
-}
-
 impl<N, Rpc> EthBlocks for BaseEthApi<N, Rpc>
 where
     N: RpcNodeCore,
@@ -45,8 +29,12 @@ where
         Self: FullEthApiTypes,
     {
         let Some(block) = self.recovered_block(block_id).await? else { return Ok(None) };
-        let timestamp_ms =
-            Self::extract_timestamp_ms(block.body(), block.number(), block.timestamp());
+        let timestamp_ms = BaseTimeUpdateTx::extract_timestamp_ms(
+            block.body().transactions(),
+            block.number(),
+            block.timestamp(),
+        )
+        .ok();
         let mut header =
             self.converter().convert_header(block.clone_sealed_header(), block.rlp_length())?;
         header.timestamp_ms = timestamp_ms;
@@ -62,8 +50,12 @@ where
         Self: FullEthApiTypes,
     {
         let Some(block) = self.recovered_block(block_id).await? else { return Ok(None) };
-        let timestamp_ms =
-            Self::extract_timestamp_ms(block.body(), block.number(), block.timestamp());
+        let timestamp_ms = BaseTimeUpdateTx::extract_timestamp_ms(
+            block.body().transactions(),
+            block.number(),
+            block.timestamp(),
+        )
+        .ok();
         let mut block = block.clone_into_rpc_block(
             full.into(),
             |tx, tx_info| self.converter().fill(tx, tx_info),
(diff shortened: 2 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/rpc/src/eth/block.rs
```diff
@@ -1,18 +1,79 @@
...
+}
+
+fn timestamp_ms<B: BlockBody>(body: &B, block_number: u64, timestamp: u64) -> Option<u64>
+where
+    B::Transaction: BaseTransaction,
+{
+    let millis = BaseTimeUpdateTx::extract_from_transactions(body.transactions(), block_number)
+        .ok()?
+        .timestamp_millis_part();
+    Some(timestamp.wrapping_mul(1_000).wrapping_add(u64::from(millis)))
```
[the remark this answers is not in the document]
0x00101010: Keeping wrapping arithmetic intentionally: BaseTime::timestamp_ms uses the same operations to match Solidity uint64 truncation semantics, so RPC should expose the value the predeploy computes rather than apply a different overflow policy.

crates/execution/rpc/src/eth/block.rs
```diff
@@ -1,18 +1,83 @@
...
+    where
+        Self: FullEthApiTypes,
+    {
+        let Some(block) = self.recovered_block(block_id).await? else { return Ok(None) };
+        let timestamp_ms = block.body().extract_timestamp_ms(block.number(), block.timestamp());
+        let mut header =
+            self.converter().convert_header(block.clone_sealed_header(), block.rlp_length())?;
+        header.set_timestamp_ms(timestamp_ms);
+        Ok(Some(header))
+    }
```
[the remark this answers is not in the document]
0x00101010: This is the existing Reth behavior rather than a new cost introduced here: the default `EthBlocks::rpc_block_header` implementation already calls `self.recovered_block(block_id)` before converting the header. This override preserves that path so its cache and pending/local-pending semantics remain unchanged, then adds `timestamp_ms`. Avoiding sender recovery would be a worthwhile upstream Reth optimization, but switching to direct provider reads here would broaden this PR and subtly change those semantics.

crates/execution/rpc/src/eth/block.rs
```diff
@@ -1,18 +1,87 @@
...
+    fn extract_timestamp_ms<B: BlockBody>(
+        body: &B,
+        block_number: u64,
+        timestamp: u64,
+    ) -> Option<u64>
+    where
+        B::Transaction: BaseTransaction,
+    {
+        let millis = BaseTimeUpdateTx::extract_from_transactions(body.transactions(), block_number)
+            .ok()?
```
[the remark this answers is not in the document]
0x00101010: This is covered by tests

crates/execution/rpc/src/eth/block.rs
```diff
@@ -1,18 +1,87 @@
...
 
+trait TimestampMsHeader {
+    fn set_timestamp_ms(&mut self, timestamp_ms: Option<u64>);
+}
+
+impl TimestampMsHeader for BaseHeaderResponse {
+    fn set_timestamp_ms(&mut self, timestamp_ms: Option<u64>) {
+        self.timestamp_ms = timestamp_ms;
+    }
+}
```
refcell: Can probably just place this method directly on `BaseHeaderResponse` right? No real need for a trait unless this is going to be used across many types, and then if so should be made `pub trait`

crates/execution/rpc/src/eth/block.rs
```diff
@@ -1,18 +1,87 @@
...
+        timestamp: u64,
+    ) -> Option<u64>
+    where
+        B::Transaction: BaseTransaction,
+    {
+        let millis = BaseTimeUpdateTx::extract_from_transactions(body.transactions(), block_number)
+            .ok()?
+            .timestamp_millis_part();
+        Some(timestamp.wrapping_mul(1_000).wrapping_add(u64::from(millis)))
+    }
```
refcell: This can probably just be a method on `BaseTimeUpdateTx` to simplify the call chain.

Then, the `rpc_block_header` method can just directly call `BaseTimeUpdateTx::extract_from_block(...)`, passing in the block instead of needing to do the whole `block.body()`, `block.number()`, `block.timestamp()` deconstruction

crates/execution/rpc/src/eth/block.rs
0x00101010: good call, updated

crates/execution/rpc/src/eth/block.rs
```diff
@@ -1,18 +1,77 @@
...
+        let Some(block) = self.recovered_block(block_id).await? else { return Ok(None) };
+        let timestamp_ms =
+            Self::extract_timestamp_ms(block.body(), block.number(), block.timestamp());
+        let mut header =
+            self.converter().convert_header(block.clone_sealed_header(), block.rlp_length())?;
+        header.timestamp_ms = timestamp_ms;
+        Ok(Some(header))
+    }
+
+    async fn rpc_block(
```
refcell, line 44: where is this method used?

0x00101010: Reth calls `EthBlocks::rpc_block` from both `eth_getBlockByHash` and `eth_getBlockByNumber` (`rpc-eth-api/src/core.rs`). Those block responses embed this converted header, so overriding it populates `timestampMs` for the standard block endpoints. The separate `rpc_block_header` override covers `eth_getHeaderByHash` and `eth_getHeaderByNumber`.

0x00101010: Reth calls that helper from both eth_getBlockByHash and eth_getBlockByNumber

crates/execution/rpc/src/eth/block.rs
0x00101010: Good call, added one other helper method in `BaseTimeUpdateTx`, didn't directly use RecoveredBlock though cuz I don't want to introduce reth dependency, and instead just used the same parameters here

crates/consensus/protocol/src/base_time.rs
```diff
@@ -92,6 +92,16 @@ impl BaseTimeUpdateTx {
...
     }
 
+    /// Extracts the block timestamp in milliseconds from its transactions and header fields.
+    pub fn extract_timestamp_ms<T: BaseTransaction>(
+        transactions: &[T],
+        block_number: u64,
+        timestamp: u64,
+    ) -> Result<u64, BaseTimeMetadataError> {
+        let base_time = Self::extract_from_transactions(transactions, block_number)?;
+        Ok(timestamp.wrapping_mul(1_000).wrapping_add(u64::from(base_time.timestamp_millis_part())))
```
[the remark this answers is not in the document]
0x00101010: The wrapping behavior is intentional protocol compatibility, not defensive omission: the BaseTime predeploy returns a Solidity `uint64`, so overflow truncates modulo 2^64. `BaseTime::timestamp_ms` uses the same `wrapping_mul`/`wrapping_add`, and `timestamp_ms_matches_solidity_uint64_truncation` explicitly tests that invariant. Using checked arithmetic here would make RPC omit a value where the predeploy returns a wrapped value.

crates/execution/rpc/src/eth/block.rs
refcell: Makes sense, thanks!

refcell (approved):

PR #4061, opened 2026-07-20
fix(execution): account for EIP-8130 native gas costs

chunter-cb:
## Summary
- price packed account-state, policy-manager, policy-slot, and revoke accesses in EIP-8130 intrinsic gas
- consolidate config changes around one evolving account-state snapshot while preserving policy-slot warming semantics
- align TxContext getter pricing with the EIP's returned-word schedule while keeping protocol publication gas-free
- thread resolved policy state through execution and txpool accounting

## Test plan
- [x] `cargo test -p base-execution-eip8130 --lib`
- [x] `cargo test -p base-precompile-storage --lib`
- [x] `cargo test -p base-common-precompiles tx_context --lib`
- [x] `cargo test -p base-common-evm eip8130 --lib`
- [x] `cargo test -p base-execution-txpool eip8130 --lib`
- [x] `cargo fmt --all -- --check`
- [x] `git diff --check`


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/eip8130/src/apply.rs
```diff
@@ -359,18 +431,55 @@ impl AccountChangeApplier {
...
+    /// independently reading or writing the packed account-state slot.
+    pub fn revoke_actor_with_account_state(
+        storage: &mut AccountConfigurationStorage<'_>,
+        account: Address,
+        actor_id: B256,
+        state: &mut AccountState,
+    ) -> Result<(), ApplyError> {
+        let config = storage.get_actor_config(account, actor_id)?;
+        if config.authenticator != Address::ZERO {
+            return Self::revoke_explicit_actor(storage, account, actor_id, config);
```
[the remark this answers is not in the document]
chunter-cb: Documented the invariant in 2993aaa3c. Thanks!

chunter-cb:
Addressed the latest review findings in 284ab5dcf. Scoped to the confirmed correctness bug and the low-risk hardening/cleanup items; the consensus-gas *repricing* items (#2 per-ConfigChange first/subsequent split, #4 inline-k1-self revoke) and the #6 `EvolvingAccountState` refactor were intentionally deferred as follow-ups since they change consensus gas / are large-surface.

- **#1 (confirmed OOG risk):** estimation now resolves the payer's policy gate from its self-actor storage — the same way the sender's is resolved — instead of hardcoding `false`. A policy-gated sponsor's estimate now includes the payer's `policy_manager` SLOAD, so a `gas_limit == estimate` submission no longer OOGs on execution, without overcharging ungated payers (the prior round's concern). Factored into a shared `resolve_actor_scope` helper.
- **#3 (defense-in-depth):** the inline-self `DEFAULT_EOA_REVOKED` re-assertion is placed in `revoke_actor_with_account_state`, where the `AccountState` handle is already held and flushed by the caller (rather than `revoke_explicit_actor`, which is shared with non-self revokes and has no state handle). It's a no-op on all in-tree install paths and adds no storage access.
- **#5:** `OUTPUT_WORD_GAS` (W_copy) and the nonce dispatcher's `CALLDATA_WORD_GAS` (G_SHA3WORD) are now each pinned to their revm source (`gas::COPY` / `gas::KECCAK256WORD`) with drift tripwire tests mirroring the schedule's, and both document the deliberate input-vs-output word-pricing split.
- **#7:** added a comment documenting that `ACTOR_SLOT_SET_COST` / `ACCOUNT_STATE_SET_COST` / `CONFIG_CHANGE_STATE_COST` intentionally share the fresh-slot value but stay distinct for independent future repricing.
- **#8:** `tload_unmetered` is now a required trait method (no defaulting to `tload`), with explicit impls on all five providers, so metering is composed rather than silently inherited.

Verified: `cargo test` for `base-execution-eip8130`, `base-precompile-storage`, `base-common-precompiles`, `base-common-evm` (eip8130), `base-execution-txpool` (eip8130); `cargo clippy` on all touched crates; `cargo fmt --all -- --check`.

chunter-cb:
Added review findings **#2** and **#4** (both consensus gas, downward) in c60f6c130.

**#2 — repeated same-account config-change state writes**
All config changes in a transaction target the same (`sender`) packed account-state slot, and a leading create bootstraps that same slot. Only the first access is a cold zero-to-nonzero write; later bumps are a warm SLOAD + reset. Added `CONFIG_CHANGE_STATE_COST_SUBSEQUENT` (`WARM_SLOAD + SSTORE_RESET`) and a positional first-touch tracker in `IntrinsicGas::compute`. Purely static — no new runtime hint.

**#4 — inline secp256k1 self revoke**
A revoke of the inline self key touches three empty (zero-to-zero) slots, not three resets. This is state-derived (an explicit non-k1 self is a real reset and stays conservatively priced), so it needs an execution-resolved count rather than a static tweak:
- `revoke_actor_with_account_state` now returns whether it hit the inline home; `apply_config_change[_with_account_state]` sum the count.
- Threaded via `AppliedTransaction.inline_self_revokes` (execution + txpool) and `apply_account_changes` (estimation) into `IntrinsicGasInput::with_inline_self_revokes`.
- The intrinsic layer subtracts `INLINE_SELF_REVOKE_DISCOUNT` (= `ACTOR_REVOKE_COST − 3·COLD_SLOT_NOOP_COST`) per resolved inline revoke, via `saturating_sub`. A zero/unresolved count keeps the conservative reset price, so this can only reduce — never under-price — the charge, and estimation/execution/txpool all resolve it from the same deterministic apply pass.

Tests: added first-vs-subsequent state-bump and create+config-change cases, inline-self-revoke discount + saturation cases, an applier count test distinguishing inline/explicit/non-self revokes, and schedule drift asserts for both new constants. `fmt` + `clippy` clean; `base-execution-eip8130`, `base-common-evm`, `base-execution-txpool` suites green.

crates/common/evm/src/eip8130.rs
```diff
@@ -615,6 +615,29 @@ impl Eip8130Executor {
...
+        let actor_config =
+            acc.get_actor_config(account, actor_id).map_err(BaseTransactionError::eip8130)?;
+        if !actor_config.authenticator.is_zero() {
+            Ok(actor_config.scope)
+        } else if actor_id == AccountConfigurationStorage::self_actor_id(account) {
+            let state = acc.get_account_state(account).map_err(BaseTransactionError::eip8130)?;
+            Ok(if state.default_eoa_revoked() { 0 } else { state.default_eoa_scope })
+        } else {
+            Ok(0)
+        }
```
lukasrosario: isn't this admin scope? seems like a scary fallback

chunter-cb: Good eye — you're right that a raw `0` scope is `SCOPE_UNRESTRICTED` (full admin), so surfacing the absent-actor case as a scope is a footgun. It was safe here because this is estimation-only and both callers only tested `& SCOPE_POLICY` (`0 & 0x02 == 0` → not policy-gated, so we just skip pricing the extra `policy_manager` SLOAD — never an authorization decision). But the overloading was needless.

Fixed in bc3887506: the helper now returns a `bool` and is renamed `resolve_actor_policy_gated`, so "unknown/ungated actor" can no longer be confused with an unrestricted scope, and the doc comment calls out that it must never be repurposed for authorization.

crates/common/evm/src/eip8130.rs
```diff
@@ -708,6 +709,24 @@ impl Eip8130Executor {
...
+                Some(payer_addr) => {
+                    let payer_actor_id = AccountConfigurationStorage::self_actor_id(payer_addr);
+                    let payer_scope = acc
+                        .resolve_actor_config(payer_addr, payer_actor_id)
+                        .map_err(BaseTransactionError::eip8130)?
+                        .scope;
+                    payer_scope & Eip8130Constants::SCOPE_POLICY != 0
+                }
+                None => false,
+            };
```
chunter-cb: Clarifying the estimation-vs-execution difference for `payer_policy_gated`, since it isn't obvious from the diff:

This flag is **purely a gas-pricing signal** — it only decides whether `auth_cost` adds one extra `COLD_SLOAD` (2100) for the `policy_manager` read that authorization performs when `scope & SCOPE_POLICY != 0`. It never makes an authorization decision.

The three call sites resolve it differently, and that's intentional:
- **Execution** / **txpool**: derived from the *authenticated* payer actor (`applied_tx.actors.payer…is_policy_gated()`), so it's exact.
- **Estimation** (here): the payer's unsigned representative blob can't be authenticated, so we can't recover a non-self payer actor. Unlike the sender (which gets an RPC `senderActorId` hint), there is no payer-actor hint. So we price the payer's **self-actor** via `resolve_actor_config` — the same effective-config path authorize uses.

Accuracy:
- **Exact** when there's no payer, when the payer signs as its self-actor (gated or not), or when self and the acting actor share the same `SCOPE_POLICY` bit — this is the common sponsorship path and the case the fix targets (previously hardcoded `false`, which OOG'd a gas-gated sponsor at `gas_limit == estimate`).
- **Can diverge** only when the payer authenticates as a **non-self** actor whose `SCOPE_POLICY` bit differs from the self-actor's:
  - self ungated / acting gated → estimate under by 2100 (OOG risk at `gas_limit == estimate`)
  - self gated / acting ungated → estimate over by 2100 (safe, just loose)

This residual gap is the same limitation as "the payer stub isn't authenticable," not a flaw in the self-actor resolution. Closing it fully would need a `payerActorId`-style hint analogous to `senderActorId`.

crates/execution/eip8130/src/apply.rs
```diff
@@ -359,18 +439,97 @@ impl AccountChangeApplier {
...
+        if !is_self || state.default_eoa_revoked() {
             return Err(ApplyError::NotAnActor { actor_id });
         }
+
+        // The inline self's `actor_config` slot is always empty; its two policy
+        // slots are populated (real resets) only when the self was policy-gated,
+        // and empty otherwise. Read the gate from the inline scope before
+        // `disable_inline_self` zeroes it.
+        let empty_slots =
+            if state.default_eoa_scope & Eip8130Constants::SCOPE_POLICY == 0 { 3 } else { 1 };
```
xenoliss: The revoke discount still seems imprecise: `SCOPE_POLICY` does not imply non-zero policy slots. The EIP permits a zero manager and/or commitment, so this can overcharge an inline-self revoke by 2,800 gas per zero policy slot (up to 5,600).

chunter-cb: Good catch — you're right, and I've fixed it in 61be8dcb.

`slice_policy` writes policy data verbatim and its own doc confirms a gated actor may carry a zero `manager` and/or `commitment`, so the gate bit isn't a reliable proxy for populated slots. `revoke_actor_with_account_state` now counts the actually-empty slots from the raw storage before clearing:

```rust
let mut empty_slots = 1; // actor_config is always empty
if storage.get_policy_manager(account, actor_id)?.is_zero()    { empty_slots += 1; }
if storage.get_policy_commitment(account, actor_id)?.is_zero() { empty_slots += 1; }
```

Ungated stays 3; a gated self is now 1–3 depending on how many policy slots were written non-zero. Added gated-with-zero and mixed (manager set / commitment zero) test cases. This is safe - estimation still pins `revoke_discount_slots = 0`, so a larger execution discount only lowers the charge below the estimate. Also corrected the misleading `get_policy`/`get_policy_commitment` doc invariant that claimed slots are non-zero iff gated.

crates/common/evm/src/eip8130.rs
chunter-cb: Superseded by the Variant B change (55d1b89): estimation and txpool now **pin** the payer policy gate to worst-case (`policy_gates(true, payer.is_some())`) rather than resolving the payer's self-actor, so the divergence described here can no longer under-estimate. Execution still charges the exact gate from the authenticated actor. Resolving.

crates/execution/eip8130/src/apply.rs
```diff
@@ -300,49 +343,86 @@ impl AccountChangeApplier {
...
         // valid selector (`require(config.authenticator >= K1_AUTHENTICATOR)`).
         if config.authenticator.is_zero() {
             return Err(ApplyError::InvalidAuthenticator);
         }
 
-        let (manager, commitment) = Self::slice_policy(config.scope, policy_data)?;
         let self_id = AccountConfigurationStorage::self_actor_id(account);
+        if actor_id != self_id {
+            return Self::authorize_non_self_actor(storage, account, actor_id, config, policy_data);
+        }
```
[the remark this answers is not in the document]
chunter-cb: Documented the `state` contract on `authorize_actor_with_account_state` in 61be8dcb: it only reads/mutates `state` for the account's own self-actor; a non-self `actor_id` is dispatched to `authorize_non_self_actor`, which leaves `state` untouched (the mixed-actor loop relies on this). Callers must not assume `state` reflects a non-self change.

crates/execution/eip8130/src/intrinsic.rs
```diff
@@ -67,13 +67,117 @@ pub struct IntrinsicGasInput {
...
+    /// - A `ConfigChange` or a call-only transaction never installs sender code,
+    ///   so it does not suppress either.
+    ///
+    /// Both estimation and admission must pin this ceiling (execution reprices it
+    /// precisely from state) for the `estimate == admission >= execution`
+    /// guarantee to hold; resolving it from current code state on one path but
+    /// not the other silently breaks that invariant.
+    #[must_use]
+    pub fn sender_auto_delegated(account_changes: &[AccountChange]) -> bool {
+        !account_changes.iter().any(|change| matches!(change, AccountChange::Delegation(_)))
```
xenoliss:
```suggestion
        !account_changes.iter().any(|change| {
            matches!(change, AccountChange::Create(_) | AccountChange::Delegation(_))
        })
```
A valid `Create` installs code on the sender before auto-delegation, so this unnecessarily over-reserves 4,600 gas.

xenoliss (approved):

cb-heimdall:
Review Error for xenoliss @ 2026-07-24 07:43:55 UTC

PR #4066, opened 2026-07-20
feat(consensus): reconcile shadow sequencer cycles

haardikk21:
## Summary

- stop private block production after the configured shadow cycle length and wait for reconciliation
- replace the private branch with an exact contiguous range of authenticated P2P payloads
- defer derivation safe-head and finalized-head updates during private production, then apply them during reconciliation before resuming
- validate the cycle anchor and private head, preserve a single in-flight reconciliation request, and reset buffered cycle state on forkchoice reset
- cap the configured cycle length to the in-memory reconciliation buffer capacity

## Relationship

Built on the shadow P2P payload buffering merged in #4063.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/consensus/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -496,6 +525,155 @@ where
...
+                max_shadow_safe_signals = EngineProcessorOptions::MAX_SHADOW_PAYLOADS,
+                "Shadow safe-head signal buffer is full"
+            );
+            return;
+        }
+        self.shadow_safe_signals.push_back(safe_signal);
+    }
+
+    const fn buffer_shadow_finalized_block_number(&mut self, finalized_block_number: u64) {
+        self.shadow_finalized_block_number = Some(finalized_block_number);
```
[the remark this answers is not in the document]
haardikk21: Fixed. Buffered finalized updates now preserve the maximum observed height, and the regression test now expects 7 after receiving 7 then 6.

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -307,6 +346,46 @@ where
...
+                    reconciliation_task = Some(tokio::spawn(async move {
+                        engine_client.reconcile_shadow(cycle_start, private_head).await
+                    }));
+                }
+                task_result = async {
+                    match reconciliation_task.as_mut() {
+                        Some(task) => task.await,
+                        None => std::future::pending().await,
+                    }
+                } => {
```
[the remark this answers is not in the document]
haardikk21: Keeping the current pattern intentionally. `OptionFuture::from(None)` resolves immediately, which would leave this select arm continuously ready while no reconciliation is in flight. The explicit pending branch disables it, while awaiting the persistent `JoinHandle` remains cancel-safe when another arm wins.

crates/consensus/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -496,6 +525,155 @@ where
...
+                "safe-head consolidation changed the reconciled unsafe head".to_string(),
+            ));
+        }
+        self.shadow_payloads.retain(|number, _| *number > private_head.block_info.number);
+        self.shadow_reconciled_head = reconciled_head;
+        self.shadow_safe_signals.clear();
+        self.shadow_finalized_block_number = None;
+        if let Some(unsafe_head_tx) = &self.unsafe_head_tx {
+            unsafe_head_tx.send_replace(reconciled_head);
+        }
```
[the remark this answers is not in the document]
haardikk21: Addressed. Reconciliation now checkpoints the resulting safe/finalized forkchoice and notifies derivation of the safe head before clearing buffers or acknowledging completion.

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -322,10 +401,46 @@ where
...
+                                if !acknowledged_envelope_matches(submitted, inserted_head)
+                                    || inserted_head.block_info.number != expected_number
+                                {
+                                    self.cancellation_token.cancel();
+                                    return Err(EngineClientError::InvalidShadowReconciliation(
+                                        "private insertion acknowledgement does not match the cycle"
+                                            .to_string(),
+                                    )
+                                    .into());
+                                }
```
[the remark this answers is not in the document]
haardikk21: Agreed. This ordering is intentional: the acknowledgement is validated while the submitted envelope is still available, and every validation failure returns from the actor immediately, so retaining the sealer until actor teardown cannot lead to a retry or publication.

crates/consensus/service/src/actors/sequencer/admin_api_impl.rs
```diff
@@ -91,11 +92,14 @@ where
                 }
             }
             SequencerAdminQuery::ResetDerivationPipeline(tx) => {
-                if tx.send(self.reset_derivation_pipeline().await).is_err() {
+                let result = self.reset_derivation_pipeline().await;
+                reset_requested = true;
```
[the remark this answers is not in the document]
haardikk21: Fixed. A reset request now only resets the sequencer cycle bookkeeping when `reset_derivation_pipeline` succeeds. The existing success/error admin-query tests now also assert the returned reset indicator.

crates/consensus/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -208,20 +222,25 @@ where
...
                 self.checkpoint_reader.as_ref(),
             )
-            .await?;
+            .await;
 
         if self.is_shadow_sequencer() {
-            self.clear_shadow_payload_buffer();
+            self.shadow_reconciled_head = self.engine.state().sync_state.unsafe_head();
         }
+        let l2_safe_head = reset_result?;
```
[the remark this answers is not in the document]
haardikk21: Fixed. Shadow payloads, deferred safe/finalized signals, the fault marker, and `shadow_reconciled_head` are now mutated only after `reset_with_checkpoint_reader` succeeds. Added a regression test that forces reset failure and verifies all reconciliation state remains intact.

crates/consensus/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -496,6 +521,162 @@ where
...
+        }
+        self.shadow_safe_signals.push_back(safe_signal);
+    }
+
+    const fn buffer_shadow_finalized_block_number(&mut self, finalized_block_number: u64) {
+        self.shadow_finalized_block_number = Some(match self.shadow_finalized_block_number {
+            Some(previous) if previous > finalized_block_number => previous,
+            _ => finalized_block_number,
+        });
+    }
```
refcell: Shouldn't this always be fine to update? The finalized block number should always be >= right

crates/consensus/service/src/actors/engine/engine_request_processor.rs
refcell, line 1: This module is getting very heavy... could we refactor the shadow logic out into a new type?

crates/consensus/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -496,6 +521,162 @@ where
...
+
+    async fn reconcile_shadow_payloads(
```
refcell: Some doc comments on these methods would make it a lot easier to follow along.

crates/consensus/service/src/actors/engine/request.rs
```diff
@@ -62,10 +78,23 @@ pub enum EngineActorRequest {
...
+    ReconcileShadowRequest(Box<ReconcileShadowRequest>),
     /// Request to reset engine forkchoice.
     ResetRequest(Box<ResetRequest>),
 }
 
+/// Request to replace a private shadow range with the active sequencer's P2P branch.
+#[derive(Debug)]
+pub struct ReconcileShadowRequest {
+    /// Canonical unsafe head at the beginning of the private cycle.
+    pub cycle_start: L2BlockInfo,
```
refcell: The "cycle" terminology is very confusing to me. As I understand it, this is effectively the canonical chain's `L2BlockInfo` for the block at which shadow sequencer potentially diverges. If so, could this be named something like `canonical_start` to be more clear?

crates/consensus/service/src/actors/engine/request.rs
```diff
@@ -62,10 +78,23 @@ pub enum EngineActorRequest {
...
+    /// Last privately built head in the cycle.
+    pub private_head: L2BlockInfo,
```
refcell: Why is this private? Should this just be `shadow_head`?

crates/consensus/service/src/actors/engine/engine_request_processor.rs
haardikk21: Agreed. Finalized updates from derivation are ordered and monotonically increasing, so the buffer now simply stores the latest value instead of defensively taking the maximum. The test now uses the valid 6 -> 7 ordering.

crates/consensus/service/src/actors/engine/engine_request_processor.rs
haardikk21: Addressed. The payload buffer, fault state, canonical anchor, deferred safe signals, and finalized update now live in `ShadowReconciliationState`. That type owns buffering, pruning, clearing, and contiguous-range validation; `EngineProcessor` retains only the EL/derivation orchestration that requires its clients and engine.

crates/consensus/service/src/actors/engine/request.rs
haardikk21: Agreed and renamed to `canonical_start`. I propagated the terminology through the request, sequencer client, actor state, and engine reconciliation path so it consistently means the canonical unsafe head from which the shadow branch diverged.

crates/consensus/service/src/actors/engine/engine_request_processor.rs
haardikk21: Added doc comments to the extracted shadow reconciliation type and each of its buffering, pruning, clearing, and range-selection methods, plus the processor reconciliation method.

crates/consensus/service/src/actors/engine/request.rs
haardikk21: Agreed and renamed to `shadow_head` throughout the request/client/engine path. It is the current head built by the shadow sequencer, not a visibility distinction.

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -307,6 +346,46 @@ where
...
+                        Some(task) => task.await,
+                        None => std::future::pending().await,
+                    }
+                } => {
+                    reconciliation_task = None;
+                    let result = task_result.map_err(|error| {
+                        EngineClientError::ResponseError(format!(
+                            "shadow reconciliation task failed: {error}"
+                        ))
+                    })?;
```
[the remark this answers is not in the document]
haardikk21: Fixed. A reconciliation task panic or unexpected cancellation now cancels the shared node token before propagating the `JoinError`, matching the explicit reconciliation-error path. I also fixed the CI format failure by running the repository’s nightly formatter; it was only import grouping that stable rustfmt had not applied.

crates/consensus/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -106,12 +278,8 @@ where
...
     /// forkchoice update before being explicitly started.
     sequencer_stopped: bool,
-    /// Whether the sequencer privately builds blocks and reconciles from P2P payloads.
-    shadow_sequencer: bool,
-    /// Active-sequencer payloads received from authenticated P2P gossip, keyed by block number.
-    shadow_payloads: BTreeMap<u64, BaseExecutionPayloadEnvelope>,
-    /// Whether a conflicting payload or buffer overflow has made reconciliation unsafe.
-    shadow_payload_buffer_faulted: bool,
+    /// Shadow reconciliation state, present only when private shadow sequencing is enabled.
+    shadow_reconciliation: Option<ShadowReconciliationState>,
```
refcell: Thinking about this more, I'm not sure the shadow tracking belongs in the engine request processor at all. The model I keep coming back to is that `EngineProcessor` should be a view over reth's engine state and little else (i.e. forkchoice, sync state, the task queue, the heads reth knows about). It's already carrying too much, and the direction I'd like to see is less state, not more (starting by paying down the task-queue debt). Adding another mode to it cuts against that.

The thing that clarifies it for me is that there are really two categories of state getting mixed here:

- engine state — things reflected in reth: the canonical/unsafe head, forkchoice, what's been consolidated and finalized.
- actor state — things that live only in our process and describe how an actor is driving the engine.

Almost everything this PR adds is the second kind. `shadow_payloads`, `shadow_safe_signals`, `shadow_finalized_block_number`, `shadow_reconciled_head`, the cycle counter. None of that is in reth. The *private* chain is in reth (it's the current unsafe head), but the buffered authoritative payloads we're holding back, and the safe/finalized signals we're deferring, are pending inputs we haven't handed to reth yet. That's actor state that happens to be parked on the processor.

Looking at it through that split, a few things stick out to me:
 
1. The processor's interaction with other actors quietly changes based on sequencer config. Derivation sends `ProcessSafeL2SignalRequest` expecting a consolidate;  Network sends unsafe payloads expecting an insert. In shadow mode we silently buffer both instead:
  ```rust
  EngineActorRequest::ProcessSafeL2SignalRequest(safe_signal) => {
      if self.is_shadow_sequencer() {
          self.buffer_shadow_safe_signal(safe_signal);
          continue;
      }
      // ...normal consolidate path
  }
  ```
  Derivation and Network have no idea they're being intercepted. The processor's behavior toward them now depends on where the *sequencer* is in its cycle. That's a lot of hidden coupling to hang off an `is_shadow_sequencer()` check in the hot path, and it's now in `reset()` too.

  2. We track the cycle anchor twice. The sequencer owns `cycle_start`, the processor owns `shadow_reconciled_head`, and `reconcile_shadow_payloads` has to assert
  they agree:
  ```rust
  if range_len == 0
      || range_len > EngineProcessorOptions::MAX_SHADOW_PAYLOADS as u64
      || cycle_start != self.shadow_reconciled_head
      || self.engine.state().sync_state.unsafe_head() != private_head
  {
      return Err(EngineClientError::InvalidShadowReconciliation(/* ... */));
  }
  ```
  With two structs owning the same data and validating it by equality, there should probably just be a single owner, sharing data through message passing not copying.

  3. `is_shadow_sequencer()` branching is now threaded through the request loop, the buffer path, and reset. Every one of those is a spot where the engine gets further and further away from having a pure reth view.
  Maybe we could do a few things to clean this up:
  Keep `EngineProcessor` as a reth-state view plus a small set of primitives and move all the shadow bookkeeping out onto the actor side. Concretely two pieces, both owned outside the processor:
  The cycle orchestration (`cycle_start`, the private-blocks counter, the pending target, the reconcile cadence) is sequencer actor state. It reacts to each seal and decides when a cycle is done, so it wants to live next to the build/seal loop. Could pull it into a small `ShadowCycle` value type the sequencer owns rather than loose locals in the `select!`. Something like `on_inserted(head) -> Continue | Reconcile(target)` and `on_reconciled(head)`. That also makes it unit-testable without standing up an engine.
  The reconciliation buffer (the held authoritative payloads, the deferred safe/finalized signals, the anchor) is also actor state, but it belongs at the engine boundary, because that's where those inputs converge from Network and Derivation. I'd make it an explicit gate the `EngineActor` owns, in front of the processor: normal mode is pass-through; shadow mode buffers the forkchoice-advancing requests and, on a reconcile, replays them through the processor's primitives. The processor goes back to not knowing shadow mode exists.

  Roughly:

  ```text
  EngineActor
    ├─ ShadowReconciliation   (actor state: payload buffer, deferred safe/finalized, anchor, fault flag)
    │     ├─ intercept(req) -> Buffered | PassThrough      // one explicit call-site
    │     └─ reconcile(target) -> composes processor primitives
    └─ EngineProcessor        (reth-state view + primitives, no shadow awareness)
  ```

  That gives the anchor a single owner (the gate), so the sequencer passes a target and the gate validates against its own state instead of two copies cross-checking. It gets the `is_shadow_sequencer()` branches out of the processor. And it keeps the "not in reth yet" state clearly on the actor side of the engine/actor line.

  One more: we already have `handle_external_unsafe_l2_block` with its own gap handling and `MAX_SEQUENCER_EXTERNAL_UNSAFE_GAP = 300` — the same 300 as the shadow cap. Shadow buffering is basically a second mechanism for "external unsafe payloads showing up while I'm sequencing." It probably makes sense for those two to live under one roof (the gate is a natural home) rather than two parallel paths carrying the same magic number.

None of this has to block the PR if we're time-boxed, but at least pulling the shadow state off `EngineProcessor` would be preferrable.

haardikk21: Addressed in 95664a5ad. `EngineProcessor` no longer owns shadow mode/config/buffers or shadow-specific request branches. The actor-owned `EngineRequestHandler` now serializes an optional `ShadowReconciliationGate` with processor dispatch; the gate is the sole reconciliation-anchor owner and holds P2P payloads plus deferred safe/finalized inputs. Reconciliation calls a reth-focused `apply_canonical_inputs` primitive. I also pulled the sequencer locals into `ShadowCycle` and removed the anchor from `ReconcileShadowRequest`. Reset transitions now reanchor/publish before acknowledgment, while uncoordinated/internal resets and mutating reconciliation failures terminate shadow mode rather than risk stale actor state. The normal external-unsafe gap and shadow buffer cap remain separate policies.

crates/consensus/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -34,11 +41,66 @@ pub trait EngineRequestReceiver: Send + Sync {
...
+    DerivationClient: EngineDerivationClient,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.processor
+    }
+}
+
 /// Classifies the bootstrap behavior for the [`EngineProcessor`].
 ///
 /// Determined once at startup from the node's configuration and (if applicable)
```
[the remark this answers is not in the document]
haardikk21: Fixed in c050f0c93. Removed the `Deref`/`DerefMut` implementations and made every handler-to-processor interaction explicit, so the ownership/delegation boundary is visible at each call site.

crates/consensus/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -429,87 +478,79 @@ where
...
+    async fn apply_canonical_inputs(
+        &mut self,
+        inputs: CanonicalReconciliationInputs,
+    ) -> Result<Option<L2BlockInfo>, EngineClientError> {
+        let CanonicalReconciliationInputs {
+            shadow_head,
+            payloads,
+            safe_signals,
+            finalized_block_number,
+        } = inputs;
```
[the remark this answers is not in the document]
haardikk21: Fixed in c050f0c93. An empty authoritative range now returns `InvalidShadowReconciliation` instead of relying on the gate invariant and panicking.

crates/consensus/service/src/actors/engine/shadow_reconciliation_gate.rs
```diff
@@ -0,0 +1,240 @@
...
+        if self.safe_signals.len() >= Self::MAX_PAYLOADS {
+            self.faulted = true;
+        } else {
+            self.safe_signals.push_back(signal);
+        }
+    }
+
+    /// Records the latest finalized block.
+    pub const fn buffer_finalized(&mut self, number: u64) {
+        self.latest_finalized = Some(number);
```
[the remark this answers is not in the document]
haardikk21: Fixed in c050f0c93. `buffer_finalized` now retains the maximum observed height, with a regression test covering out-of-order updates.

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -307,6 +419,45 @@ where
...
+                        None => std::future::pending().await,
+                    }
+                } => {
+                    reconciliation_task = None;
+                    let result = task_result.map_err(|error| {
+                        self.cancellation_token.cancel();
+                        EngineClientError::ResponseError(format!(
+                            "shadow reconciliation task failed: {error}"
+                        ))
+                    })?;
```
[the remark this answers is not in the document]
haardikk21: This is already handled in the current code: the `JoinError` mapping calls `self.cancellation_token.cancel()` before returning the mapped error. This comment appears to have analyzed a stale version of the hunk after the force-push.

crates/consensus/service/src/actors/engine/shadow_reconciliation_gate.rs
```diff
@@ -0,0 +1,252 @@
...
+        if self.payloads.len() >= Self::MAX_PAYLOADS {
+            self.faulted = true;
+            return;
+        }
+        self.payloads.insert(number, envelope);
+    }
+
+    /// Buffers a deferred safe signal.
+    pub fn buffer_safe_signal(&mut self, signal: ConsolidateInput) {
+        if self.safe_signals.len() >= Self::MAX_PAYLOADS {
```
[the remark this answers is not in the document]
haardikk21: Addressed in be966cffd. Safe signals now have an independent 1024-entry bound rather than sharing the 300-payload cycle cap, with coverage showing derivation catch-up can exceed the payload capacity without faulting the gate. The gate still fails closed at its own bound rather than dropping consolidation inputs.

crates/consensus/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -909,73 +947,156 @@ where
...
+                                    .send(Err(EngineClientError::ResetForkchoiceError(
+                                        error.to_string(),
+                                    )))
+                                    .await
+                                    .is_err()
+                                {
+                                    warn!(target: "engine", "Sending reset response failed");
+                                }
+                                continue;
+                            }
```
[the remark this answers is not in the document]
haardikk21: Cleaned up in be966cffd. The reset error response is sent once, followed by the shadow-specific terminal return or normal-mode continuation.

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -37,6 +40,74 @@ use crate::{
 /// Sender stashed by `stop_sequencer` when waiting for an in-flight seal pipeline to drain.
 pub type PendingStopSender = oneshot::Sender<Result<B256, SequencerAdminAPIError>>;
 
+/// Progress through one private shadow build and reconciliation cycle.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ShadowCycle {
+    Building { next_height: u64, built: u64 },
+    AwaitingReconciliation { shadow_head: L2BlockInfo },
+}
```
refcell: This is nice

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -283,6 +362,17 @@ where
...
+            Some(ShadowCycle::building(self.engine_client.get_unsafe_head().await?)?)
+        } else {
+            None
+        };
+        let mut reconciliation_task: Option<
+            JoinHandle<Result<Option<L2BlockInfo>, EngineClientError>>,
+        > = None;
+        let mut reconciliation_ticker =
+            tokio::time::interval(Duration::from_secs(self.rollup_config.block_time));
+        reconciliation_ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
```
refcell, line 360: Can we clean this up and place a doc comment here? This is the main sequencer actor loop and it's important to keep it legible. Maybe we could refactor this into a `self.start_reconciliation()` method?

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -291,13 +381,35 @@ where
...
+                        shadow_cycle = Some(ShadowCycle::building(canonical_anchor)?);
+                        next_payload_to_seal = if self.is_active {
+                            self.builder.build_on(canonical_anchor).await?
+                        } else {
+                            None
+                        };
+                        if self.is_active && next_payload_to_seal.is_none() {
+                            build_ticker.reset_immediately();
+                        }
+                    }
```
refcell, line 390: Doc comment explaining why we do this would be a nice to have

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -429,6 +611,14 @@ where
     }
 }
 
+fn acknowledged_envelope_matches(
+    envelope: &BaseExecutionPayloadEnvelope,
+    head: L2BlockInfo,
+) -> bool {
+    envelope.execution_payload.block_number() == head.block_info.number
+        && envelope.execution_payload.block_hash() == head.block_info.hash
+}
```
refcell: Could this be a method on `BaseExecutionPayloadEnvelope` to avoid a free floating method in the sequencer actor module?

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -307,6 +419,45 @@ where
...
+                            shadow_cycle.as_mut().expect("shadow reconciliation requires cycle state").reconcile(head)?;
+                            if self.is_active {
+                                next_payload_to_seal = self.builder.build_on(head).await?;
+                                if next_payload_to_seal.is_none() {
+                                    build_ticker.reset_immediately();
+                                }
+                            }
+                        }
+                        Ok(None) => reconciliation_ticker
+                            .reset_after(Duration::from_secs(self.rollup_config.block_time)),
```
refcell, line 438: Why do we do this?

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -322,10 +473,41 @@ where
...
+                                    || inserted_head.block_info.number != expected_number
+                                {
+                                    self.cancellation_token.cancel();
+                                    return Err(EngineClientError::InvalidShadowReconciliation(
+                                        "private insertion acknowledgement does not match the cycle"
+                                            .to_string(),
+                                    )
+                                    .into());
+                                }
+                            }
```
refcell, line 464: Could this be refactored into a method? Also, if this errors it will exit the sequencer process entirely. I think that's intentional here and in other places but just calling that out.

crates/consensus/service/src/actors/sequencer/actor.rs
```diff
@@ -322,10 +473,41 @@ where
...
+                                let cycle = shadow_cycle.as_mut().expect("shadow mode has cycle state");
+                                cycle.inserted(
+                                    inserted_head,
+                                    self.shadow_blocks_per_cycle.expect("shadow mode checked").get(),
+                                )?;
+                                if cycle.reconciliation_target().is_some() {
+                                    next_payload_to_seal = None;
+                                    reconciliation_ticker.reset_immediately();
+                                }
+                            }
```
refcell, line 476: Also could be refactored out

crates/consensus/service/src/actors/sequencer/actor.rs
refcell: Can we move this type out of `actor.rs` into a new module to keep `actor.rs` a bit more digestible since it contains the critical sequencer actor loop?

jowparks (approved):
Looks good and is low risk as this doesn't affect active sequencer code path.

cb-heimdall:
Review Error for jowparks @ 2026-07-21 20:06:30 UTC

crates/consensus/service/src/actors/sequencer/actor.rs
haardikk21: Moved reconciliation startup and its task type into the documented `ShadowCycle::start_reconciliation` API, so the actor loop only starts the returned task.

crates/consensus/service/src/actors/sequencer/actor.rs
haardikk21: Added `reset_shadow_cycle_after_admin` with docs explaining that the coordinated engine reset invalidates the in-flight seal, reconciliation task, and actor-side cycle state, all of which must be discarded before rebuilding on the canonical head.

crates/consensus/service/src/actors/sequencer/actor.rs
haardikk21: Moved this onto `PayloadSealer::matches_inserted_head`. That is the local owner of the submitted envelope and avoids making the lower-level envelope crate depend on the consensus-layer `L2BlockInfo` type.

crates/consensus/service/src/actors/sequencer/actor.rs
haardikk21: Added a comment here: `None` means the gate is not ready because one or more canonical payloads have not arrived over P2P yet. Resetting by one block interval prevents immediately spawning another reconciliation task in a hot loop.

crates/consensus/service/src/actors/sequencer/actor.rs
haardikk21: Extracted this into the documented cycle validation path plus the actor cancellation wrapper. And yes, termination is intentional: an acknowledgement mismatch means the actor and engine disagree about the private branch, so continuing could build from corrupted/stale cycle state.

crates/consensus/service/src/actors/sequencer/actor.rs
haardikk21: Done. `ShadowCycle`, reconciliation startup, transition validation, and focused tests now live in `sequencer/shadow_cycle.rs`.

crates/consensus/service/src/actors/sequencer/actor.rs
haardikk21: Extracted the insertion transition into `ShadowCycle::record_insertion`, with the actor helper retaining the fail-closed node cancellation behavior.

refcell (approved):
Think this is in a solid place to merge. There's a good amount of refactoring that can still be done at the sequencer actor level but nothing critical/blocking.

refcell (approved):

jowparks (approved):

cb-heimdall:
Review Error for jowparks @ 2026-07-21 20:34:06 UTC

PR #4093, opened 2026-07-22
fix(txpool): EIP-8130 mempool guard follow-ups (stale balance, LRU, expiry sweep, metrics, lock reuse)

chunter-cb:
Second-pass review follow-ups for the EIP-8130 mempool guard. Independent, self-contained commits, each tightening correctness, cache health, expiry timeliness, ops signal, code reuse, or (#8) the trusted-payer safety bar. EIP-8130 traffic only, dormant until Beryl activates.

## Commits

**#1 + #2 — `fix(txpool): correct stale payer-balance seed and LRU eviction bias`**
- **#1** A trusted payer whose balance dropped before its first tracked tx would seed a fresh `PayerBook` from the stale validation snapshot (`on_balance_changed` only corrects payers that already have a book). Fix: advance the classification generation on a balance diff **for a known-trusted payer**, forcing an admission whose validation predates the change to re-validate against fresh balance. The bump is restricted to trusted payers — the only ones that use the balance book, and already classified into the cache by the pending tx's own validation — so ordinary balance churn (every block) no longer bounces unrelated admissions. Uses a non-promoting cache lookup so it doesn't perturb LRU recency.
- **#2** `invalidate_code`/`invalidate_slot` switched from the promoting `get_mut` to non-promoting `peek_mut`, so an account with frequent code/config churn no longer pins itself at MRU and biases eviction against fresher entries.

**#3 — expiry horizon (wall-clock sweep reverted)**
- Dropped the one-block lookahead from the expiry horizon (was `(now + BUCKET) / BUCKET`, evicting up to ~3s early); horizon is now `now / BUCKET`, so a bucket fires only once its window has started.
- The initially-explored 500ms wall-clock sweep was **reverted**: it was a best-effort hygiene supplement, not an inclusion-safety mechanism (the builder's manifest precheck already drops past-expiry txs against the exact build timestamp, and block-driven `expire_due_buckets` frees reservations/slots every canonical update). Relying on block-driven eviction also moots the `spawn_critical_task` concern. No dedicated sweep task remains.

**#4 — `refactor(txpool): split reorg vs feed-gap invalidation metric label`**
- Reorg (common) and feed lag (rare, alert-worthy) previously collapsed into one `feed_gap` label. Introduces `InvalidationCause { Reorg, FeedGap }` with `as_label`, threads it through `invalidate_all_tracked`, and replaces the stringly-typed helper with `record_bulk_invalidations`.

**#6 — `refactor(txpool): derive account lock via lock_status helper`**
- `account_lock` now routes through `AccountState::lock_status` instead of re-deriving from raw flags. Behavior preserved exactly (`unlocks_at` only for a pending unlock; hard-lock `UNLOCKS_AT_MAX` sentinel never surfaces as a timed bucket).

**#8 — `feat(txpool): require hard lock >=1h for high-rate payer classification`** (now included; originally slated for separate discussion)
- A high-rate (balance-bounded) payer must now be *hard*-locked with a configured unlock delay of at least one hour (`MIN_HIGH_RATE_PAYER_LOCK_SECS = 3600`). An account with a pending (initiated) unlock — regardless of how far in the future — or an unlocked account is never treated as high-rate.
- Rationale (security tightening): once an unlock is initiated the payer could soon move ETH while still holding `PayerBook` reservations, breaking the eth-movement guarantee the balance book relies on. The hard-lock → pending-unlock transition writes the account-state slot, which the existing slot watch already invalidates, so admitted txs drain naturally with no timed bucket needed. **This narrows the trusted set** vs. prior behavior; re-review of this specific change is welcome.
- Also reused authorization reads for lock/trust classification (`66f2089`) to avoid duplicate SLOADs on the validation hot path.

Not included: #5 (byte-cap accounting — confirmed non-load-bearing; count caps + PayerBook are the real backstop) and #7 (CLI/governance shape — held for separate discussion).

## Testing
- `cargo test -p base-execution-txpool` — passing (adds: trusted-only generation bump, LRU non-promotion, hard-lock ≥1h high-rate classification).
- `cargo clippy -p base-execution-txpool -p base-node-core --all-targets` — clean.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/txpool/src/state_diff_maintain.rs
```diff
@@ -14,13 +14,38 @@ use tracing::{debug, warn};
...
 use crate::AccountStateDiff;
 
+/// Why every guarded transaction was flushed in one shot. Both paths are
+/// fail-safe (the guard cannot prove any admission current), but they differ
+/// wildly in frequency, so they carry distinct metric labels.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum InvalidationCause {
+    /// A canonical chain reorg. The committed segment omits state changes caused
+    /// solely by rolling back the reverted segment, so targeted invalidation is
+    /// not possible. Common in normal operation.
```
haardikk21: non blocking nit, this is not at all common in normal operation - canonical chain reorgs should never be happening on base, smth has gone severely wrong otherwise

crates/execution/txpool/src/state_diff_maintain.rs
```diff
@@ -14,13 +14,38 @@ use tracing::{debug, warn};
...
+    Reorg,
+    /// A gap in the canonical-state feed (the broadcast stream lagged and
+    /// dropped notifications). Rare, and a signal worth alerting on.
+    FeedGap,
```
haardikk21, line 31: this is probs more likely than reorg lol

haardikk21 (approved):

PR #4096, opened 2026-07-22
fix(payload): Bound SSZ Payload List Decode

refcell:
## Summary

An unauthenticated peer on the public P2P mesh could send a small (~0.5 MiB) snappy frame on the v4 blocks topic whose SSZ `transactions` list declared ~2.6M entries, forcing the single-threaded gossip task to allocate ~125 MiB of heap before any signature check ran (CWE-770). The transaction and withdrawal lists were decoded into raw unbounded `Vec`s, so the decoder honored the attacker's element count instead of the protocol maximum.

This bounds the v4 `transactions` and `withdrawals` SSZ lists to `MAX_TRANSACTIONS_PER_PAYLOAD` and `MAX_WITHDRAWALS_PER_PAYLOAD`, rejecting an oversized declared count at the offset read before any per-element allocation. It adds two regression tests: one asserting the decoder rejects the count bomb, and one that measures decode allocation via a counting global allocator and holds it well under the amplified size.

Note this covers the v4 decode path only; v1/v2/v3 delegate to `alloy_rpc_types_engine`'s unbounded decoder and still need equivalent bounding as a follow-up.

### DIFF
--- a/crates/common/rpc-types-engine/src/envelope.rs
+++ b/crates/common/rpc-types-engine/src/envelope.rs
@@ -579,12 +579,6 @@ impl PayloadHash {
 
 #[cfg(test)]
 mod tests {
-    #[cfg(feature = "std")]
-    use std::{
-        alloc::{GlobalAlloc, Layout, System},
-        cell::Cell,
-    };
-
     use alloy_primitives::b256;
     #[cfg(feature = "std")]
     use alloy_primitives::hex;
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/rpc-types-engine/src/payload/v4.rs
```diff
@@ -78,6 +78,63 @@ impl BaseExecutionPayloadV4 {
     }
 }
 
+/// The maximum number of transactions in an execution payload, per
+/// `MAX_TRANSACTIONS_PER_PAYLOAD` in the Ethereum consensus specs.
+pub const MAX_TRANSACTIONS_PER_PAYLOAD: usize = 1 << 20;
```
BrianBland, line 83: This is just over 1M transactions, and at the bare minimum this would consume around 22Ggas (11Ggas/s)

crates/common/rpc-types-engine/src/payload/v4.rs
```diff
@@ -78,6 +78,63 @@ impl BaseExecutionPayloadV4 {
...
+
+/// The maximum number of withdrawals in an execution payload, per
+/// `MAX_WITHDRAWALS_PER_PAYLOAD` in the Ethereum consensus specs.
+pub const MAX_WITHDRAWALS_PER_PAYLOAD: usize = 1 << 4;
```
BrianBland: Can we just set this to 0 for Base?

BrianBland (approved):

cb-heimdall:
Review Error for BrianBland @ 2026-07-22 21:26:18 UTC

crates/common/rpc-types-engine/src/payload/v4.rs
refcell: Yes

crates/common/rpc-types-engine/src/envelope.rs
```diff
@@ -713,4 +719,114 @@ mod tests {
...
+        unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+            // SAFETY: `ptr` and `layout` come straight from the caller and
+            // originate from `System.alloc`, satisfying `System::dealloc`.
+            unsafe { System.dealloc(ptr, layout) };
+        }
+    }
+
+    #[cfg(feature = "std")]
+    #[global_allocator]
+    static COUNTING_ALLOCATOR: CountingAllocator = CountingAllocator;
```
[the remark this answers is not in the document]
refcell: Good catch — moved the allocation test and its `#[global_allocator]` out of the crate's `#[cfg(test)] mod tests` and into a dedicated integration test binary (`tests/decode_allocation_bound.rs`). The counting allocator now governs only that one-test binary, so it can't collide with a future `#[global_allocator]` elsewhere in the crate and no other unit test pays the bump overhead. Done in e996d3a.

crates/common/rpc-types-engine/src/payload/v4.rs
refcell: Right — `1 << 20` (~1.05M txs) is intentionally the Ethereum consensus-spec `MAX_TRANSACTIONS_PER_PAYLOAD`, i.e. a generous defensive upper bound rather than a tight per-block limit. As you note it's already ~22 Ggas, roughly a block's worth at 11 Ggas/s, so it comfortably covers any legitimate payload while still rejecting the ~2.6M-entry bomb a max decompressed frame can declare. Real gas/size limits are enforced downstream during block validation; this bound only exists to cap pre-authentication allocation. Keeping it at the spec value.

crates/common/rpc-types-engine/src/payload/v4.rs
refcell: Done — set `MAX_WITHDRAWALS_PER_PAYLOAD` to 0 in e996d3a. Base sets the withdrawals root directly and requires the SSZ `withdrawals` list to be present but empty, so any non-empty list is now rejected during decode.

BrianBland (approved):

PR #4103, opened 2026-07-22
feat(protocol): bind EIP-8130 authenticator length to actor presence on decode

xenoliss:
# feat(protocol): bind EIP-8130 authenticator length to actor presence on decode

## Summary

The span-batch EIP-8130 decoder did not enforce any relationship between an
actor being configured and the length of its authenticator column. The two
sides could disagree freely:

- `sender = Some(..)` but `sender_authenticator` empty or not 20 bytes;
- `sender = None` but `sender_authenticator` non-empty;
- the same mismatches for `payer` / `payer_authenticator`.

The authenticator column is meant to hold the leading 20-byte account address
on the configured-actor path and to be empty on the EOA / self-pay path (see
the struct docs and `split_auth`). Without a check on the decode side, a
malformed span batch flows into `to_tx`, where `join_auth` silently
reconstructs a corrupt auth blob instead of failing.

This binds an actor's presence to its authenticator length at decode time,
which is the counterpart of the split `split_auth` performs on the encode side.

## Changes

- Add `SpanBatchEip8130TransactionData::validate_authenticator`, enforcing:
  - configured path (`sender`/`payer` is `Some`): authenticator is exactly
    20 bytes;
  - EOA / self-pay path (`None`): authenticator is empty.
- Call it for both the sender and payer columns in `Decodable::decode`,
  right after the fields are read, so a mismatch is rejected with a clear
  RLP error before `to_tx`/`join_auth` can run.
- Add `decode_rejects_authenticator_actor_mismatch`, which starts from a
  consistent configured-sender / self-pay tx and breaks exactly one
  presence↔length invariant per case (sender empty, sender >20 bytes,
  EOA sender with a non-empty authenticator, absent payer with a non-empty
  authenticator), asserting each is rejected while the untampered tx decodes.

## Notes

- No wire-format change: encoding is unchanged; this only tightens what the
  decoder accepts. Any well-formed batch produced by the encoder already
  satisfies the invariant.
- Mirrors the equivalent fix in the OP Stack fork (`op-node`,
  `checkEip8130Authenticator` in `span_batch_tx.go`, branch `bo/eip8130`).

## Testing

- `cargo test -p base-protocol --lib batch::tx_data::eip8130` — 4 passed.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

roger-bai-coinbase (dismissed):

chunter-cb (approved):

PR #4113, opened 2026-07-22
fix(consensus): preserve L1 finality across derivation resets

### LINKED ISSUE #3895: Preserve L2 finalization progress
We observed this issue across several of our RPC nodes. The investigation was initially triggered by unbounded memory growth and repeated OOM kills. We traced the root cause to lost finalization progress, applied this patch to our nodes, and confirmed it resolves the problem.

You can find the code fix in https://github.com/base/base/pull/3894

The L2Finalizer loses finalization progress in two independent ways:

try_finalize_next() silently drops a finalized L1 signal when the awaiting_finalization queue is empty. The L1 watcher does not re-send signals, so the EL's finalized head can stay stuck indefinitely; nodes restarted during sync report finalized=0 for days. Since reth's changeset cache eviction keys off the finalized block, the wedge also causes unbounded memory growth and eventual OOM.

Every pipeline reset clears the whole queue. An entry only becomes finalizable once the finalized L1 overtakes its L1 origin (~2 L1 epochs, ~13 min). When resets occur more frequently than that (e.g. transient BlockNotFound resets every few minutes), entries never survive long enough, and finalization only advances in multi-hour batches.

Fix:

Remember the highest finalized L1 block number observed and preserve it across clear(). After each enqueue, retry finalization against the stored signal via try_finalize_pending().

On Signal::Reset, drop only the queue entries above the safe head instead of clearing the whole queue. Blocks at or below the safe head survive the reset unchanged, so their entries remain valid.

Neither change relaxes the finality rule: an L2 block is only ever finalized when its derived_from L1 block is itself finalized. The changes only stop already-satisfied finalizations from being lost.

niran:
## Summary
- Retain the latest finalized L1 signal in `L2Finalizer` across derivation resets, matching op-node's `onReset` / `finalizedL1` behavior.
- Retry finalization after the engine confirms rebuilt safe-head data so L2 finality resumes without waiting for a duplicate L1 finality poll.
- Add unit and action-harness regressions for empty-queue signals, reset retention, and finalization without a new L1 signal.

Fixes #3895

## Test plan
- [x] `cargo nextest run -p base-consensus-node -- finalizer`
- [x] `cargo nextest run -p base-action-harness -- finalization`
- [x] `cargo +nightly fmt --all -- --check`
- [x] `RISC0_SKIP_BUILD_KERNELS=1 BASE_SUCCINCT_ELF_STUB=1 cargo clippy -p base-consensus-node --all-features --all-targets -- -D warnings`
- [x] `RISC0_SKIP_BUILD_KERNELS=1 BASE_SUCCINCT_ELF_STUB=1 cargo clippy -p base-action-harness --all-targets -- -D warnings`
- [x] Fresh local Docker/Podman single-sequencer QA on `6584cc930` (`base:latch-l1-finality-6584cc930fa9`), verifier `BASE_NODE_VERIFIER_L1_CONFS=0`

## Devnet QA
- **Result:** PASS (tier 4 fresh full-stack single-sequencer; no HA/ZK/prover)
- **Image:** `base:latch-l1-finality-6584cc930fa9` @ `6584cc930`
- **Trigger observed:** natural tip-adjacent derivation `Reset` storm on `base-client` at startup (`BlockNotFound` / 9× reset signals) with L1 confs=0
- **Outcome:** L2 finalized advanced after those resets (`Updated finalized head` 18 → 38 → 50 → … → 194+; 12 finalized-head updates); builder/client latest hash parity
- **Limitations:** second calm window had no additional Reset (could not hold `finalized_l1` constant); relied on natural startup resets rather than controlled post-steady L1 fault injection; local L1 finality cadence is not Sepolia-identical

type=routine
risk=low
impact=sev5

### DIFF
--- a/actions/harness/src/node.rs
+++ b/actions/harness/src/node.rs
@@ -144,7 +144,7 @@ pub struct TestRollupNode<P: Pipeline + SignalReceiver + Debug + Send = Verifier
     /// Current L2 finalized head.
     finalized_head: L2BlockInfo,
     /// Most recently signalled finalized L1 block number.
-    finalized_l1_number: u64,
+    finalized_l1_number: Option<u64>,
     /// Most recently signalled L1 safe block number.
     safe_l1_number: u64,
     /// History of safe head updates paired with their L1 origin number.
@@ -202,7 +202,7 @@ impl<P: Pipeline + SignalReceiver + Debug + Send> TestRollupNode<P> {
             safe_head,
             unsafe_head: safe_head,
             finalized_head: safe_head,
-            finalized_l1_number: 0,
+            finalized_l1_number: None,
             safe_l1_number: 0,
             safe_head_history: Vec::new(),
             derived_tx_counts: Vec::new(),
@@ -340,17 +340,25 @@ impl<P: Pipeline + SignalReceiver + Debug + Send> TestRollupNode<P> {
         Ok(())
     }
 
-    /// Update the L2 finalized head by scanning safe-head history.
-    ///
-    /// Finds the highest L2 safe block whose L1 origin number is ≤
-    /// `head.number` and promotes it to `finalized_head`.
+    /// Update the retained finalized L1 signal and retry finalization.
     pub async fn act_l1_finalized_signal(&mut self, head: BlockInfo) {
-        self.finalized_l1_number = head.number;
+        if self.finalized_l1_number.is_some_and(|previous| head.number < previous) {
+            return;
+        }
+        self.finalized_l1_number = Some(head.number);
+        self.try_finalize_pending();
+    }
+
+    /// Retry finalization using the retained finalized L1 signal.
+    fn try_finalize_pending(&mut self) {
+        let Some(finalized_l1_number) = self.finalized_l1_number else {
+            return;
+        };
         let candidate = self
             .safe_head_history
             .iter()
             .rev()
-            .find(|(_, l1_origin_number)| *l1_origin_number <= self.finalized_l1_number)
+            .find(|(_, l1_origin_number)| *l1_origin_number <= finalized_l1_number)
             .map(|(l2_info, _)| *l2_info);
         if let Some(l2) = candidate
             && l2.block_info.number > self.finalized_head.block_info.number
@@ -363,7 +371,8 @@ impl<P: Pipeline + SignalReceiver + Debug + Send> TestRollupNode<P> {
     ///
     /// Sends a [`ResetSignal`] to the pipeline, resets all head pointers,
     /// clears all per-block metadata, and replaces the engine with a fresh
-    /// instance that will re-execute from genesis on the new fork.
+    /// instance that will re-execute from genesis on the new fork. The finalized L1 signal is
+    /// retained so re-derived safe heads can be finalized without a duplicate signal.
     pub async fn act_reset(&mut self, l2_safe_head: L2BlockInfo) {
         self.pipeline
             .signal(ResetSignal { l2_safe_head }.signal())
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

niran:
Pushed a CI fix for **Base Std Interface Tests**: after the Reth v2.4 / MSRV 1.95 bump, the Beryl job was still compiling patched Base crates under base-anvil's pinned `rust-toolchain` (1.94.1). The workflow now forces `RUSTUP_TOOLCHAIN` from this checkout for that build.

.github/workflows/base-std-fork-tests.yml
```diff
@@ -103,6 +103,11 @@ jobs:
         run: |
           set -euo pipefail
 
+          # Patched Base crates require Base's MSRV. The pinned base-anvil snapshot may
+          # ship an older rust-toolchain (e.g. 1.94.1); override it for this build.
+          base_toolchain=$(cd "$GITHUB_WORKSPACE" && rustup show active-toolchain | awk '{print $1}')
+          export RUSTUP_TOOLCHAIN="$base_toolchain"
```
refcell: I think the MSRV in base-anvil should be updated, but this works for now 😄

refcell (dismissed):
Makes sense

.github/workflows/base-std-fork-tests.yml
refcell: These jobs are actually erroring since I believe we bumped reth and alloy dependencies while base-anvil hasn't

niran:
Follow-up CI fix for **Base Std Interface Tests**:

The MSRV override alone was not enough. Path-patching this checkout into `base-anvil` `v1.1.1` fails with duplicate `alloy-evm` / `revm-precompile` types after the Reth v2.4 bump (Base is on alloy-evm 0.37; anvil still resolves 0.36).

The workflow now builds the pinned Beryl anvil snapshot against its own `base/base` pin (no path patches) so the historical suite can run again. Path patches should be restored once `base-anvil` tracks Base’s alloy/revm stack.

niran:
Reverted the Base Std fork-test workflow changes out of this PR — that CI break is unrelated to the finalizer fix (Reth v2.4 / alloy-evm skew vs pinned `base-anvil` v1.1.1) and belongs in a dedicated CI/`base-anvil` update.

wlawt (approved):

PR #4118, opened 2026-07-23
test(systems): add upgrade signal system tests with mock ProtocolVersions

PelleKrab:
## Summary

Adds end-to-end system test coverage for the L1 upgrade signal (ProtocolVersions) integration across consensus and execution nodes, backed by a new mutable MockProtocolVersions contract in base-test-utils.

The system test stack gains optional upgrade signal support: the builder deploys and seeds the mock contract on L1, and the in-process nodes read the schedule at startup and poll it live, mirroring the standalone CLIs (fail-closed runtime validation).

## Changes

- MockProtocolVersions (base-test-utils): mutable L1 contract so tests can reschedule upgrades live.
- Stack wiring (etc/systems): UpgradeSignalStackOptions with independent consensus and execution modes (startup-apply, metrics-only, runtime-admin). Both consensus nodes — including follow-mode client consensus — and the in-process execution client consume the configured signal.
- Runtime registry hygiene: runtime-admin overrides are scoped to the owning test stack and cleared on teardown, so a later stack reusing the same chain ID can't inherit stale fork activations.

## Tests

7 integration tests (etc/systems/tests/upgrade_signal.rs) plus 3 unit tests:

- startup-apply activates the L1 schedule on the rollup config served by both consensus nodes (validator and follow mode)
- startup-apply activates the scheduled fork on the execution client chain spec while metrics-only consensus stays untouched
- metrics-only leaves the rollup config and runtime registry untouched
- runtime-admin auto re-applies a live L1 reschedule through the RuntimeUpgradeRegistry (consensus and execution paths)
```
cargo test -p base-system-tests --lib                              # 3 passed
cargo test -p base-system-tests --test upgrade_signal              # 7 passed
```


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

etc/systems/src/smoke.rs
```diff
@@ -22,20 +28,69 @@ use crate::{
...
+/// Exclusive ownership of the process-global [`RuntimeUpgradeRegistry`] entries for one L2
+/// chain ID, cleared when dropped.
+///
+/// A runtime-admin [`SystemTestStack`] acquires this guard before its nodes start so that its
+/// live schedule overrides never outlive the stack and never contaminate a later stack reusing
+/// the same chain ID. Ownership is exclusive: acquiring a chain ID that another live guard
+/// already owns fails, because the registry is shared by every in-process node in the test
+/// binary and two concurrent writers for the same chain cannot be isolated from each other.
+#[derive(Debug)]
+pub struct RuntimeUpgradeSignalGuard {
```
refcell, line 57: This is neat

etc/systems/src/upgrade_signal.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// Explicit `(upgrade, activation timestamp)` entries seeded into the mock contract.
+    ///
+    /// Nodes always read and apply the full contract schedule, and a `0` contract timestamp
+    /// explicitly clears an upgrade. The stack therefore seeds the contract with a baseline
+    /// derived from the deployed rollup config (see [`Self::baseline_schedule`]) and these
+    /// entries override the baseline per upgrade.
+    pub schedule: Vec<(BaseUpgrade, u64)>,
+    /// Minimum protocol version seeded into the mock contract (packed semver, must be nonzero
+    /// while any activation timestamp is positive).
+    pub minimum_protocol_version: U256,
```
refcell, line 40: Are we only doing this as a one-off in system tests because there won't ever be different versions of our client software running locally in system tests?

In production, we'll need to have this mapped from upgrade -> minimum protocol version I believe

etc/systems/tests/upgrade_signal.rs
```diff
@@ -0,0 +1,262 @@
...
+use base_system_tests::{SystemTestStack, SystemTestStackBuilder, UpgradeSignalStackOptions};
+use base_upgrade_signal::UpgradeSignalMode;
+use eyre::{Result, WrapErr};
+use jsonrpsee::http_client::HttpClientBuilder;
+use tokio::time::{sleep, timeout};
+
+/// Initial Cobalt activation timestamp seeded into the mock contract (2100-01-01).
+const COBALT_ACTIVATION_TIMESTAMP: u64 = 4_102_444_800;
+/// Rescheduled Cobalt activation timestamp for live updates (2101-01-01).
+const COBALT_RESCHEDULED_TIMESTAMP: u64 = 4_133_980_800;
```
refcell, line 18: Are these not already defined elsewhere? It feels like we're duplicating timestamps here.

etc/systems/tests/upgrade_signal.rs
```diff
@@ -0,0 +1,262 @@
...
+
+/// Reads the rollup config served by a consensus node's RPC.
+async fn rollup_config_via_rpc(rpc_url: &url::Url) -> Result<RollupConfig> {
+    let client = HttpClientBuilder::default()
+        .build(rpc_url.as_str())
+        .wrap_err("Failed to build consensus RPC client")?;
+    RollupNodeApiClient::rollup_config(&client)
+        .await
+        .wrap_err("Failed to fetch rollup config via RPC")
+}
```
refcell, line 28: This could be a method on `RollupNodeApiClient` something like `RollupNodeApiClient::rollup_config_from_url(...)`

etc/systems/tests/upgrade_signal.rs
```diff
@@ -0,0 +1,262 @@
...
+        }
+    })
+    .await
+    .map_err(|_| {
+        eyre::eyre!(
+            "live L1 Cobalt schedule was not re-applied to the runtime registry within {}s",
+            LIVE_APPLY_TIMEOUT.as_secs()
+        )
+    })
+}
```
refcell: Could this be placed on `SystemTestStack`? Ideally we try not to have free floating methods in test modules to keep test modules test-specific and reduce boilerplate code colocated to tests.

etc/systems/tests/upgrade_signal.rs
```diff
@@ -0,0 +1,262 @@
...
+        None,
+        "dropping a runtime-admin stack should clear its runtime overrides"
+    );
+    Ok(())
+}
+
+/// Runtime-admin mode applies the L1 schedule at startup and automatically re-applies observed
+/// live L1 schedule changes to the process-local runtime registry.
+#[tokio::test]
+async fn test_upgrade_signal_runtime_admin_reapplies_live_schedule_change() -> Result<()> {
```
refcell, line 203: Nice tests here 💯 
Very concise and legible

refcell (commented):
Few comments but looks very solid

etc/systems/src/upgrade_signal.rs
PelleKrab: Yes, we would need to update the pined version in the mock contract for a full production test. Right now it is more or less just checking the logic. After talking with Danyal, we might want to ditch the version all together and have each release have a supported upgrade instead, i.e have `MIN_SUPPORTED_UPGRADE` const. This would be better if there is ever another client for base.

etc/systems/tests/upgrade_signal.rs
PelleKrab: No, these are needed for testing as Cobalt doesn't have timestamps yet. This group of test also relies on the upgrade never happening, since we have nodes configured asymmetrically with metrics, startup, and runtime modes.

A good followup PR could be to have a separate system test where the fork actually triggers from a new dynamic timestamp. We could also refactor to use Beryl and grab the first timestamp from the static upgrade config.

etc/systems/tests/upgrade_signal.rs
PelleKrab: We could, but if we want to keep it as test glue we could also move it to `SystemTestRpcClient` in etc/systems/src/rpc.rs.

refcell: Either way is fine by me, can be a follow on :)

refcell (dismissed):

refcell (approved):

PR #4120, opened 2026-07-23
feat(b20_asset): scaffold AssetV2 at Cobalt

robriks:
## Summary
- introduce the append-only ERC-8056 ABI and packed slot-4 storage layout
- resolve `AssetV2` at Cobalt and gate new selectors to preserve V1's historical unknown-selector behavior
- add erroring defaults and a delegation-only V2 without changing frozen `logic/v1.rs`

This is PR 1 of 2. The ERC-8056 behavior is intentionally deferred to stacked PR #4121.

## Test plan
- [x] `cargo test -p base-common-precompiles`
- [x] `cargo clippy -p base-common-precompiles --all-targets`
- [x] `cargo fmt -p base-common-precompiles -- --check`
- [x] `git diff origin/main -- crates/common/precompiles/src/b20_asset/logic/v1.rs` is empty
- [x] V1 rejects the new selectors with their historical `UnknownFunctionSelector`
- [x] Cobalt routes the new selectors to the erroring V2 defaults


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompile-macros/src/accounting.rs
```diff
@@ -323,6 +323,28 @@ fn expand_asset(input: DeriveInput) -> syn::Result<TokenStream> {
...
+            }
+
+            fn set_pending(
+                &mut self,
+                multiplier: u128,
+                effective_at: u64,
+            ) -> ::base_precompile_storage::Result<()> {
+                self.asset.set_pending_multiplier(multiplier)?;
+                self.asset.set_pending_effective_at(effective_at)
+            }
```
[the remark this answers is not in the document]
robriks: the two sstores are on a single call frame, so any Err reverts the whole precompile call and both writes roll back atomically

crates/common/precompiles/src/b20_asset/accounting.rs
```diff
@@ -17,6 +17,26 @@ pub trait AssetAccounting: TokenAccounting {
     /// Writes a new multiplier.
     fn set_multiplier(&mut self, multiplier: U256) -> Result<()>;
 
+    /// Returns the pending scheduled multiplier target (ERC-8056)
+    ///
+    /// Packs with [`Self::pending_effective_at`] into a single slot. Introduced with the
+    /// scheduled-multiplier feature (`AssetV2`); earlier versions never touch this slot.
+    fn pending_multiplier(&self) -> Result<u128> {
+        Err(BasePrecompileError::UnknownFunctionSelector([0u8; 4]))
```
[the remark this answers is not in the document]
robriks: yes if the scenario described was hit, this would break solidity's function selector computation and thus be a hard fork from the EVM. all functions *must* be overridden by inheritors; shipping a precompile with default behavior like this should be impossible

crates/common/precompile-macros/src/accounting.rs
```diff
@@ -323,6 +323,28 @@ fn expand_asset(input: DeriveInput) -> syn::Result<TokenStream> {
...
 
+            fn pending_multiplier(&self) -> ::base_precompile_storage::Result<u128> {
+                self.asset.pending_multiplier()
+            }
+
+            fn pending_effective_at(&self) -> ::base_precompile_storage::Result<u64> {
+                self.asset.pending_effective_at()
+            }
+
+            fn set_pending(
```
rayyan224: I think this should be named set_pending_and_effective_at()

crates/common/precompile-macros/src/accounting.rs
```diff
@@ -323,6 +323,28 @@ fn expand_asset(input: DeriveInput) -> syn::Result<TokenStream> {
...
+                &mut self,
+                multiplier: u128,
+                effective_at: u64,
+            ) -> ::base_precompile_storage::Result<()> {
+                self.asset.set_pending_multiplier(multiplier)?;
+                self.asset.set_pending_effective_at(effective_at)
+            }
+
+            fn clear_pending(&mut self) -> ::base_precompile_storage::Result<()> {
```
rayyan224: clear_pending_multiplier_and_effective_at()

crates/common/precompiles/src/b20_asset/logic/v2.rs
```diff
@@ -0,0 +1,304 @@
...
+impl<S: AssetAccounting, A: PolicyAccounting> Asset<S, A> for AssetV2 {
+    fn transfer(
+        &self,
+        token: &mut B20AssetToken<S, A>,
+        caller: Address,
+        to: Address,
+        amount: U256,
+        privileged: bool,
+    ) -> Result<()> {
+        AssetV1.transfer(token, caller, to, amount, privileged)
```
rayyan224: Originally the thought was LogicV2 dupplicates the logic ? What are your thoughts ?

crates/common/precompiles/src/b20_asset/dispatch.rs
```diff
@@ -26,6 +26,22 @@ use crate::{
     macros::decode_precompile_call,
 };
 
+/// Selectors introduced with the ERC-8056 scheduled-multiplier surface at `AssetV2` (Cobalt).
+///
+/// Versions that do not advertise scheduling ([`AssetVersion::supports_scheduled_multiplier`] is
+/// false) must treat these as unknown so they fall through to the same `UnknownFunctionSelector`
+/// path as before the shared ABI enum grew to preserve byte-identical Beryl (V1) behavior.
+const SCHEDULED_MULTIPLIER_SELECTORS: [[u8; 4]; 8] = [
```
rayyan224: Hey, I think this UnknownFunctionSelector should be in line with the matching logic, and throwing at that level. Scaling it this way would get complicated.

crates/common/precompiles/src/b20_asset/dispatch.rs
```diff
@@ -346,11 +365,21 @@ impl<S: AssetAccounting, A: PolicyAccounting> B20AssetToken<S, A> {
...
+            SC::uiMultiplier(_) => logic.ui_multiplier(self)?.abi_encode().into(),
+            SC::newUIMultiplier(_) => logic.new_ui_multiplier(self)?.abi_encode().into(),
+            SC::effectiveAt(_) => logic.effective_at(self)?.abi_encode().into(),
             SC::toScaledBalance(c) => {
                 logic.to_scaled_balance(self, c.rawBalance)?.abi_encode().into()
             }
             SC::toRawBalance(c) => logic.to_raw_balance(self, c.scaledBalance)?.abi_encode().into(),
             SC::scaledBalanceOf(c) => logic.scaled_balance_of(self, c.account)?.abi_encode().into(),
+            SC::balanceOfUI(c) => logic.balance_of_ui(self, c.account)?.abi_encode().into(),
+            SC::totalSupplyUI(_) => logic.total_supply_ui(self)?.abi_encode().into(),
```
rayyan224, line 376: for exmple for above comment 
`if (fork < COBAL) throw else logic.balance_of_ui(self, c.account)?.abi_encode().into()`

crates/common/precompiles/src/b20_asset/dispatch.rs
rayyan224: Ideally only worth grouping if we want very spefici logic given these selctors, in this case we don't. The logic is just hard fork gating

crates/common/precompiles/src/b20_asset/logic/v2.rs
rayyan224: I'm not super against this version also

rayyan224: @stephancill @meyer9 Thoughts ?

crates/common/precompiles/src/b20_asset/storage.rs
```diff
@@ -23,6 +23,14 @@ pub struct B20AssetExtensionStorage {
     pub used_announcement_ids: Mapping<String, bool>, // slot 2
     /// Extra metadata values by metadata key.
     pub extra_metadata: Mapping<String, String>, // slot 3
+    /// Pending scheduled multiplier target (ERC-8056). Introduced with `AssetV2`
+    #[accessor]
```
rayyan224, line 27: btw we are storing these in the same slot, there could be an optimization in the accounting module to write this slot in 1 ? thoughts ?

crates/common/precompiles/src/b20_asset/versions.rs
```diff
@@ -28,10 +30,20 @@ impl AssetVersion {
...
     {
         static V1: AssetV1 = AssetV1;
+        static V2: AssetV2 = AssetV2;
         match self {
             Self::V1 => &V1,
+            Self::V2 => &V2,
         }
     }
+
+    /// Whether this version advertises the ERC-8056 scheduled-multiplier surface
```
rayyan224: It seems like this is here for throwing on the dispatch, if we move them inline we can remove this, and simplify this gating logic

crates/common/precompiles/src/b20_asset/versions.rs
```diff
@@ -44,8 +56,16 @@ pub struct AssetVersions;
 impl AssetVersions {
     /// Returns the version active at `upgrade`, or `None` before the introduction
     /// fork (Beryl), where the asset precompile is not installed at all.
+    ///
+    /// V1 is active from Beryl; V2 supersedes it from Cobalt.
     pub fn from_base_upgrade(upgrade: BaseUpgrade) -> Option<AssetVersion> {
-        if upgrade >= BaseUpgrade::Beryl { Some(AssetVersion::V1) } else { None }
```
rayyan224, line 48: let's use a match statement here ?

crates/common/precompiles/tests/b20_asset_v1_golden.rs
```diff
@@ -245,7 +245,30 @@ fn domain_separator(storage: &mut HashMapStorageProvider) -> B256 {
 fn resolver_maps_forks_to_versions() {
     assert_eq!(AssetVersions::from_base_upgrade(BaseUpgrade::Azul), None);
     assert_eq!(AssetVersions::from_base_upgrade(BaseUpgrade::Beryl), Some(AssetVersion::V1));
-    assert_eq!(AssetVersions::from_base_upgrade(BaseUpgrade::Cobalt), Some(AssetVersion::V1));
+    assert_eq!(AssetVersions::from_base_upgrade(BaseUpgrade::Cobalt), Some(AssetVersion::V2));
+}
+
+#[test]
```
rayyan224, line 251: Nice!

crates/common/precompiles/src/b20_asset/logic/v2.rs
```diff
@@ -0,0 +1,304 @@
...
+use crate::{
+    Asset, AssetAccounting, AssetV1, B20AssetToken, Eip712Domain, IB20, PermitArgs,
+    PolicyAccounting,
+};
+
+/// Second B-20 Asset precompile implementation, introduced at Cobalt.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct AssetV2;
+
+impl<S: AssetAccounting, A: PolicyAccounting> Asset<S, A> for AssetV2 {
```
refcell, line 204: This is a lot of duplicate logic/scaffold code that's a good candidate to be refactored. v2.rs is 304 lines of pure AssetV1.foo(...) forwarding with zero divergence in this PR. I think there are a few ways we could go about this:

1. Hoist V1 logic into the `Asset` trait default methods, leaving `AssetV1`/`AssetV2` as near-empty, overriding only what changes. This is the smallest diff, but it conflicts with the frozen-per-version model. Moving the mutations into shared defaults means a future V3 editing a default silently mutates V1's historical behavior. It also strands V1's private helpers (`transfer_inner`, `*_unchecked`, `ensure_*`, `OPERATOR_ROLE`), and `grant_role_unchecked` is deliberately kept off the trait. I don't think we want this one.
2. A thin delegation macro. e.g. `delegate_asset!(AssetV2 => AssetV1, overrides { … })` that emits the forwarding impl and takes an explicit override list. Collapses this PR to ~one line, keeps each version a distinct frozen impl, and makes V2-specific logic the only thing in the `v2.rs` module. We already lean on precompile-macros, so we're not introducing any new crates with this. The main downside is it obfuscates the selector surface, which makes it a bit more involved to trace.
3. Leave it explicit. The freeze is already enforced by the golden suite's coverage checklist, so the boilerplate is buying explicitness, not correctness.

My preference would be with #2 to keep precompile code concise.

crates/common/precompiles/src/b20_asset/abi.rs
```diff
@@ -127,9 +201,45 @@ impl IB20Asset::IB20AssetCalls {
...
+            for (a, s) in acc.iter_mut().zip(selector) {
+                *a ^= *s;
+            }
+        }
+        FixedBytes::new(acc)
+    }
 
-    use crate::{IB20, IB20Asset};
+    #[test]
+    fn erc8056_interface_ids_match_selectors() {
```
refcell, line 215: TY for testing these

crates/common/precompiles/src/b20_asset/abi.rs
```diff
@@ -3,8 +3,23 @@
 //! [`IB20Asset`] defines only the asset-specific surface.
 //! All inherited selectors come from [`crate::IB20`] defined in `b20/abi.rs`.
 
+use alloy_primitives::FixedBytes;
 use alloy_sol_types::sol;
 
+/// ERC-165 plus the three ERC-8056 interface IDs advertised by `supportsInterface` from `AssetV2`.
```
refcell: I'm not sure I like grouping erc-165 in with erc-8056 and then calling the list `ERC8056_INTERFACE_IDS`

crates/common/precompiles/src/b20_asset/accounting.rs
```diff
@@ -17,6 +17,26 @@ pub trait AssetAccounting: TokenAccounting {
...
+    }
+    /// Returns the timestamp at which the pending multiplier becomes effective; `0` means none.
+    fn pending_effective_at(&self) -> Result<u64> {
+        Err(BasePrecompileError::UnknownFunctionSelector([0u8; 4]))
```
refcell: Something we could do here to make it clear why we are returning unknown function selector is create a macro that expands to `Err(BasePrecompileError::UnknownFunctionSelector([0u8; 4]))`.

Something like:

```rust
/// Rejects a call as an unknown selector, freezing the observable behavior of any version that
/// predates it.
///
/// These live as **shared trait defaults**, so every `Asset` / `AssetAccounting` version inherits
/// them until it explicitly overrides. That is the point: a version introduced *before* a selector
/// existed must reject it exactly as it did at its activation fork. Returning
/// [`BasePrecompileError::UnknownFunctionSelector`] — rather than an `Ok` or a real value —
/// guarantees that re-executing a historical block on a newer binary produces the same result the
/// chain already committed. Any other default would retroactively change an older version's state
/// transition: a silent hardfork. Locked down by `golden_v2_selectors_unknown_at_v1`.
///
/// The `[0u8; 4]` is a placeholder; the dispatcher already validated and logged the real selector
/// before routing here.
macro_rules! reject_frozen_selector {
    () => {
        ::core::result::Result::Err(
            ::base_precompile_storage::BasePrecompileError::UnknownFunctionSelector([0u8; 4]),
        )
    };
}
```

refcell (commented):
Left a couple comments. Overall looks good

crates/common/precompiles/src/b20_asset/logic/v2.rs
refcell: I personally do not see a reason to duplicate the logic. If `v1.rs` is "frozen" and V2 contains the same verbatim logic as v1, duplication buys us nothing but more risk of changing behavior with a typo or whatnot. If anything, I'd suggest even less LOC in this file by using a macro that handles the delegation (see my comment below).

crates/common/precompiles/src/b20_asset/abi.rs
robriks: I agree, this was only grouped that way because erc-165 is an explicit requirement of erc-8056. I can separate it out

crates/common/precompiles/src/b20_asset/logic/v2.rs
robriks: I've implemented what I believe you mean by #2 in commit 5d223f9c1, have a look to make sure I captured your intent

in the past I've avoided macros since my main language is solidity and don't always find macros to be super intuitive, but I like this approach a lot!

crates/common/precompiles/src/b20_asset/dispatch.rs
robriks: makes sense. I moved the hard fork gating logic inline (commit 715d6daf0), ensuring it happens pre-decoding so that we don't break parity with Beryl (post-decoding would allow `AbiDecodeFailed` to be thrown before reaching `UnknownFunctionSelector` in the match)

crates/common/precompiles/src/b20_asset/logic/v2.rs
robriks: went with the `delegate_asset` macro approach, let me know what you guys think!

crates/common/precompiles/src/b20_asset/versions.rs
robriks: makes sense, removed as part of inline migration in 715d6daf0

crates/common/precompiles/src/b20_asset/accounting.rs
robriks: thanks, this is a great pattern to establish for precompile interfaces going forward! implemented in 53c5259bc

crates/common/precompiles/src/b20_asset/storage.rs
robriks: yes, good callout! optimized down to a single SSTORE (rather than 2 as before) in 02b47dd28

crates/common/precompile-macros/src/accounting.rs
robriks: you got it! 548e32787

crates/common/precompile-macros/src/accounting.rs
robriks: you got it! 548e32787

crates/common/precompiles/src/b20_asset/abi.rs
robriks: separated out into `ERC165_INTERFACE_ID` in 50889d27d

crates/common/precompiles/src/b20_asset/versions.rs
robriks: sounds good.  switched to a match in ba2bf3df3 

One nuance worth calling out: I used ordered guards rather than explicit per-variant arms, because:
- `BaseUpgrade` is `#[non_exhaustive]`, so a wildcard arm is mandatory
- A fork newer than Cobalt must inherit the latest version (V2) until one supersedes it.
  With per-variant arms the wildcard would be `_ => None`, which would treat a
  future fork as "precompile not installed." The new `>=` guards preserve the "inherit-latest" behavior

crates/common/precompiles/src/b20_asset/dispatch.rs
```diff
@@ -107,6 +107,24 @@ impl<S: AssetAccounting, A: PolicyAccounting> B20AssetToken<S, A> {
...
+                    || selector == IB20Asset::newUIMultiplierCall::SELECTOR
+                    || selector == IB20Asset::effectiveAtCall::SELECTOR
+                    || selector == IB20Asset::balanceOfUICall::SELECTOR
+                    || selector == IB20Asset::totalSupplyUICall::SELECTOR
+                    || selector == IB20Asset::setUIMultiplierCall::SELECTOR
+                    || selector == IB20Asset::cancelScheduledMultiplierCall::SELECTOR
+                    || selector == IB20Asset::supportsInterfaceCall::SELECTOR)
+            {
+                return Err(BasePrecompileError::UnknownFunctionSelector(selector));
+            }
```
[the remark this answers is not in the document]
robriks: The mentioned drift risk is fair, but this has already been discussed- with the resulting decision that an array of selectors as suggested has already been deleted and is *not* desirable

crates/common/precompiles/src/b20_asset/logic/mod.rs
```diff
@@ -15,6 +15,394 @@ pub use interface::Asset;
...
+/// it to the version it extends. This macro generates that forwarding from a single method-name
+/// list, so a version's module contains only the methods whose behavior actually diverges.
+///
+/// `delegate_asset!(NewVersion => prior, { method_a, method_b, ... })` forwards each named method
+/// to `prior` (a unit-struct value such as [`AssetV1`]). A method the new version overrides is
+/// omitted from the list and given its real implementation instead. Rust allows only one
+/// `impl Asset for` a given type, so an override cannot live in a second `impl` block; the macro
+/// gains an `overrides { ... }` section (spliced into this generated `impl`) in the follow-up PR
+/// that introduces the first diverging version.
+macro_rules! delegate_asset {
```
rayyan224: Hmmmm, for each new function we need to update the macro ? It seems this way call stack may grow linearly on the hot path 🤔 ? Could this be more generalized to copy the code for the delegation, since macros generate on build.

crates/common/precompiles/src/b20_asset/mod.rs
```diff
@@ -1,7 +1,29 @@
...
+/// activation fork. Returning `UnknownFunctionSelector` rather than an `Ok` or a real value keeps
+/// re-executing a historical block on a newer binary byte-identical to what the chain already
+/// committed; any other default would retroactively change an older version's state transition, i.e.
+/// a silent hardfork.
+///
+/// The `[0u8; 4]` is a placeholder: the dispatcher rejects these selectors on the raw calldata,
+/// before any version method is reached (see the fork gate in `route`), so a reachable call never
+/// actually returns this value. Pinned by `golden_v2_selectors_unknown_at_v1`.
+macro_rules! reject_frozen_selector {
+    () => {
```
rayyan224, line 18: Let's move to common this is going to be used throught all new versions

crates/common/precompiles/src/b20_asset/logic/mod.rs
rayyan224: Not blocking though, we can think on the best approach here

crates/common/precompiles/src/b20_asset/abi.rs
```diff
@@ -3,8 +3,23 @@
 //! [`IB20Asset`] defines only the asset-specific surface.
 //! All inherited selectors come from [`crate::IB20`] defined in `b20/abi.rs`.
 
+use alloy_primitives::FixedBytes;
 use alloy_sol_types::sol;
 
+/// The ERC-165 `supportsInterface` interface id (`IERC165`, `0x01ffc9a7`).
```
rayyan224: Can we shorten these coments seems over kill

crates/common/precompiles/src/b20_asset/dispatch.rs
rayyan224, line 127: Hey, question is there a reason this can't be inline with the dispatching loic for the selector itself, so instead of this:
`  SC::uiMultiplier(_) => logic.ui_multiplier(self)?.abi_encode().into(),` 
we would have 
```SC::uiMultiplier(_) => if (version <  AssetVersion::V2) Err(BasePrecompileError::UnknownFunctionSelector(selector)) else logic.ui_multiplier(self)?.abi_encode().into(),```

Comment (rayyan224) [reply crates/common/precompiles/src/b20_asset/dispatch.rs:127]: At this dispathing logic 
https://github.com/base/base/pull/4120/changes#r3638933775

File: crates/common/precompiles/src/b20_asset/abi.rs [hunk 20]
```diff
@@ -16,9 +31,22 @@ sol! {
         /// `updateExtraMetadata` was called with an empty metadata key.
         error InvalidMetadataKey();
 
-        /// `updateMultiplier` was called with a zero multiplier.
+        /// A multiplier setter (`setUIMultiplier` / `updateMultiplier`) was called with a
+        /// multiplier of zero or above the `type(uint128).max` overflow guard.
```
rayyan224, line 32: Can we add a devnote added for V2 for all new features we add. Just a small tag infront

crates/common/precompiles/src/b20_asset/logic/v2.rs
stephancill: @refcell fwiw the intention with this architecture was to be more explicit. the plan is to see how the cobalt changes shape up and then identify opportunities to refactor if necessary once we have a clearer picture

crates/common/precompiles/src/b20_asset/logic/mod.rs
robriks: resolving after discussion on slack where we confirmed call stack doesn't grow linearly

crates/common/precompiles/src/b20_asset/abi.rs
robriks: Added a [V2] tag to each new ERC-8056 error/event/function (d97cb7f27). Left `updateMultiplier/MultiplierUpdated` untagged since they're V1 items whose V2 behavior is already noted inline. Say the word if you want those marked too !

crates/common/precompiles/src/b20_asset/dispatch.rs
robriks: yes, the reason for pre-decode is: 
`handle_asset_call` runs after `abi_decode_validate`, so a pre-Cobalt call with a new selector + malformed args would hit `AbiDecodeFailed` before any arm. historically in Beryl `UnknownFunctionSelector` would have been hit. this is diverging node execution during Beryl blocks once cobalt lands (hardfork). 

Happy to make the raw-selector list a matches! if the || chain reads poorly, but placement needs to stay pre-decode to avoid a silent hard fork

crates/common/precompiles/src/b20_asset/abi.rs
rayyan224: I like just tagging new ones thanks!

rayyan224 (dismissed):

crates/common/precompiles/src/b20_asset/mod.rs
robriks: Moved delegate_asset! into the shared macros.rs (cccf359ea) so it has a common home alongside decode_precompile_call

crates/common/precompiles/src/b20_asset/logic/v2.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    fn update_multiplier(
+        &self,
+        token: &mut B20AssetToken<S, A>,
+        caller: Address,
+        new_multiplier: U256,
+        privileged: bool,
+    ) -> Result<()> {
+        AssetV1.update_multiplier(token, caller, new_multiplier, privileged)
```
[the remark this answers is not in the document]
robriks: follow up

crates/common/precompiles/src/b20_asset/logic/v2.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    fn approve(
+        &self,
+        token: &mut B20AssetToken<S, A>,
+        caller: Address,
+        spender: Address,
+        amount: U256,
+    ) -> Result<()> {
+        AssetV1.approve(token, caller, spender, amount)
```
rayyan224: @robriks Hey check this one maybe for implementation https://github.com/base/base/blob/main/crates/common/precompiles/src/policy/logic/v2.rs

crates/common/precompiles/src/b20_asset/logic/v2.rs
robriks: it was decided that the `delegate_asset` macro pattern should be reverted [6818ce5](https://github.com/base/base/pull/4120/commits/6818ce552014cc2f953318b2fadb89b44f24543d)

crates/common/precompiles/src/b20_asset/logic/v2.rs
robriks: [6818ce5](https://github.com/base/base/pull/4120/commits/6818ce552014cc2f953318b2fadb89b44f24543d)

crates/common/precompiles/src/b20_asset/logic/v2.rs
robriks: [ba1d73a](https://github.com/base/base/pull/4120/commits/ba1d73a76a801df26e5eb746c4d9450924f754ff)

rayyan224 (approved):

PR #4065, opened 2026-07-20
feat(load-tests): open-loop load generation + B20/EVM payloads (umbrella)

meyer9:
## Summary

Umbrella PR for the **open-loop load-testing feature** used to benchmark builder throughput (ETH vs. B20 token transfers). It is intentionally large and **will be split into smaller reviewable PRs** — opening as a draft to surface the full feature set for discussion first.

## What's included

- **Open-loop load generation** (`run_open_loop`): all transactions are pre-signed and fired without waiting for confirmations, so the builder is what's measured, not client submission speed. The measurement clock is **reset after the pre-signing phase** so the (potentially long) pre-sign step is excluded from the run window.
- **B20 payloads**: native-precompile and deployed-EVM-bytecode (`B20EvmTransferPayload`) token-transfer workloads, plus **fresh-recipient** support (send to brand-new vs. existing accounts).
- **Precompile-exclude env toggle** (`crates/common/precompiles`): run the same token via the native precompile or its EVM bytecode.
- **32MB flashblock decode cap** (`crates/common/flashblocks`).

## Planned split

1. Open-loop mode + pre-sign measurement-clock fix
2. B20 native + EVM-bytecode payloads
3. Fresh-recipient support
4. Precompile-exclude toggle
5. Flashblock decode cap

## Notes

- Ported onto `main`; builds (`cargo build --release --bin base-load-tester`) and `cargo clippy -p base-load-tests` are clean.
- Companion base-bench changes (benchmark harness) are in base/benchmark#204 (report) and base/benchmark#205 (saturated-window GPS metric).
- The `state-populate` state-seeding tool used alongside this is already up as #4045.

PR #4133, opened 2026-07-23
feat(load-tests): open-loop load generation, adaptive pacing, and B20 payloads

meyer9:
## Summary

Open-loop load-testing feature for benchmarking builder throughput (ETH vs. B20 token transfers), split out of the umbrella #4065 with two additional pieces that hadn't landed there yet.

## What's included

- **Single execution mode (open-loop only)**: the load tester now runs exclusively in the adaptive open-loop path. The old closed-loop fixed-rate path and its `RateLimiter`-driven submission loop were removed, and there is no longer an `open_loop` config switch.
- **Adaptive margin-based pacing**: `OpenLoopHeadroomTarget` is the pacing controller, using an EWMA-smoothed confirmed-TPS estimator that continuously retargets in-flight headroom.
- **Optional gas-rate safety cap (`max_target_gps`)**: renamed from `target_gps` and repurposed. It no longer drives a fixed-rate sleeper. Instead, when set, it clamps the adaptive in-flight target so implied gas throughput does not exceed the cap; when omitted, pacing is fully adaptive/unbounded and gated only by confirmation backpressure.
- **Duration-based only (count-based mode removed)**: the run is always duration-bounded now (`duration` config field). The earlier `prefill_per_sender` fixed-transaction-count mode was removed entirely — throughput measurement is the only use case this tool needs, and the count-based path added complexity (a second termination condition threaded through the producer, progress tracking, and every log site) without enough value to justify keeping both modes.
- **Outstanding-aware pacing gate fix**: the in-flight headroom gate previously only counted RPC-accepted transactions, ignoring transactions already handed to the submission pipeline but not yet accepted. This let the configured target overshoot by up to 2.4x before the gate engaged. Added `total_outstanding()` (in_flight + queued) and used it throughout the gate.
- **B20 payload**: native-precompile token-transfer workload, using a fixed 1-wei transfer amount to avoid balance exhaustion during long unbounded runs.
- **Fresh-recipient support**: `fresh_recipient_ratio` for account-create fan-out workloads. The recipient key stream is always positioned with a runtime-random seed/offset (never the deterministic `config.seed`), so repeated runs never regenerate the same "fresh" addresses, regardless of the configured ratio. The runner logs `randomized_recipient_seed`/`randomized_recipient_offset` at startup for out-of-band recovery.
- **`skip_drain`** (CLI flag + `TestConfig`/YAML field): skip returning leftover ETH to the funder after a run, for disposable benchmark accounts that get funded fresh each time (useful for harness-driven runs via `base-bench` that only pass a config path).
- **Configurable `funding_batch_size`**: the funding/setup-phase submission batch size (previously a hardcoded `16`) is now a YAML-configurable `TestConfig`/`LoadConfig` field, defaulting to 16, so it can be tuned against different target txpool per-sender slot limits.
- Minor: `sort_by_key(Reverse)` cleanup in the metrics aggregator's top-failure-reasons sort.

## Relationship to other PRs

- Supersedes the load-testing portion of the #4065 umbrella (open-loop mode, B20 payload, fresh-recipient support), plus adds adaptive pacing and configurable funding-batch-size that hadn't been ported into #4065 yet.
- Does **not** include #4065's `B20EvmTransferPayload` (deployed-EVM-bytecode variant), precompile-exclude-via-env-var toggle, or the 32MiB flashblock decode cap — the EVM-bytecode payload was removed as no longer needed, and the exclude toggle / flashblock cap are temporary benchmarking workarounds that will get a proper configurable design in a separate follow-up PR.
- Companion state-populate tool is #4045 (separate).

## Notes

- Builds and tests clean: `cargo build/test/clippy -p base-load-tests -p base-load-tester-bin`, `cargo +nightly fmt --check`.
- Rebased onto current `main`.
- The `Base Std Interface Tests` / `Base Std Interface Tests (Beryl)` checks fail because this PR is opened from a fork; that workflow requires `secrets.BASE_STD_TOKEN` to clone `base/base-std`, which GitHub withholds from `pull_request`-triggered workflows on cross-repo PRs. Unrelated to this diff — same-repo PRs pass this check normally.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

meyer9:
Closed in favor of #4156

PR #4152, opened 2026-07-24
test(system): fuzz the batch-encode/derive span-batch round-trip

montycheese:
## Description

Covers the second chain-halting **agreement boundary** in-process: **batch-encode vs. derive**. The batcher encodes L2 blocks into a span batch; the derivation pipeline must reconstruct exactly those blocks. A mismatch splits the chain.

`fuzz_derive_roundtrip` drives Base's own span-batch codec (`base_protocol`) end to end, fully in-process (no L1, no devnet):

1. fuzzed user transactions — a mix of 1559/2930 and **EIP-8130** — are grouped into blocks;
2. appended into a `SpanBatch` the way the batcher does (`append_singular_batch`);
3. serialized (`to_raw_span_batch` → `encode`);
4. read back the way derivation does (`RawSpanBatch::decode` → `derive`).

The reconstructed blocks must match the originals: **transactions byte-for-byte**, plus epoch and timestamp. Deposits are excluded, as they are never carried in span batches.

## Notes

- This exercises the real reconstruction logic — span-batch transaction decomposition and the `full_txs` rebuild, including Base's EIP-8130 transaction data — not a trivial encode/decode inverse.
- Pure codec, ~0.07s, so it adds no meaningful cost to the per-PR gate (which already targets `binary(fuzz_build_validate)`).
- Reuses the suite's existing transaction generators and signing helpers.

Stacked on the EIP-8130 PR; base will retarget as the stack merges down to `main`.

### DIFF
--- a/etc/systems/tests/fuzz_build_validate.rs
+++ b/etc/systems/tests/fuzz_build_validate.rs
@@ -13,6 +13,12 @@
 //! implementation invariants (gas within limit), and a second test asserts
 //! determinism: the same seed builds byte-identical blocks twice.
 //!
+//! Alongside build-vs-validate, `fuzz_derive_roundtrip` covers the other chain-
+//! halting agreement boundary: batch-encode vs. derive. It runs fuzzed
+//! transactions through Base's own span-batch codec (encode as the batcher does,
+//! read back as the derivation pipeline does) and requires the reconstructed
+//! blocks to match byte-for-byte.
+//!
 //! Runs fully in-process (no devnet, no Docker), forces exactly the generated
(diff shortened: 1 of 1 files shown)


etc/systems/tests/fuzz_build_validate.rs
refcell, line 1: We should be able to do write these as action tests which are lighter weight and run in CI whereas system tests spin up components in docker, are heavier, and don't run in ci.

refcell: Action tests live in [the actions directory](https://github.com/base/base/tree/main/actions).

PR #4168, opened 2026-07-27
refactor(BOP-441): extract move_balance from AssetV2 + StablecoinV2 transfer_inner

stephancill:
## Summary

Behavior-preserving refactor that extracts the balance debit/credit + `Transfer` emission out of `transfer_inner` into a shared `move_balance` primitive in **both** `AssetV2` and `StablecoinV2`. `transfer_inner` keeps the zero-address and sender/receiver policy checks and delegates the move.

This is **PR1** of the Transfer Blocked / seize work ([BOP-441](https://linear.app/coinbase/issue/BOP-441)); it lands the shared mover the transfer-based seize entrypoint (`transferFromBlockedWithMemo`, a later PR) will reuse — so seize never goes through the policy-bearing `transfer_inner` nor the factory `privileged` bypass. Mirrors the base-std reference ([base-std#177](https://github.com/base/base-std/pull/177)).

## Stacked PR

Stacked on **#4170** (`StablecoinV2` scaffold), which this needs so the stablecoin extraction has a mutable V2 to live in. Review/merge #4170 first; this PR's diff is just the two `move_balance` extractions.

## Scope

- Asset **V2** and stablecoin **V2** logic only. V1 is frozen and untouched.
- No functional/ABI change; no new selectors.

## Testing

- `cargo test -p base-common-precompiles --features test-utils`: 620 unit + all goldens pass.
- `cargo clippy --all-targets --features test-utils` clean; `cargo +nightly fmt --check` clean.

PR #4186, opened 2026-07-28
feat(BOP-441): add seize plumbing (scope, role, event, storage, guard)

stephancill:
## Summary

PR3 of the seize stack (BOP-441). Adds the **additive base/base plumbing** for the seize operation class, mirroring the base-std reference ([base-std#177](https://github.com/base/base-std/pull/177)). This is inert plumbing: **no new selectors, no dispatch arms, and no change to any existing selector's behavior.** The entrypoints, ABI getters, `SEIZE` pause vector, and per-version gating land in PR4 (the first PR where seize is callable).

Was stacked on #4168 (`move_balance`), now merged; this branch is rebased onto `main`.

## What's included

- **Policy scope** (`policy_type.rs`): `SEIZABLE_ACCOUNT_POLICY` constant + `B20PolicyType::SeizableAccount`, wired through `from_id`/`id`.
- **Role** (`roles.rs`): `TRANSFER_FROM_SEIZABLE_ROLE` constant + `B20TokenRole::TransferFromSeizable` + `RoleManaged::transfer_from_seizable_role()` accessor.
- **Event** (`abi.rs`): `TransferredFromSeizable(caller, from, to, amount)` (event only — adds no call-enum variant).
- **Storage** (`core_storage.rs`): `seizable_policy_id` at **offset 14** (right after `nonces`). base-std keeps its mock-only `initialized` flag trailing at offset 15; this impl stores no such flag (factory-init is checked via deployed marker bytecode), so it omits that trailing slot and seize lands at offset 14 on both sides — no filler.
- **Accounting macro** (`precompile-macros`): wire the `SeizableAccount` scope to the new storage field in the generated `policy_id`/`set_policy_id`.
- **Guard** (`guards.rs`): `ensure_seizable` — mirror of `ensure_blocked` on the seizable scope (`!isAuthorized` semantics).

## Design notes

- **Storage offsets are stable.** Offsets 0–13 are untouched; `seizable_policy_id` is appended at offset 14, so already-deployed tokens are undisturbed. An offset-pin test asserts offset 14 and the byte offsets within the slot.
- **Keccak parity.** New tests assert the scope/role ids equal `keccak256` of the exact strings base-std uses, so the two implementations cannot silently diverge.
- **Versioning.** The seizable scope is recognized globally (matching the base-std reference); per-version gating (so frozen V1 keeps rejecting it while V2 accepts it) is folded into the in-flight ABI-versioning work and PR4.

## Deferred to PR4

`transferFromSeizableWithMemo` / `burnBlockedWithMemo` entrypoints + dispatch arms + logic; the `TRANSFER_FROM_SEIZABLE_ROLE()` / `SEIZABLE_ACCOUNT_POLICY()` getters; the `SEIZE` `PausableFeature` member; golden-matrix coverage. Those all add shared call-enum variants / change arg-enum decode and require the version gating, so they belong with the callable surface.

## Testing

- `cargo test -p base-common-precompiles --features test-utils`: all pass, incl. frozen V1 goldens (asset 102, stablecoin 80) unchanged.
- New unit tests: keccak parity (scope + role), storage-offset pin, `ensure_seizable` behavior (authorized => not seizable; independent of transfer-sender policy).
- `cargo +nightly fmt --check` and `cargo clippy --all-targets --features test-utils` clean.

### DIFF
--- a/crates/common/precompile-macros/src/accounting.rs
+++ b/crates/common/precompile-macros/src/accounting.rs
@@ -251,6 +251,7 @@ fn expand_token(input: DeriveInput) -> syn::Result<TokenStream> {
                         self.b20.transfer_executor_policy_id()
                     }
                     crate::B20PolicyType::MintReceiver => self.b20.mint_receiver_policy_id(),
+                    crate::B20PolicyType::Seizable => self.b20.seizable_policy_id(),
                 }
             }
 
@@ -271,6 +272,9 @@ fn expand_token(input: DeriveInput) -> syn::Result<TokenStream> {
                     }
                     crate::B20PolicyType::MintReceiver => {
                         self.b20.set_mint_receiver_policy_id(policy_id)
+                    }
+                    crate::B20PolicyType::Seizable => {
+                        self.b20.set_seizable_policy_id(policy_id)
                     }
                 }
             }
--- a/crates/common/precompile-macros/src/accounting.rs
+++ b/crates/common/precompile-macros/src/accounting.rs
@@ -251,7 +251,7 @@ fn expand_token(input: DeriveInput) -> syn::Result<TokenStream> {
                         self.b20.transfer_executor_policy_id()
                     }
                     crate::B20PolicyType::MintReceiver => self.b20.mint_receiver_policy_id(),
-                    crate::B20PolicyType::Seizable => self.b20.seizable_policy_id(),
+                    crate::B20PolicyType::SeizableAccount => self.b20.seizable_policy_id(),
                 }
             }
 
@@ -273,7 +273,7 @@ fn expand_token(input: DeriveInput) -> syn::Result<TokenStream> {
                     crate::B20PolicyType::MintReceiver => {
                         self.b20.set_mint_receiver_policy_id(policy_id)
                     }
-                    crate::B20PolicyType::Seizable => {
+                    crate::B20PolicyType::SeizableAccount => {
                         self.b20.set_seizable_policy_id(policy_id)
                     }
                 }
--- a/crates/common/precompiles/src/common/core_storage.rs
+++ b/crates/common/precompiles/src/common/core_storage.rs
@@ -78,9 +78,9 @@ pub struct B20CoreStorage {
     #[accessor(name = nonce, keys(owner))]
     #[mutator(name = set_nonce, keys(owner), value = nonce)]
     pub nonces: Mapping<Address, U256>, // offset 13
-    /// Seizable-account policy ID, consulted by the seize operations. base-std keeps its mock-only
-    /// `initialized` flag trailing (offset 15), which this impl omits, so seize lands at offset 14
-    /// on both sides.
+    // The base-std mock keeps an `initialized` bootstrap flag as its last field; this impl checks
+    // factory-init via deployed marker bytecode instead, so it stores no such field.
+    /// Seizable-account policy ID, consulted by the seize operations.
     #[accessor]
     #[mutator]
     pub seizable_policy_id: u64, // slot 14, offset 0
--- a/crates/common/precompiles/src/common/core_storage.rs
+++ b/crates/common/precompiles/src/common/core_storage.rs
@@ -78,15 +78,14 @@ pub struct B20CoreStorage {
     #[accessor(name = nonce, keys(owner))]
     #[mutator(name = set_nonce, keys(owner), value = nonce)]
     pub nonces: Mapping<Address, U256>, // offset 13
-    /// Unused full-slot hole where base-std pins its `initialized` flag; kept empty so
-    /// `seizable_policy_id` lands at offset 15 as in base-std.
-    pub initialized_hole: FixedBytes<32>, // offset 14 (unused; base-std `initialized`)
-    /// Seizable-account policy ID, consulted by the seize operations.
+    /// Seizable-account policy ID, consulted by the seize operations. base-std keeps its mock-only
+    /// `initialized` flag trailing (offset 15), which this impl omits, so seize lands at offset 14
+    /// on both sides.
     #[accessor]
     #[mutator]
-    pub seizable_policy_id: u64, // slot 15, offset 0
-    /// Reserved padding to close slot 15.
-    pub seize_reserved: FixedBytes<24>, // slot 15, offset 8
+    pub seizable_policy_id: u64, // slot 14, offset 0
+    /// Reserved padding to close slot 14.
+    pub seize_reserved: FixedBytes<24>, // slot 14, offset 8
 }
 
 #[cfg(test)]
@@ -130,10 +129,9 @@ mod tests {
         assert_eq!(__packing_b20_core_storage::PAUSED_LOC.offset_slots, 11);
         assert_eq!(__packing_b20_core_storage::SUPPLY_CAP_LOC.offset_slots, 12);
         assert_eq!(__packing_b20_core_storage::NONCES_LOC.offset_slots, 13);
-        assert_eq!(__packing_b20_core_storage::INITIALIZED_HOLE_LOC.offset_slots, 14);
-        assert_eq!(__packing_b20_core_storage::SEIZABLE_POLICY_ID_LOC.offset_slots, 15);
+        assert_eq!(__packing_b20_core_storage::SEIZABLE_POLICY_ID_LOC.offset_slots, 14);
         assert_eq!(__packing_b20_core_storage::SEIZABLE_POLICY_ID_LOC.offset_bytes, 0);
-        assert_eq!(__packing_b20_core_storage::SEIZE_RESERVED_LOC.offset_slots, 15);
+        assert_eq!(__packing_b20_core_storage::SEIZE_RESERVED_LOC.offset_slots, 14);
         assert_eq!(__packing_b20_core_storage::SEIZE_RESERVED_LOC.offset_bytes, 8);
     }
 }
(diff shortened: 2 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompile-macros/src/accounting.rs
```diff
@@ -251,6 +251,7 @@ fn expand_token(input: DeriveInput) -> syn::Result<TokenStream> {
                         self.b20.transfer_executor_policy_id()
                     }
                     crate::B20PolicyType::MintReceiver => self.b20.mint_receiver_policy_id(),
+                    crate::B20PolicyType::Seizable => self.b20.seizable_policy_id(),
```
stephancill: why is this just `Seizable` and not `SeizableAccount`?

crates/common/precompiles/src/common/core_storage.rs
```diff
@@ -121,5 +136,12 @@ mod tests {
         assert_eq!(__packing_b20_core_storage::PAUSED_LOC.offset_slots, 11);
         assert_eq!(__packing_b20_core_storage::SUPPLY_CAP_LOC.offset_slots, 12);
         assert_eq!(__packing_b20_core_storage::NONCES_LOC.offset_slots, 13);
+        // Offset 14 is the unused `initialized` hole (base-std pins its bootstrap flag here); the
+        // seizable policy id is appended after it at offset 15, byte 0, mirroring base-std.
```
stephancill: remove

crates/common/precompiles/src/common/core_storage.rs
```diff
@@ -78,6 +78,21 @@ pub struct B20CoreStorage {
...
+    /// Do not collapse this hole, or the mock and precompile storage layouts diverge and fork tests
+    /// break.
+    pub initialized_hole: FixedBytes<32>, // offset 14 (unused; base-std `initialized`)
+    /// Seizable-account policy ID, consulted by the seize operations. Appended after the
+    /// `initialized` hole so no pre-existing field offset shifts on already-deployed tokens.
+    #[accessor]
+    #[mutator]
+    pub seizable_policy_id: u64, // slot 15, offset 0
+    /// Reserved padding to fill the remainder of slot 15.
+    pub seize_reserved: FixedBytes<24>, // slot 15, offset 8 (fills remaining 24 bytes)
```
stephancill: less verbose comments

crates/common/precompiles/src/common/ops/guards.rs
```diff
@@ -83,6 +83,22 @@ impl B20Guards {
             Ok(())
         }
     }
+
+    /// Ensures `account` is seizable, i.e. blocked by the current seizable-account policy.
+    ///
+    /// Mirrors [`Self::ensure_blocked`] but consults `SEIZABLE_ACCOUNT_POLICY` instead of the
```
rayyan224, line 89: /// Ensures `account` is seizable, i.e. blocked by the current seizable-account policy. 

I think we can keep comment to this just that

crates/common/precompiles/src/common/core_storage.rs
```diff
@@ -78,6 +78,15 @@ pub struct B20CoreStorage {
     #[accessor(name = nonce, keys(owner))]
     #[mutator(name = set_nonce, keys(owner), value = nonce)]
     pub nonces: Mapping<Address, U256>, // offset 13
+    /// Unused full-slot hole where base-std pins its `initialized` flag; kept empty so
+    /// `seizable_policy_id` lands at offset 15 as in base-std.
+    pub initialized_hole: FixedBytes<32>, // offset 14 (unused; base-std `initialized`)
```
rayyan224: What is this for ?

stephancill: AI: It's a storage-layout alignment placeholder, not a functional field (never read or written).

base-std pins a `bool initialized` bootstrap flag in its own slot at offset 14. This Rust impl doesn't store that flag (it tracks the factory-init window a different way), but the seize work appends `seizable_policy_id` immediately after it, which base-std places at offset 15.

If we omitted the hole, `seizable_policy_id` would collapse to offset 14 and the base/base ↔ base-std layouts would diverge — breaking the mock/precompile mirror and the fork-test cross-validation. Reserving offset 14 keeps `seizable_policy_id` at 15, matching base-std. The offset-pin test asserts exactly this.

rayyan224: Hmm so the reasoning is too match parity with mock tests ?

stephancill: AI: Some additional context on _why_ it only became necessary now.

base-std (`MockB20Storage.sol`) and this layout share the same ERC-7201 namespace/root (`base.b20` → `0xc78b71fee795…434000`), and the factory writes token init state directly via slot arithmetic into both impls, so every field must sit at the same offset on both sides (pinned by `b20_core_offsets_match_mock_b20_storage`).

Pre-seize, base-std had `bool initialized` at offset 14 as its **last** field. The Rust impl doesn't store that flag (it demarcates the factory-init window differently), and because `initialized` was trailing, base/base could just omit it — nothing followed it, so dropping it shifted no other offset. That was free.

Seize breaks that: base-std #177 appends `seizePolicyIds` at offset **15**, after `initialized`. It can't reuse 14 or move `initialized` — deployed tokens have those slots baked in, and moving the flag would make live tokens read as uninitialized (see `MockB20Storage.sol:150-160`). So base-std pins seize at 15.

Now the Rust side must also land `seizable_policy_id` at 15. If we just appended after `nonces` (13), the `Storable` derive would place it at 14 — one slot early — and the factory (writing slot 15) would hit the wrong field, breaking fork cross-validation. The `initialized_hole` reserves offset 14 so `seizable_policy_id` stays at 15. Once a field sits *after* `initialized`, the omission that used to be free has to become an explicit reserved slot.

(Nuance: the hole is `FixedBytes<32>` vs base-std's `bool` — both occupy exactly one slot; only the slot index, 14, matters for cross-impl agreement.)

rayyan224: `Pre-seize, base-std had bool initialized at offset 14 as its last field. The Rust impl doesn't store that flag (it demarcates the factory-init window differently), and because initialized was trailing, base/base could just omit it — nothing followed it, so dropping it shifted no other offset. That was free.`

@stevieraykatz The base-std don't need to be backwards compatible so can we move this to the end of the mock tests and then remove the place holder from this ? Seems wrong to have slots so that we can match our tests.

stephancill: AI: Update — removed the hole entirely (commit 9b4e364a4). Since `initialized` is mock-only (base-std) and this impl has no such field, base-std #177 was reordered to keep `initialized` trailing (offset 15) and place `seizePolicyIds` before it, so `seizable_policy_id` now lands at offset 14 on both sides with no filler. This resolves the original question — there is no longer an `initialized_hole`.

crates/common/precompiles/src/common/core_storage.rs
```diff
@@ -78,15 +78,14 @@ pub struct B20CoreStorage {
     #[accessor(name = nonce, keys(owner))]
     #[mutator(name = set_nonce, keys(owner), value = nonce)]
     pub nonces: Mapping<Address, U256>, // offset 13
-    /// Unused full-slot hole where base-std pins its `initialized` flag; kept empty so
-    /// `seizable_policy_id` lands at offset 15 as in base-std.
-    pub initialized_hole: FixedBytes<32>, // offset 14 (unused; base-std `initialized`)
-    /// Seizable-account policy ID, consulted by the seize operations.
+    /// Seizable-account policy ID, consulted by the seize operations. base-std keeps its mock-only
+    /// `initialized` flag trailing (offset 15), which this impl omits, so seize lands at offset 14
+    /// on both sides.
```
stephancill: let's remove this to avoid confusion. make a more generic note that mock has initialized flag and why we don't need to track it here

crates/common/precompiles/src/common/core_storage.rs
robriks: seeing as the reserved slot exists just to keep offset parity with base-std's `MockB20Storage`, moving initialized around is fine. 

generally I'd prefer for all corresponding storage slots across rust/solidity to maintain parity so that we don't accidentally introduce storage collisions in the future, but it may not be worth it just for these test mocks.

robriks (commented):
I believe these changes do affect existing v1 behavior because `B20PolicyType::from_id` is shared by all versions including V1 (through `ensure_supported_policy_type`. So there is concrete behavior diff in v1 from this.  

it will be resolved by the per-version scope gating planned for abis & interfaces, so it's not technically a blocker. but we need to be sure that the follow-up version gating reverts us to parity with V1 behavior

stephancill:
@robriks correct, this PR is aware of the versioning issue and is being tracked to verify once we move to versioned ABI

robriks (approved):
LGTM !

PR #4190, opened 2026-07-28
feat(txpool): two-tier high-rate payer recognition for upgradeable proxies

amiecorso:
Prototype for discussion. Adds a second tier to EIP-8130 high-rate-payer mempool
recognition so a **constrained upgradeable proxy** account can qualify for the
balance-bounded payer book, not just immutable ERC-1167 clones. The two allowlists
default empty, so the new path is dormant and behavior is identical to today until
a shell + implementations are finalized and pinned.

## Problem

Today high-rate trust is single-tier: the payer's own codehash must be the canonical
ERC-1167 clone of a trusted implementation. That is deliberately **free** to check,
the balance/fee check already loaded the payer's `basic_account`, so `bytecode_hash`
is in hand and recognition is an O(1) set lookup with no extra reads. The cost is
rigidity: an upgradeable account can never qualify, and even an immutable one is
frozen to a single implementation forever.

## Design

`high_rate_payer_class` returns `HighRatePayerClass { None, Immutable, Upgradeable }`:

- **Immutable (tier A, unchanged):** codehash is a trusted ERC-1167 clone. Free.
- **Upgradeable (tier B, new):** codehash is a trusted *constrained proxy shell*
  (one that freezes upgrades while locked) whose current ERC-1967 implementation is
  trusted.

**Admission does zero extra state reads** (per review): tier B is a pure in-memory
cache lookup, never an impl-slot `SLOAD`. A cold/unclassified upgradeable payer is
conservatively count-limited, never over-admitted.

**Classification is moved to the once-per-block state-diff path** and computed purely
from data the node already produced for the block (no reads anywhere):

- `AccountStateDiff` carries the account's new `code_hash` and, when the ERC-1967 slot
  changed that block, its new raw value (from the `BundleState` in `collect()`).
- `invalidate_limit_class_cache` classifies trusted shells: slot `0` is the baked
  default (trusted); a well-formed address is checked against the trusted-implementation
  allowlist; a malformed value is untrusted; a code change to a shell is treated as
  creation (baked default). The result is cached and advances the classification
  generation so pending admissions re-evaluate.

Because the diff exposes the implementation only by **address** (the slot value) and the
classifier must not read the implementation account, the inner allowlist is address-keyed
(`trusted_implementation_addresses`) rather than codehash-keyed.

An `Upgradeable` payer additionally watches its ERC-1967 slot (an upgrade demotes
already-admitted txs), and a slot write invalidates/reclassifies its cached class.
EIP-7702-delegated accounts still never qualify.

## Config

`with_trusted_upgradeable_proxies(shell_code_hashes, implementation_addresses)`;
`Eip8130Contracts::ERC1967_IMPLEMENTATION_SLOT` added. Defaults empty (provisional,
like the other Base Sepolia values in `Eip8130Contracts`).

## Known tradeoff

After a cache clear / node restart, an already-locked upgradeable payer with no
subsequent state change stays count-limited until a relevant diff reclassifies it
(safe/conservative). A startup backfill is possible follow-up.

## Testing

`cargo test -p base-execution-txpool --lib` — 205 pass (200+ existing unchanged + new);
`base-common-consensus` 93 pass; clippy + rustfmt clean. New tests: cold-cache is None,
commit classifies trusted then admission trusts, untrusted impl, slot-zero baked default,
malformed slot, immutable tier, impl-slot invalidation.

## Open questions

- Address-based inner allowlist is forced by the no-read constraint (diff carries the
  impl address, not its codehash). Acceptable given CREATE2-deterministic immutable impls?
- Cold-start after restart: acceptable as-is, or add a bounded startup backfill?
- Should the trusted sets be a companion-ERC canonical set (like authenticators) rather
  than node config?

### DIFF
--- a/crates/common/consensus/src/transaction/eip8130/addresses.rs
+++ b/crates/common/consensus/src/transaction/eip8130/addresses.rs
@@ -90,6 +90,21 @@ impl Eip8130Contracts {
     pub const CANONICAL_HIGH_RATE_PAYER_PROXY_CODE_HASH: B256 =
         b256!("0x3e37c64c39476e47c52408ea45eb3ae0f07e3ca0fd1d713acbcf17bf3b51312c");
 
+    /// The EIP-1967 implementation storage slot
+    /// (`keccak256("eip1967.proxy.implementation") - 1`), where an upgradeable
+    /// ERC-1967 proxy stores its current implementation address.
+    ///
+    /// Read by the two-tier high-rate payer recognition (see the txpool
+    /// validator): behind a *trusted upgradeable-proxy shell* codehash, the node
+    /// resolves the effective implementation from this slot and checks it against
+    /// the trusted-implementation allowlist. This lets an *upgradeable* account
+    /// qualify for the high-rate tier (provided the shell is a constrained proxy
+    /// that freezes upgrades while locked), whereas the single-tier path
+    /// ([`Self::CANONICAL_HIGH_RATE_PAYER_PROXY_CODE_HASH`]) only recognizes
+    /// immutable ERC-1167 clones.
+    pub const ERC1967_IMPLEMENTATION_SLOT: B256 =
+        b256!("0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc");
+
     // ─────────────────────────────────────────────────────────────────────────
     // Canonical authenticators (accepted on the EIP-8130 block-validation path)
     // ─────────────────────────────────────────────────────────────────────────
--- a/crates/execution/txpool/src/pool.rs
+++ b/crates/execution/txpool/src/pool.rs
@@ -46,6 +46,10 @@ pub struct AccountStateDiff {
     pub code_changed: bool,
     /// Changed contract storage slots.
     pub changed_slots: Vec<B256>,
+    /// New account code hash, when the account is present in the bundle.
+    pub code_hash: Option<B256>,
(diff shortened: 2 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4194, opened 2026-07-28
example(policy): add abi/v3.rs to show what a new wire surface costs

rayyan224:
**Worked example, not intended to merge.** Stacked on #4193 (`feat/policy-abi-versioning`), which introduces the `abi/` tree this builds on. Answers "what does adding an ABI v3 with a new method look like."

Activated at `Zombie`, an existing post-Cobalt variant already documented for experimental features, so the fork ladder and `base-common-genesis` stay untouched.

## The case shown: a wire-only fork

`policyCount()` reads the existing counter, so `PolicyRegistryV2` still backs Zombie and **nothing appears under `logic/`**. Total cost:

| file | change |
|---|---|
| `abi/v3.rs` | new; Cobalt's surface plus `policyCount()` |
| `abi/mod.rs` | `mod v3`, canonical retargeted from v2, one `as_label` arm |
| `versions.rs` | `PolicyVersion::V3`, `PolicyAbi::V3`, three match arms, the `Zombie` fork arm |
| `logic/interface.rs` | `policy_count` with a shared default body |
| `dispatch.rs` | one route arm, one view fast-path selector |
| both goldens | one coverage arm plus the test it names |

## What building it surfaced

The two axes were described as growing independently. That independence runs one direction only.

A logic-only fork reuses an existing surface. A **wire-only fork still earns a new `PolicyVersion`**, because `PolicyVersion::abi()` is a function of the version alone; leaving Zombie on `V2` would ask one version to answer with two surfaces. So `implementation()` gets `Self::V2 | Self::V3 => &V2` while `abi()` gets `Self::V3 => PolicyAbi::V3`. Pinned by `zombie_moves_the_wire_without_moving_the_logic`.

## Old forks reject the new selector for free

Neither the Beryl nor Cobalt surface declares `policyCount`, so `valid_selector` is false and dispatch returns `UnknownFunctionSelector`. No `if version < V3` anywhere. That gate is also why `policy_count` can be a plain shared default on the trait rather than a per-version implementation: older versions inherit a correct body they can never be asked for.

## The compiler drove the wiring

Adding the `sol!` function broke the exhaustive matches in `as_label`, `route`, and **both golden coverage checklists**. That last one is the useful part. `v1_op_coverage_checklist` carries the comment *"Adding an ABI op fails the build until a golden is added"*, so the example could not compile until goldens pinned that Beryl and Cobalt reject the new selector. The composite work had already established that pattern; the new arms follow it.

## Verification

947 tests pass, 0 fail. clippy and fmt clean.

Generated with Claude Code

### DIFF
--- a/crates/common/precompiles/src/b20_asset/logic/v1.rs
+++ b/crates/common/precompiles/src/b20_asset/logic/v1.rs
@@ -1080,6 +1080,15 @@ mod tests {
             self.initialized = true;
             Ok(())
         }
+
+        // This fake does not exercise composite policies.
+        fn read_children(&self, _policy_id: u64) -> Result<Vec<u64>> {
+            Ok(Vec::new())
+        }
+
+        fn write_children(&mut self, _policy_id: u64, _child_policy_ids: &[u64]) -> Result<()> {
+            Ok(())
+        }
     }
 
     type Tok = B20AssetToken<FakeAccounting, FakePolicyAccounting>;
--- a/crates/common/precompiles/src/b20_asset/logic/v2.rs
+++ b/crates/common/precompiles/src/b20_asset/logic/v2.rs
@@ -1275,6 +1275,15 @@ mod tests {
             self.initialized = true;
             Ok(())
         }
+
+        // This fake does not exercise composite policies.
+        fn read_children(&self, _policy_id: u64) -> Result<Vec<u64>> {
+            Ok(Vec::new())
+        }
+
+        fn write_children(&mut self, _policy_id: u64, _child_policy_ids: &[u64]) -> Result<()> {
+            Ok(())
+        }
     }
 
     type Tok = B20AssetToken<FakeAccounting, FakePolicyAccounting>;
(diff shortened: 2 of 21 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4202, opened 2026-07-29
perf(protocol): encode batch frames in place

haardikk21:
## Summary
- add non-allocating frame encoding into an existing buffer
- pre-size batch transaction output exactly and encode every frame directly into it
- preserve the existing allocating `Frame::encode` API

## Evidence
Previously `BatchTransaction::to_bytes` allocated one temporary vector per frame, then copied each temporary into a separately growing accumulator. The new path performs one exact-size allocation for the complete batch, eliminating N temporary allocations and accumulator growth for N frames.

## Validation
- `cargo fmt --check`
- `cargo test -p base-protocol batch::tx`
- `cargo check -p base-protocol --all-targets`
- `git diff --check`

### DIFF
--- a/crates/consensus/protocol/src/batch/tx.rs
+++ b/crates/consensus/protocol/src/batch/tx.rs
@@ -29,13 +29,12 @@ impl BatchTransaction {
 
     /// Returns the [`BatchTransaction`] as a [`Bytes`].
     pub fn to_bytes(&self) -> Bytes {
-        self.frames
-            .iter()
-            .fold(Vec::new(), |mut acc, frame| {
-                acc.append(&mut frame.encode());
-                acc
-            })
-            .into()
+        let capacity = self.frames.iter().map(Frame::encoded_len).sum();
+        let mut encoded = Vec::with_capacity(capacity);
+        for frame in &self.frames {
+            frame.encode_into(&mut encoded);
+        }
+        encoded.into()
     }
 }
 
--- a/crates/consensus/protocol/src/frame.rs
+++ b/crates/consensus/protocol/src/frame.rs
@@ -166,7 +166,8 @@ pub struct Frame {
 }
 
 impl Frame {
-    /// Number of metadata bytes in an encoded frame.
+    /// Number of metadata bytes in an encoded frame: 16-byte channel ID, 2-byte frame number,
+    /// 4-byte payload length, and 1-byte final-frame flag.
     pub const ENCODED_OVERHEAD: usize = 16 + 2 + 4 + 1;
 
     /// Overhead estimation for frame metadata and tagging information.
--- a/crates/consensus/protocol/src/frame.rs
+++ b/crates/consensus/protocol/src/frame.rs
@@ -166,6 +166,9 @@ pub struct Frame {
 }
 
 impl Frame {
+    /// Number of metadata bytes in an encoded frame.
+    pub const ENCODED_OVERHEAD: usize = 16 + 2 + 4 + 1;
+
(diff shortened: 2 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

haardikk21:
Addressed the review consistency note in the latest commit: decoding now uses `Frame::ENCODED_OVERHEAD` as the single source of truth. The full `base-protocol` test suite (235 tests) and formatting checks pass.

crates/consensus/protocol/src/batch/tx.rs
```diff
@@ -29,13 +29,12 @@ impl BatchTransaction {
...
-                acc.append(&mut frame.encode());
-                acc
-            })
-            .into()
+        let capacity = self.frames.iter().map(Frame::encoded_len).sum();
+        let mut encoded = Vec::with_capacity(capacity);
+        for frame in &self.frames {
+            frame.encode_into(&mut encoded);
+        }
+        encoded.into()
```
refcell, line 37: Can we add a benchmark to show this improves perf?

crates/consensus/protocol/src/frame.rs
```diff
@@ -166,6 +166,9 @@ pub struct Frame {
 }
 
 impl Frame {
+    /// Number of metadata bytes in an encoded frame.
+    pub const ENCODED_OVERHEAD: usize = 16 + 2 + 4 + 1;
```
refcell, line 171: Maybe we could comment what these #s are so it's clear to the reader

crates/consensus/protocol/src/batch/tx.rs
haardikk21: Added a Criterion benchmark comparing the previous temporary-buffer implementation with in-place encoding over 32 × 128 KiB frames. On my machine, median time improved from 103.94 µs to 56.33 µs (~45.8% lower), with throughput increasing from 37.59 to 69.36 GiB/s. The benchmark is in `crates/consensus/protocol/benches/batch_transaction.rs`.

crates/consensus/protocol/src/frame.rs
haardikk21: Added the wire-field breakdown to the constant documentation: 16-byte channel ID, 2-byte frame number, 4-byte payload length, and 1-byte final-frame flag.

refcell (approved):

PR #4205, opened 2026-07-29
feat(simulation): inline resource metering

wlawt:
- Beginning of `eth_sendRawTx`, start a worker thread that calls `meterBundle`
- Forwarder requires `MeterBundleResponse` present before calling `base_insertValidatedTx` for a tx
- `base_insertValidatedTx` will update the builder's metering cache in the same RPC call
- If metering is not present in time, the forwarder will consider the tx on the next forwarding batch


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/txpool/src/forwarder/task.rs
```diff
@@ -257,6 +275,35 @@ where
...
+                None => {
+                    metering.submit(buffered.tx_hash, buffered.transaction.raw.clone());
+                    ForwarderMetrics::txs_deferred_metering(Arc::clone(&self.url_label))
+                        .increment(1);
+                    waiting.push(buffered);
+                }
+            }
+        }
+        (ready, waiting)
+    }
```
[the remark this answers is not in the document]
wlawt: From what I recall, the failure modes of meterBundle have been addressed:
- nonce gap errors ✅ 
- unable to get l1 block info ✅ 
- timeouts (but this is from the ingress-rpc side)

Cursoragent confirmed that these errors are handled in the `meter_bundle` API

crates/execution/txpool/src/forwarder/task.rs
```diff
@@ -158,6 +159,11 @@ where
 
             let closed = tokio::select! {
                 _ = self.cancel.cancelled() => break,
+                // Retry metering gate while buffered txs are waiting on a response.
+                _ = time::sleep(std::time::Duration::from_millis(50)),
+                    if self.config.require_metering && !self.buffer.is_empty() => {
+                    false
+                }
                 result = self.receiver.recv() => {
```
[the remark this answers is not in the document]
wlawt: This is probably not even needed. If we think about the two scenarios p50 and p999 on the Consumer POV:

- p50: the tx from the Consumer is immediately sent to the Forwarder
- p999: txs are batched and sent from the Consumer every 10ms

A tx in the Forwarder that did not have its metering info in time will at most wait another 10ms before checking the cache again

crates/execution/cli/src/standard_node.rs
```diff
@@ -394,6 +407,12 @@ impl StandardBaseRethNode {
             gas_limit: args.metering.metering_gas_limit,
             da_bytes: args.metering.metering_da_bytes,
         };
+        // Inline metering is only for mempool nodes that both meter and forward.
+        // Builders can enable metering RPC without allocating the inline service.
+        let inline_metering_slot = (args.metering.enable_metering
```
niran: nit: `slot` isn't a pattern we use elsewhere for `OnceLock` / late-bound handles. I think we could either just name it after what it holds (e.g. `inline_metering`) or with a more common suffix, like `…_cell` from the common `OnceCell` pattern. Same for the `with_inline_metering_slot` builders.

crates/execution/metering/src/inline.rs
```diff
@@ -0,0 +1,139 @@
...
+            let Ok(_permit) = semaphore.acquire_owned().await else {
+                // Closed between the pre-check and acquire (shutdown). Leave the
+                // hash in in_flight so submit() will not respawn.
+                return;
+            };
+
+            let bundle = Bundle { txs: vec![raw], ..Default::default() };
+            match api.meter_bundle(bundle).await {
+                Ok(response) => {
+                    debug!(tx_hash = %tx_hash, "inline meterBundle completed");
```
niran, line 106: Ingress emits a txobs trail for this path today: `SIMULATION_STARTED` / `_SUCCEEDED` / `_FAILED` (with a metering summary) plus `INGRESS_METERING_SEND_*` when pushing `base_setMeteringInformation`.

This PR only logs + `txs_deferred_metering`. As we move metering onto the mempool node, we should keep sim coverage in the event stream — at least emit `SIMULATION_*` from here with `TransactionEventProducer::BaseRethNode` (those names aren't ingress-specific).

`INGRESS_METERING_SEND_*` can stay ingress-only and eventually be deleted.

crates/execution/rpc/src/eth/transaction.rs
```diff
@@ -66,6 +66,12 @@ where
             return Err(BaseInvalidTransactionError::Eip8130NotAccepted.into());
         }
 
+        // Kick off in-process meterBundle before pool insert so the forwarder can
+        // wait on a MeterBundleResponse without delaying the RPC response.
```
niran, line 70: Totally optional: it'd be neat if we had a baseline for how parallelizing this work affects how long it takes for the transaction to get forwarded. In theory, the current implementation should produce lower latency to the builder than calling meter_bundle sequentially. Knowing by how much would be informative, but we could always get those numbers later if/when we actually need them.

crates/execution/cli/src/standard_node.rs
```diff
@@ -406,19 +425,27 @@ impl StandardBaseRethNode {
                 .clone()
                 .map_or_else(MeteringConfig::enabled, MeteringConfig::with_flashblocks)
                 .with_resource_limits(resource_limits)
-                .with_metered_opcodes(metered_opcodes);
+                .with_metered_opcodes(metered_opcodes)
+                .with_inline_max_concurrent(args.metering.inline_metering_max_concurrent);
```
haardikk21: nit: `inline_max_concurrent` is not representative of what this is for. the use of the term `inline` as well only makes sense with previous context. imo we should just rename this to `with_metering_max_processes` to make it clearer and rename the actually CLI arg accordingly as well

crates/execution/metering/src/extension.rs
```diff
@@ -269,6 +315,11 @@ pub struct MeteringConfig {
     pub target_flashblocks_per_block: Option<usize>,
     /// Opcodes and precompiles to track for gas metering.
     pub metered_opcodes: MeteredOpcodes,
+    /// Max concurrent in-process meterBundle workers for mempool inline simulation.
+    pub inline_max_concurrent: usize,
+    /// Slot shared with `EthApi` / tx-forwarding for the inline metering handle.
+    pub inline_metering_slot:
+        Option<std::sync::Arc<std::sync::OnceLock<base_bundles::SharedInlineMetering>>>,
```
haardikk21: please import and use shorter readable names here rather than the FQDN names of the imports re: `Arc`, `OnceLock`, and `SharedInlineMetering`

crates/execution/metering/src/inline.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        {
+            let mut in_flight = self.in_flight.lock();
+            // Cache + in_flight under one lock so a concurrent completion cannot
+            // race a second spawn past a stale contains_key check.
+            if self.cache.contains_key(&tx_hash) || !in_flight.insert(tx_hash) {
+                return;
+            }
+        }
```
haardikk21: im not a big fan of this implicit lock freeing by creating a scope. lets get rid of the scope and explicitly drop/free the lock where necessary

crates/execution/metering/src/inline.rs
wlawt: i can make this a fast follow PR

PR #4192, opened 2026-07-28
feat(BOP-441): add seize entrypoints to asset + stablecoin V2

stephancill:
## Summary

PR4 of the seize stack (BOP-441). **Makes seize callable** across asset V2 + stablecoin V2, mirroring the base-std reference ([base-std#177](https://github.com/base/base-std/pull/177)). Stacked on #4186 (plumbing); review/merge that first.

New V2-only surface, version-gated so frozen V1 keeps rejecting it byte-for-byte.

## What's included

- **ABI** (`abi.rs`): `SEIZE` `PausableFeature` member; `transferFromSeizableWithMemo` + `burnBlockedWithMemo`; `TRANSFER_FROM_SEIZABLE_ROLE()` / `SEIZABLE_ACCOUNT_POLICY()` getters; `as_label` arms. `SEIZE` also wired into `B20PausableFeature::ensure_valid`.
- **Logic** (asset + stablecoin V2): `transfer_from_seizable_with_memo` and `burn_blocked_with_memo`, overriding `reject_frozen_selector!` defaults on the shared trait (V1 inherits reject). Explicit guard sequence `SEIZE pause -> role -> ensure_seizable(from) -> to != 0 -> balance`, moving balance via the shared `move_balance` (from PR1). Emit order `Transfer -> Memo -> {TransferredFromSeizable | BurnedBlocked}` per the IB20 memo invariant.
- **Dispatch** (asset + stablecoin): handler arms + a **pre-decode fork gate** rejecting the four seize selectors before Cobalt with `UnknownFunctionSelector` (same pattern as the ERC-8056 gate), so pre-Cobalt calldata is byte-identical.
- **Golden** (asset + stablecoin V1): `golden_seize_selectors_unknown_at_v1` + coverage-checklist arms pin that the frozen V1 op surface rejects the new selectors.
- **Unit tests** (asset + stablecoin V2): happy path, event order, revert paths (`AccountNotBlocked`, `InvalidReceiver`, `ContractPaused(SEIZE)`, role), and the decoupling invariants (ignores receiver policy on `to`, no allowance required).
- **etc/systems**: include `SEIZE` in the RPC `unpause`-all helper (mask `0x0f`) so it truly clears every vector.

## Design notes

[...]

PR #4206, opened 2026-07-29
refactor(factory): decode calldata against the wire surface frozen at each fork

rayyan224:
Ports the shape from #4193 to `b20_factory`, and closes a live gap in the `B20Variant` ordinal binding that #4193's own pin suite would not have caught.

## This one is prophylactic — with one exception

Unlike the policy precompile, **there is no bug here and no gate to delete.** `b20_factory/dispatch.rs` has never carried a hand-written fork gate — zero `version ==` / `version <` conditions — and `FactoryVersion` has only `V1`, mapping Beryl and Cobalt alike. So `FactoryAbi` has exactly one variant and gates nothing today. The value is that the next fork to move this surface becomes a compile error in two exhaustive matches instead of a hand-written selector list. No third `B20Variant` or `FactoryV2` is currently planned.

The exception is the ordinal gap below, which is real and live.

## The change

`abi.rs` becomes `abi/{mod,v1}.rs`. `v1.rs` holds the full frozen interface; `mod.rs` aliases canonical onto the newest frozen surface:

```rust
mod v1;
pub use v1::{IB20Factory, IB20Factory as IB20FactoryV1};
```

Canonical *is* the v1 type rather than a third copy, so every external `IB20Factory` reference is untouched and each generated nested name (`IB20FactoryCalls`, `B20Variant`, `createB20Call`, `B20StablecoinCreateParams`) keeps its exact spelling.

`versions.rs` gains the second axis:

```rust
pub enum FactoryVersion { V1 }   // logic axis; ONLY thing from_base_upgrade returns
pub enum FactoryAbi     { V1 }   // wire axis; grows only when the wire moves

impl FactoryVersion { pub const fn abi(self) -> FactoryAbi { … } }
```

`FactoryAbi` has no `from_base_upgrade` on purpose: a second resolver over the same ladder is two maps that must agree, which is the failure mode this module exists to prevent.

**The interface's Rust name is consensus data.** `SolInterface::abi_decode_validate` short-circuits `data.len() < MIN_DATA_LENGTH + 4` with `Error::type_check_fail(data, Self::NAME)`, and `AbiDecodeFailed` encodes as `selector ‖ utf8(error)`. The frozen file therefore keeps declaring `interface IB20Factory` — naming it `IB20FactoryV1` would have moved Beryl's bytes on every truncated-calldata revert. Pinned by `surface_interface_name_is_frozen`.

**One deliberate divergence from `PolicyAbi::decode`:** it decodes once, not twice. Policy re-decodes against canonical because there frozen V1 ≠ canonical V2; here canonical aliases v1, so the second decode would be provably dead. The `match self` arm marks where a later surface adds it.

## The ordinal gap this closes

`test_abi_enum_ordinals_match_solidity` did not check Solidity ordinals. All four of its assertions were Rust-side, and `IB20Factory::B20Variant::ASSET as u8` appeared **nowhere in the repo**. That ordinal is byte `[10]` of every deployed token address — spliced in as plaintext, not hashed:

```
address = 0xb2 ‖ 0x00 × 9 ‖ discriminant ‖ keccak256(abi_encode(creator, salt))[..9]
```

Reordering the `sol!` enum makes `createB20(0, …)` decode to `STABLECOIN`, which writes byte `1`. Every token remaps, silently.

**#4193's fingerprint shape would not catch this either** — it hashes selectors, event topic0s, error selectors and `COUNT`, and a reorder changes none of them, since Solidity encodes enums as `uint8` so no signature moves. The fingerprint here additionally hashes each named variant's ordinal, and `abi_ordinals_are_the_address_discriminants` pins the chain end to end: ABI ordinal → Rust discriminant → derived address → recovered variant.

**Verified, not assumed.** Reordering `ASSET`/`STABLECOIN` and running the whole module:

```
49 passed; 2 failed
  ✗ abi_ordinals_are_the_address_discriminants
      "ABI ordinal for Asset moved; every token it deploys would remap"  left: 1, right: 0
  ✗ v1_abi_fingerprint_is_pinned
```

`selector_set_is_frozen`, `b20_variant_is_frozen` and all 47 pre-existing tests passed under the reorder — confirming both that the gap was real and that the ported pins alone would not have closed it.

## Behavior change

None. `FactoryAbi::decode` replaced `decode_precompile_call!` on the execution path; the macro decodes first and consults `valid_selector` only on the error path, while `decode` checks the selector first. The orderings are observationally identical:

| input | macro | `FactoryAbi::decode` |
|---|---|---|
| `len < 4` | `UnknownFunctionSelector([0;4])` | same |
| unknown selector | `UnknownFunctionSelector(sel)` | same |
| known selector, bad args | `AbiDecodeFailed{sel, err}` | same |
| known selector, `4 ≤ len < MIN_DATA_LENGTH + 4` | `AbiDecodeFailed` | same |
| valid call | `Ok(call)` | same |

Pinned by `decode_preserves_legacy_macro_semantics`. No revert bytes move.

## Verification

- Full suite green — 691 lib + 267 integration tests, 0 failures
- **All 21 factory goldens pass without `BLESS_GOLDEN`**; no storage root or gas pin moved
- `base-metering`, `base-load-tests` and the benches compile untouched
- clippy (`--all-targets --all-features`) and nightly fmt clean

## Follow-ups, not in this PR

- **`create_b20` (`dispatch.rs:75`) still hard-pins `FactoryV1`**, bypassing version resolution — the last place a concrete version is named outside `versions.rs`.
- **`metrics.rs` `factory_call`** selector-matches with a `_ => unknown()` wildcard, so a fifth op would silently label as `unknown` rather than failing the build.
- **The golden `v1_op_coverage_checklist`** is wildcard-free over `IB20FactoryCalls` but keyed on functions, not `B20Variant` values, so a third variant would not fail it.
- **`b20_asset` and `b20_stablecoin` still carry hand-written gates** — `b20_asset` has an 8-selector gate over the Cobalt multiplier surface, and open #4192 adds a 4-selector seize gate to both over the *shared* `IB20`. Treating those means versioning `common/abi.rs`, which is a wider blast radius than this PR.

Generated with Claude Code


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/b20_factory/abi/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// Keccak of sorted call selectors, then sorted event topic0s, then sorted error selectors,
+    /// then `B20Variant::COUNT`, then each named variant's ordinal. Order is fixed so a single pin
+    /// catches any wire-surface edit.
+    ///
+    /// The trailing ordinals are this precompile's addition to the shape used by
+    /// [`crate::PolicyAbi`]'s fingerprint. A `B20Variant` reorder leaves selectors, topic0s, error
+    /// selectors and `COUNT` all untouched — Solidity encodes enums as `uint8`, so no signature
+    /// moves — yet it remaps every token address. Hashing the ordinals is what makes the reorder
+    /// visible here.
+    fn abi_fingerprint(
```
rayyan224: Let's move this to a common, since it's also used in policy registry

crates/common/precompiles/src/b20_factory/versions.rs
```diff
@@ -28,6 +32,70 @@ impl FactoryVersion {
...
+    /// call with `expect`, so a non-validating decode here would turn an out-of-range variant byte
+    /// from a revert into a node panic.
+    pub fn decode(self, calldata: &[u8]) -> Result<IB20Factory::IB20FactoryCalls> {
+        let Some(selector) = calldata.first_chunk::<4>().copied() else {
+            return Err(BasePrecompileError::UnknownFunctionSelector([0u8; 4]));
+        };
+        if !self.valid_selector(selector) {
+            return Err(BasePrecompileError::UnknownFunctionSelector(selector));
+        }
+        match self {
```
rayyan224: Maybe use self.decode_validate_here ?

crates/common/precompiles/src/b20_factory/abi/mod.rs
rayyan224: Done in c48dda95a — moved to `common/abi_fingerprint.rs` as `AbiFingerprint::compute`, gated `cfg(any(test, feature = "test-utils"))` next to `common::test_utils` so it stays out of production builds. Both `policy/abi/mod.rs` and `b20_factory/abi/mod.rs` now call it.

One thing worth a look: the shared signature keeps the `enum_ordinals` term the factory added, and **policy passes `[]`**. That is a byte-level no-op, so both merged policy fingerprints keep the values they were blessed with — confirmed, their pins are untouched and still green.

I left it that way deliberately rather than feeding policy's ordinals in, because `PolicyType` has the same exposure the factory does: the discriminant rides the top byte of every policy ID via `PolicyRegistryV1::make_id`, and `shared_policy_type_discriminants_agree_across_surfaces` only catches a reorder of *one* surface, not a simultaneous reorder of both. Passing the ordinals would close that — at the cost of re-blessing `V1_ABI_FINGERPRINT` and `V2_ABI_FINGERPRINT`. Silently moving a just-merged canary inside a "move helper to common" commit felt like the wrong call, so it is documented at the call site as a follow-up. Say the word and I'll do it here instead.

crates/common/precompiles/src/b20_factory/versions.rs
rayyan224: Good call — done in c48dda95a, and it resolves the dead-code finding on this same method in one go.

`decode` now selector-gates and delegates:

```rust
pub fn decode(self, calldata: &[u8]) -> Result<IB20Factory::IB20FactoryCalls> {
    let Some(selector) = calldata.first_chunk::<4>().copied() else {
        return Err(BasePrecompileError::UnknownFunctionSelector([0u8; 4]));
    };
    if !self.valid_selector(selector) {
        return Err(BasePrecompileError::UnknownFunctionSelector(selector));
    }
    self.abi_decode_validate(calldata, selector)
}
```

To keep that a *single* decode I changed `abi_decode_validate` to return the decoded call rather than discard it with `.map(|_| ())`. Delegating while still discarding would have meant validating and then decoding the same bytes again — on `createB20` that means parsing the full `params` and `initCalls` blobs twice.

Returning the call is also closer to alloy, whose own `abi_decode_validate` returns it. `PolicyAbi`'s wrapper only discards it because policy's dispatch uses it as a pure gate ahead of the activation check (`policy/dispatch.rs:74`); the factory has no such caller. The two methods keep distinct jobs — one decodes, the other adds the selector gate.

No behavior change: all 21 goldens still pass without `BLESS_GOLDEN`.

crates/common/precompiles/src/b20_factory/versions.rs
```diff
@@ -28,6 +32,70 @@ impl FactoryVersion {
...
+    /// the decoded call.
+    pub fn abi_decode_validate(self, calldata: &[u8], selector: [u8; 4]) -> Result<()> {
+        match self {
+            Self::V1 => IB20FactoryV1::IB20FactoryCalls::abi_decode_validate(calldata).map(|_| ()),
+        }
+        .map_err(|error| BasePrecompileError::AbiDecodeFailed {
+            selector,
+            error: error.to_string(),
+        })
+    }
```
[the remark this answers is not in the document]
rayyan224: Confirmed and fixed in c48dda95a — you were right that it was dead. `decode` was calling the *generated* `IB20FactoryV1::IB20FactoryCalls::abi_decode_validate` directly, not `self.abi_decode_validate`, so the wrapper had zero call sites.

Rather than remove it, `decode` now delegates to it (per review on the line below), which was the intent you guessed at. To avoid the double decode that delegation would otherwise cause, `abi_decode_validate` now returns the decoded call instead of discarding it — matching alloy's own signature.

crates/common/precompiles/src/policy/abi/mod.rs
```diff
@@ -57,54 +56,29 @@ mod tests {
...
-        buf.push(policy_type_count as u8);
-        keccak256(&buf)
-    }
-
+    /// These two surfaces pass no enum ordinals to [`AbiFingerprint`], so the pinned constants
+    /// above keep the values they were blessed with. `PolicyType` ordinals *are* load-bearing —
+    /// the discriminant rides the top byte of every policy ID via `PolicyRegistryV1::make_id` —
+    /// and `shared_policy_type_discriminants_agree_across_surfaces` below only catches a reorder
+    /// of one surface, not a simultaneous reorder of both. Feeding the ordinals in here would
+    /// close that, at the cost of re-blessing both constants; deliberately left as a follow-up.
```
robriks, line 65: makes sense, do we have a linear ticket for this follow up?

crates/common/precompiles/src/b20_factory/versions.rs
```diff
@@ -28,6 +32,71 @@ impl FactoryVersion {
...
+
+/// A frozen wire (ABI) surface of the `B20Factory` precompile. Reached only through
+/// [`FactoryVersion::abi`].
+///
+/// Deliberately has no `from_base_upgrade`: a second resolver over the same fork ladder would be
+/// two maps that must agree, which is the failure mode this module exists to prevent. With one
+/// resolver, adding a version is a compile error in both [`FactoryVersion::implementation`] and
+/// [`FactoryVersion::abi`] until someone fills them in.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FactoryAbi {
```
robriks, line 52: awesome, will base the multiplier scheduling abi versioning on this!

crates/common/precompiles/src/b20_factory/abi/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+        );
+    }
+
+    /// The ABI ordinal of each variant *is* byte `[10]` of every token that variant deploys
+    /// (`B20Variant::address_prefix`), spliced in as plaintext rather than hashed. Reordering the
+    /// `sol!` enum would keep every selector, topic0 and error selector identical while silently
+    /// remapping the entire deployed token address space, so the binding is pinned end to end:
+    /// ABI ordinal -> Rust discriminant -> derived address -> recovered variant.
+    #[test]
+    fn abi_ordinals_are_the_address_discriminants() {
```
robriks, line 81: good catch on pinning this !

robriks (dismissed):

robriks (approved):
LGTM

PR #4207, opened 2026-07-29
refactor(BOP-441): version the shared IB20 ABI per hardfork

stephancill:
Base of the BOP-441 stack. The seize PR (#4192) is stacked on top of this one.

## Why

`IB20` is shared by the asset and stablecoin B-20 precompiles and is unversioned in Solidity, but the wire it accepts need not be frozen forever. A hardfork may widen a `sol!` enum or add selectors, and a widened enum is decoded **before** version dispatch — so, unlike a brand-new selector (which a caller cannot dial until a surface declares it), a new enum discriminant would otherwise decode against the canonical surface at every historical fork. This introduces a per-fork frozen wire surface so calldata is decoded against the surface active at the block's fork.

This is the general mechanism, mirroring the policy-registry versioning in #4193. The immediate consumer is the seize surface (#4192): once `SEIZE` is added to v2 only, the frozen v1 rejects that discriminant automatically, with no hand-written gate.

## The change

- `common/abi.rs` becomes `common/abi/{mod,v1,v2}.rs`. `v1.rs` is the frozen copy; `v2.rs` is the canonical live surface. **In this PR v1 and v2 are byte-identical** (the currently shipped surface) — the split carries no behavior change on its own.
- Both declare `interface IB20`, so the Rust name (`IB20Calls`) — which reaches consensus data via `AbiDecodeFailed` on short calldata — never moves. Canonical aliases the newest surface; all external `IB20` references are untouched.
- New `B20Abi {V1, V2}` (shared by both precompiles): `valid_selector` / `abi_decode_validate` / `decode` (validates against the frozen surface, then re-decodes canonical so `route` matches one call type). Selected per version via `AssetVersion::abi` / `StablecoinVersion::abi`.
- Both dispatchers decode `IB20` via `version.abi().decode(...)` instead of the shared decode macro.

## Behavior

Behavior-preserving: `v1 == v2`, so every existing golden passes unchanged and no storage root moves.

## Verification

- `cargo test -p base-common-precompiles --features test-utils`: all pass (lib 668, asset 102, stablecoin 80, factory 21, policy 30/32).
- clippy and fmt clean.

### DIFF
--- a/crates/common/precompiles/src/b20_asset/versions.rs
+++ b/crates/common/precompiles/src/b20_asset/versions.rs
@@ -9,7 +9,7 @@
 
 use base_common_genesis::BaseUpgrade;
 
-use crate::{Asset, AssetAccounting, AssetV1, AssetV2, PolicyAccounting};
+use crate::{Asset, AssetAccounting, AssetV1, AssetV2, B20Abi, PolicyAccounting};
 
 /// An activated version of the asset B-20 precompile logic.
 ///
@@ -37,6 +37,14 @@ impl AssetVersion {
         match self {
             Self::V1 => &V1,
             Self::V2 => &V2,
+        }
+    }
+
+    /// Returns the shared `IB20` wire surface frozen for this version.
+    pub const fn abi(self) -> B20Abi {
+        match self {
+            Self::V1 => B20Abi::V1,
+            Self::V2 => B20Abi::V2,
         }
     }
 }
--- a/crates/common/precompiles/src/b20_stablecoin/versions.rs
+++ b/crates/common/precompiles/src/b20_stablecoin/versions.rs
@@ -9,7 +9,9 @@
 
 use base_common_genesis::BaseUpgrade;
 
-use crate::{PolicyAccounting, Stablecoin, StablecoinAccounting, StablecoinV1, StablecoinV2};
+use crate::{
+    B20Abi, PolicyAccounting, Stablecoin, StablecoinAccounting, StablecoinV1, StablecoinV2,
+};
 
 /// An activated version of the stablecoin B-20 precompile logic.
 ///
@@ -35,6 +37,18 @@ impl StablecoinVersion {
         match self {
             Self::V1 => &V1,
             Self::V2 => &V2,
+        }
+    }
+
+    /// Returns the shared `IB20` wire surface frozen for this version.
+    ///
+    /// The seize surface — the `SEIZE` pause feature and the seize functions/getters — arrived at
+    /// Cobalt with `V2`. `V1` decodes against the frozen Beryl surface so that discriminant and
+    /// those selectors stay unreachable, matching the pre-Cobalt binary.
+    pub const fn abi(self) -> B20Abi {
+        match self {
+            Self::V1 => B20Abi::V1,
+            Self::V2 => B20Abi::V2,
         }
     }
 }
--- a/crates/common/precompiles/src/common/abi/v1.rs
+++ b/crates/common/precompiles/src/common/abi/v1.rs
@@ -1,4 +1,7 @@
-//! ABI definition for the `IB20` interface.
+//! The shared `IB20` wire surface frozen at Beryl, the fork where the B-20 token precompiles
+//! activate.
+//!
+//! A new wire surface goes in a new `abi/vN.rs`; see [`super`].
 
 use alloy_sol_types::sol;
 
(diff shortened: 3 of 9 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/common/abi/v1.rs
```diff
@@ -1,4 +1,10 @@
-//! ABI definition for the `IB20` interface.
+//! The shared `IB20` wire surface frozen at Beryl, the fork where the B-20 token precompiles
+//! activate. This surface predates the seize surface: `PausableFeature` has no `SEIZE` variant, and
+//! the seize functions/getters are absent (the `TransferredFromSeizable` event already existed here,
+//! inert). Decoding pre-Cobalt calldata against it keeps the `SEIZE` discriminant (index 3) and the
+//! seize selectors unreachable, exactly as they were before Cobalt.
```
stephancill: remove

crates/common/precompiles/src/b20_asset/versions.rs
```diff
@@ -39,6 +39,18 @@ impl AssetVersion {
             Self::V2 => &V2,
         }
     }
+
+    /// Returns the shared `IB20` wire surface frozen for this version.
+    ///
+    /// The seize surface — the `SEIZE` pause feature and the seize functions/getters — arrived at
+    /// Cobalt with `V2`. `V1` decodes against the frozen Beryl surface so that discriminant and
+    /// those selectors stay unreachable, matching the pre-Cobalt binary.
```
stephancill: remove

rayyan224:
There's a versioned IB20, and then IB20Stablecoin right, and IB20Asset, since we are versioning this, now per contract. Let's remove it from the common what do you think

stephancill:
@rayyan224 not a fan of that tbh. duplicating the b20 logic between asset and stablecoin only adds more surface area for mistakes. recommending we also version the extension ABIs in lockstep with the shared abi

wdyt @robriks

robriks:
> @rayyan224 not a fan of that tbh. duplicating the b20 logic between asset and stablecoin only adds more surface area for mistakes. recommending we also version the extension ABIs in lockstep with the shared abi
> 
> wdyt @robriks

aligned, I prefer single-source-of-truth in `common` here. plus, I think it already lets us version per-contract (opt-in via the `abi()` map)

crates/common/precompiles/src/b20_stablecoin/versions.rs
```diff
@@ -37,6 +39,18 @@ impl StablecoinVersion {
             Self::V2 => &V2,
         }
     }
+
+    /// Returns the shared `IB20` wire surface frozen for this version.
+    ///
+    /// The seize surface — the `SEIZE` pause feature and the seize functions/getters — arrived at
+    /// Cobalt with `V2`. `V1` decodes against the frozen Beryl surface so that discriminant and
+    /// those selectors stay unreachable, matching the pre-Cobalt binary.
```
stephancill, line 47: this shouldn't know or care about seize

crates/common/precompiles/src/common/abi/v1.rs
```diff
@@ -137,83 +140,26 @@ sol! {
...
+
+    /// See [`super`] — the interface name reaches consensus data via `AbiDecodeFailed` on short
+    /// calldata, so it is pinned rather than left to a future rename. Both surfaces must keep the
+    /// `IB20` Rust name so this string never moves.
+    #[test]
+    fn interface_name_is_frozen() {
+        assert_eq!(IB20::IB20Calls::NAME, "IB20Calls");
+    }
 
+    /// Beryl's `PausableFeature` has exactly three variants. The `SEIZE` discriminant (index 3)
```
stephancill, line 157: this file shouldn't know or care about seize

stephancill:
Superseded by #4221, which lands the versioned common `IB20` surface (`common/abi/{v1,v2,b20_abi}` + `B20Abi`). Closing.

PR #4208, opened 2026-07-29
fix(batcher): align shadow parity streams by L2 timestamp at startup

xenoliss:
## Summary

`compare_ready` paired canonical and shadow batches positionally, so it assumed both queues started at the same point in L2 history. They do not: the shadow batcher begins submitting from its own cursor, so the two streams are offset at startup and every subsequent pair is compared against the wrong counterpart. The monitor then reports sustained divergence for batches that are in fact identical, which is what zeronet has been showing since the shadow batcher was enabled.

Batches are now paired by `start_timestamp`. Before the streams meet, the older side's prefix is discarded as a pre-alignment skip and no verdict is recorded, so `is_aligned()` stays unknown rather than false. Once a common timestamp is found, a gap on either side is a genuinely missing batch and is reported as such.

## Changes

- Pair decoded batches by `start_timestamp` instead of queue position, tracking whether the streams have met via `ParityState::has_aligned`.
- Skip and count the pre-alignment prefix on either side without recording a divergence or setting `last_result`.
- After alignment, report a timestamp gap as `missing_canonical` / `missing_shadow`, mark the comparison diverged, and log a warning; a side that is merely waiting for its next batch still records nothing.
- Add four counters (`canonical_pre_alignment_skips_total`, `shadow_pre_alignment_skips_total`, `missing_canonical_batches_total`, `missing_shadow_batches_total`) and an info log when the streams first align.
- Add tests for prefix skipping on both sides, staying unknown until a common timestamp arrives, waiting on a delayed batch, and reporting a missing batch on either side.

### DIFF
--- a/crates/batcher/service/src/shadow_parity.rs
+++ b/crates/batcher/service/src/shadow_parity.rs
@@ -340,6 +340,11 @@ impl ShadowParityMonitor {
                 evicted_channels = %evicted_channels,
                 "evicted stale shadow parity channels"
             );
+
+            let canonical = self.state.canonical.drain_ready_channels(&self.config.rollup_config);
+            let shadow = self.state.shadow.drain_ready_channels(&self.config.rollup_config);
+            ParitySide::Canonical.record_ingested(canonical);
+            ParitySide::Shadow.record_ingested(shadow);
         }
 
         let stats = self.state.compare_ready(block_number);
@@ -537,12 +542,7 @@ impl ShadowParityMonitor {
     /// Parse and ingest frame data from one DA payload.
     pub fn ingest_payload(&mut self, side: ParitySide, payload: &[u8], block_info: BlockInfo) {
         match self.state.ingest_payload(side, payload, block_info, &self.config.rollup_config) {
-            Ok(decoded) => {
-                if decoded.complete_channels > 0 {
-                    side.increment_complete_channels(decoded.complete_channels as u64);
-                    side.increment_batches(decoded.batches as u64);
-                }
-            }
+            Ok(decoded) => side.record_ingested(decoded),
             Err(e) => {
                 BatcherServiceMetrics::extraction_errors_total().increment(1);
                 warn!(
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/batcher/service/src/shadow_parity.rs
```diff
@@ -605,17 +620,63 @@ impl ParityState {
...
+            let canonical_start = self
+                .canonical
+                .batches
+                .front()
+                .expect("canonical queue is non-empty")
+                .start_timestamp;
+            let shadow_start =
+                self.shadow.batches.front().expect("shadow queue is non-empty").start_timestamp;
+
+            match canonical_start.cmp(&shadow_start) {
```
leopoldjoy, line 643: This merge assumes each queue is sorted by `start_timestamp`, but `drain_ready_channels` can append a later ready channel before an earlier incomplete one, so could we order or buffer batches by timestamp here to avoid false startup skips or missing-batch reports?

xenoliss: Good catch. I fixed this at the channel boundary rather than sorting batches after the fact. `drain_ready_channels` now drains only the contiguous ready prefix and stops at the first incomplete channel.

leopoldjoy (approved):

PR #4209, opened 2026-07-29
feat(proof): pin proofs to L2-activated upgrade schedules

PelleKrab:
## Summary

Align proof schedule commitments with `ProtocolVersions.activatedScheduleId(gameL2Timestamp)`.

Previously, the prover committed to the full upgrade schedule from the rollup config. Registering a future upgrade could therefore change the prover’s schedule ID for an existing game. This change commits only to the schedule prefix activated at the game’s deterministic L2 timestamp.

- Derive the schedule timestamp from the L2 genesis timestamp, genesis block, and block time.
- Compute the ProtocolVersions hash chain through the highest activated upgrade.
- Clear upgrades above that activated prefix before derivation.
- Add an optional `schedule_l2_block_number` to TEE and ZK proof requests.
  - Full-game/proposer requests default to the claimed L2 block.
  - Challenger subrange requests use the game’s `l2SequenceNumber()`.
- Thread the schedule block through the prover service, TEE host, and Succinct witness paths.
- Move the Beryl activation-admin requirement after schedule pinning so pre-Beryl proofs do not require it.
- Bump proposer and challenger proof-session namespaces to avoid reusing journals created with the previous schedule encoding.

## Validation

The boot loader now rejects schedules that cannot faithfully represent the contract commitment:

- zero L2 genesis timestamps;
- zero L2 block times;
- claims before genesis;
- schedule blocks before the claim;
- schedule timestamp arithmetic overflow;
- schedules with an activated Zenith upgrade.

Contract timestamp `0` remains the representation for an unscheduled upgrade, which is represented as `None` in the Rust rollup config.

## Testing

- `cargo test -p base-proof`
- `cargo clippy -p base-proof --all-targets --no-deps -- -D warnings`
- Targeted `cargo check --all-targets` for the affected prover-service, infra, and system-test packages
- `cargo fmt --all -- --check`
- `git diff --check`
- E2E tested activated-schedule pinning on a local devnet using the no-Nitro prover path.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proof/src/schedule_id.rs
```diff
@@ -1,27 +1,79 @@
...
 impl ScheduleId {
-    /// Derive the schedule ID from upgrade timestamps in canonical field order.
+    /// Returns the activated schedule prefix length at `l2_timestamp`.
+    pub fn activated_count(
+        upgrades: &UpgradeConfig,
+        l2_timestamp: u64,
+    ) -> Result<usize, ScheduleIdError> {
+        if let Some((upgrade, _)) = upgrades.iter().find(|(_, timestamp)| *timestamp == Some(0)) {
+            return Err(ScheduleIdError::ZeroActivationTimestamp { upgrade });
+        }
```
jackchuma: Normalize genesis-active zero timestamps. When an online host uses the standard Base rollup configuration, `regolith_time` is `Some(0)` on both mainnet and Sepolia, and that raw config is sent through `L2_ROLLUP_CONFIG_KEY`. This new validation returns `ZeroActivationTimestamp` before boot can complete, so normal TEE and ZK proof generation fails before derivation begins. The new test helper masks this by rewriting zero values to the genesis timestamp; production boot loading needs the same normalization (or another representation matching the contract schedule) instead of rejecting the config

crates/infra/zk-fork-dispute/src/checkpoint.rs
```diff
@@ -147,6 +147,7 @@ impl Checkpoint {
                 sequence_window: None,
                 l1_head: Some(l1_head),
                 intermediate_root_interval: Some(self.block_count),
+                schedule_l2_block_number: None,
```
jackchuma: Pin fork-dispute requests to the game's final block. When `--invalid-index` selects a checkpoint before the game’s final L2 block and an upgrade activates between those blocks, `None` makes boot pin the schedule to `self.target_block()` rather than the game block. This differs from `DisputeProofManager::build_zk_request`, which now uses `candidate.info.l2_block_number`, so this tool produces a proof with a different schedule commitment from the game it submits to and its `challenge`/`nullify` transaction cannot validate. Fetch and pass the game’s L2 block number here; the `v1` derived session should also be versioned so an old cached request is not reused or rejected as a payload collision

crates/proof/proof/src/schedule_id.rs
```diff
@@ -1,27 +1,91 @@
...
-            // unwrap_or_default() maps None (unscheduled) to 0, matching the onchain
-            // ProtocolVersions.scheduleId() which uses 0 for unscheduled entries.
-            schedule_id = Self::next_link(schedule_id, index as u64, timestamp.unwrap_or_default());
+        for (index, (_, configured_timestamp)) in upgrades.iter().enumerate().take(count) {
+            let timestamp = configured_timestamp.unwrap_or_default();
+            schedule_id = Self::next_link(schedule_id, index as u64, timestamp);
+        }
+
+        Ok(schedule_id)
+    }
```
jackchuma: Could we collapse this into `ScheduleId::pin`?

  `pin` is the only non-test consumer of `activated_count` and `from_upgrades`; the two methods only split up `pin`’s implementation and expand the public API. By the time `pin` calls them, it has already normalized zero timestamps for every contract-backed upgrade. `BootInfo::load` also rejects a zero genesis timestamp and Zenith before reaching this path, so the `ScheduleIdError` checks are not reachable in production.

  Keeping prefix selection, pruning, and hash construction inside `pin` would remove the extra public helpers, repeated full-schedule scans, error plumbing, and re-export. The existing tests can exercise the same behavior through `pin`

crates/proof/proof/src/schedule_id.rs
```diff
@@ -1,27 +1,57 @@
...
+    /// `link[i + 1] = keccak256(abi.encode(link[i], i, timestamp_i))`, committing the activated
+    /// schedule prefix in contract registration order with 0 for unscheduled entries.
+    ///
+    /// Genesis-active upgrades stored with the local-only zero timestamp convention are first
+    /// normalized to the config's genesis timestamp, matching the contract schedule
+    /// representation. Boot loading rejects a zero genesis timestamp before reaching this, so a
+    /// zero timestamp cannot survive normalization.
+    pub fn pin(config: &mut RollupConfig, l2_timestamp: u64) -> B256 {
+        for &upgrade in &BaseUpgrade::CONTRACT_VARIANTS {
+            if config.upgrades.activation_timestamp(upgrade) == Some(0) {
```
leopoldjoy, line 21: Could we avoid normalizing every genesis-active `Some(0)` to the same timestamp, since configs such as Zeronet contain multiple such forks while `ProtocolVersions` requires strictly increasing nonzero timestamps, making this hash impossible to match onchain?

crates/proof/prover-service/protocol/src/types.rs
```diff
@@ -185,6 +185,9 @@ pub struct ZkProofRequest {
     /// Optional intermediate output root interval.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub intermediate_root_interval: Option<u64>,
+    /// L2 block used to pin the upgrade schedule; defaults to the claimed block.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub schedule_l2_block_number: Option<u64>,
```
leopoldjoy, line 190: Could we gate schedule-pinned jobs on a protocol capability/version, since an older worker can deserialise this request while silently dropping `schedule_l2_block_number` and return a legacy full-schedule journal during a rolling deployment?

crates/infra/basectl/src/rpc/prover.rs
```diff
@@ -74,6 +74,7 @@ impl ProofFinalizeRequest {
                     sequence_window: self.sequence_window,
                     l1_head: self.l1_head,
                     intermediate_root_interval: self.intermediate_root_interval,
+                    schedule_l2_block_number: None,
```
leopoldjoy, line 79: Could we also bump Basectl's proof-session namespace, since `None` is omitted and leaves the deterministic request unchanged, allowing a redeployment that retains either prover-service DB or SP1 cluster state to replay a pre-pinning journal?

crates/proof/proof/src/boot.rs
```diff
@@ -353,7 +348,63 @@ impl BootInfo {
...
+        }
+        let blocks_since_genesis = schedule_l2_block_number - rollup_config.genesis.l2.number;
+        let l2_schedule_timestamp = blocks_since_genesis
+            .checked_mul(rollup_config.block_time)
+            .and_then(|offset| rollup_config.genesis.l2_time.checked_add(offset))
+            .ok_or(OracleProviderError::L2ScheduleTimestampOverflow {
+                schedule_block: schedule_l2_block_number,
+            })?;
+
+        if rollup_config.upgrades.base.zenith.is_some() {
```
leopoldjoy: Could we reject Zenith only when it is active at `l2_schedule_timestamp` and clear a future Zenith instead, since merely scheduling Zenith currently makes existing pre-Zenith games unprovable?

crates/proof/proof/src/schedule_id.rs
PelleKrab: I am about to push a PR for `ProtocolVersions` to make the rule for new timestamps >= instead of >.

crates/proof/proof/src/boot.rs
PelleKrab: Good catch

crates/proof/proof/src/schedule_id.rs
PelleKrab: https://github.com/base/contracts/pull/389

crates/proof/host/src/kv/boot.rs
```diff
@@ -45,6 +45,9 @@ impl KeyValueStore for BootKeyValueStore {
                 Some(self.cfg.request.intermediate_block_interval.to_be_bytes().to_vec())
             }
             L1_HEAD_NUMBER_KEY => Some(self.cfg.request.l1_head_number.to_be_bytes().to_vec()),
+            L2_SCHEDULE_BLOCK_NUMBER_KEY => {
+                Some(self.cfg.request.schedule_l2_block_number.unwrap_or(0).to_be_bytes().to_vec())
```
refcell, line 50: Right now this works since all base chains use `0` as the default genesis L2 block number. I believe there isn't anything preventing this from being the case though. RollupConfig doesn't have any restrictions on non-zero l2 genesis block numbers. Perhaps it's worth leaving a comment here with a SAFETY note. e.g.

```rust
                // SAFETY: All Base Chains use `0` as their genesis L2 block number.
                Some(self.cfg.request.schedule_l2_block_number.unwrap_or(0).to_be_bytes().to_vec())
```

crates/proof/proof/src/boot.rs
```diff
@@ -353,7 +348,68 @@ impl BootInfo {
...
+        // Missing or zero values default to the claimed block.
+        let schedule_l2_block_number = match oracle
+            .get(PreimageKey::new_local(L2_SCHEDULE_BLOCK_NUMBER_KEY.to()))
+            .await
+        {
+            Ok(bytes) => {
+                let value = u64::from_be_bytes(
+                    bytes.as_slice().try_into().map_err(OracleProviderError::SliceConversion)?,
+                );
+                if value == 0 { l2_claim_block } else { value }
```
refcell, line 360: I think there is one critical implication with this PR that's worth calling out.

If there are old prover-service servers serving legacy requests, and an upgrade is scheduled, the proof will fail onchain. Consider:
- An upgrade activates at block 950
- Correct schedule at block 1000 includes the upgrade
- subrange 800-900 is being proved, so block 900 _excludes_ the upgrade
- the old prover server *isn't* aware of the `schedule_l2_block_number` field (before this PR)
- this defaults to `0`, which is the claimed block `900`
- the fallback of `900` is correct for proposer proofs, but not challenger subranges
- the schedule id is computed incorrectly

I may be smooth-braining this but there are a few issues from what I can tell:
- there is no check for whether a queued job is legacy or new
- new workers can claim old jobs, but an old challenger job lacks the game schedule block

refcell: Ostensibly, we can mitigate this by performing a careful rollout:
- Stop accepting legacy requests
- Prevent ProtocolVersions registration/schedule changes
- Let old servers and old workers drain all pending/running legacy jobs
- submit or consume those proofs while legacy stores are still valid
- clear any unfinished legacy jobs
- upgrade all servers
- upgrade workers and proof images
- deploy v2 request producers
- reissue unfinished work under v2 session IDs

But I'd need Jack and Leopold to confirm this since I have an incomplete understanding of the proofs system

jackchuma: This makes sense as a best-practice way to roll this change out, although the risk is relatively low since we don't expect frequent challenger proof jobs (if any at all). what about if we persist a job version in the prover service db and then hosts specify what version they are trying to claim? Then we can roll this out by just deploying the prover service first followed by everything else without worrying about additional ordering too much

jackchuma (dismissed):

jackchuma (approved):

PR #4211, opened 2026-07-29
feat(policy): add compositePolicyChildIds view for composite child sets

rayyan224:
## Summary

Composite policies (UNION/INTERSECT) shipped in #4159 with no on-chain read path for their child set. The only ways to recover it today are decoding the `CompositePolicyUpdated` event or a raw `eth_getStorageAt` against slot 4. This adds the missing getter to the V2 (Cobalt) wire surface.

```solidity
function compositePolicyChildIds(uint64 policyId) external view returns (uint64[] memory);
```

Spec counterpart (interface + reference mock + tests): **base/base-std#183**.

## Semantics

Total — it never reverts, matching the rest of the read surface. An empty array is the array-typed analogue of `Address::ZERO`:

| Input | Returns |
|---|---|
| Existing composite | Stored child set, in the order last written |
| Simple policy (ALLOWLIST / BLOCKLIST) | `[]` |
| Built-in sentinel | `[]` |
| Well-formed but uncreated ID | `[]` |
| Malformed ID (type byte > INTERSECT) | `[]` |

The gate byte alone decides whether a child set can exist, so `is_composite` short-circuits every non-composite before any storage read. Returning `[]` rather than reverting `IncompatiblePolicyType` is safe because **an empty return is unambiguous in practice**: both write paths reject a set outside `[2, 4]`, and there is no clear-the-list path, so a composite that exists always holds at least two children. Tests pin that invariant on both sides.

## Why this extends V2 in place rather than minting a V3

Adding a selector changes `V2_ABI_FINGERPRINT` (`047d1fce` → `cc5f508d`) and moves the frozen selector count 13 → 14, which normally would mean a new frozen `abi/v3.rs`. That isn't required here: **Cobalt has never activated on any network.** `BaseUpgrade::LATEST` is `Beryl`, and every real chain config leaves `cobalt` unset. No historical execution depends on the old value, and the fingerprint exists to catch *unintended* drift, not to freeze an unactivated fork.

The re-blessing is documented inline at the constant, including the rule that once Cobalt ships, a change here means `abi/v3.rs` instead.

Worth a reviewer's explicit sign-off, since it's the one consensus-relevant judgement call in the PR.

## Fork gating

The V1 surface does not declare the selector, so Beryl rejects it as `UnknownFunctionSelector`. This matters more than usual for a view: views bypass the *activation* gate but not the *wire* gate, and `golden_composite_child_ids_selector_unknown_in_v1` pins that. The `PolicyRegistryLogic` default impl reverts to match its sibling composite methods; it is unreachable for the same reason.

## Independence from #4206

This branches off `main`, not off `feat/factory-abi-versioning`. The `AbiFingerprint` extraction in that PR is behavior-preserving, so `cc5f508d` is identical whether computed by main's local `abi_fingerprint` helper or the extracted `AbiFingerprint::compute` — verified by running the suite green on both bases. The two PRs touch `policy/abi/mod.rs` and will need a trivial merge resolution depending on order, but neither blocks the other.

## Testing

`cargo test -p base-common-precompiles --all-features` — 958 passed, 0 failed. `cargo clippy --all-features --all-targets` clean. `cargo +nightly fmt --check` clean.

New coverage:
- 5 unit tests in `logic/v2.rs`: creation-order readback, `update_composite` tracking, view/event agreement, the four empty cases, and the never-empty invariant.
- 3 V2 golden tests: wire-level readback, empty-for-non-composites, and a read before activation (confirming the view bypasses the activation gate).
- 1 V1 golden test: the selector is unknown at Beryl.
- Both selector-coverage exhaustiveness maps updated.

Generated with Claude Code

### DIFF
--- a/crates/common/precompiles/src/policy/abi/mod.rs
+++ b/crates/common/precompiles/src/policy/abi/mod.rs
@@ -55,10 +55,6 @@ mod tests {
         b256!("1ae189209c8c4875de2caa707322ea74f0d1f3e74a1104ecee6884e8984415da");
 
     /// Absolute wire fingerprint for Cobalt's (canonical) surface.
-    ///
-    /// Re-blessed when `compositePolicyChildIds` joined the surface — legitimate only because
-    /// Cobalt has never activated (`BaseUpgrade::LATEST` is Beryl, every real chain config leaves
-    /// `cobalt` unset). Once Cobalt ships, a change here means a new `abi/v3.rs` instead.
     const V2_ABI_FINGERPRINT: B256 =
         b256!("cc5f508d2cacd9b729ddbd7a7d9ed3788929fdbb237c286e5a9d8961a90cf656");
 
--- a/crates/common/precompiles/src/policy/abi/mod.rs
+++ b/crates/common/precompiles/src/policy/abi/mod.rs
@@ -28,6 +28,7 @@ impl IPolicyRegistry::IPolicyRegistryCalls {
             Self::policyExists(_) => "policy.policyExists",
             Self::policyAdmin(_) => "policy.policyAdmin",
             Self::pendingPolicyAdmin(_) => "policy.pendingPolicyAdmin",
+            Self::compositePolicyChildIds(_) => "policy.compositePolicyChildIds",
         }
     }
 }
@@ -54,8 +55,13 @@ mod tests {
         b256!("1ae189209c8c4875de2caa707322ea74f0d1f3e74a1104ecee6884e8984415da");
 
     /// Absolute wire fingerprint for Cobalt's (canonical) surface.
+    ///
+    /// Re-blessed when `compositePolicyChildIds` joined the surface. Legitimate only because
+    /// Cobalt has never activated on any network (`BaseUpgrade::LATEST` is Beryl, and every real
+    /// chain config leaves `cobalt` unset), so no historical execution depends on the old value.
+    /// Once Cobalt ships, a change here means a new `abi/v3.rs` instead.
     const V2_ABI_FINGERPRINT: B256 =
-        b256!("047d1fceaa9beac82fd473d474bb7f3b6dcd720c9e62c86facc75ef6dd611631");
+        b256!("cc5f508d2cacd9b729ddbd7a7d9ed3788929fdbb237c286e5a9d8961a90cf656");
 
     /// Keccak of sorted call selectors, then sorted event topic0s, then sorted error selectors,
     /// then `PolicyType::COUNT`. Order is fixed so a single pin catches any wire-surface edit.
--- a/crates/common/precompiles/src/policy/abi/mod.rs
+++ b/crates/common/precompiles/src/policy/abi/mod.rs
@@ -56,10 +56,9 @@ mod tests {
 
     /// Absolute wire fingerprint for Cobalt's (canonical) surface.
     ///
-    /// Re-blessed when `compositePolicyChildIds` joined the surface. Legitimate only because
-    /// Cobalt has never activated on any network (`BaseUpgrade::LATEST` is Beryl, and every real
-    /// chain config leaves `cobalt` unset), so no historical execution depends on the old value.
-    /// Once Cobalt ships, a change here means a new `abi/v3.rs` instead.
+    /// Re-blessed when `compositePolicyChildIds` joined the surface — legitimate only because
+    /// Cobalt has never activated (`BaseUpgrade::LATEST` is Beryl, every real chain config leaves
+    /// `cobalt` unset). Once Cobalt ships, a change here means a new `abi/v3.rs` instead.
     const V2_ABI_FINGERPRINT: B256 =
         b256!("cc5f508d2cacd9b729ddbd7a7d9ed3788929fdbb237c286e5a9d8961a90cf656");
 
--- a/crates/common/precompiles/src/policy/abi/v2.rs
+++ b/crates/common/precompiles/src/policy/abi/v2.rs
@@ -66,6 +66,9 @@ sol! {
         function policyExists(uint64 policyId) external view returns (bool);
         function policyAdmin(uint64 policyId) external view returns (address);
         function pendingPolicyAdmin(uint64 policyId) external view returns (address);
+        /// Introduced in V2 (Cobalt). Returns the live child set of a composite policy, or an
+        /// empty array for simple policies, built-in sentinels, unknown IDs, and malformed IDs.
+        function compositePolicyChildIds(uint64 policyId) external view returns (uint64[] memory);
     }
 }
 
@@ -117,10 +120,11 @@ mod tests {
             IPolicyRegistry::policyExistsCall::SELECTOR,
             IPolicyRegistry::policyAdminCall::SELECTOR,
             IPolicyRegistry::pendingPolicyAdminCall::SELECTOR,
+            IPolicyRegistry::compositePolicyChildIdsCall::SELECTOR,
         ];
         expected.sort_unstable();
 
-        assert_eq!(selectors.len(), 13);
+        assert_eq!(selectors.len(), 14);
         assert_eq!(selectors, expected);
     }
 }
--- a/crates/common/precompiles/src/policy/abi/v2.rs
+++ b/crates/common/precompiles/src/policy/abi/v2.rs
@@ -66,8 +66,7 @@ sol! {
         function policyExists(uint64 policyId) external view returns (bool);
         function policyAdmin(uint64 policyId) external view returns (address);
         function pendingPolicyAdmin(uint64 policyId) external view returns (address);
-        /// Introduced in V2 (Cobalt). Returns a composite policy's child set, in the order it
-        /// was last written. Empty for anything else. Never reverts.
+        /// Introduced in V2 (Cobalt). Read function for composite policy child IDs.
         function compositePolicyChildIds(uint64 policyId) external view returns (uint64[] memory);
     }
 }
--- a/crates/common/precompiles/src/policy/abi/v2.rs
+++ b/crates/common/precompiles/src/policy/abi/v2.rs
@@ -66,8 +66,8 @@ sol! {
         function policyExists(uint64 policyId) external view returns (bool);
         function policyAdmin(uint64 policyId) external view returns (address);
         function pendingPolicyAdmin(uint64 policyId) external view returns (address);
-        /// Introduced in V2 (Cobalt). Returns the live child set of a composite policy, or an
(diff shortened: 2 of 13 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/policy/abi/mod.rs
```diff
@@ -54,8 +55,12 @@ mod tests {
         b256!("1ae189209c8c4875de2caa707322ea74f0d1f3e74a1104ecee6884e8984415da");
 
     /// Absolute wire fingerprint for Cobalt's (canonical) surface.
+    ///
```
rayyan224: Lol "Re-Blessed" 😆

crates/common/precompiles/src/policy/abi/v2.rs
```diff
@@ -66,6 +66,9 @@ sol! {
         function policyExists(uint64 policyId) external view returns (bool);
         function policyAdmin(uint64 policyId) external view returns (address);
         function pendingPolicyAdmin(uint64 policyId) external view returns (address);
+        /// Introduced in V2 (Cobalt). Returns a composite policy's child set, in the order it
+        /// was last written. Empty for anything else. Never reverts.
```
rayyan224: Just Intrdouced V2 Cobalt is a good enough comment

crates/common/precompiles/src/policy/logic/interface.rs
```diff
@@ -94,4 +94,11 @@ pub trait PolicyRegistryLogic<S: PolicyAccounting> {
 
     /// Returns the staged pending admin for `policy_id`, or `Address::ZERO` if none.
     fn pending_policy_admin(&self, storage: &S, policy_id: u64) -> Result<Address>;
+
+    /// Returns a composite policy's child set, empty for anything else.
+    /// Composite support is a V2 feature; the V1 surface omits the selector, so this default is
+    /// unreachable and reverts to match the other composite defaults above.
+    fn composite_policy_child_ids(&self, _storage: &S, _policy_id: u64) -> Result<Vec<u64>> {
+        Err(BasePrecompileError::Revert(Bytes::new()))
```
rayyan224: There should be a macro called "unreachable" or "Unsupported used for others use that"

crates/common/precompiles/src/policy/logic/interface.rs
```diff
@@ -94,4 +94,11 @@ pub trait PolicyRegistryLogic<S: PolicyAccounting> {
 
     /// Returns the staged pending admin for `policy_id`, or `Address::ZERO` if none.
     fn pending_policy_admin(&self, storage: &S, policy_id: u64) -> Result<Address>;
+
+    /// Returns a composite policy's child set, empty for anything else.
```
rayyan224: /// (V2) Returns a composite policy's child set, empty for anything else. keep comment to this

crates/common/precompiles/src/policy/logic/v2.rs
```diff
@@ -569,6 +569,16 @@ impl<S: PolicyAccounting> PolicyRegistryLogic<S> for PolicyRegistryV2 {
         }
         storage.read_pending_admin(policy_id)
     }
+
+    fn composite_policy_child_ids(&self, storage: &S, policy_id: u64) -> Result<Vec<u64>> {
+        // Never reverts, matching the rest of the read surface. The gate byte alone decides
```
rayyan224: Remove this comment

crates/common/precompiles/src/policy/logic/v2.rs
```diff
@@ -1349,6 +1359,87 @@ mod tests {
...
+        let c = create_allowlist(&mut rt);
+        let id = create_intersect(&mut rt, vec![a, b]);
+        assert_eq!(LOGIC.composite_policy_child_ids(&rt, id).unwrap(), vec![a, b]);
+        set_caller(&mut rt, ADMIN);
+        LOGIC.update_composite(&mut rt, id, vec![b, c]).unwrap();
+        // Full replacement, not a merge — `a` is gone rather than retained as a stale tail.
+        assert_eq!(LOGIC.composite_policy_child_ids(&rt, id).unwrap(), vec![b, c]);
+    }
+
+    /// The view mirrors the event exactly, so indexers reconciling a log against a live read
```
rayyan224: Remove comment self explanatory

crates/common/precompiles/src/policy/logic/v2.rs
```diff
@@ -1349,6 +1359,87 @@ mod tests {
...
+        let before = rt.events.len();
+        let id = create_union(&mut rt, vec![a, b]);
+        let emitted =
+            IPolicyRegistry::CompositePolicyUpdated::decode_log_data(&rt.events[before + 2])
+                .unwrap()
+                .childPolicyIds;
+        assert_eq!(LOGIC.composite_policy_child_ids(&rt, id).unwrap(), emitted);
+    }
+
+    /// Every non-composite input collapses to an empty set rather than reverting, matching the
```
rayyan224: Remove comment

crates/common/precompiles/src/policy/logic/v2.rs
```diff
@@ -1349,6 +1359,87 @@ mod tests {
...
+            uncreated_union,
+        ] {
+            assert!(
+                LOGIC.composite_policy_child_ids(&rt, policy_id).unwrap().is_empty(),
+                "expected an empty child set for policy {policy_id}"
+            );
+        }
+    }
+
+    /// A simple policy and a never-created composite are indistinguishable from the return value
```
rayyan224: This is retesting, MIN and MAX polices, if there's already a test for that, remove this one

crates/common/precompiles/src/policy/abi/mod.rs
rayyan224: Fair. Dropped the whole paragraph — the fingerprint const speaks for itself.

crates/common/precompiles/src/policy/abi/v2.rs
rayyan224: Done — cut to `/// Introduced in V2 (Cobalt).`

crates/common/precompiles/src/policy/logic/interface.rs
rayyan224: Done, used your wording verbatim.

crates/common/precompiles/src/policy/logic/interface.rs
rayyan224: Done — using `reject_frozen_selector!()`.

Heads up on two side effects, happy to revert either:

1. The macro was defined locally in `b20_asset/mod.rs` with no `pub(crate) use`, so it was only visible inside the `b20_asset` subtree. Promoted it to the shared `macros.rs` next to `base_precompile!`/`decode_precompile_call!` and added explicit imports at the two existing call sites (`b20_asset/accounting.rs`, `b20_asset/logic/interface.rs`). Also degeneralized its doc comment, which named `[Asset]`/`[AssetAccounting]` and `golden_v2_selectors_unknown_at_v1` — none of which resolve from the crate root.

2. This changes the error, not just the spelling: `UnknownFunctionSelector([0u8;4])` instead of `Revert(Bytes::new())`. Unreachable either way — `dispatch.rs:68` rejects the selector on raw calldata at V1 — and `UnknownFunctionSelector` is what dispatch actually returns, so it is arguably more honest. But it leaves `create_composite_policy` and `update_composite` above still hand-rolling the empty revert. Want those switched too, or leave them since they are outside this PR?

crates/common/precompiles/src/policy/logic/v2.rs
rayyan224: Removed.

crates/common/precompiles/src/policy/logic/v2.rs
rayyan224: Removed.

crates/common/precompiles/src/policy/logic/v2.rs
rayyan224: Removed.

crates/common/precompiles/src/policy/logic/v2.rs
rayyan224: Good catch — `create_composite_child_count_out_of_range_reverts` already covers both bounds. Deleted the test.

crates/common/precompiles/src/policy/logic/interface.rs
```diff
@@ -94,4 +94,9 @@ pub trait PolicyRegistryLogic<S: PolicyAccounting> {
 
     /// Returns the staged pending admin for `policy_id`, or `Address::ZERO` if none.
     fn pending_policy_admin(&self, storage: &S, policy_id: u64) -> Result<Address>;
+
+    /// (V2) Returns a composite policy's child set, empty for anything else.
+    fn composite_policy_child_ids(&self, _storage: &S, _policy_id: u64) -> Result<Vec<u64>> {
```
robriks, line 97: should the sibling methods `create_composite_policy` & `update_composite` use `reject_frozen_selector!()` as well?

robriks: (if so, non-blocking for this PR and can be a follow-up; glad we got this in!)

robriks (dismissed):

crates/common/precompiles/src/policy/logic/interface.rs
rayyan224: yes it should!

robriks (dismissed):

robriks (approved):

PR #4214, opened 2026-07-30
feat(txpool): make the builder wire transaction type pluggable

refcell:
## Motivation

`ValidatedTransaction` is the payload of `base_insertValidatedTransaction`, the endpoint mempool nodes use to forward transactions to a builder. Its shape is fixed, so a downstream node build cannot carry additional per-transaction data over that hop without forking the forwarder (~600 lines, since `RateLimiter` and `BufferedTransaction` are private and would drift permanently).

This adds a seam so the payload type is chosen by whoever assembles the node.

## What changed

`ValidatedTransaction<E = NoExtensions>` gains a flattened extension payload. Downstream builds implement `ValidatedTransactionExtensions<T>` — one method per direction, `extract` on the forwarder side and `apply` on the builder side — and register the generic monomorphizations.

Nothing about any specific extension is named in this crate; it only ever mentions `NoExtensions`.

Threaded through `BuilderApi<E>` / `BuilderApiImpl<P, E = NoExtensions>` and `Forwarder<T, E>` / `SpawnedForwarder::spawn_with_extensions`.

This follows the shape already established by `CandidateSource<T>` + `FlashblocksServiceBuilder<S = DefaultCandidateSource>`: seam trait, defaulted type parameter, existing call sites untouched.

## Wire compatibility

`NoExtensions` encodes to **exactly the same bytes** as the previous layout. `no_extensions_encoding_matches_legacy_layout` asserts this against a struct mirroring the old field list. Both compat directions are covered: an old reader accepts a payload carrying extensions, and an extension-aware reader accepts a payload without them.

## Two non-obvious constraints, documented in-place

- **`new` is defined only on the `NoExtensions` monomorphization.** A single generic constructor leaves `E` unconstrained at every call site (`E0282`) — type-parameter defaults do not drive inference through associated-function calls. `with_extensions` covers the generic case.
- **`spawn` delegates to `spawn_with_extensions`.** Generic parameters cannot carry defaults on a function.

Also: the flattened field is deliberately *not* `#[serde(default)]`, which would force an `E: Default` bound onto the generated `Deserialize` impl and break the `#[rpc]` expansion.

## No call sites changed

`crates/builder/core/src/extension.rs:13` and `etc/systems/src/l2/in_process_builder.rs:155` still read `BuilderApiImpl::new(ctx.pool().clone())` verbatim. `ForwarderConfig` is left non-generic, and the batch send is untouched — it addresses the method by name, so only `Serialize` is required of the payload.

## Testing

- `cargo test -p base-execution-txpool --lib` — 16 passed, 0 failed. 6 new wire tests (encoding, both compat directions, `u64::MAX` through flatten buffering) and 3 new RPC tests proving the hook is *invoked*: a failing extension surfaces as `InvalidParams` before reaching the pool, a passing one falls through to insertion, and both monomorphizations build `RpcModule`s.
- `cargo clippy` clean on `base-execution-txpool`, `base-builder-core`, `base-tx-forwarding`.
- `cargo check --all-targets` clean on those three plus `base-system-tests` and `base-builder-bin`.
- Full-workspace check could not complete locally: `risc0-sys` fails its build script with `xcrun: error: unable to find utility "metal"`. Environmental (no full Xcode), pre-existing, unrelated — `base-execution-txpool` checks clean before it.

## Follow-up, not in this PR

This makes the **wire** type pluggable. For extension data to persist into block building, `BasePooledTransaction` also needs a metadata slot — its fields are private and it has no such slot today. That is a separate change to a type used workspace-wide, and it needs an audit of `crates/execution/payload/src/builder.rs:778`, where `into_consensus()` discards the pooled-transaction wrapper inside the build loop.

### DIFF
--- a/crates/builder/core/tests/rpc.rs
+++ b/crates/builder/core/tests/rpc.rs
@@ -6,7 +6,7 @@ use alloy_primitives::{Address, Bytes, Signature, TxKind, U256};
 use alloy_rpc_client::RpcClient;
 use base_builder_core::BuilderApiExtension;
 use base_common_consensus::{BaseTransactionSigned, BaseTypedTransaction, TxDeposit};
-use base_execution_txpool::ValidatedTransaction;
+use base_execution_txpool::{NoExtensions, ValidatedTransaction};
 use base_node_runner::test_utils::TestHarness;
 use base_test_utils::Account;
 
@@ -69,6 +69,7 @@ async fn test_insert_validated_deposit_tx() -> eyre::Result<()> {
         max_block_number: None,
         min_timestamp: None,
         max_timestamp: None,
+        extensions: NoExtensions {},
     };
 
     let result: Result<(), _> =
@@ -99,6 +100,7 @@ async fn test_insert_validated_eip1559_tx() -> eyre::Result<()> {
         max_block_number: None,
         min_timestamp: None,
         max_timestamp: None,
+        extensions: NoExtensions {},
     };
 
     // EIP-1559 transactions are supported by the pool
@@ -122,6 +124,7 @@ async fn test_insert_invalid_tx_fails() -> eyre::Result<()> {
         max_block_number: None,
         min_timestamp: None,
         max_timestamp: None,
+        extensions: NoExtensions {},
     };
 
     let result: Result<(), _> =
--- a/crates/execution/txpool/README.md
+++ b/crates/execution/txpool/README.md
@@ -10,6 +10,44 @@ Extends Reth's transaction pool with Base-specific validation and ordering for t
 Also includes a `Consumer` for processing mempool events, a `Forwarder` for relaying transactions,
 and a `BuilderApiImpl` for builder-specific pool management.
 
+### Pluggable builder wire format
+
+`ValidatedTransaction<E>` is the payload of `base_insertValidatedTransaction`, the endpoint mempool
+nodes use to forward transactions to a builder. `E` carries additional wire fields and defaults to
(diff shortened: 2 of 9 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

BrianBland (approved):

PR #4218, opened 2026-07-30
refactor(asset): decode calldata against the wire surface frozen at each fork

robriks:
Ports the version-scoped ABI shape from #4193 (policy) and #4206 (factory) to `b20_asset`, replacing the hand-written fork gate for the ERC-8056 scheduled-multiplier surface (BOP-429) with per-version frozen wire surfaces reached through a single version key.

## Not a bug fix

Unlike #4193 (which fixed a live enum-widening divergence), **this changes no behavior.** The pre-decode gate at `dispatch.rs` already returned the correct historical `UnknownFunctionSelector` bytes for the 8 scheduled-multiplier selectors at V1. This is a pure structural refactor: the gate is deleted and rejection instead falls out of the frozen V1 surface not declaring those selectors. **No golden storage root moved and no revert byte changed for any input** — every existing test passes with zero re-blessing.

## The change

`abi.rs` becomes `abi/{mod,v1,v2}.rs`. `v2.rs` is the canonical Cobalt surface (20 fns); `v1.rs` is the frozen Beryl surface (12 fns). `mod.rs` is pure glue (re-exports + docs); logic lives in the `vN` modules.

```rust
pub enum AssetAbi { V1, V2 }
impl AssetVersion { pub const fn abi(self) -> AssetAbi { … } }
```

`route()` decodes the asset branch via `version.abi().decode_asset(calldata)` and the hand-written fork gate is gone. At V1 the 8 scheduled selectors are absent from the frozen surface, so the asset branch is skipped and they fall through to the **disjoint** inherited `IB20` decode → byte-identical `UnknownFunctionSelector`.

## `IB20` scope note

The inherited `IB20` surface is **not** versioned here — it has not moved across a fork. A fingerprint tripwire (`ib20_abi_fingerprint_is_pinned`) fails CI the first time `IB20` — or its `PausableFeature` enum — grows. Follow-up common composition lives in a stacked PR on this branch.

## Verification

- `cargo test -p base-common-precompiles --features test-utils` — all pass, **0 golden re-bless**.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/b20_asset/abi/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Keccak of sorted call selectors, then sorted event topic0s, then sorted error selectors,
+    /// then an enum-count byte, then enum ordinals. Order is fixed so a single pin catches any
+    /// wire-surface edit.
+    ///
+    /// `IB20Asset` declares no enum, so the count byte is `0` and there are no ordinals. The layout
+    /// deliberately matches the shared `AbiFingerprint::compute` helper introduced in #4206 (called
+    /// here as if with `enum_count = 0` and empty ordinals); once that lands these fingerprints
+    /// should call it directly, and the matching layout means no re-bless is needed.
+    fn abi_fingerprint(
```
rayyan224: Let's re use the common function for this

robriks: whoops I think yhou're right, Stephan tells me we don't want logic in mod.rs files

crates/common/precompiles/src/b20_asset/abi/mod.rs
stephancill, line 1: logic in mod.rs is an anti pattern for this repo afaik

rayyan224: Yep no logic in mod

crates/common/precompiles/src/b20_asset/abi/mod.rs
robriks: oh I see what you mean. it's a placeholder until 4206 merges; will rebase once it's available in main branch !

crates/common/precompiles/src/b20_asset/abi/mod.rs
robriks: [be45e9b](https://github.com/base/base/pull/4218/commits/be45e9b47ff118ce7a1dfa5b5e732fffa71d0fc1)

crates/common/precompiles/src/b20_asset/abi/mod.rs
robriks: [1d029e3](https://github.com/base/base/pull/4218/commits/1d029e32135af7c72a871ecdfd4449dc7f92a771) !!

robriks:
Superseding this in favor of #4221, which is git-stacked on this branch and carries the same `b20_asset` wire-surface versioning **plus** the AbiPair pattern — composing the frozen common `IB20`/`B20Abi` with the versioned asset extension via `AssetVersion::abi() -> AssetAbiPair`, and routing through the crate-private `AssetCall`. Closing to avoid a duplicate/competing change on the same files.

Follow-up (tracked): #4221 does not yet include the V2 double-decode elision from this PR — its `AssetAbi::decode` still decodes twice on the current-fork (V2) hot path, where the frozen surface *is* canonical. I'll port that as a small follow-up onto `main` once #4221 lands. The two doc-comment trims and the shared-`AbiFingerprint` wiring are already covered by #4221.

PR #4221, opened 2026-07-30
refactor(asset): compose frozen common IB20 into AssetAbiPair

rayyan224:
## Summary

- Freeze shared `IB20` as `B20Abi::V1` under `common/abi/{mod,v1}.rs` (no independent `from_base_upgrade`).
- Join each asset version's extension + common surface via `AssetAbiPair` from `AssetVersion::abi()`.
- Decode asset-first, then common, with no fallthrough after a malformed known selector; dispatch routes through crate-private `AssetCall`.

Alloy `sol!` cannot inherit interfaces, so the composite dialable surface is composed in Rust. Stablecoin still uses live canonical `IB20`.

Stacked on #4218.

## Test plan

- [x] `cargo test -p base-common-precompiles --features test-utils --lib versions`
- [ ] `cargo test -p base-common-precompiles --features test-utils`
- [ ] Confirm no golden re-bless / revert-byte changes

PR #4229, opened 2026-07-31
perf(asset): elide the dead V2 canonical re-decode in the wire surfaces

robriks:
## Summary

Follow-up to #4221 (which subsumed #4218). Both `AssetAbi::decode` and `B20Abi::decode` validate `calldata` against the frozen surface and then re-decode against the canonical surface. That double-decode is only load-bearing where the frozen surface **differs** from canonical (`V1`). At **V2** the frozen surface *is* canonical — `IB20Asset` aliases `IB20AssetV2`, and `IB20` aliases `IB20V2` — so every V2 call was decoding the identical type twice on the current-fork hot path.

This gates the frozen pre-validation behind an exhaustive `match` so only `V1` pays for it:

```rust
match self {
    Self::V1 => self.abi_decode_validate(calldata, selector)?,
    Self::V2 => {}
}
IB20…::abi_decode_validate(calldata).map_err(…)
```

The `match` is exhaustive on purpose: a future version whose frozen surface diverges from canonical is forced to add its own pre-validating arm (rather than silently inheriting a `V2 => {}` skip once canonical moves off V2).

Applied to **both** surfaces because the redundancy is identical, and `B20Abi::decode` is the hotter of the two — it covers every inherited ERC-20 call (`transfer`/`balanceOf`/`approve`/`mint`/…) on an asset token at Cobalt+, not just the asset-specific multiplier ops.

## Behavior

Behavior-preserving. At V2 the pre-validate and the canonical decode were the *same function on the same type*, so skipping the first cannot change the returned call or the `AbiDecodeFailed` error bytes. This is the elision originally in #4218 that #4221 didn't carry over, now applied to the merged `AbiPair` structure.

## Verification

- `cargo test -p base-common-precompiles --features test-utils` — all pass (623 lib + integration), **0 failures, 0 golden re-bless**.
- `cargo clippy -p base-common-precompiles --all-targets --features test-utils` clean; `cargo +nightly fmt --check` clean.

## Test plan

- [x] Unit + golden suite green without `BLESS_GOLDEN`
- [x] clippy + nightly fmt clean

---

Linear: BOP-490

### DIFF
--- a/crates/common/precompiles/src/b20_asset/versions.rs
+++ b/crates/common/precompiles/src/b20_asset/versions.rs
@@ -100,6 +100,14 @@ impl AssetAbi {
     }
 
     /// Decodes `calldata` into a routable asset call, gated on this extension surface.
+    ///
+    /// Where the frozen surface differs from canonical (`V1`), the calldata is validated against the
+    /// frozen surface first so a rejection carries that version's consensus error bytes, then
+    /// re-decoded into the canonical enum for routing. Where the frozen surface *is* canonical
+    /// (`V2`, which `IB20Asset` aliases), that pre-validation would decode the identical type twice,
+    /// so it is skipped and the single canonical decode suffices. The `match` is exhaustive so a
+    /// later version whose frozen surface diverges from canonical must add its own pre-validating
+    /// arm.
     fn decode(self, calldata: &[u8]) -> Result<IB20Asset::IB20AssetCalls> {
         let Some(selector) = calldata.first_chunk::<4>().copied() else {
             return Err(BasePrecompileError::UnknownFunctionSelector([0u8; 4]));
@@ -107,7 +115,10 @@ impl AssetAbi {
         if !self.valid_selector(selector) {
             return Err(BasePrecompileError::UnknownFunctionSelector(selector));
         }
-        self.abi_decode_validate(calldata, selector)?;
+        match self {
+            Self::V1 => self.abi_decode_validate(calldata, selector)?,
+            Self::V2 => {}
+        }
 
         IB20Asset::IB20AssetCalls::abi_decode_validate(calldata).map_err(|error| {
             BasePrecompileError::AbiDecodeFailed { selector, error: error.to_string() }
--- a/crates/common/precompiles/src/b20_asset/versions.rs
+++ b/crates/common/precompiles/src/b20_asset/versions.rs
@@ -103,11 +103,7 @@ impl AssetAbi {
     ///
     /// Where the frozen surface differs from canonical (`V1`), the calldata is validated against the
     /// frozen surface first so a rejection carries that version's consensus error bytes, then
-    /// re-decoded into the canonical enum for routing. Where the frozen surface *is* canonical
-    /// (`V2`, which `IB20Asset` aliases), that pre-validation would decode the identical type twice,
-    /// so it is skipped and the single canonical decode suffices. The `match` is exhaustive so a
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

rayyan224:
Should we do this. for policy and the others also ?

crates/common/precompiles/src/b20_asset/versions.rs
```diff
@@ -100,14 +100,21 @@ impl AssetAbi {
...
         let Some(selector) = calldata.first_chunk::<4>().copied() else {
             return Err(BasePrecompileError::UnknownFunctionSelector([0u8; 4]));
         };
         if !self.valid_selector(selector) {
             return Err(BasePrecompileError::UnknownFunctionSelector(selector));
         }
-        self.abi_decode_validate(calldata, selector)?;
+        match self {
+            Self::V1 => self.abi_decode_validate(calldata, selector)?,
+            Self::V2 => {}
```
rayyan224, line 116: Why not for v2 ?

crates/common/precompiles/src/common/abi/b20_abi.rs
```diff
@@ -52,7 +52,10 @@ impl B20Abi {
         if !self.valid_selector(selector) {
             return Err(BasePrecompileError::UnknownFunctionSelector(selector));
         }
-        self.abi_decode_validate(calldata, selector)?;
+        match self {
```
rayyan224, line 55: What's the difference between this and calling self.abi_decode_validate ?

robriks: in the `v2` context, `abi_decode_validate` was running twice. this change allows it just to run once, eliminating the duplication

crates/common/precompiles/src/b20_asset/versions.rs
robriks: it doesn't need to run twice for v2 since it happens canonically using the current ruleset in the line below (119). this is where the optimization lies

v1 on the other hand runs the decode twice, first to apply its historical (stricter) ruleset and then second to produce the canonically-decoded output type `IB20Asset::IB20AssetCalls` that is consumable by the dispatcher

rayyan224 (approved):

PR #4250, opened 2026-08-03
fix(b20-asset): version-gate announce inner-call error handling — wrap Panic as InternalCallFailed at AssetV2 (Cobalt), keep AssetV1 (Beryl) raw

robriks:
## Summary

Resolves the `announce` inner-call error-handling divergence from BOP-477 by changing the **Rust precompile** at **Cobalt (AssetV2)** to remap an inner-call `Panic` to the dedicated `InternalCallFailed` typed revert, matching the base-std mock and `IB20Asset.announce` natspec. **AssetV1 (Beryl) is consensus-frozen and unchanged** — it keeps propagating system errors raw, byte-for-byte (the `v1.1.1` Beryl-launch tag already ships that behavior, so it cannot change).

Must land before Cobalt freezes `AssetV2` (AssetV2 already exists and is still mutable, so this rides it — no new version needed).

## What changed

- New `Asset` trait method `map_announce_internal_error(&self, call: &Bytes, err) -> BasePrecompileError`, mirroring the existing version-specific `begin_announce`/`end_announce`. It's **required (no default)**, so any future version is forced to declare its own inner-call classification (same "force a choice" spirit as #4229's exhaustive match).
  - **AssetV1** (`logic/v1.rs`, FROZEN): `if err.is_system_error() { err } else { InternalCallFailed }` — byte-identical to the previous inline mapping.
  - **AssetV2** (`logic/v2.rs`): `Panic` **and** ordinary reverts → `InternalCallFailed`; `OutOfGas` / `Fatal` / `SlotOverflow` still propagate **raw**.
- The dispatcher's `announce` loop (`dispatch.rs`) now calls `logic.map_announce_internal_error(call, err)` instead of the inline closure — the mapping moves onto the version axis, no hand-written fork gate.

## Scoping (security-relevant)

Only `Panic` (Solidity `0x11`/`0x21`/`0x32`) is remapped. `OutOfGas` **must** halt — converting it to a catchable typed revert would let a caller swallow OOG and continue, breaking EVM gas semantics. `Fatal`/`SlotOverflow` are unrecoverable internal faults that must not be masked. Both are kept raw at **both** versions and pinned by test.

## Tests

- **V1 (kept):** `announce_inner_system_error_propagates_unchanged` — inner overflow → raw `Panic(0x11)`.
- **V2 (new, end-to-end):** `announce_inner_panic_wraps_as_internal_call_failed_at_v2` — same overflowing `toScaledBalance` inner call → `InternalCallFailed`.
- **V1/V2 mapper units (new):** `map_announce_internal_error_*` pin the full classification, including that `OutOfGas`/`Fatal`/`SlotOverflow` stay raw at V2 (guards against over-wrapping the hard faults).

The `announce` wire surface (selectors/events/errors) is unchanged, so no frozen-surface/fingerprint pins move and there is **no golden re-bless** (revert-only paths).

## Verification

- `cargo test -p base-common-precompiles --features test-utils` — all pass (636+ lib + integration + goldens), 0 failures, 0 golden re-bless.
- `cargo clippy -p base-common-precompiles --all-targets --features test-utils` clean; `cargo +nightly fmt --check` clean.

## Follow-up (separate, base-std repo)

The per-fork base-std reference work is **not** in this PR (different repo): fix the `IB20Asset.announce` natspec (Panic → `InternalCallFailed` at Cobalt; OOG halts), verify/fix the Beryl-line mock to propagate raw `Panic`, and add fork + smoke pins for the inner-Panic case on both fork lines. Tracked under BOP-485 / BOP-477.

Linear: BOP-485

### DIFF
--- a/crates/common/precompiles/src/b20_asset/dispatch.rs
+++ b/crates/common/precompiles/src/b20_asset/dispatch.rs
@@ -461,17 +461,8 @@ impl<S: AssetAccounting, A: PolicyAccounting> B20AssetToken<S, A> {
             if call_bytes[..4] == IB20Asset::announceCall::SELECTOR {
                 return Err(BasePrecompileError::revert(IB20Asset::AnnouncementInProgress {}));
             }
-            self.route(ctx, call_bytes, version, privileged, NoopPrecompileCallObserver).map_err(
-                |err| {
-                    if err.is_system_error() {
-                        err
-                    } else {
-                        BasePrecompileError::revert(IB20Asset::InternalCallFailed {
-                            call: call.clone(),
-                        })
-                    }
-                },
-            )?;
+            self.route(ctx, call_bytes, version, privileged, NoopPrecompileCallObserver)
+                .map_err(|err| logic.map_announce_internal_error(call, err))?;
         }
 
         logic.end_announce(self, id)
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

robriks:
Closing per design decision (see BOP-485): we're keeping the Rust precompile's raw `Panic` propagation at both AssetV1 (Beryl) and AssetV2 (Cobalt) rather than wrapping at Cobalt.

base-std `main` already reflects the raw behavior via base-std#185 (BOP-477), so the Solidity reference and the precompile agree on the live/Cobalt line with no version-gated divergence. The raw `Panic` is a correct, atomic revert (OOG still halts); the only thing forgone is the self-describing `InternalCallFailed(call)` payload — an acceptable DX trade-off here.

No base/base change lands; `main` is unaffected.

PR #4273, opened 2026-08-04
feat(builder): Share Execution Cache With Builder

refcell:
## Summary

Preserve Reth's optional engine-owned execution cache through the custom flashblocks payload job, establishing the shared-cache plumbing needed by transaction prewarming without changing transaction selection or block-building behavior.

## What changed

- Carry the optional `SavedCache` from `BuildNewPayload` through `BlockPayloadJob` and `BuildArguments`.
- Wrap the canonical state provider in `CachedStateProvider` so block execution can reuse warmed entries while remaining lookup-only and leaving cache misses unpopulated.
- Separate payload construction from publication: `build_payload` returns the completed payload, allowing its execution state, provider, and shared-cache handles to be released through RAII before `try_build` wakes the payload resolver.
- Use ordinary return paths for flashblock finalization instead of loop-valued `break` expressions or manual `drop(...)` ordering.
- Cover cache propagation, lookup-only behavior, and cache-handle release in unit tests.

## Validation

- `cargo test -p base-builder-core --lib -- --test-threads=1`
- `cargo clippy -p base-builder-core --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

### DIFF
--- a/crates/builder/core/src/flashblocks/payload.rs
+++ b/crates/builder/core/src/flashblocks/payload.rs
@@ -31,6 +31,7 @@ use base_observability_events::{GlobalTransactionEventWriter, TransactionEventTy
 use eyre::WrapErr as _;
 use reth_basic_payload_builder::BuildOutcome;
 use reth_evm::{ConfigureEvm, execute::BlockBuilder};
+use reth_execution_cache::{CachedStateMetrics, CachedStateMetricsSource, CachedStateProvider};
 use reth_execution_types::ChangedAccount;
 use reth_node_api::{Block, BuiltPayloadExecutedBlock, PayloadBuilderError};
 use reth_payload_primitives::PayloadAttributes;
@@ -272,7 +273,13 @@ where
         payload_tx: &watch::Sender<Option<BaseBuiltPayload>>,
     ) -> Result<(), PayloadBuilderError> {
         let block_build_start_time = Instant::now();
-        let BuildArguments { mut cached_reads, config, cancel: block_cancel, publish_guard } = args;
+        let BuildArguments {
+            mut cached_reads,
+            execution_cache,
+            config,
+            cancel: block_cancel,
+            publish_guard,
+        } = args;
 
         // We log only every Nth block based on sampling ratio to reduce usage
         let block_number = config.parent_header.number + 1;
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/builder/core/src/flashblocks/payload.rs
```diff
@@ -477,7 +492,7 @@ where
         let mut executed_sender_nonces: HashMap<Address, u64> = HashMap::default();
 
         // Process flashblocks in a blocking loop
-        loop {
+        let final_payload = loop {
```
BrianBland: Do we want to move this into a helper function with proper returns instead of having `break` return the payload value?

crates/builder/core/src/flashblocks/payload.rs
```diff
@@ -387,6 +401,7 @@ where
         }
 
         if skip_flashblocks_building {
+            drop(state);
```
BrianBland: If we have to drop before send in order to avoid a cache mismatch then this could be a footgun.

crates/builder/core/src/flashblocks/payload.rs
refcell: Yep can do -- motivation here was to go with a minimal diff/least invasive approach.

crates/builder/core/src/flashblocks/payload.rs
refcell: Agreed, let me clean this up.

refcell: Addressed in 6c4d838b9f. Publication now lives in `PayloadBuilder::try_build`, outside the `build_payload` construction helper. The helper owns the execution state/provider and must return—releasing the shared-cache handle through RAII—before `send_replace` can wake the resolver. The explicit production and test `drop(...)` calls were removed.

crates/builder/core/src/flashblocks/payload.rs
refcell: Addressed in 6c4d838b9f. `build_payload` now returns `BaseBuiltPayload`, and each loop completion path uses a regular `return self.finalize_payload(...)`. The `PayloadBuilder::try_build` implementation owns the separate publication step.

BrianBland (approved):

PR #4277, opened 2026-08-04
feat(builder): Add Speculative Transaction Executor

refcell:
## Summary

Adds a cache-filling speculative execution primitive for builder transaction prewarming. Each transaction executes against a fresh parent-state provider with nonce and balance checks relaxed, while execution results and state changes are discarded. This is stacked on #4273 and remains dormant until a later PR adds bounded scheduling and builder integration.

### DIFF
--- a/crates/builder/core/src/flashblocks/prewarmer.rs
+++ b/crates/builder/core/src/flashblocks/prewarmer.rs
@@ -94,8 +94,8 @@ where
 
         let mut warmed = 0;
         for transaction in transactions {
-            let transaction = transaction.clone_into_consensus();
-            match evm.transact(&transaction) {
+            let transaction = transaction.consensus_ref();
+            match evm.transact(transaction) {
                 Ok(_) => warmed += 1,
                 Err(error) => {
                     trace!(
--- a/crates/builder/core/src/flashblocks/prewarmer.rs
+++ b/crates/builder/core/src/flashblocks/prewarmer.rs
@@ -1,0 +1,305 @@
+//! Speculative transaction execution for warming the shared execution cache.
+
+use alloy_primitives::B256;
+use base_common_consensus::BaseTransactionSigned;
+use base_common_evm::BaseSpecId;
+use base_execution_evm::BaseEvmConfig;
+use reth_evm::{ConfigureEvm, Evm, EvmEnv};
+use reth_execution_cache::{CachedStateProvider, ExecutionCache};
+use reth_provider::StateProviderFactory;
+use reth_revm::database::StateProviderDatabase;
+use reth_transaction_pool::PoolTransaction;
+use tracing::trace;
+
+/// Context for speculatively executing transactions against parent state.
+///
+/// Each transaction receives a fresh state provider and EVM. State changes and execution results
+/// are discarded; only reads populated through the shared [`ExecutionCache`] are retained.
+#[derive(Clone, Debug)]
+pub struct PrewarmingExecutionContext<Client> {
+    client: Client,
+    parent_hash: B256,
+    evm_config: BaseEvmConfig,
+    evm_env: EvmEnv<BaseSpecId>,
+    execution_cache: ExecutionCache,
+}
+
+impl<Client> PrewarmingExecutionContext<Client>
+where
+    Client: StateProviderFactory,
+{
+    /// Creates a speculative execution context for one payload build.
+    pub const fn new(
+        client: Client,
+        parent_hash: B256,
+        evm_config: BaseEvmConfig,
+        evm_env: EvmEnv<BaseSpecId>,
+        execution_cache: ExecutionCache,
+    ) -> Self {
+        Self { client, parent_hash, evm_config, evm_env, execution_cache }
+    }
+
+    /// Executes `transaction` against parent state to populate the shared execution cache.
+    ///
+    /// Nonce and balance checks are disabled so transactions that depend on earlier block state can
+    /// still warm useful reads. The execution result and all state changes are discarded.
+    ///
+    /// Returns `true` when speculative execution completed and `false` when the provider could not
+    /// be opened or the EVM rejected execution. Failures are deliberately non-fatal because
+    /// prewarming is only an optimization.
+    #[must_use]
+    pub fn prewarm_transaction<T>(&self, transaction: &T) -> bool
+    where
+        T: PoolTransaction<Consensus = BaseTransactionSigned>,
+    {
+        let state_provider = match self.client.state_by_block_hash(self.parent_hash) {
+            Ok(provider) => provider,
+            Err(error) => {
+                trace!(
+                    target: "payload_builder",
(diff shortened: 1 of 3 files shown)


crates/builder/core/src/flashblocks/prewarmer.rs
```diff
@@ -0,0 +1,305 @@
...
+                    parent_hash = ?self.parent_hash,
+                    transaction_hash = ?transaction.hash(),
+                    "failed to open parent state for transaction prewarming",
+                );
+                return false;
+            }
+        };
+        let state_provider =
+            CachedStateProvider::new_prewarm(state_provider, self.execution_cache.clone());
+        let database = StateProviderDatabase::new(state_provider);
```
BrianBland, line 88: How heavy is this initialization process, and can we share this across repeated calls to `prewarm_transaction`?

crates/builder/core/src/flashblocks/prewarmer.rs
```diff
@@ -0,0 +1,305 @@
...
+            Err(error) => {
+                trace!(
+                    target: "payload_builder",
+                    error = %error,
+                    transaction_hash = ?transaction.tx_hash(),
+                    "failed to speculatively execute transaction for prewarming",
+                );
+                false
+            }
+        }
```
BrianBland: Do we want all of the side effects to be fully discarded? Or can we chain prewarmed transactions to depend on each other?

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/builder/core/src/flashblocks/prewarmer.rs
refcell: Yep good callout. Reworked prewarming to apply across all transactions by constructing the evm once and then looping over `evm.transact()`.

crates/builder/core/src/flashblocks/prewarmer.rs
refcell: I do think we want side effects to be fully discarded. reth's prewarming always transacts without commit. If they were chained, we would lose the ability to warm useful reads due to nonce/balance checks.

crates/builder/core/src/flashblocks/prewarmer.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Returns `true` when speculative execution completed and `false` when the provider could not
+    /// be opened or the EVM rejected execution. Failures are deliberately non-fatal because
+    /// prewarming is only an optimization.
+    #[must_use]
+    pub fn prewarm_transaction<T>(&self, transaction: &T) -> bool
+    where
+        T: PoolTransaction<Consensus = BaseTransactionSigned>,
+    {
+        let state_provider = match self.client.state_by_block_hash(self.parent_hash) {
+            Ok(provider) => provider,
```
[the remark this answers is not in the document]
refcell: This is already resolved by the batching fix pushed in c1b741c (see the reply on the sibling thread from @BrianBland about the same cost). `prewarm_transactions` now opens the state provider and builds the EVM exactly once per call, then loops `evm.transact()` over every transaction in the batch — so once the scheduler lands and calls this once per block with the full candidate set, there is a single `state_by_block_hash` per block rather than one per transaction.

I went with sharing one owned `CachedStateProvider`/EVM across the whole batch rather than the `Arc<dyn StateProvider>` plus per-transaction EVM split you sketched. It sidesteps the `Box<dyn StateProvider>: !Clone` problem entirely since there is only ever one owned provider for the batch, and it matches reth's own precedent in `reth_engine_tree`'s `prewarm.rs`, which builds one EVM per worker and reuses it across every transaction that worker executes (never rebuilding per transaction, never committing between them). Since transaction execution here is sequential per batch rather than parallel-per-transaction, that is simpler than introducing an `Arc` adapter and gets the same amortization. Agreed this is worth re-benchmarking once the bounded scheduler and parallel execution land.

crates/builder/core/src/flashblocks/prewarmer.rs
```diff
@@ -0,0 +1,353 @@
...
+        let database = StateProviderDatabase::new(state_provider);
+
+        let mut evm_env = self.evm_env.clone();
+        evm_env.cfg_env.disable_nonce_check = true;
+        evm_env.cfg_env.disable_balance_check = true;
+        let mut evm = self.evm_config.evm_with_env(database, evm_env);
+
+        let mut warmed = 0;
+        for transaction in transactions {
+            let transaction = transaction.clone_into_consensus();
```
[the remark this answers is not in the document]
refcell: Good callout, and yes it's fixable — reth's own prewarming (`reth_engine_tree::prewarm`) avoids this by pre-building `TxEnv` once outside the hot loop and passing the already-built env into `evm.transact()`, so it never clones the consensus transaction inside the loop either.

For this API, `consensus_ref()` (returning `Recovered<&BaseTransactionSigned>`) does satisfy `Evm::transact`'s `IntoTxEnv` bound: `ToTxEnv<TxEnv>` is implemented for `Recovered<T>` whenever `TxEnv: FromRecoveredTx<T>`, and there's a blanket `impl<TxEnv, T> FromRecoveredTx<&T> for TxEnv where TxEnv: FromRecoveredTx<T>` — so since our EVM's tx type already implements `FromRecoveredTx<BaseTransactionSigned>`, it automatically implements `FromRecoveredTx<&BaseTransactionSigned>` too, and `Recovered<&BaseTransactionSigned>` converts straight into `TxEnv` without ever calling `.clone()` on the transaction. `BasePooledTransaction::consensus_ref()` confirms this — it's `self.inner.transaction().as_recovered_ref()`, no clone.

Swapped `clone_into_consensus()` for `consensus_ref()` in the loop (`3643b04`); tests still pass and clippy is clean.

PR #4287, opened 2026-08-05
fix(txpool): bound delegate-signed high-throughput senders and sponsors

chunter-cb:
## Summary

A "high-throughput" EIP-8130 account is one whose owner config is **locked** — that lock exempts it from the per-account `signature_limit`, granting effectively unlimited in-flight signatures. The justification is baked into the guard: *"locked owner config ⇒ stable auth surface ⇒ unlimited dimension."* It applies to both the **sender** dimension and the sponsored-**payer** dimension.

That premise is false when the actor authenticates via the **delegate authenticator**. A delegate resolves `actorId = bytes20(delegate_account)` and re-enters auth against a **foreign** account whose config is mutable and outside the signer's control. Locking the signer's own config does not freeze that surface — so **the delegate could be removed and the stable signature surface is lost**, mass-invalidating the account's unbounded in-flight backlog. That's a third-party-controlled mempool DoS lever, and it exists on both the sender and the payer sides.

This mirrors an asymmetry already handled on the balance side: `is_high_rate_account` refuses to trust mutable (EIP-7702) code because the delegating EOA can re-point it at any time. The signature dimensions had no equivalent guard; the delegate authenticator is their 7702 equivalent.

## Approach

Treat each delegate account as an **independent gating surface**, consistent with the guard's existing "one signature dimension per gating account" model (sender, payer):

- A delegate-signed **sender** charges its delegate account a signature slot unless that delegate account is **itself locked**.
- A delegate-signed sponsored **payer** does the same (empty for self-pay, where the sender-delegate dimension already covers it).
- Locked signer + locked delegate ⇒ fully stable ⇒ keeps unlimited throughput.
- Locked signer + unlocked delegate ⇒ bounded by the delegate's `signature_limit` (caps the revocation blast radius).
- Each delegate's account-state slot is added to the tx `WatchSet`, so an unlock drains transactions admitted as exempt.

Purely additive: two new dimensions threaded through `LimitClass` → `Admission` → `AdmissionRecord`, charged/rolled-back/released exactly like the existing sender/payer signature dimensions. A shared `classify_delegate_lock` helper handles both sides.

## Scope / follow-ups

- Delegate exemption keys off the same `locked + unlock-horizon` gate as the sender/payer lock classification (plain lock, not the stricter `qualifies_as_high_rate_lock` hard-lock). Open to tightening to a hard lock if we want delegates held to the same bar as a balance-bounded payer.

## Test plan

- [x] `cargo test -p base-execution-txpool --lib` (203 passed)
- [x] Guard tests, sender + payer: delegate charged when unlocked, exempt when locked, released on removal, rolled back on a later-dimension rejection, and self-pay ignores the payer-delegate dimension
- [x] `cargo clippy -p base-execution-txpool --all-targets` clean; `cargo fmt` applied
- [ ] Follow-up: end-to-end validator test asserting delegate-signed sender/payer pick up the delegate account + watch slot


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4289, opened 2026-08-05
fix(BOP-498): version-gate B20 seize policy scope validation to Cobalt (V2)

stephancill:
## Summary

Audit ([BOP-498](https://linear.app/coinbase/issue/BOP-498)) found that **B20 V1 (Beryl) incorrectly accepts the seize policy scopes**.

The seize scopes (`SEIZE_HOLDER_POLICY` / `SEIZE_RECEIVER_POLICY`) were introduced at Cobalt (V2). But the `updatePolicy(bytes32,uint64)` and `policyId(bytes32)` selectors are **common** (dialable on V1), and their scope validation routed through the shared, un-versioned `B20PolicyType::from_id`, which recognizes all six scopes. Because a `bytes32` argument is not range-checked by ABI decode, V1 accepted the V2-only seize scopes and wrote/read the seize storage slot, instead of reverting `UnsupportedPolicyType`. At the frozen base-base tags (v1.1.0/v1.2.0), `B20PolicyType` has only the four pre-Cobalt scopes, confirming the correct V1 behavior is to reject.

No live seize on V1 (the `seizeWithMemo` entrypoint stays V2-only), so a stored value is dormant. But it is a frozen-surface divergence, and the fork harness missed it because it never exercises `updatePolicy`/`policyId` with the seize scope args.

Contrast: the enum-typed sibling (`SEIZE` pause feature) is already safe — `pause`/`isPaused` take a `PausableFeature` enum, so the V1 wire decode (`IB20V1`, three variants) rejects `SEIZE`=3 before any shared code runs.

## Change

Move the supported-scope decision into each version's `ensure_supported_policy_type` (asset + stablecoin):
- **V1**: accept only `TransferSender`/`TransferReceiver`/`TransferExecutor`/`MintReceiver`; wildcard-reject everything else (so any scope added to `B20PolicyType` later stays frozen out of V1 automatically).
- **V2**: exhaustive match over all six scopes — a future scope will fail to compile until V2's stance is decided explicitly, closing this class of silent widening going forward.

## Tests

- `golden_update_policy_rejects_seize_scopes_at_v1` / `golden_policy_id_rejects_seize_scopes_at_v1` (asset + stablecoin) — the fixed leak.
- `golden_seize_pause_feature_unknown_at_v1` (asset + stablecoin) — defense-in-depth regression pinning that `pause`/`unpause`/`isPaused` reject `SEIZE` on V1 via wire decode.
- Registered in the exhaustive `covered(&[...])` matrices. Full `base-common-precompiles` suite green; clippy + nightly fmt clean.

## Notes

- Verified against the frozen base-base tags (v1.1.0/v1.2.0): `B20PolicyType` there has only the four pre-Cobalt scopes, so rejecting is the frozen-correct V1 behavior.
- A stacked PR ([BOP-500](https://linear.app/coinbase/issue/BOP-500)) builds on this branch: it adds failed-path SLOAD testing and fixes an asset `updatePolicy` revert-path SLOAD divergence surfaced during the same audit.

### DIFF
--- a/crates/common/precompiles/src/b20_asset/logic/v1.rs
+++ b/crates/common/precompiles/src/b20_asset/logic/v1.rs
@@ -170,7 +170,12 @@ impl AssetV1 {
     /// The seize scopes (`SEIZE_HOLDER_POLICY` / `SEIZE_RECEIVER_POLICY`) were introduced at Cobalt
     /// (V2), so they are rejected here — matching the base-std `v1.0.0` reference. This is what keeps
     /// the common `updatePolicy` / `policyId` selectors (dialable on V1) from reaching a V2-only
-    /// scope. The wildcard arm also freezes V1 against any scope added to `B20PolicyType` later.
+    /// scope.
+    ///
+    /// The match is exhaustive by design (no `_` arm): every `B20PolicyType` variant is named, so a
+    /// scope added for a future fork fails to compile here until an engineer explicitly places it in
+    /// the supported or rejected arm, reconciling V1 against the base-std `v1.0.0` reference. A
+    /// wildcard would silently reject a newly added scope that should be valid at V1.
     fn ensure_supported_policy_type(policy_scope: B256) -> Result<()> {
         match B20PolicyType::from_id(policy_scope) {
             Some(
@@ -179,9 +184,12 @@ impl AssetV1 {
                 | B20PolicyType::TransferExecutor
                 | B20PolicyType::MintReceiver,
             ) => Ok(()),
-            _ => Err(BasePrecompileError::revert(IB20::UnsupportedPolicyType {
-                policyScope: policy_scope,
-            })),
+            // Cobalt (V2)-only scopes, and any unrecognized id, are not part of the frozen V1 surface.
+            Some(B20PolicyType::SeizeHolder | B20PolicyType::SeizeReceiver) | None => {
+                Err(BasePrecompileError::revert(IB20::UnsupportedPolicyType {
+                    policyScope: policy_scope,
+                }))
+            }
         }
     }
 
--- a/crates/common/precompiles/src/b20_asset/logic/v1.rs
+++ b/crates/common/precompiles/src/b20_asset/logic/v1.rs
@@ -164,14 +164,24 @@ impl AssetV1 {
         Ok(())
     }
 
-    /// Ensures `policy_scope` names a built-in B-20 policy slot.
+    /// Ensures `policy_scope` names a built-in B-20 policy slot available on the frozen V1 (Beryl)
+    /// common surface.
+    ///
+    /// The seize scopes (`SEIZE_HOLDER_POLICY` / `SEIZE_RECEIVER_POLICY`) were introduced at Cobalt
+    /// (V2), so they are rejected here — matching the base-std `v1.0.0` reference. This is what keeps
+    /// the common `updatePolicy` / `policyId` selectors (dialable on V1) from reaching a V2-only
+    /// scope. The wildcard arm also freezes V1 against any scope added to `B20PolicyType` later.
     fn ensure_supported_policy_type(policy_scope: B256) -> Result<()> {
-        if B20PolicyType::from_id(policy_scope).is_some() {
-            Ok(())
-        } else {
-            Err(BasePrecompileError::revert(IB20::UnsupportedPolicyType {
+        match B20PolicyType::from_id(policy_scope) {
+            Some(
+                B20PolicyType::TransferSender
+                | B20PolicyType::TransferReceiver
+                | B20PolicyType::TransferExecutor
+                | B20PolicyType::MintReceiver,
+            ) => Ok(()),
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/b20_asset/logic/v1.rs
```diff
@@ -164,14 +164,24 @@ impl AssetV1 {
...
-        } else {
-            Err(BasePrecompileError::revert(IB20::UnsupportedPolicyType {
+        match B20PolicyType::from_id(policy_scope) {
+            Some(
+                B20PolicyType::TransferSender
+                | B20PolicyType::TransferReceiver
+                | B20PolicyType::TransferExecutor
+                | B20PolicyType::MintReceiver,
+            ) => Ok(()),
+            _ => Err(BasePrecompileError::revert(IB20::UnsupportedPolicyType {
```
robriks: I believe the use of _ here will swallow both `None || Some(newVariant)`. So when a scope is later added to B20PolicyType, the compile-time forcing function only fires on one of the two versions. Rejecting the seize scopes on frozen V1 is correct; the concern is a future scope that should be valid at V1 getting silently rejected. This can be resolved if we explicitly prompt engineers at compile time to reconcile against the base-std v1.0.0 reference.

To address this there is already a "no _ arm" rule from the b20-fork-testing skill, under Version-control / security invariants:

```
base/base enforces this with golden/manifest tests (…), a compile-time op-coverage match (no _ arm), and reject_frozen_selector!.
```

robriks (changes requested):

cb-heimdall:
Review Error for robriks @ 2026-08-05 17:30:52 UTC

crates/common/precompiles/src/b20_asset/logic/v1.rs
stephancill: good callout thank you!

crates/common/precompiles/src/b20_asset/logic/v1.rs
```diff
@@ -164,14 +164,32 @@ impl AssetV1 {
...
+    /// the supported or rejected arm, reconciling V1 against the base-std `v1.0.0` reference. A
+    /// wildcard would silently reject a newly added scope that should be valid at V1.
     fn ensure_supported_policy_type(policy_scope: B256) -> Result<()> {
-        if B20PolicyType::from_id(policy_scope).is_some() {
-            Ok(())
-        } else {
-            Err(BasePrecompileError::revert(IB20::UnsupportedPolicyType {
-                policyScope: policy_scope,
-            }))
+        match B20PolicyType::from_id(policy_scope) {
```
rayyan224, line 180: What are your thought's on having, a B20PolicyType::from_supported_ids(policy_scope, [...list of supported policies]).  It seems to be what we are duplicating throughout.

stephancill: can you elaborate? these are all separate versions so unless we put it in the common b20 we can't really refactor

rayyan224: B20PolicyType::from_supported_ids(policy_scope, setOfSupported) 

So for the below example could be like this 
```
B20PolicyType::from_supported_ids(policy_scope, [B20PolicyType::TransferSender, B20PolicyType::TransferReceiver, B20PolicyType::TransferExecutor, B20PolicyType::MintReceiver])?
```

stephancill: not too much gained imo

rayyan224: It seems we are doing this same logic, in all 4 places right ?

rayyan224 (approved):

robriks (approved):
concerns addressed, great stuff !

PR #4296, opened 2026-08-05
fix(BOP-500): add failed-path SLOAD testing + restore asset updatePolicy ordering

stephancill:
Stacked on #4289 (BOP-498). Review/merge that first; this PR's base is the BOP-498 branch, so the diff here is only the BOP-500 delta.

## Motivation ([BOP-500](https://linear.app/coinbase/issue/BOP-500))

Golden gas-footprint coverage (`golden_gas_footprints`) only exercises **success** paths. So a divergence in the **revert-path** storage-access footprint (SLOAD count = gas = consensus) is invisible to the existing suite and to the fork harness (which compares logs/reverts, not gas). This PR adds failed-path SLOAD testing and fixes the divergence it surfaces.

## The divergence

The unreleased `logic/` refactor changed **asset** `updatePolicy` to read the old policy id **before** the `policy_exists` check. The frozen asset surface (base-base `token.rs` at v1.1.0/v1.2.0) reads it **after**, only on the success path — so the `PolicyNotFound` revert path does **zero** old-id SLOADs. Reading before the check adds an SLOAD on that reachable revert path = gas/consensus divergence from frozen Beryl.

Per-variant (verified against the frozen tags):

| Variant | Frozen v1.1.0/v1.2.0 | main before | main after |
| --- | --- | --- | --- |
| Asset | read-after (0 SLOAD on PolicyNotFound) | read-before (1) | read-after (0) |
| Stablecoin | read-before (1 SLOAD) | read-before (1) | read-before (1) |

## Change

- Restore asset V1 `update_policy` to read-after (frozen Beryl ordering); mirror in asset V2 so asset's gas profile is unchanged across Beryl→Cobalt.
- **Stablecoin is intentionally left unchanged** — it reads before the check on both the frozen surface and main; the two variants legitimately differ, and that difference is itself frozen.

## Tests (the BOP-500 deliverable)

Failed-path SLOAD pins on the `PolicyNotFound` revert path:
- asset V1/V2 assert **0** old-id SLOADs;
- stablecoin V1 asserts **1**, locking each variant's frozen ordering so they can't be accidentally unified.

Full `base-common-precompiles` suite green; clippy + nightly fmt clean.

## Reviewer notes

- The refactor that introduced this is unreleased, so no live chain is affected — but it must land before Cobalt.
- base-std reads-before for asset (disagrees with frozen base-base asset, which reads-after); harmless for cross-validation (fork harness ignores gas), but worth reconciling.
- Follow-up worth considering: extend failed-path SLOAD/SSTORE footprint checks to every op (not just `updatePolicy`), since the refactor could have introduced other revert-path gas divergences.

### DIFF
--- a/crates/common/precompiles/src/b20_asset/logic/v1.rs
+++ b/crates/common/precompiles/src/b20_asset/logic/v1.rs
@@ -586,11 +586,6 @@ impl<S: AssetAccounting, A: PolicyAccounting> Asset<S, A> for AssetV1 {
         if !privileged {
             B20Guards::ensure_token_role(token, caller, B20TokenRole::DefaultAdmin)?;
         }
-        // Frozen Beryl ordering (see `token.rs` at v1.1.0/v1.2.0): validate the scope (pure), then
-        // the existence check, and only read the old id on the success path. Reading the old id
-        // before `policy_exists` would add an SLOAD on the `PolicyNotFound` revert path, changing
-        // gas — a consensus divergence from the frozen asset surface. (Stablecoin reads before the
-        // check; that difference is itself frozen, so the two variants intentionally differ here.)
         Self::ensure_supported_policy_type(policy_scope)?;
         if !token.policy().policy_exists(token.policy_storage(), new_policy_id)? {
             return Err(BasePrecompileError::revert(IB20::PolicyNotFound {
--- a/crates/common/precompiles/src/b20_asset/logic/v1.rs
+++ b/crates/common/precompiles/src/b20_asset/logic/v1.rs
@@ -586,12 +586,18 @@ impl<S: AssetAccounting, A: PolicyAccounting> Asset<S, A> for AssetV1 {
         if !privileged {
             B20Guards::ensure_token_role(token, caller, B20TokenRole::DefaultAdmin)?;
         }
-        let old_policy_id = self.policy_id(token, policy_scope)?;
+        // Frozen Beryl ordering (see `token.rs` at v1.1.0/v1.2.0): validate the scope (pure), then
+        // the existence check, and only read the old id on the success path. Reading the old id
+        // before `policy_exists` would add an SLOAD on the `PolicyNotFound` revert path, changing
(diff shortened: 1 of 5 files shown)


crates/common/precompiles/src/b20_asset/logic/v1.rs
```diff
@@ -586,12 +586,18 @@ impl<S: AssetAccounting, A: PolicyAccounting> Asset<S, A> for AssetV1 {
         if !privileged {
             B20Guards::ensure_token_role(token, caller, B20TokenRole::DefaultAdmin)?;
         }
-        let old_policy_id = self.policy_id(token, policy_scope)?;
+        // Frozen Beryl ordering (see `token.rs` at v1.1.0/v1.2.0): validate the scope (pure), then
+        // the existence check, and only read the old id on the success path. Reading the old id
+        // before `policy_exists` would add an SLOAD on the `PolicyNotFound` revert path, changing
+        // gas — a consensus divergence from the frozen asset surface. (Stablecoin reads before the
+        // check; that difference is itself frozen, so the two variants intentionally differ here.)
```
stephancill: remove

crates/common/precompiles/src/b20_asset/logic/v2.rs
```diff
@@ -662,12 +662,17 @@ impl<S: AssetAccounting, A: PolicyAccounting> Asset<S, A> for AssetV2 {
         if !privileged {
             B20Guards::ensure_token_role(token, caller, B20TokenRole::DefaultAdmin)?;
         }
-        let old_policy_id = self.policy_id(token, policy_scope)?;
+        // Preserve asset's frozen Beryl `update_policy` ordering into Cobalt: validate the scope
+        // (pure) and run the existence check before reading the old id, so the `PolicyNotFound`
+        // revert path does no old-id SLOAD (matches AssetV1 / `token.rs`). Keeps asset's gas
+        // profile unchanged across the Beryl->Cobalt boundary.
```
stephancill: remove

stephancill:
closing in favour of #4305

PR #4274, opened 2026-08-04
fix(batcher): guarantee gap-free ordered L2 block delivery

xenoliss:
## Context

The batcher could skip L2 blocks. On restart it decoded L1 frames to guess the
highest submitted L2 block and advanced the backfill cursor past it, silently
missing blob txs and channels opened before the scan window. At runtime the
hybrid source raced a subscription against polling and deduped by hash, so a
delayed or dropped subscription event left a gap that was never backfilled.

## Changes

- Anchor the L2 source to a trusted `safe_head` and always fetch the exact next
  height; subscription events are hints, gaps and closures fall back to
  `block_by_number(tip + 1)`. `PollingSource` collapses to `block_by_number`.
- Detect reorgs on same-height/different-hash and parent-hash mismatch;
  `L2BlockEvent::Reorg` is now a unit variant.
- Replace frame decoding with `recent_tx_sync_target`: a nonce binary-search
  that only picks an L1 sync target, never the L2 cursor. Requires
  `--wait-node-sync`.
- Safe-head feed becomes an ordered `mpsc<BlockInfo>`; `prune_safe` validates
  against the buffered chain and a regression/conflict resets to the safe head.
- Shadow mode reads the safe head from the parity validator's `safe` label.
- Update source/core/encoder unit tests and the `gap_filling` harness test.

## Testing

- `cargo test -p base-batcher-source -p base-batcher-core -p base-batcher-encoder`
- `cargo test -p base-action-harness --test batcher`

PR #4316, opened 2026-08-07
fix(batcher): validate and prune encoder state on safe-head advance

xenoliss:
## Summary

Confirmation drove both channel removal and block pruning, so blocks could be
dropped as soon as their frames confirmed on L1 — before the node had actually
derived them. Make the safe-head advance the single owner of that lifecycle:
`confirm` only records inclusion, and `prune_safe` validates the reported safe
head against the buffered chain before removing anything.

## Changes

- `BatchPipeline::prune_safe` now takes a `BlockInfo` and returns a verdict:
  `false` when the safe head leaves a gap below the buffer, lies above it, or its
  hash does not match the boundary block. The driver resets to the safe head when
  the verdict is `false`.
- `confirm` becomes bookkeeping only; `prune_safe` is the sole remover of blocks
  and channels. It rewinds staged span state, drops fully-covered ready channels
  and rebases pending references, clears a covered open channel, clamps the
  encoding cursor and anchors the tip.
- Track per-frame `FrameState` (ready/pending/confirmed) instead of a cursor plus
  counters; drop `ReadyChannel::block_range`. `tip` becomes `Option<B256>`.
- Exclude fully-confirmed channels from the DA backlog estimate; keep
  `has_ready_submission` in the trait and its mocks.
- Separately, fix pre-Fjord channel compression by making the algorithm
  configurable and correcting zlib stream output and draining reads.

## Testing

- `cargo test -p base-batcher-encoder -p base-batcher-core`

Fifth of five PRs splitting #4274.


crates/batcher/encoder/src/encoder.rs
```diff
@@ -913,14 +863,16 @@ impl BatchPipeline for BatchEncoder {
...
+
         if self.deferred_step_error.is_some() {
             return;
         }
-        if let Err(error) = self.check_channel_timeout() {
+
+        if advanced && let Err(error) = self.check_channel_timeout() {
             self.defer_step_error(error, "advance_l1_head");
         }
         self.invalidate_expired_ready_channels();
```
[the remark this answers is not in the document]
xenoliss: The driver calls `advance_l1_head` after every receipt, and the expiry check now runs on every call, even when the block number does not move the head forward.

crates/batcher/encoder/src/encoder.rs
```diff
@@ -955,26 +907,85 @@ impl BatchPipeline for BatchEncoder {
...
+            prune_count,
+            safe_l2_number = safe_l2.number,
+            "pruning safe blocks from input queue"
+        );
+
+        // Rewind staged span state when pruning crosses into its buffered blocks.
+        let span_start = self.block_cursor.saturating_sub(self.span_accumulator.len());
+        if !self.span_accumulator.is_empty() && prune_count > span_start {
+            self.block_cursor = span_start;
+            self.clear_span_accumulator();
```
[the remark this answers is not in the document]
xenoliss: Yep, a safe head cutting into the staged span intentionally drops the whole accumulator and the remaining non-safe blocks are re-encoded on the next step, which keeps duplicate data off L1 at the cost of a little CPU.

crates/batcher/encoder/src/encoder.rs
```diff
@@ -955,26 +905,85 @@ impl BatchPipeline for BatchEncoder {
...
+            return true;
+        };
+
+        let oldest_number = oldest.header.number;
+        let next_safe = safe_l2.number.saturating_add(1);
+        if next_safe < oldest_number {
+            return false;
+        }
+
+        let prune_count = (next_safe - oldest_number) as usize;
```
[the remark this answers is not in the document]
xenoliss: Even if the addition saturates, the next guard returns false because `prune_count` is larger than `blocks.len()`, so the value is never used. We also do not support 32-bit targets.

crates/batcher/encoder/src/encoder.rs
```diff
@@ -517,9 +503,18 @@ impl BatchEncoder {
...
+            };
+            let inclusion_span =
+                channel.last_confirmed_l1_block.unwrap_or(first).saturating_sub(first);
+            if inclusion_span >= channel_timeout {
+                return true;
+            }
+
+            let incomplete =
+                channel.frame_states.iter().any(|state| *state != FrameState::Confirmed);
+            incomplete && self.l1_head.saturating_sub(first) >= channel_timeout
```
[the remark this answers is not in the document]
xenoliss: Intentional (same as the op bacther). It also keeps fully submitted channels until the safe head moves past them. If derivation stalls, the block queue grows first, and DA throttling plus monitoring keep that bounded.

crates/batcher/encoder/src/encoder.rs
```diff
@@ -955,26 +905,85 @@ impl BatchPipeline for BatchEncoder {
...
+                .flat_map(|channel| &channel.frame_states)
+                .filter(|state| **state == FrameState::Ready)
+                .count();
+            self.ready_channels.drain(..ready_channels_to_prune);
+            self.pending.retain(|_, pending| {
+                if pending.channel_idx < ready_channels_to_prune {
+                    return false;
+                }
+                pending.channel_idx -= ready_channels_to_prune;
+                true
```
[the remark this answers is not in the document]
xenoliss: Correct on all branches. If a channel only contains blocks that are already safe, its remaining frames are dropped on purpose, because data that is already safe never needs to be submitted again.

bin/batcher/src/cli.rs
```diff
@@ -311,6 +311,7 @@ impl BatcherArgs {
             batch_type: self.batch_type.into(),
             da_type: self.da_type,
             approx_compr_ratio: self.approx_compr_ratio,
+            compression_algo: base_batcher_encoder::CompressionAlgo::Brotli10,
```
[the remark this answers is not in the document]
xenoliss: All deployments run after Fjord, so the binary always uses brotli.

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/batcher/encoder/src/encoder.rs
```diff
@@ -517,9 +503,18 @@ impl BatchEncoder {
...
+            let Some(first) = channel.first_confirmed_l1_block else {
+                return false;
+            };
+            let inclusion_span =
+                channel.last_confirmed_l1_block.unwrap_or(first).saturating_sub(first);
+            if inclusion_span >= channel_timeout {
+                return true;
+            }
+
+            let incomplete =
```
refcell: I think this leaves a fully confirmed channel retained forever if the node processes its L1 inclusion without advancing the safe head through the channel’s L2 blocks.

The op-batcher detects this condition once `CurrentL1` has moved past the channel’s maximum inclusion block and resets its local channel state. Here, incomplete disables expiry as soon as every frame confirms, so the channel can only be removed by a safe-head update that may never cover it. `da_backlog_bytes` also excludes fully confirmed channels, so throttling will not bound the retained state.

Could we reset to the safe head and replay these blocks when the rollup node’s `CurrentL1` has passed the channel’s last inclusion block but its safe head is still below the channel’s last L2 block?

bin/batcher/src/cli.rs
refcell, line 315: @xenoliss can we leave a comment here mentioning that the reason we can hardcode this is because all chains are post-fjord? Adds a bit of nice-to-have context for future readers

crates/batcher/core/src/driver.rs
```diff
@@ -355,25 +360,47 @@ where
...
         }
 
-        self.pipeline.prune_safe(head.number);
-        debug!(safe_l2_number = %head.number, "pruned safe L2 blocks");
+        match self
+            .pipeline
+            .reconcile_derivation(head, status.current_l1.map(|current_l1| current_l1.number))
+        {
+            DerivationReconciliation::Consistent => {
+                debug!(safe_l2_number = %head.number, "reconciled pipeline with derivation");
```
refcell: We can probably remove this as it will spam, or demote to a `trace!`

crates/batcher/encoder/src/encoder.rs
```diff
@@ -719,86 +817,89 @@ impl BatchPipeline for BatchEncoder {
...
-                                    "frame exceeds max_l1_tx_size_bytes; submitting anyway"
-                                );
-                            }
-                            total += frame_size;
-                            n += 1;
+            let Some(frame_start) =
+                channel.frame_states.iter().position(|state| *state == FrameState::Ready)
+            else {
+                continue;
+            };
```
refcell: This probably would be negligible in practice, but each time we look at the channel we need to iter the `frame_states` vec to find the earliest ready frame. Previously we had a cursor so we didn't have to iterate. Is there a reason we can't use a cursor index or it's just to accept the tradeoff of simplicity for performance?

crates/batcher/encoder/src/encoder.rs
```diff
@@ -719,86 +817,89 @@ impl BatchPipeline for BatchEncoder {
...
+                                frame_size,
+                                max_l1_tx_size_bytes = max_size,
+                                "frame exceeds max_l1_tx_size_bytes; submitting anyway"
+                            );
+                        }
+                        total += frame_size;
+                        n += 1;
                     }
+                    n.max(1)
                 } else {
```
refcell, line 856: All of this seems like logic that could be refactored onto the `channel` type and unit tested rather than placed here in the `encoder.rs` module which is already too large

crates/batcher/encoder/src/encoder.rs
```diff
@@ -808,73 +909,27 @@ impl BatchPipeline for BatchEncoder {
...
     }
 
     fn requeue(&mut self, id: SubmissionId) {
-        // Invariant: each `ReadyChannel` owns its own frame cursor. This
-        // encoder keeps at most one `current_channel` open at a time; when it
-        // closes (by size or timeout) it moves to `ready_channels` as the
-        // newest entry. `pending_ref.channel_idx` therefore always points to
-        // a specific, independent slot in `ready_channels`. Resetting the
-        // cursor on that slot does not affect any other channel and FIFO
-        // ordering across channels is preserved by construction.
```
refcell, line 875: Did we remove this invariant? maybe worth updating the comment if so for posterity

crates/batcher/encoder/src/encoder.rs
refcell, line 1: I'd really like to break this module up. We should be targeting for ~ 300 LOC in a module to keep them legible, and `encoder.rs` is 10x this target

crates/batcher/service/src/derivation_status_poller.rs
```diff
@@ -0,0 +1,174 @@
...
+                )
+            })?;
+        Ok(DerivationStatus::from_safe_l2(BlockInfo {
+            hash: block.header.hash,
+            number: block.header.number,
+            parent_hash: block.header.parent_hash,
+            timestamp: block.header.timestamp,
+        }))
+    }
+}
```
refcell, line 50: I'm fine with this abstraction if it lets us test the `DerivationStatusPoller` 👍

refcell (commented):
Left some comments here and there. Overall like the direction this is heading with simplifying a lot of the batcher code

refcell (approved):

PR #4321, opened 2026-08-08
refactor(nitro): stamp batch-delete TEE signer in the enclave (phase 2 of #4129)

0xth4nh:
Phase 2 of batch-delete-of-proofs-by-TEE-signer, shipped in #4129.

The true signer lives inside the enclave. Phase 1 could not ask for it, because modifying the enclave binary changes the EIF and therefore PCR0, which forces a coordinated onchain rollout — build the new EIF, derive the new hash, re-register and authorize the enclave signer and attestation, and activate an `AggregateVerifier` carrying the matching `TEE_IMAGE_HASH`. That is fork-class coordination, not something to attach to a feature that was otherwise pure host and prover-service work. So phase 1 landed batch delete on the best signer available without touching the enclave: recover it host-side from the proof's own signature.

This PR is the enclave-side half, landing now that an enclave change is on the table. The enclave stamps its signer directly, which removes the reconstruction phase 1 had to do — and with it the config that only existed to make that reconstruction possible.

## What changed?

- The enclave stamps its own signer into `ProofResult::Tee`, so the address recorded against a proof is the very key that signed it.
- The host forwards that address instead of recovering it; `signer_recovery.rs` is deleted (−185 lines), along with the `alloy-signer-local` dev-dependency and `NitroHostError::SignerRecovery`.
- `ProofRequest.image_hash` and the proposer config that fed it (`--tee-image-hash` / `BASE_PROPOSER_TEE_IMAGE_HASH`, `ProposerConfig`, `DriverConfig`, `ProofDispatcherConfig`) are deleted — they existed only so the host could rebuild the journal.

## Why phase 2 is worth taking

Host-side recovery rebuilds the `ProofJournal` the enclave signed and ecrecovers the address from it. Two of that journal's fields are not carried on the `Proposal`, so the host supplies its own copies from the originating request: `proposer` and `tee_image_hash`. Both are approximations of enclave-internal state, and that is inherent to solving this without the enclave's cooperation.

The consequence is that **ecrecover does not fail on a wrong journal — it returns a different, valid-looking address.** #4129 documented this as a property rather than an error path (`wrong_context_recovers_a_different_address`: "Recovering with a mismatched image hash yields some other address, never the signer"). That address is stored and indexed as the proof's signer, and it is the key `delete_proofs_by_tee_signer` matches on. Two independent ways the substituted fields can disagree with what the enclave signed:

1. **`tee_image_hash`** — the enclave signs `keccak256(its own PCR0)` (`nitro-enclave/src/server.rs:84`). The host supplies the proposer's configured hash. Separate values, nothing keeping them equal.
2. **`proposer`** — the enclave signs `boot_info.proposer`, which defaults to `Address::ZERO` when the preimage is absent (`proof/src/boot.rs:302`, "defaults to zero for backwards compatibility"). The host supplies the real proposer address from the request.

The gap matters most in the scenario batch delete was built for. A backlog needing clearing appears when the enclave image rotates: new EIF, new PCR0, submissions start reverting `InvalidSigner`, and the proposer calls `delete_proofs_by_tee_signer`. That is also the moment the proposer's configured image hash is most likely to still be the previous one — so the recorded signer is a phantom address and the delete clears the wrong set or zero rows while the real backlog stays put.

Phase 1 already anticipated this; `proof_collector.rs:409` carries the comment *"A zero-row delete means the just-submitted proof was not matched (e.g. its recorded signer diverged from the reported one)."* Stamping in the enclave closes that divergence class by construction: the signer is returned alongside the signature it just produced, from the same `&self`, so the two cannot disagree. The zero-row guard stays useful for other cases, but this no longer causes it.

### Why not just read `TEE_IMAGE_HASH` onchain?

An earlier revision of this PR kept host-side recovery and replaced the CLI flag with an onchain `AggregateVerifier.TEE_IMAGE_HASH()` read. That is the right *source* — `TEEProverRegistry` reads the same value to enforce `PCR0Mismatch` — but it does not close the gap. It is read once at proposer startup, so it goes stale on verifier rotation: the same drift, relocated. And it only addresses vector 1; the `proposer` vector stays open regardless. Once the enclave stamps the signer, nothing reads the value at all, so the field went away instead.

## Rollout

This PR carries the coordinated enclave rollout that phase 1 was structured to avoid. Deployment must build the new EIF, derive the new hash, register/authorize the enclave signer and attestation, and activate an `AggregateVerifier` configured with the matching `TEE_IMAGE_HASH`. Removing proposer configuration does not make the enclave image self-authorizing.

**Host and enclave must ship lock-step.** `ProofResult::Tee` gained a required field and crosses the vsock as length-prefixed **bincode** (`nitro-enclave/src/transport.rs`), which is non-self-describing and has no version negotiation. A new host against a pre-change EIF fails every `Frame::read` with a bincode EOF → `NitroHostError::FrameRead`. `serde(default)` does not help — bincode reads fields positionally and never reports one as "missing". The failure is loud and total rather than silent, and registering the new PCR0 onchain already forces lock-step, but `Dockerfile.nitro-host` ships only the host binary, so the two artifacts must be rolled together.

**Breaking config change.** `--tee-image-hash` / `BASE_PROPOSER_TEE_IMAGE_HASH` is gone. An unknown *env* var is ignored, but an unknown *flag* is fatal in clap — deployments passing `--tee-image-hash` will fail to start. Remove it from proposer env/args before rolling out.

**No back-fill for persisted `image_hash`.** `image_hash` had `serde(default)` but no `skip_serializing_if`, so it always serialized into `request_payload`. TEE session ids derive deterministically from the claimed output root, so replaying a pre-deploy row against an incoming payload without `image_hash` would fail JSON equality and wedge on `IdCollision { field: "request_payload" }` (the proposer only self-heals the `l1_head` collision field). This ships without a compatibility shim for that case: nitro-prover, prover-service, and proposer are redeploying together, so no pre-change row survives to be replayed. ZK sessions are unaffected either way — `image_hash` only ever serialized under `TeeProofRequest.proof`.

## How to test?

```
cargo test -p base-proof-primitives -p base-proof-contracts -p base-proposer \
  -p base-challenger -p base-proof-tee-nitro-host -p base-prover-service-protocol \
  -p base-prover-service-db --lib
```
310 passed, 0 failed. `cargo clippy` clean across the same set plus `base-proof-succinct-host-utils`.

Targeted checks:

- `cargo test -p base-proof-tee-nitro-host --lib proof_submitter` — `tee_proof_request_forwards_enclave_signer` asserts the address on `WorkerSubmitProofRequest` is the one the enclave stamped, not a recovered one.
- `cargo test -p base-prover-service-db --lib failed_requeue_payload_match_allows_only_l1_head_fields` — requeue matching still rejects a non-L1-head field change (now exercised via `proposer`, since `image_hash` no longer exists).
- Grep that nothing reads the removed value: `grep -rn "image_hash" --include="*.rs" crates/` should only hit `ProofJournal.tee_image_hash` (enclave-derived and signed), the enclave's own PCR0 derivation, and the unrelated SP1 `imageHash` in `crates/proof/succinct/utils/client/src/types.rs`.

End-to-end check of the batch-delete behaviour this PR is sharpening:

1. Run the proposer against an enclave whose PCR0 does **not** match the `AggregateVerifier`'s `TEE_IMAGE_HASH` (i.e. mid-rotation), and let it accumulate two or more succeeded TEE proofs.
2. Submit; the game reverts `InvalidSigner` and the proposer calls `delete_proofs_by_tee_signer`.
3. Expect `deleted_count` in the "Deleted invalid TEE signer proof requests" log to equal the whole backlog for that enclave. With host-side recovery this is where a phantom address yields a zero-row or partial delete and the backlog survives.
4. Cross-check that the `tee_signer` recorded on those rows equals the address the enclave reports via its signer query.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

jackchuma (approved):
well done!

PR #4325, opened 2026-08-09
feat(eip8130): reject codeless create in the enshrined apply path

chunter-cb:
## Summary

Mirrors the finalized `Keystore`'s `_buildDeploymentCode` `revert EmptyBytecode()` in the enshrined execution path.

- `build_deployment_code` now rejects empty bytecode with a new `ApplyError::EmptyBytecode` (before the size check), so `apply_create` / `compute_address` refuse a **codeless create**.
- A codeless create would otherwise install actor config at an address with no runtime code (nor a delegation), breaking the invariant that a key-bearing account is either a real EOA or a contract/delegation.
- Ordering matches the contract (deployment code is built before the initial-actor checks), so a codeless entry is rejected even with a well-formed actor set.
- The txpool structural gate already rejects `create.code.is_empty()` at admission; this adds the execution-layer defense-in-depth plus its pool rejection reason.

> Stacked on #4324.

## Test plan

- [x] `cargo test -p base-execution-eip8130` (new `create_rejects_codeless_account` + `build_deployment_code` empty/too-large coverage)
- [x] `cargo test -p base-execution-txpool -p base-common-evm`
- [x] `cargo clippy --all-targets` clean

### DIFF
--- a/crates/execution/eip8130/src/apply.rs
+++ b/crates/execution/eip8130/src/apply.rs
@@ -92,10 +92,30 @@ pub enum ApplyError {
     #[error("create initial actors must be strictly ascending by actor id")]
     ActorsNotSortedOrDuplicate,
 
+    /// A create entry's bytecode is empty. Mirrors `_buildDeploymentCode`'s
+    /// `revert EmptyBytecode()`: a codeless create is rejected because an account
+    /// carrying actor config but no runtime code (nor a delegation) would break
+    /// the EOA invariant — a key would exist for an address that is not an EOA.
+    #[error("create bytecode is empty")]
+    EmptyBytecode,
+
     /// A create entry's bytecode exceeds the 0xFFFF deployment limit. Mirrors
     /// `require(n <= 0xFFFF)`.
     #[error("create bytecode exceeds the 65535-byte limit")]
     BytecodeTooLarge,
+
+    /// A create entry's runtime code cannot be deployed. Mirrors `createAccount`'s
+    /// `AccountDeploymentFailed`: on the contract, `CREATE2` returns `address(0)`
+    /// when the runtime code is over the EIP-170 [`Eip8130Constants::MAX_CODE_SIZE`]
+    /// cap or leads with the EIP-3541 reserved `0xEF` byte. The enshrined apply
+    /// path deploys the code directly (`set_code`) rather than via `CREATE2`, so it
+    /// must reject those payloads explicitly — a leading-`0xEF` payload would
+    /// otherwise panic in `Bytecode::new_raw` instead of failing the transaction.
+    #[error("account {account} runtime code is not deployable (EIP-170 size / EIP-3541 prefix)")]
+    AccountDeploymentFailed {
+        /// The counterfactual address whose runtime code cannot be deployed.
+        account: Address,
+    },
 
     /// The account targeted by a create entry already has EIP-8130 state. Mirrors
     /// the CREATE2 collision that makes `createAccount` unrepeatable.
@@ -139,6 +159,18 @@ pub enum ApplyError {
     #[error("delegation cannot replace non-delegation code at account {account}")]
     NonDelegatableCode {
         /// The account whose existing code cannot be replaced by a delegation.
+        account: Address,
+    },
+
+    /// A delegation targeted an empty-code account that is
+    /// [`Eip8130Constants::FLAG_CONTRACT_ESTABLISHED`]. Empty code on a
+    /// keystore-established account (e.g. after an EIP-6780 same-transaction
+    /// `SELFDESTRUCT`) is not proof of a key-backed EOA, so it must not be
+    /// (re)delegated as if it were one — doing so would resurrect a CREATE2
+    /// address that no private key controls.
+    #[error("delegation cannot target empty-code contract-established account {account}")]
+    ContractEstablishedCodeless {
+        /// The keystore-established account whose empty code cannot be delegated.
         account: Address,
     },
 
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/eip8130/src/apply.rs
```diff
@@ -714,9 +750,14 @@ impl AccountChangeApplier {
 
     /// Builds an account's deployment code: a 14-byte EVM loader header that
     /// returns the trailing `bytecode` as the account's runtime code. Mirrors
-    /// `_buildDeploymentCode`.
+    /// `_buildDeploymentCode`: rejects empty `bytecode` with
+    /// [`ApplyError::EmptyBytecode`] (codeless creates are invalid) and bytecode
+    /// over `0xFFFF` bytes with [`ApplyError::BytecodeTooLarge`].
     pub fn build_deployment_code(bytecode: &[u8]) -> Result<Vec<u8>, ApplyError> {
         let n = bytecode.len();
+        if n == 0 {
```
xenoliss, line 783: I think this validation still allows malformed 7702 prefixed runtime code such as `0xef01`, which later panics in `Bytecode::new_raw` instead of returning an execution error. We should also reject 3541 code and enforce `MAX_CODE_SIZE` here as well.

crates/execution/eip8130/src/apply.rs
```diff
@@ -192,13 +211,25 @@ impl DelegationEffect {
...
         })?;
         if !can_replace {
             return Err(ApplyError::NonDelegatableCode { account: self.account });
         }
 
+        // Empty code on a keystore-established account is not proof of a key-backed
+        // EOA (e.g. an EIP-6780 same-transaction SELFDESTRUCT leaves EIP-8130 state
+        // behind empty code), so it must not be (re)delegated as if it were one.
+        // Reading the AccountConfiguration flag mirrors `Keystore.isContractEstablished`.
+        if code_is_empty
```
xenoliss, line 240: Is it expected that this check only applies to explicit delegations while `auto_delegate_codeless_sender` still delegates any codeless sender without checking `FLAG_CONTRACT_ESTABLISHED`?

crates/execution/eip8130/src/apply.rs
chunter-cb: Good catch — fixed. The enshrined path deploys the runtime code via `set_code(Bytecode::new_raw(..))` rather than `CREATE2`, so it now replicates the two checks `CREATE2` would enforce (and that make `createAccount` revert `AccountDeploymentFailed`): a leading `0xEF` (EIP-3541) and code over `MAX_CODE_SIZE` (EIP-170). I put the check in `apply_create` (new `ApplyError::AccountDeploymentFailed`) rather than `build_deployment_code`, to keep `build_deployment_code`/`compute_address` at parity with the reference `_buildDeploymentCode` (which only rejects empty / `> 0xFFFF` and still returns an address for oversize or `0xEF`-leading code). Both create paths (execution + estimation) route through `apply_create`, so the previous `0xef01` payload now fails the tx cleanly instead of panicking in `Bytecode::new_raw`. Also added the `0xEF` prefix reject to the txpool structural gate (which already capped `MAX_CODE_SIZE`), and unit coverage for the `0xEF` lead, the over-`MAX_CODE_SIZE` case, and the exact-`MAX_CODE_SIZE` boundary.

crates/execution/eip8130/src/apply.rs
chunter-cb: You're right, that was an inconsistency. `auto_delegate_codeless_sender` now carries the same guard as `DelegationEffect::install`: if the codeless sender is `FLAG_CONTRACT_ESTABLISHED` it fails closed with `ApplyError::ContractEstablishedCodeless` instead of being delegated to `DEFAULT_ACCOUNT`. This is reachable — e.g. an imported account (established regardless of code shape) whose delegation was later cleared, sending via a configured non-self actor — so both delegation entrypoints now refuse to resurrect an empty-code keystore-established account. Added a unit test asserting a plain codeless EOA is still auto-delegated while an established one is rejected.

chunter-cb: Thanks good call.

crates/execution/eip8130/src/apply.rs
chunter-cb: The enshrined deploy path already rejects this before it can reach `Bytecode::new_raw`: `apply_create` (apply.rs:635, added in this PR) returns `AccountDeploymentFailed` when `entry.code` leads with the reserved `0xEF` byte (covers `0xef01`) or exceeds `MAX_CODE_SIZE`, and the guard runs before any `set_code(Bytecode::new_raw(..))` (evm/eip8130.rs:1632 and :953). `build_deployment_code` here is intentionally kept as a faithful mirror of the contract's `_buildDeploymentCode` (empty + `0xFFFF` only) so `compute_address` derives the same CREATE2 address the contract would — moving the 3541/170 rejection into this helper would diverge address derivation for edge-case payloads. Net effect matches the contract: CREATE2 would return `address(0)`, we return `AccountDeploymentFailed`.

crates/execution/eip8130/src/apply.rs
chunter-cb: Symmetric now: `auto_delegate_codeless_sender` (evm/eip8130.rs:1662-1669) checks `is_contract_established(sender)` on the codeless branch and fails closed with `ApplyError::ContractEstablishedCodeless` instead of auto-delegating, matching the explicit-delegation guard here.

xenoliss (approved):

PR #4335, opened 2026-08-10
fix(sequencer): Prefetch Next L1 Origin

refcell:
## Summary

Move next-origin lookups off the sequencer build path into an owned `Idle` / `InFlight` / `Ready` state machine while bounding inline current-origin resolution. Speculative work stays tied to the current origin and observed L1 chain view, so resets and reorgs abort or discard stale work. This PR is stacked on #4334 and intentionally prefetches only `BlockInfo`; header and receipt reuse remain for the next PR.


crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
```diff
@@ -61,15 +96,68 @@ impl<P: L1OriginSelectorProvider + Send + Sync> OriginSelector for L1OriginSelec
...
 
         // Start building on the next L1 origin block if the next L2 block's timestamp is
         // greater than or equal to the next L1 origin's timestamp.
-        if let Some((next, _)) = self.next
+        if let Some(next) = next
             && next_l2_timestamp >= next.timestamp
         {
-            return Ok(next);
+            return Ok(*next);
         }
```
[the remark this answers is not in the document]
refcell: Fixed this by checking that the current origin exists before considering a ready successor. Added a regression test that prepares a successor, removes the current origin, enters recovery mode, and verifies `OriginNotFound` is returned instead of the cached successor.

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
```diff
@@ -93,124 +181,208 @@ impl<P: L1OriginSelectorProvider + Send + Sync> OriginSelector for L1OriginSelec
...
+                        "Background next-origin fetch failed; retrying on next tick"
                     );
-                    return;
+                    None
                 }
-            };
+                Err(_) => {
+                    Metrics::sequencer_l1_origin_fetch_timeouts_total("by_number").increment(1);
+                    None
+                }
```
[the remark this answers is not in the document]
refcell: Added a structured warning for by-number background fetch timeouts with the block number and configured timeout duration. The existing timeout metric remains in place.

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
```diff
@@ -61,21 +96,73 @@ impl<P: L1OriginSelectorProvider + Send + Sync> OriginSelector for L1OriginSelec
         is_recovery_mode: bool,
     ) -> Result<BlockInfo, L1OriginSelectorError> {
         self.select_origins(&unsafe_head, is_recovery_mode).await?;
+        self.choose_origin(&unsafe_head)
+    }
+}
+
+impl<P: L1OriginSelectorProvider> L1OriginSelector<P> {
+    /// Default deadline for resolving the current L1 origin on the build critical path.
+    pub const DEFAULT_L1_FETCH_TIMEOUT: Duration = Duration::from_secs(2);
```
0x00101010: curious why don't make this 12s?

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
```diff
@@ -93,124 +180,214 @@ impl<P: L1OriginSelectorProvider + Send + Sync> OriginSelector for L1OriginSelec
...
+                    }
+                };
+                match self.adopt_next(current_hash, chain_view, fetched) {
+                    NextSlot::Idle => self.l1.chain_view().map_or(NextSlot::Idle, |live_view| {
+                        self.spawn_next(current_hash, current_number, live_view)
+                    }),
+                    adopted => adopted,
+                }
+            }
+            in_flight @ NextSlot::InFlight { .. } => in_flight,
```
0x00101010: **Unpaced retry loop when drift-exceeded and the successor fetch is pending** (should-fix, partly pre-existing)

Once `max_sequencer_drift` (1800s post-Fjord) is exceeded and the selector is obligated to advance the origin, an unready successor busy-loops the actor:

```rust
in_flight @ NextSlot::InFlight { .. } => in_flight,        // poll_next: no await
next.copied().ok_or(L1OriginSelectorError::NotEnoughData(current))  // choose_origin + warn!
return Ok(None);                                            // build.rs + warn!
None => self.reset_immediately(),                           // ticker: "next tick" is now
```

Previously the inline by-number fetch paced each lap by an RPC round trip; now nothing does — the actor spins at CPU speed, emitting two `warn!`s per lap and spamming by-number fetches, exactly during an L1 incident (origin stuck ≥30 min). It self-heals once a successor is adopted, and the loop pre-exists in kind (with `confirmation_depth > 0` the old path was also unpaced), so not a blocker — but since this path is being touched: consider backing the ticker off ~block time when selection defers, instead of `reset_immediately()`.

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
refcell: Because it blocks the sequencer hot path

meyer9 (dismissed):

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
BrianBland: In that case, should this be < block time? 1s or less?

refcell: Yes it should be, this PR will need to be rebased on https://github.com/base/base/pull/4363 which uses `500ms` as the default

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
refcell: So this is correct but happens after 1800s or 30 minutes of l1 rpc downtime. I'm going to add a backoff for this code path but I think it's already very low risk of us ever losing l1 rpc connection for > 30 minutes.

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
refcell: Fixed so this now uses the configured sequencer l1 rpc timeout

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
```diff
@@ -93,124 +180,214 @@ impl<P: L1OriginSelectorProvider + Send + Sync> OriginSelector for L1OriginSelec
...
+            );
+            Metrics::sequencer_l1_origin_fetch_timeouts_total("by_hash").increment(1);
+            Err(L1OriginSelectorError::Timeout)
+        })
+    }
+
+    /// Advances the next-origin state machine without awaiting an unfinished fetch.
+    async fn poll_next(&mut self, current_hash: B256, current_number: u64) {
+        let Some(chain_view) = self.l1.chain_view() else {
+            self.next = NextSlot::Idle;
```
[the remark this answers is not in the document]
refcell: After tracing this closely, I opted not to change it. `chain_view()` is backed by a `watch` channel initialized to `None` and only ever `send_replace(Some(..))` afterward — it never transitions back to `None`, so `None` here means "before the first L1 head is observed" (startup), not a transient dropout between updates. Clearing to `Idle` in that state is also the exact invariant enforced everywhere else: `invalidate_next_if_chain_view_changed` treats `Some(stored) != live` (which includes `live == None`) as invalidation. Preserving a `Ready` slot under an unknown chain view would let `choose_origin`/`take_ready_if` promote it without the chain-view recheck that `adopt_next` guarantees — a regression. And by the time `poll_next` runs, the earlier `invalidate_*` call in the same `select_origins` invocation (no await between them on the non-recovery path) has already reduced `next` to `Idle`, so nothing is actually discarded here.

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
```diff
@@ -93,124 +180,214 @@ impl<P: L1OriginSelectorProvider + Send + Sync> OriginSelector for L1OriginSelec
...
-                self.next = next.map(|next| (next, chain_view));
+    /// Waits until an in-flight fetch can be settled through the production polling path.
+    #[cfg(test)]
+    async fn wait_for_inflight_completion(&self) {
+        loop {
+            if matches!(&self.next, NextSlot::InFlight { handle, .. } if handle.is_finished()) {
+                return;
             }
+            tokio::task::yield_now().await;
         }
```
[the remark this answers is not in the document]
refcell: Good catch — fixed. Added a `debug_assert!(matches!(&self.next, NextSlot::InFlight { .. }), ...)` at the top of `wait_for_inflight_completion` so a caller that enters without an in-flight fetch fails loudly in test builds instead of spinning forever on `yield_now()`. Kept it as an assertion rather than an early return so a misuse surfaces as a test failure instead of silently passing without actually waiting.

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
refcell: Added the backoff. The ticker now paces this path via `ScheduledTicker::schedule_l1_origin_retry`: the first `MAX_IMMEDIATE_L1_ORIGIN_RETRIES` (5) laps fire immediately to catch the common case where the background fetch is about to complete, after which retries are spaced by `L1_ORIGIN_RETRY_DELAY` (200ms) rather than `reset_immediately()`. The budget is reset via `reset_l1_origin_retry_budget` as soon as selection advances past the drift wait (or on an active transition). So the actor no longer hot-loops at CPU speed during an extended L1 outage — it settles into a bounded ~200ms cadence while still recovering promptly once a successor is adopted.

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
```diff
@@ -93,124 +154,193 @@ impl<P: L1OriginSelectorProvider + Send + Sync> OriginSelector for L1OriginSelec
...
-}
-
-impl<P: L1OriginSelectorProvider> L1OriginSelector<P> {
-    /// Creates a new [`L1OriginSelector`].
-    pub const fn new(cfg: Arc<RollupConfig>, l1: P) -> Self {
-        Self { cfg, l1, current: None, next: None }
+        // An in-flight background fetch is intentionally not exposed as `next`. Past the drift
+        // limit, the caller must defer the build until the verified successor becomes ready rather
+        // than continue building on an expired origin.
+        next.copied().ok_or(L1OriginSelectorError::NotEnoughData(current))
```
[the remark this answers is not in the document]
refcell: Correct — simplified. At line 166 any `Some(next)` is already handled earlier: a successor at or behind `next_l2_timestamp` returns at the first check, and one ahead returns the current origin, so `next` is provably `None` here. Replaced `next.copied().ok_or(..)` with a direct `Err(NotEnoughData(current))` and added a `debug_assert!(next.is_none(), ...)` plus a comment to make the invariant explicit rather than leaving the dead path.

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
```diff
@@ -93,124 +154,193 @@ impl<P: L1OriginSelectorProvider + Send + Sync> OriginSelector for L1OriginSelec
...
+                        None
+                    }
+                };
+                match self.adopt_next(current_hash, chain_view, fetched) {
+                    NextSlot::Idle => self.l1.chain_view().map_or(NextSlot::Idle, |live_view| {
+                        self.spawn_next(current_hash, current_number, live_view)
+                    }),
+                    adopted => adopted,
+                }
+            }
```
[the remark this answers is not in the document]
refcell: Agreed this is safe, and the behavior is intentional. The re-query at line 269 deliberately grabs the freshest `chain_view` for the *new* fetch rather than reusing the poll's captured value, because `adopt_next` returning `Idle` may itself be caused by the view changing during the fetch (the guard at line 295), so the captured view can be stale precisely here. The TOCTOU window only ever costs one extra speculative retry: an in-flight fetch records the view it started under, and the next `poll_next` invalidates + re-spawns on a `fetch_view != chain_view` mismatch. Adoption is double-guarded on parent hash and live view, so a stale-view fetch can never be adopted, and it all runs off the block-building path. `test_completed_fetch_retries_with_latest_chain_view` covers exactly this. No change needed.

crates/consensus/service/src/actors/sequencer/l1_origin/selector.rs
refcell: Fixed. Now it will retry 5 times and then backoff to the configured sequencer l1 rpc timeout

0x00101010 (approved):

PR #4372, opened 2026-08-11
feat(shadow-indexer): add shadow-indexer ExEx for block/tx metadata persistence

wbj-cb:
## Summary

Adds `base-shadow-indexer`, an execution extension (`ExEx`) that captures committed
execution blocks and persists per-block metadata plus the full executed block and
receipts (as JSONB) to Postgres for out-of-band block verification. **Disabled by
default** and a no-op unless explicitly enabled; when off it installs nothing on the
node.

The ExEx follows the established `base-execution-exex` pattern: it drains chain
notifications (`ChainCommitted` / `ChainReorged` / `ChainReverted`), emits the
mandatory `FinishedHeight` event, and hands rows to a background writer task over a
bounded channel so persistence never blocks execution.

## What's included

**`base-shadow-indexer`** (the ExEx)
- Builds one block row per committed block. Identity/bookkeeping fields
  (`number`, `hash`, `reorged_out`, `canonical_hash`, `created_at`) are materialized
  as real columns; the full `RecoveredBlock` + receipts are stored in a single JSONB
  `payload` so downstream consumers can evolve the shape they read without a schema
  migration.
- Reorg handling: blocks reorged out are flagged `reorged_out = true` and stamped with
  the new `canonical_hash` at their height (`None` when the new chain has no block at
  that height). `ChainReverted` marks unwound blocks `reorged_out = true` with no
  canonical hash.
- Background writer: batches rows (100 / 1s) on a bounded channel, **bounded retry**
  (3 attempts, 200ms backoff), then drops the batch with a single structured error log
  rather than looping forever. Rows are timestamped once on receipt so retries preserve
  build time. The configured builder version is injected into each payload before
  persistence.

**`base-shadow-indexer-db`** (access layer)
- `ShadowDbConfig::init_pool()` connects and runs embedded `sqlx` migrations.
- `ShadowBlockRepo`: deduped batch upsert (`INSERT ... ON CONFLICT (number, hash)
  DO UPDATE`) + a `list_by_number_range` query. Single table `shadow_blocks`, primary
  key `(number, hash)`, `payload JSONB NOT NULL`, indexed on `number DESC`.
- In-batch dedup (last-write-wins on `(number, hash)`) so a commit-then-reorg within one
  flush window doesn't trip Postgres error 21000 (cannot touch the same conflict key
  twice in one statement).

**Node wiring** — registered via the `BaseNodeExtension` seam on both the standard
execution node and the `base-builder` binary, gated by `--enable-shadow-indexer`. The
writer runs as a critical task: pool init failure panics at startup (fail-fast); a DB
outage during operation retries then drops rows and keeps the node running.

## Configuration

| Flag | Env | Default |
| --- | --- | --- |
| `--enable-shadow-indexer` | `ENABLE_SHADOW_INDEXER` | `false` |
| `--shadow-indexer.database-url` | `SHADOW_INDEXER_DATABASE_URL` | (required when enabled) |
| `--shadow-indexer.max-connections` | `SHADOW_INDEXER_MAX_CONNECTIONS` | `5` |
| `--shadow-indexer.connection-timeout` | `SHADOW_INDEXER_CONNECTION_TIMEOUT` | `30s` |

## Operational notes

- **Not block-production-sensitive.** The ExEx is an observability sidecar that consumes
  `ChainCommitted`/`ChainReorged`/`ChainReverted` after the fact; it does not participate
  in payload building or block production.
- **Shadow DB is a startup liveness dependency.** If the DB is unreachable/misconfigured
  when the node starts, pool init panics and the node fails fast (intentional for canary
  nodes). Mid-run DB outages do **not** take the node down — affected batches are retried
  3× then dropped with an error log, so shadow rows can be lost silently during an outage.

## Test plan

- [x] Unit — writer (bounded retry, stamp-once) and ExEx notification tests
      (commit + reorg `canonical_hash` paths + revert path).
- [x] Unit — repo dedup (last-write-wins, distinct-hash-at-same-height).
- [x] Unit — CLI arg parsing / `enable`-flag requirements.
- [x] E2E — `etc/systems/tests/shadow_indexer.rs`: boots a node stack against an
      ephemeral Postgres (testcontainers) and asserts committed blocks land in the DB.
- [x] `cargo +nightly fmt --check`, `clippy -D warnings`, `cargo deny`, `zepter` clean.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/shadow-indexer-db/src/repo.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /// Insert a batch of shadow block rows.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the insert fails.
+    pub async fn insert_batch(&self, rows: &[ShadowBlockRow]) -> Result<usize> {
+        // 11 columns per row are bound. Postgres caps a single statement at 65_535
+        // bind parameters, so keep chunks below 65_535 / 11 ≈ 5_957 rows.
```
jowparks: Instead of hardcoding this, can we infer based on number of number of properties on schemas, just wondering if this will ever break if we add columns during hardforks.

crates/execution/shadow-indexer/src/exex.rs
```diff
@@ -0,0 +1,459 @@
...
+                        new_block_hash = ?new.tip().hash(),
+                        "ChainReorged notification received"
+                    );
+                    self.handle_chain_reorged(old, new).await?
+                }
+                ExExNotification::ChainReverted { old } => {
+                    info!(
+                        target: "base::shadow-indexer",
+                        old_block_number = old.tip().header().number(),
+                        old_block_hash = ?old.tip().hash(),
```
jowparks, line 57: What is chain reverted? Wondering if we should put a todo in here somewhere

crates/execution/shadow-indexer-db/src/repo.rs
```diff
@@ -0,0 +1,275 @@
...
+                    .push_bind(entry.gas_used)
+                    .push_bind(&entry.state_root)
+                    .push_bind(entry.reorged_out)
+                    .push_bind(&entry.canonical_hash)
+                    .push_bind(&entry.builder_version)
+                    .push_bind(entry.created_at);
+            });
+
+            query_builder.push(
+                " ON CONFLICT (number, hash) DO UPDATE SET \
```
jowparks, line 60: Should this just be on conflict of number, not hash? If a block is reorged the hash will be different and only the number will be the same, this wouldn't cause conflict trigger.

crates/execution/shadow-indexer/src/writer.rs
```diff
@@ -0,0 +1,338 @@
...
+impl ShadowWriter {
+    /// Spawns the writer task on the provided executor.
+    pub fn spawn(
+        executor: TaskExecutor,
+        rx: mpsc::Receiver<ShadowBlockRecord>,
+        db_config: ShadowDbConfig,
+        builder_version: String,
+    ) {
+        let writer = Self { rx, db_config, builder_version };
+        executor.spawn_critical_task("shadow-indexer-writer", async move {
```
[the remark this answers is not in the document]
jowparks: This is not a production node, it is specifically built for canary so I think it actually might make sense to bring the whole thing down in a fault. @wbj-cb can nomment on whether this assumption is correct.

crates/execution/shadow-indexer-db/src/repo.rs
wbj-cb: This is meant for inserts of the same number + hash to be idempotent, not to handle re-orgs.

crates/execution/shadow-indexer/src/exex.rs
wbj-cb: I made it so that on chain revert it just marks all of the blocks in the revert as re-orged out without a canonical block header

crates/execution/shadow-indexer-db/migrations/0001_init.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE shadow_blocks(
+  number BIGINT NOT NULL,
+  hash TEXT NOT NULL,
+  reorged_out BOOL NOT NULL DEFAULT false,
+  canonical_hash TEXT,
+  created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+  payload JSONB NOT NULL,
+  PRIMARY KEY(number, hash)
+);
+CREATE INDEX idx_shadow_blocks_number ON shadow_blocks(number DESC);
```
haardikk21: i'm unsure if sqlx magically handles this for us somehow but perhaps we should include `IF NOT EXISTS` in these statements

crates/execution/shadow-indexer-db/src/models.rs
```diff
@@ -0,0 +1,63 @@
...
+/// Row representation for a shadow indexer block.
+///
+/// Only the columns required for identity, reorg bookkeeping, and range scans are
+/// materialized as real columns. All descriptive block and transaction metadata
+/// lives in a single JSONB `payload`, so downstream consumers can evolve the shape
+/// they read without a database migration.
+#[derive(Clone, Debug, sqlx::FromRow)]
+pub struct ShadowBlockRow {
+    /// Block number.
+    pub number: i64,
```
haardikk21, line 15: minor nit: we can use unsigned numbers here, it should never be <0

crates/execution/shadow-indexer-db/src/models.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// Block hash.
+    pub hash: String,
```
haardikk21: minor nit, non blocking: wonder if this should just be bytes since we know these are always going to be a 32 byte hash. strings will double the overhead

crates/execution/shadow-indexer-db/src/models.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// Whether the block was reorged out.
+    pub reorged_out: bool,
+    /// Canonical block hash at the same height after reorg.
+    pub canonical_hash: Option<String>,
```
haardikk21: same as `hash` re: storage type

crates/execution/shadow-indexer-db/src/models.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// Row creation time.
+    pub created_at: DateTime<Utc>,
+    /// Flexible block and transaction metadata persisted as JSONB.
+    #[sqlx(json)]
+    pub payload: ShadowBlockPayload,
```
haardikk21, line 26: instead of JSON encoded `ShadowBlockPayload` - i think we should store literally the entire RPC block payload as returned via `getBlockByHash` or `getBlockByNumber`. I forget the exact type name but there should be a `BaseBlockRpcPayload` or something like that being exported from one of the `common` crates in this repo

crates/execution/shadow-indexer-db/src/models.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub sender: Option<String>,
+    /// EIP-2718 transaction type byte (0x7e denotes an OP deposit).
+    pub tx_type: u8,
+    /// Effective priority fee per gas (tip) in wei, as a base-10 string to preserve full u128 range.
+    pub effective_priority_fee_per_gas: Option<String>,
+    /// Block base fee per gas in wei.
+    pub base_fee_per_gas: Option<u64>,
+    /// Gas consumed by this transaction.
+    pub gas_used: u64,
+}
```
haardikk21: dont need these and related dependents if we can just store raw block payload entirely following above

crates/execution/shadow-indexer/src/exex.rs
haardikk21: reorg = chain replaces previously committed blocks
revert = chain just rolls back to a previously committed block without replacements

practically we shouldnt really have reverts

crates/execution/shadow-indexer/src/exex.rs
```diff
@@ -0,0 +1,449 @@
...
+        while let Some(notification) = ctx.notifications.try_next().await? {
+            let fully_processed = match &notification {
+                ExExNotification::ChainCommitted { new } => {
+                    debug!(
+                        target: "base::shadow-indexer",
+                        block_number = new.tip().header().number(),
+                        block_hash = ?new.tip().hash(),
+                        "Committed chain notification received"
+                    );
+                    self.emit_canonical_blocks(new).await?
```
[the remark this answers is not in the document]
haardikk21: the bot's comment here is maybe reasonable. terminating the exex in case pg is unreachable or smth maybe bad, wonder if we can retry in a way that doesn't block emitting the `FinishedHeight` event? though probably not super important and non-blocking, i wouldnt expect pg to go down

crates/execution/shadow-indexer/src/writer.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        let pool = match self.db_config.init_pool().await {
+            Ok(pool) => pool,
+            Err(error) => {
+                error!(
+                    target: "base::shadow-indexer",
+                    error = ?error,
+                    "Failed to initialize shadow indexer database pool"
+                );
+                panic!("failed to initialize shadow indexer database pool: {error:?}");
```
haardikk21: i guess we can just merge these into a single `panic` rather than error then panic?

crates/infra/shadow-metrics/migrations/001_shadow_metrics.sql
```diff
@@ -1,32 +1,10 @@
...
-  parent_hash TEXT NOT NULL,
-  timestamp BIGINT NOT NULL,
-  tx_count INT NOT NULL,
-  gas_used BIGINT NOT NULL,
-  da_bytes BIGINT NOT NULL,
-  state_root TEXT NOT NULL,
   reorged_out BOOL NOT NULL DEFAULT false,
   canonical_hash TEXT,
-  builder_version TEXT NOT NULL,
   created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
```
haardikk21, line 13: isnt this entire file the same as `0001_init.sql` in `crates/execution/shadow-indexer-db`?

if there is duplication going on it implies part of this or the shadow-indexer-db (or in full) needs to be moved into a more common place that can be imported and reused as necessary. perhaps `crates/utilities` ?

haardikk21 (changes requested):

cb-heimdall:
Review Error for haardikk21 @ 2026-08-13 15:06:11 UTC

crates/infra/shadow-metrics/migrations/001_shadow_metrics.sql
wbj-cb: the intention is that shadow-metrics will depend on shadow db for db ops. going to revert this change so its just untouched and we can do a clean cutover to the crate dep

crates/execution/shadow-indexer-db/src/models.rs
wbj-cb: postgres doesn't have unsigned types and sqlx doesn't support binding a signed postgres type to an unisnged rust type. We could add additional code to the do the conversion but doesn't feel necessary now

jowparks (dismissed):

haardikk21 (approved):
LGTM! :D

PR #4424, opened 2026-08-14
perf(builder): optionally distribute predicate blockers by tx hash

BrianBland:
## Summary

Optionally prevent predicate ordering from concentrating parked transactions under the same state key.

A transaction can have multiple failing predicates but only needs to be indexed under one of them. Always choosing the first failing predicate can create a large hot bucket when many transactions list the same high-value state first.

This PR adds an opt-in policy that derives a deterministic starting offset from the transaction hash and scans predicates circularly from that position. This distributes parked transactions across their failing predicates without mutating predicate lists, allocating shuffled copies, or introducing nondeterminism. The existing first-failing behavior remains the default.

It also adds metrics for evaluating blocker topology and rescan costs under both policies, plus skewed benchmarks covering up to 100,000 parked transactions.

## Configuration

Hash rotation is disabled by default and can be enabled with either:

- CLI: `--builder.hash-rotate-predicate-blockers`
- Environment: `BUILDER_HASH_ROTATE_PREDICATE_BLOCKERS=true`

The corresponding `BuilderConfig` field is `hash_rotate_predicate_blockers`.

## Metrics

All predicate-parking metrics carry a `policy` label of either `first_failing` or `hash_rotated`:

- `validity_predicate_parked_total` — transactions parked due to an unsatisfied predicate, additionally labeled by `predicate_count` (`17+` is the overflow bucket)
- `validity_predicate_rescan_transactions` — parked transactions considered after each relevant state change
- `validity_predicate_rescan_duration` — predicate rescan duration in seconds
- `validity_predicate_parked_transactions` — transactions remaining predicate-parked after a selection pass
- `validity_predicate_blocker_keys` — distinct blocker state keys after a selection pass
- `validity_predicate_max_blocker_bucket_size` — largest blocker bucket after a selection pass

Together, these expose parking volume, hot-key concentration, state-change fanout, and rescan cost for comparing both policies in production.

## Benchmark topology

The benchmark models an intentionally extreme state-skew scenario:

- every transaction has multiple failing predicates;
- predicate zero reads the same shared “hot” state key;
- the remaining predicates read transaction-specific “cold” keys;
- changing the hot key wakes every transaction indexed under it.

The wakeup benchmark measures only the time to identify affected parked transactions. It excludes subsequent predicate reads, promotion, and re-indexing.

## Hot-key wakeup performance

With first-failing selection, all 100,000 transactions are indexed under the shared predicate. Hash rotation places approximately `1 / P` of them there.

| Predicates per tx | First-failing candidates | First-failing lookup | Rotated candidates | Rotated lookup |
|---:|---:|---:|---:|---:|
| 2 | 100,000 | ~190µs | 50,000 | ~95µs |
| 8 | 100,000 | ~189µs | 12,500 | ~24µs |
| 64 | 100,000 | ~189µs | 1,563 | ~3µs |

Lookup time scales with the number of transactions returned, producing the expected approximately `1 / P` reduction in hot-key fanout.

## Initial parking tradeoff

Distribution is not free. In this benchmark, rotation often starts with a transaction-specific cold key instead of the shared hot key.

| Workload | Blocker policy | Full selection + parking |
|---|---:|---:|
| 100k txs × 2 predicates | Always first failing predicate | ~102ms |
| | Hash-rotated failing predicate | ~130ms |
| 100k txs × 8 predicates | Always first failing predicate | ~110ms |
| | Hash-rotated failing predicate | ~142ms |

This is a roughly 27–29% initial-parking penalty in the deliberately adversarial unique-cold-state topology.

Whether that tradeoff wins in production depends on actual predicate counts, state-key skew, and how frequently hot state changes. The benchmarks establish worst-case boundaries; the new metrics allow production workloads to validate the tradeoff before enabling the policy by default.

## Correctness

Rotation changes only which failing predicate is used as the current blocker.

When its state changes, all predicates are still re-evaluated. The transaction is promoted only when every predicate passes; otherwise, it is re-indexed under another failing predicate. With the flag disabled, predicate scanning preserves the existing order.

### DIFF
--- a/crates/builder/cli/src/args.rs
+++ b/crates/builder/cli/src/args.rs
@@ -242,6 +242,14 @@ pub struct Args {
     )]
     pub manifest_precheck_enabled: bool,
 
+    /// Distribute predicate-parked transactions across unsatisfied predicate state keys.
+    #[arg(
+        long = "builder.hash-rotate-predicate-blockers",
+        env = "BUILDER_HASH_ROTATE_PREDICATE_BLOCKERS",
+        default_value = "false"
+    )]
+    pub hash_rotate_predicate_blockers: bool,
+
     /// Flashblocks configuration
     #[command(flatten)]
     pub flashblocks: FlashblocksArgs,
@@ -298,6 +306,7 @@ impl Default for Args {
             rejection_cache_ttl_secs: 1800,
             sampling_ratio: 100,
             manifest_precheck_enabled: true,
+            hash_rotate_predicate_blockers: false,
             flashblocks: FlashblocksArgs::default(),
             transaction_events: TransactionEventsArgs::default(),
             shadow_indexer: ShadowIndexerArgs::default(),
@@ -351,6 +360,7 @@ impl Args {
             rejected_tx_channel_size: self.rejected_tx_channel_size,
             max_rejected_txs_per_block: self.max_rejected_txs_per_block,
             manifest_precheck_enabled: self.manifest_precheck_enabled,
+            hash_rotate_predicate_blockers: self.hash_rotate_predicate_blockers,
         })
     }
 }
@@ -392,6 +402,7 @@ mod tests {
         assert_eq!(config.block_time, Duration::from_millis(1000));
         assert!(config.max_gas_per_txn.is_none());
         assert!(config.manifest_precheck_enabled);
+        assert!(!config.hash_rotate_predicate_blockers);
     }
 
     #[test]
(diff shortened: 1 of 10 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/2</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4432, opened 2026-08-14
feat(txpool): Add Block Number And Flashblock Index Predicates

refcell:
## Summary

Adds two new validity predicate types, `block_number` and `flashblock_index`, to the experimental transaction submission API. Unlike the existing `balance` and `storage` predicates that read database state, these compare against the coordinates of the block and flashblock currently being built. The builder now threads a `PredicateContext` through both predicate evaluation sites so these block-level conditions are enforced during flashblock construction. Wire encoding and RPC ingress required no changes since they handle predicates generically through serde.

Implements BASE-275.

### DIFF
--- a/crates/execution/txpool/src/validity.rs
+++ b/crates/execution/txpool/src/validity.rs
@@ -490,7 +490,7 @@ mod tests {
     }
 
     #[test]
-    fn block_number_predicate_round_trips() {
+    fn flashblock_index_predicate_round_trips() {
         let predicate = ValidityPredicate::FlashblockIndex {
             op: ValidityOperator::LessThanOrEqual,
             value: U256::from(7),
@@ -503,6 +503,30 @@ mod tests {
                 "params": {
                     "op": "<=",
                     "value": "0x7",
+                },
+            })
+        );
+        assert_eq!(
+            serde_json::from_value::<ValidityPredicate>(serde_json::to_value(&predicate).unwrap())
+                .unwrap(),
+            predicate
+        );
+    }
+
+    #[test]
+    fn block_number_predicate_round_trips() {
+        let predicate = ValidityPredicate::BlockNumber {
+            op: ValidityOperator::GreaterThanOrEqual,
+            value: U256::from(0x1234),
+        };
+
+        assert_eq!(
+            serde_json::to_value(&predicate).unwrap(),
+            json!({
+                "type": "block_number",
+                "params": {
+                    "op": ">=",
+                    "value": "0x1234",
                 },
             })
         );
--- a/crates/execution/txpool/src/validity.rs
+++ b/crates/execution/txpool/src/validity.rs
@@ -427,10 +427,9 @@ mod tests {
 
     #[test]
     fn deserializes_block_number_and_flashblock_index_predicates() {
-        let block_number: ValidityPredicate = serde_json::from_str(
-            r#"{"type":"block_number","params":{"op":">=","value":"0x64"}}"#,
-        )
-        .unwrap();
+        let block_number: ValidityPredicate =
+            serde_json::from_str(r#"{"type":"block_number","params":{"op":">=","value":"0x64"}}"#)
+                .unwrap();
         let flashblock_index: ValidityPredicate = serde_json::from_str(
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/txpool/src/validity.rs
```diff
@@ -104,6 +140,12 @@ impl ValidityPredicate {
                 let storage = db.storage(*address, *slot)? & *mask;
                 Ok(op.matches(storage, *value))
             }
+            Self::BlockNumber { op, value } => {
+                Ok(op.matches(U256::from(context.block_number), *value))
```
BrianBland, line 144: Minor, but perhaps ValidityOperator should be made generic over numeric types so we can bind BlockNumber and FlashblockIndex values to u64 instead of expanding these to U256

BrianBland, line 144: There could be an argument to be made for having `matches` return something other than a `bool` so we can handle `valid`, `temporarily invalid`, and `permanently invalid`.
FWIW doing this at the individual predicate level might be suboptimal, as block number and flashblock index may need to be evaluated in tandem to determine permanent invalidity.

BrianBland (dismissed):

crates/execution/txpool/src/validity.rs
```diff
@@ -377,4 +424,92 @@ mod tests {
...
+        };
+
+        assert!(block_number.matches(&mut db, &context).unwrap());
+        assert!(!block_number_too_low.matches(&mut db, &context).unwrap());
+        assert!(flashblock_index.matches(&mut db, &context).unwrap());
+        assert!(!flashblock_index_mismatch.matches(&mut db, &context).unwrap());
+    }
+
+    #[test]
+    fn block_number_predicate_round_trips() {
```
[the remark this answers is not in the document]
refcell: Fixed in 1725077d6: renamed this test to `flashblock_index_predicate_round_trips` and added a separate `block_number_predicate_round_trips` that actually round-trips a `BlockNumber` predicate.

crates/execution/txpool/src/validity.rs
refcell: Good thinking. I left the values as `U256` deliberately for the wire-format to serializes as a `U256` hex string today, and balance/storage genuinely need `U256`. Binding the block number/flashblock index to `u64` would either make their `value` serialize as a JSON number (inconsistent with the other predicate types) or require custom serde to keep the hex encoding. Happy to revisit if we decide per-type value encodings are worth the inconsistency, but leaning toward keeping the single `U256` representation for now.

refcell: Yep that makes sense. In #4441 I do have `ValidityPredicate::is_batch_expired(predicates, ctx)` which does check `block_number` and `flashblock_index` together to determine permanent invalidity. Otherwise, the match here is just a bool for current validity. I'll leave it out of this PR for now though

BrianBland (approved):

PR #4433, opened 2026-08-14
feat(shadow-metrics): emit metrics from shadow candidate blocks

jowparks:
## Summary

The consensus shadow sequencer privately builds blocks, inserts them into its isolated execution layer, then forcibly reorgs them out when the real sequencer's gossiped blocks arrive. The shadow-indexer ExEx persists both branches, so shadow candidates are exactly the rows where `reorged_out = true AND canonical_hash IS NOT NULL`.

This adds a polling reader to the (until now noop) `shadow-metrics` service that inspects those rows and emits Prometheus metrics.

Closes CHAIN-4794 (total gas), CHAIN-4795 (transaction count) and CHAIN-4797 (priority-fee ordering). CHAIN-4793 is emitted elsewhere and CHAIN-4796 is already covered by `base_builder.flashblock_count`, so both are out of scope.

## What changed

`shadow-indexer-db` gains an `updated_at` column, a composite `(updated_at, number, hash)` index, and the `shadow_metrics_cursor` table, plus the read side: `ShadowBlockCursor`, `ShadowBlockRawRow`, `list_reorged_since`, `max_cursor` and `ShadowMetricsCursorRepo`. It owns every statement against its own schema.

`shadow-metrics` becomes a read-only consumer. Its own migrator is deleted, `sink.rs` becomes `store.rs`, and the crate gains `metrics.rs`, `stats.rs` and `reader.rs`. The binary runs the reader instead of a heartbeat.

## Design notes worth reviewing

**The cursor is `updated_at`, not block number.** A shadow row is written twice: at insert, where it is indistinguishable from a canonical block, and again at reconciliation. A block-number cursor would read it during the first window and permanently misclassify a shadow block as canonical. Prometheus samples cannot be retracted.

**The cursor is composite, not a bare timestamp.** Postgres `now()` is transaction-start time, so every row from one `insert_batch` shares an `updated_at`. When `LIMIT` slices through such a tie group, a bare-timestamp cursor advances to that timestamp and the strict `>` skips the rest of the group, silently and permanently. The systems test seeds 55 rows in one transaction and drains them with a cap of 10; reducing the predicate to a bare timestamp makes it emit 10 and lose 45.

**Unwinds are partitioned in Rust, not filtered in SQL.** `ChainReverted` produces `reorged_out = true` with `canonical_hash = NULL`. Those rows are counted on their own series and never inspected, but they still advance the cursor. Filtering them out in SQL would make them uncountable and would strand the cursor behind them.

**Reads use a raw JSON payload.** The typed row decodes JSONB during sqlx row decoding, so one malformed payload aborts the entire fetch — the reader would never see the row, could not count it, and could not advance past it. Per-row decoding turns a wedged cursor into a skipped row and a counter.

**Emission happens before the cursor is persisted.** Duplicate histogram samples barely move a distribution; holes are permanent.

## Testing

`etc/systems/tests/shadow_metrics_reader.rs` — ten cases against a Postgres testcontainer, covering shadow-versus-canonical selection, exactly-once emission, unreconciled rows, pipeline unwinds, a poison payload between two good rows, deposit handling, sawtooth and descending fee curves, the row cap, tie-group draining, restart resumption, and the monotonic cursor guard.

The binary was also run by hand: the exporter reports the `shadow_metrics_*` series, confirming the `metrics` feature activation is real and not a `NoopMetric` no-op.

Note system tests do not run in PR CI (`run_system_tests: false`); they run in the merge queue.

## Not in this PR

Wiring `ShadowIndexerArgs` and `ShadowIndexerExtension` into `base-builder` is a separate ticket and **blocks all end-to-end verification** — `base-builder` does not currently accept `--enable-shadow-indexer`. Postgres credentials for `base-public-3`, the metrics port and Datadog annotation in `protocols/shadow-metrics`, and the `protocols/base-metal` env change are likewise separate. Until those land the reader computes correctly and emits into the void.

The shadow-indexer ExEx it reads from landed in #4372.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

bin/shadow-metrics/src/main.rs
```diff
@@ -167,28 +203,28 @@ async fn shutdown_signal() {
...
 
 async fn healthz_handler() -> &'static str {
     "ok\n"
 }
 
 async fn readyz_handler(State(state): State<HealthState>) -> Response {
-    let readiness = match &state.sink {
-        Some(sink) => sink.check_schema_ready().await,
+    if state.reader_alive.as_ref().is_some_and(|alive| !alive.load(Ordering::Acquire)) {
+        error!("shadow-metrics reader is no longer running");
```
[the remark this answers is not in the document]
jowparks: Good catch — dropped the log entirely in debe98dc.

Agreed it carried no information: `spawn_reader` already logs the real cause once, at the moment the task actually stops, and the 503 is the signal the probe consumes. Left a comment on the branch so it doesn't get re-added.

Kept the `error!` in the `check_schema_ready` arm below, which does have the same per-probe cadence. The difference is that one carries the underlying `sqlx` error and is the only place that failure surfaces, so removing it would lose the diagnostic entirely rather than de-duplicate it.

bin/shadow-metrics/README.md
```diff
@@ -20,13 +36,25 @@ container networking.
...
-| Env var                              | Default | Description                                  |
-| ------------------------------------ | ------- | -------------------------------------------- |
-| `SHADOW_METRICS_POSTGRES_URL`        | (unset) | Postgres URL. When unset, DB is disabled.    |
-| `SHADOW_METRICS_POSTGRES_MAX_CONNECTIONS` | `10` | Max Postgres pool connections.               |
-| `SHADOW_METRICS_HTTP_PORT`           | `9101`  | Health server port.                          |
-| `SHADOW_METRICS_HEARTBEAT_INTERVAL_SECS` | `30` | Idle heartbeat log interval in seconds.      |
-| `SHADOW_METRICS_METRICS_PORT`        | `9003`  | Prometheus metrics port.                     |
+| Env var                                   | Default | Description                                    |
+| ----------------------------------------- | ------- | ---------------------------------------------- |
+| `SHADOW_METRICS_POSTGRES_URL`             | (unset) | Postgres URL. When unset, DB is disabled.      |
```
wbj-cb, line 41: what does "db is disabled" mean in this context? Shouldn't it kill the service because it can't do anything?

wbj-cb (approved):

bin/shadow-metrics/README.md
jowparks: Yea that is fair, will ship as is so we can test but will write TODO for myself to remove in follow up.

wbj-cb (approved):

dguenther (approved):

cb-heimdall:
Review Error for dguenther @ 2026-08-17 17:38:57 UTC

PR #4436, opened 2026-08-14
feat(txpool): Validate Validity Predicate Payloads At Ingress

refcell:
## Summary

Enforces the validity-predicate payload rules that do not depend on the transaction type at the base_sendRawTransactionValidity ingress. The ingress now rejects an empty validity list (a predicate-less submission has no advanced semantics) and rejects storage predicates whose comparison value has bits set outside its mask, since those bits could never match. The checks live in a reusable ValidityPredicate::validate_batch helper so the same gate can back future ingress paths, and the existing predicate-count bound is folded into it.

Implements BASE-276 (child of BASE-258).

### DIFF
--- a/crates/execution/txpool/src/validity.rs
+++ b/crates/execution/txpool/src/validity.rs
@@ -147,13 +147,22 @@ impl ValidityPredicate {
         U256::MAX
     }
 
-    /// Returns whether this predicate's parameters are internally consistent.
+    /// Validates that this predicate's parameters are internally consistent.
     ///
     /// A storage predicate's comparison value must not set any bits outside its
-    /// mask, since the loaded value is masked before comparison.
-    #[must_use]
-    pub fn has_valid_params(&self) -> bool {
-        !matches!(self, Self::Storage { mask, value, .. } if (*value & !*mask) != U256::ZERO)
+    /// mask, since the loaded value is masked before comparison. `index` is the
+    /// predicate's position within its submitted batch and is embedded into any
+    /// returned error so callers can report which predicate failed. Returning the
+    /// specific [`ValidityPredicateError`] keeps the diagnosis with the predicate
+    /// itself, so new failure modes surface as distinct errors instead of
+    /// collapsing into a single caller-assigned variant.
+    pub fn validate_params(&self, index: usize) -> Result<(), ValidityPredicateError> {
+        if let Self::Storage { mask, value, .. } = self
+            && (*value & !*mask) != U256::ZERO
+        {
+            return Err(ValidityPredicateError::StorageValueOutsideMask { index });
+        }
+        Ok(())
     }
 
     /// Validates a batch of predicates submitted at ingress.
@@ -168,9 +177,7 @@ impl ValidityPredicate {
             return Err(ValidityPredicateError::TooMany { count: predicates.len(), max });
         }
         for (index, predicate) in predicates.iter().enumerate() {
-            if !predicate.has_valid_params() {
-                return Err(ValidityPredicateError::StorageValueOutsideMask { index });
-            }
+            predicate.validate_params(index)?;
         }
         Ok(())
     }
@@ -247,11 +254,7 @@ impl ValidatedTransactionExtensions<BasePooledTransaction> for TransactionValidi
     /// receives ordinary transactions that carry no predicates.
     fn apply(self, tx: BasePooledTransaction) -> Result<BasePooledTransaction, ExtensionError> {
         for (index, predicate) in self.validity.iter().enumerate() {
-            if !predicate.has_valid_params() {
-                return Err(ExtensionError(format!(
-                    "storage predicate at index {index} has value bits set outside its mask"
-                )));
-            }
+            predicate.validate_params(index).map_err(|e| ExtensionError(e.to_string()))?;
         }
         Ok(tx.with_validity_predicates(self.validity))
     }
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/txpool/src/validity.rs
```diff
@@ -9,6 +9,31 @@ use crate::{BasePooledTransaction, ExtensionError, ValidatedTransactionExtension
...
+        count: usize,
+        /// Maximum number of predicates permitted.
+        max: usize,
+    },
+    /// A storage predicate's comparison value has bits set outside its mask.
+    ///
+    /// Because the loaded storage value is masked before comparison, bits set in
+    /// `value` outside `mask` could never match and indicate a malformed request.
+    #[error("storage predicate value has bits set outside its mask")]
+    StorageValueOutsideMask,
```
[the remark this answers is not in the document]
refcell: Done in 388007a98. `StorageValueOutsideMask` now carries an `index: usize` and its message reads "storage predicate at index {index} has value bits set outside its mask". `validate_batch` reports the failing predicate's position via `.enumerate()`, and the builder-path `apply` formats the same index into its error.

crates/execution/txpool/src/validity.rs
```diff
@@ -89,6 +114,39 @@ impl ValidityPredicate {
...
+            return Err(ValidityPredicateError::TooMany {
+                count: predicates.len(),
+                max: MAX_VALIDITY_PREDICATES,
+            });
+        }
+        for predicate in predicates {
+            predicate.validate()?;
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
refcell: Good catch — fixed in 388007a98. `TransactionValidity::apply` (builder RPC path) now validates each predicate's params via `ValidityPredicate::has_valid_params` as defense-in-depth, in addition to the count bound. I intentionally did *not* add the non-empty check there: the builder legitimately receives ordinary transactions carrying no predicates, so empty-rejection stays specific to the `base_sendRawTransactionValidity` mempool ingress. Added a doc comment on `apply` spelling out that distinction, plus `apply_accepts_empty_predicates` and `apply_rejects_storage_value_outside_mask` tests.

crates/execution/txpool/src/validity.rs
```diff
@@ -89,6 +118,37 @@ impl ValidityPredicate {
...
+            return Err(ValidityPredicateError::Empty);
+        }
+        if predicates.len() > MAX_VALIDITY_PREDICATES {
+            return Err(ValidityPredicateError::TooMany {
+                count: predicates.len(),
+                max: MAX_VALIDITY_PREDICATES,
+            });
+        }
+        for (index, predicate) in predicates.iter().enumerate() {
+            if !predicate.has_valid_params() {
```
BrianBland: This couples this specific returned error with any not `has_valid_params()`, which could drift as implementations change.

This should probably instead call something like `predicate.validate_params() -> Result`, which would return `StorageValueOutsideMask` as applicable

crates/execution/txpool/src/validity.rs
```diff
@@ -128,13 +188,29 @@ impl ValidatedTransactionExtensions<BasePooledTransaction> for TransactionValidi
...
+    /// endpoint on mempool nodes.
     fn apply(self, tx: BasePooledTransaction) -> Result<BasePooledTransaction, ExtensionError> {
         if self.validity.len() > MAX_VALIDITY_PREDICATES {
             return Err(ExtensionError(format!(
                 "too many validity predicates: {} (maximum {MAX_VALIDITY_PREDICATES})",
                 self.validity.len()
             )));
         }
+        for (index, predicate) in self.validity.iter().enumerate() {
+            if !predicate.has_valid_params() {
```
BrianBland: Similar to the above comment, we can just stringify the returned error

crates/execution/txpool/src/validity.rs
refcell: Good call. Fixed in 0ed4a3473. Replaced the boolean `has_valid_params` with `validate_params(index) -> Result<(), ValidityPredicateError>`

crates/execution/txpool/src/validity.rs
refcell: Done in 0ed4a3473.

BrianBland (dismissed):

BrianBland (approved):

xenoliss (approved):

PR #4441, opened 2026-08-14
feat(builder): Expire Validity Txs By Position

refcell:
## Summary

Adds beta expiry for validity transactions based on build position instead of a timestamp. Because beta rides EIP-1559 there is no signed EIP-8130 expiry field, so a transaction expires once the build has monotonically passed the point where its block_number / flashblock_index predicates could ever be satisfied again. The builder now treats such position predicates as terminal (evicted) rather than parking and rescanning them, while balance and storage predicates remain recoverable and continue to park. A new ValidityPredicate::is_batch_expired helper computes terminality conservatively from upper-bound comparisons, and the builder emits a distinct validity_predicate_expired rejection reason.

Stacked on #4432 (block_number / flashblock_index predicates). Implements BASE-280 (child of BASE-258).

### DIFF
--- a/crates/builder/core/src/flashblocks/context.rs
+++ b/crates/builder/core/src/flashblocks/context.rs
@@ -855,6 +855,14 @@ impl BasePayloadBuilderCtx {
                 // position predicate is terminal too — no later position can satisfy it — so both
                 // are dropped rather than parked; only recoverable state mismatches are parked.
                 if predicate_read_failed || predicate_expired {
+                    // A passed position bound can never be satisfied in any later
+                    // block, so an expired predicate is permanently terminal:
+                    // record it for the rejection cache and pool eviction so it is
+                    // not re-evaluated on subsequent flashblock rebuilds. A read
+                    // failure is only terminal for this scan, so it is not cached.
+                    if predicate_expired {
+                        diag.permanently_rejected_txs.push(tx_hash);
+                    }
                     best_txs.mark_invalid(tx.sender(), tx.nonce());
                 } else if let Some(blocking_predicate) = blocking_predicate
                     && best_txs.park_current()
--- a/crates/builder/core/src/flashblocks/context.rs
+++ b/crates/builder/core/src/flashblocks/context.rs
@@ -23,7 +23,7 @@ use base_execution_payload_builder::{
 };
 use base_execution_txpool::{
     BasePooledTx, BundleTransaction, GuardMetrics, PredicateContext, TimestampedTransaction,
-    estimated_da_size::DataAvailabilitySized,
+    ValidityPredicate, estimated_da_size::DataAvailabilitySized,
 };
 use base_observability_events::TransactionEventType;
 use reth_basic_payload_builder::PayloadConfig;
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/builder/core/src/flashblocks/context.rs
```diff
@@ -792,11 +792,24 @@ impl BasePayloadBuilderCtx {
...
                 num_txs_considered += 1;
                 let ordering_position = num_txs_considered;
+                // A position predicate (block_number / flashblock_index) whose
+                // upper bound the build has passed can never be satisfied again,
+                // so the transaction is expired rather than merely unsatisfied.
+                let predicate_expired = !predicate_read_failed
+                    && ValidityPredicate::is_batch_expired(
+                        tx.validity_predicates(),
+                        &predicate_context,
+                    );
```
refcell, line 805: I suspect we can refactor this for better performance but will keep things simple for now

crates/execution/txpool/src/validity.rs
```diff
@@ -148,6 +148,75 @@ impl ValidityPredicate {
...
+        if block_impossible || block_upper.is_some_and(|max| max < current_block) {
+            return true;
+        }
+        if flashblock_impossible {
+            return true;
+        }
+        // The flashblock index resets each block, so a passed flashblock bound is
+        // terminal only when no later block is allowed either.
+        let pinned_to_current_block = block_upper == Some(current_block);
+        pinned_to_current_block && flashblock_upper.is_some_and(|max| max < current_flashblock)
```
BrianBland, line 272: I think this might be easier to read and reason about if we were to just `return true` as soon as possible. This negates the need for an `impossible` pointer, as `< 0` can just immediately `return true`

refcell: Yea agreed.

crates/builder/core/src/flashblocks/context.rs
```diff
@@ -835,8 +848,11 @@ impl BasePayloadBuilderCtx {
...
-                // its predicate may have already been satisfied at its first position.
-                let should_invalidate = predicate_read_failed || !best_txs.park_current();
+                // its predicate may have already been satisfied at its first position. An expired
+                // position predicate is terminal too — no later position can satisfy it — so both
+                // are dropped rather than parked; only recoverable state mismatches are parked.
+                let should_invalidate =
+                    predicate_read_failed || predicate_expired || !best_txs.park_current();
                 if should_invalidate {
                     best_txs.mark_invalid(tx.sender(), tx.nonce());
                 }
```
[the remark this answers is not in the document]
refcell: Good catch — done in 618aa2b3d. Position-expired predicates now push the tx hash into `diag.permanently_rejected_txs`, so the rejection cache (and pool eviction) skip them on subsequent flashblock rebuilds rather than re-running `is_batch_expired` each time. This is safe because a passed position bound is terminal for every later block, not just the current scan.

crates/execution/txpool/src/validity.rs
refcell: Done in 618aa2b3d. `is_batch_expired` now returns `true` immediately when a `< 0` upper bound is seen, dropping the `block_impossible`/`flashblock_impossible` flags and the `&mut` impossible pointer entirely.

crates/builder/core/src/flashblocks/context.rs
refcell: Keeping the structure simple as noted, but 618aa2b3d removes most of the redundant cost this was pointing at: position-expired txs are now cached as permanently rejected, so `is_batch_expired` isn't re-run for them on later flashblock rebuilds. Resolving.

BrianBland (dismissed):

xenoliss (approved):

PR #3850, opened 2026-07-06
refactor(challenger): clean up bond

jackchuma:
### What changed? Why?

Refactors `base-challenger` bond claiming from a tracked lifecycle state machine into a stateless recent-game scanner. Each discovery pass scans the configured lookback window from scratch, reads the current onchain state, and immediately submits the next ready maintenance transaction: `resolve()`, `claimCredit()` to unlock, or `claimCredit()` to withdraw.

Claimable game detection still matches configured claim addresses against `bondRecipient` and nonzero `zkProver` so unresolved challenges can be recovered. Before acting on resolved games, the scanner re-reads `bondRecipient` and skips games whose current recipient is not in the claim set.

Withdraw readiness now uses the actual `DelayedWETH.withdrawals(game, recipient).timestamp` plus the onchain `delay()` instead of estimating unlock time locally. The `DelayedWETH` client exposes the withdrawals lookup, and the bond manager caches the delay after the first successful read.

The driver no longer runs a startup recovery scan, maintains an in-memory tracked-bond map, or registers a bond after a successful challenge submission. Bond claiming is recovered entirely by the periodic lookback scan. Anchor and bond maintenance transactions now use `ChallengeSubmitter<T>` directly through `send_bond_tx`, which removes the separate `BondTransactionSubmitter` test trait and the `derive_more` dependency.

### Notes to reviewers

Most of the functional change is in `crates/proof/challenge/src/bond.rs`. The important behavior change is that progress is derived from onchain state every scan rather than from local phases, so restarts and missed ticks are handled by rescanning the lookback window.

`crates/proof/contracts/src/delayed_weth.rs` adds the `withdrawals(owner, guy)` binding used to decide whether a withdraw is ready. Metrics for tracked in-memory bonds and not-claimable removals were dropped because those states no longer exist.

Tests were updated to assert the scanner submits resolve/unlock/withdraw actions through the real `ChallengeSubmitter` path and waits on the `DelayedWETH` withdrawal timestamp before withdrawing.

### How has it been tested?

- `cargo test -p base-challenger`
- `cargo clippy -p base-challenger --all-targets`

PR #1894, opened 2026-04-03
feat(challenger): add bond lifecycle management to challenger

jackchuma:
## Summary

- Introduces a `BondManager` that tracks dispute games through a multi-phase credit claim lifecycle: **resolve → unlock → DelayedWETH delay → withdraw**.
- After a successful `challenge()` submission, the manager registers the game and automatically resolves it (once the dispute period elapses) and performs two `claimCredit()` calls (with the `DelayedWETH` withdrawal delay in between) to recover the bond. Games are tracked whenever the onchain `bondRecipient` matches a configured claim address, regardless of whether the game resolves as `CHALLENGER_WINS` or `DEFENDER_WINS`.
- Adds a startup scan that recovers in-flight games after restarts by checking both `bondRecipient` and `zkProver` against the configured `--bond-claim-addresses`.

## Changes

- **`crates/proof/challenge/src/bond.rs`** — new `BondManager` with `BondPhase` state machine, `TrackedGame` tracking, `startup_scan`, and `poll` loop.
- **`crates/proof/challenge/src/cli.rs`** / **`config.rs`** — new `--bond-claim-addresses` CLI flag and config field.
- **`crates/proof/challenge/src/driver.rs`** — integrates bond polling into the driver tick; registers games after successful challenges. Introduces `DriverComponents` to bundle injected dependencies.
- **`crates/proof/challenge/src/service.rs`** — constructs the `BondManager` when claim addresses are configured, runs the startup scan, and passes it into `DriverComponents`.
- **`crates/proof/challenge/src/submitter.rs`** — implements `BondTransactionSubmitter` for `ChallengeSubmitter<T>`.
- **`crates/proof/challenge/src/metrics.rs`** — adds bond-related metrics (resolve outcomes, claimCredit outcomes, latency, tracked count, completed count).
- **`crates/proof/challenge/src/scanner.rs`** — exposes `STATUS_CHALLENGER_WINS` and `STATUS_DEFENDER_WINS` constants.
- **`crates/proof/contracts/src/delayed_weth.rs`** — new `DelayedWETHClient` / `DelayedWETHContractClient` for reading the withdrawal delay.
[...]

PR #4445, opened 2026-08-14
fix(challenger): resolve dispute games in ascending index order

robriks:
One-word fix plus a regression test. Port of [#4442](https://github.com/base/base/pull/4442) (opened against `l3` branch used for the EVM Ledger project, where this caused a failure) — the same line and the same constraint exist here.

## The bug

`bond_actions` collected game evaluations with `buffer_unordered`, so actions arrived in **completion** order, and `discover_claimable_games` then processed them serially in that arbitrary order. `resolve()` reverts `ParentGameNotResolved` unless the parent game has already resolved, so any child processed before its parent is a guaranteed revert.

The idiom was copied from `GameScanner` — the original commit ([#1894](https://github.com/base/base/pull/1894)) said so in a comment:

```rust
// Evaluate all games concurrently using the same pattern as the
// game scanner (`buffer_unordered`).
```

`GameScanner` is safe precisely because it tuples the index with each result (`|i| async move { (i, …) }`) and rebuilds order downstream in a `BTreeMap`. `bond_actions` dropped the index and consumes the actions in the order returned, so the property that made the pattern correct did not survive the copy. [#3850](https://github.com/base/base/pull/3850) later extracted the helper and carried the primitive across unchanged.

`buffered` keeps the identical concurrency and restores ascending order. `anchor.rs` in the same crate already uses `buffered` for the same reason.

## Impact here

Latent in steady state: a scan window normally holds at most one *unresolved* game, since resolved games yield claim actions rather than resolve actions.

It surfaces during **recovery**. After a challenger outage the rescan (lookback defaults to 1000 games) finds several unresolved games and attempts them in arbitrary order, so it advances roughly one game per scan instead of draining the backlog — slowest exactly when the system is already behind. The reverts come from `eth_estimateGas`, so the cost is latency and RPC churn rather than gas.

On the `l3` branch, where a much smaller dispute-game interval puts many resolvable games in every scan, the same line produced **525 `ParentGameNotResolved` reverts against 45 successful resolves** and stalled the resolve chain entirely.

## Test plan

- [x] `bond_actions_are_ordered_by_game_index` — a factory whose `game_at_index` completes in reverse index order. **Verified it fails on the pre-change code**, returning `[4,3,2,1,0]`, so it reproduces the defect rather than asserting current behaviour.
- [x] `cargo test -p base-challenger --lib` passes.
- [x] `cargo clippy -p base-challenger --all-targets -- --deny=warnings` clean.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4457, opened 2026-08-16
fix(websocket-proxy): strip 0x prefix before substring search in address filter

### LINKED ISSUE #4369: bug(websocket-proxy): address filter uses two incompatible match semantics, making the diff.transactions branch unreachable
# bug(websocket-proxy): address filter uses two incompatible match semantics, making the `diff.transactions` branch unreachable

## Summary

`FilterType::contains_any_address` in `crates/infra/websocket-proxy/src/filter.rs` searches three places for a subscriber's address. Two of them require the address to be `0x`-prefixed; the third requires it to be bare hex. Since all three read the same `HashSet<String>`, no input format satisfies all of them.

In practice subscribers pass `0x`-prefixed addresses, so the `diff.transactions` branch can never match and the flashblock is silently dropped for that subscriber.

## The two semantics

`new_account_balances` and receipt logs use `HashSet::contains`, an **exact** match against `0x`-prefixed JSON values:

```rust
// keys look like "0x4200000000000000000000000000000000000010"
if addresses.contains(&account.to_lowercase()) {

// log `address` fields look like "0x4200000000000000000000000000000000000010"
if addresses.contains(&addr_str.to_lowercase())
```

`diff.transactions` uses `str::contains`, a **substring** search over a raw EIP-2718 transaction:

```rust
for tx in transactions {
    if let Some(tx_str) = tx.as_str() {
        let tx_lower = tx_str.to_lowercase();
        for address in addresses {
            if tx_lower.contains(address) {
                debug!(tx = %tx_str, "Found address in transaction");
                return true;
```

A raw transaction is a single hex string whose only `0x` is the leading one. An embedded address appears as bare hex, so `tx_lower.contains("0x4200…0015")` is false unless the address happens to occupy the first 20 bytes of the transaction body, which for a typed transaction never happens.

The `debug!("Found address in transaction")` line cannot be reached.

## Reproduction using the repo's own test fixture

`test_with_real_data` in the same file already contains a payload that demonstrates this. Its transaction hex includes:

```
… dead000194 4200000000000000000000000000000000000015 8080830f4240 …
```

`94` is the RLP prefix for a 20-byte string, so `0x4200000000000000000000000000000000000015` (the `L1Block` predeploy) is the transaction's `to` address. That address appears **only** inside `diff.transactions` — `new_account_balances` holds `0x336f…78db` and `0x4200…0007`, and the single log holds `0x4200…0010`.

Where `0x4200000000000000000000000000000000000015` occurs in that payload:

| Location | Present |
| --- | --- |
| `diff.transactions[0]` (raw tx hex, bare) | yes |
| `metadata.new_account_balances` keys | no |
| receipt log `address` fields | no |
| receipt log `topics` | no |
| `0x`-prefixed form, anywhere in the payload | no |

No assertion in `test_with_real_data` uses that address, so the dead branch is invisible to the existing suite. Adding a test that reuses the same payload literal (verified byte-identical, so the only difference from `test_with_real_data` is which address is filtered on):

```rust
#[test]
fn address_only_in_transaction_payload_matches() {
    let payload = /* exact payload literal from `test_with_real_data` */;

    // Sanity: the bare form is present, the 0x-prefixed form is not.
    let payload_str = std::str::from_utf8(&payload).unwrap();
    assert!(payload_str.contains("4200000000000000000000000000000000000015"));
    assert!(!payload_str.contains("0x4200000000000000000000000000000000000015"));

    let filter = FilterType::new_addresses(vec![
        "0x4200000000000000000000000000000000000015".to_string(),
    ]);
    assert!(
        filter.matches(&payload, false),
        "address appearing only inside diff.transactions must match an address filter"
    );
}
```

Both sanity assertions pass; the third fails. Verbatim output with the filter logic byte-for-byte as on `main`:

```
running 1 test
test filter::tests::address_only_in_transaction_payload_matches ... FAILED

failures:

---- filter::tests::address_only_in_transaction_payload_matches stdout ----
thread 'filter::tests::address_only_in_transaction_payload_matches' (3636229) panicked at crates/infra/websocket-proxy/src/filter.rs:505:9:
address appearing only inside diff.transactions must match an address filter
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

failures:
    filter::tests::address_only_in_transaction_payload_matches

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 23 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p websocket-proxy --lib`
```

## Reachability

- `crates/infra/websocket-proxy/src/server.rs:141` — `create_filter_from_query` takes the `?addresses=` query string verbatim; `parse_comma_separated` only trims, and `new_combined_with_mode` only lowercases. Nothing strips or adds a `0x` prefix.
- `crates/infra/websocket-proxy/src/registry.rs:67` — `filter.matches(msg_bytes, compressed)` gates every broadcast message per subscriber; `false` drops the flashblock for that client.
- The publisher sends `Message::Binary` (`bin/websocket-proxy/src/main.rs:234`), matching the `Message::Binary` arm at `registry.rs:64`, so the filter runs on real payloads.

## Impact

Silent, partial data loss for subscribers. A client subscribed with `?addresses=0xMyContract` misses every flashblock where its address appears only in transaction bytes — for example a call that reverts and emits no logs, or any transaction whose effects are not reflected in `new_account_balances`. The stream looks healthy; there is no error and no gap indicator.

The two working branches mask this, which is presumably why it went unnoticed: the common cases (balance change, emitted log) are covered, so the filter appears to work.

## Suggested resolution

Two reasonable options, and I do not want to presume which you prefer:

**Fix the branch** — strip the prefix before the substring search:

```rust
let needle = address.strip_prefix("0x").unwrap_or(address);
if tx_lower.contains(needle) {
```

This makes the branch live, but a bare 20-byte substring search has no field alignment: it can match a calldata argument, a topic word, or an unaligned byte window, not just `to` / `from`. It trades false negatives for false positives.

**Decode instead** — if the branch is meant to match `to` and `from` specifically, decoding the transaction is the only way to do that correctly. More work, but it is the only option that gives well-defined semantics.

**Or remove the branch** — if the intent is that address filtering covers only balances and logs, deleting it makes the actual semantics explicit and removes a misleading `debug!` line.

Only the remedy is debatable. The finding stands regardless: the branch costs a full scan of every transaction on every message and can never return `true`.

I would be glad to open a PR for whichever direction you choose, along with the regression test, if this is assigned to me.

## Notes

- This is not a regression. The code has been this way since the crate was merged (`e431c7656`, #726). The evidence of accident is the inconsistency inside the single function, not the history.
- There is no README or doc comment describing the filter's semantics, so I am citing the function's internal contradiction rather than an external contract.

## Environment

- `base` @ `main`
- rustc 1.97.1, macOS (aarch64)

memosr:
Closes #4369

## What changed? Why?

`FilterType::contains_any_address` searched raw transaction hex for a `0x`-prefixed address. A raw EIP-2718 transaction has only one `0x`, the leading one, so an embedded address appears as bare hex and the `diff.transactions` branch could never match. Subscribers filtering by address silently missed any flashblock where their address appeared only in transaction bytes.

This strips the prefix before the substring search, and adds a regression test reusing the existing `test_with_real_data` payload — `0x4200000000000000000000000000000000000015` is that deposit tx's `to` and appears nowhere else in the payload.

## Alternatives

#4369 lists two other directions: decoding the transaction, or removing the branch entirely. I took the minimal fix. Happy to switch to either if you'd prefer — a bare substring search has no field alignment, so this trades false negatives for possible false positives.

## How has it been tested?

- `cargo test -p websocket-proxy` — 24 passed, 0 failed
- `cargo fmt --check` and `cargo clippy` clean
- The new test fails on `main`, passes here

## Note

I'm not on `.github/VOUCHED.td`, so this may be auto-closed. Opening anyway since the work is done and the diff is small. If the process needs a vouch first, I'll close it and wait.

### DIFF
--- a/crates/infra/websocket-proxy/src/filter.rs
+++ b/crates/infra/websocket-proxy/src/filter.rs
@@ -195,7 +195,14 @@ impl FilterType {
                 if let Some(tx_str) = tx.as_str() {
                     let tx_lower = tx_str.to_lowercase();
                     for address in addresses {
-                        if tx_lower.contains(address) {
+                        // Subscribed addresses are `0x`-prefixed (the balances and
+                        // log-address branches above compare them against
+                        // `0x`-prefixed JSON fields), but a raw EIP-2718 transaction
+                        // is one hex string whose only `0x` is the leading one, so an
+                        // embedded address appears un-prefixed. Search for the bare
+                        // hex body.
+                        let needle = address.strip_prefix("0x").unwrap_or(address);
+                        if tx_lower.contains(needle) {
                             debug!(tx = %tx_str, "Found address in transaction");
                             return true;
                         }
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4458, opened 2026-08-16
fix(trusted-proxy): fall back to outermost trusted hop on unparseable entry

### LINKED ISSUE #4352: bug(trusted-proxy): forwarded_client_ip errors instead of returning the outermost trusted hop when an unparseable entry precedes it
# bug(trusted-proxy): `forwarded_client_ip` errors instead of returning the outermost trusted hop when an unparseable entry precedes it

## Summary

`TrustedProxyConfig::forwarded_client_ip` in `crates/utilities/trusted-proxy/src/trusted_proxy.rs` documents this contract:

> All header occurrences are concatenated in order into the forwarding chain `client, proxy1, proxy2, ...`, then scanned **right to left**, skipping any address in `trusted_proxy_cidrs`. [...] **If every hop is a trusted proxy, the outermost (leftmost) entry is used.**

The implementation does not honor the last sentence. While scanning right to left, an entry that fails to parse as an `IpAddr` returns `ForwardedClientIpError::InvalidIp` immediately, even when `outermost_trusted` has already been populated by trusted hops to its right and the result is therefore already determined.

## Affected code

`crates/utilities/trusted-proxy/src/trusted_proxy.rs`, `forwarded_client_ip`:

```rust
for entry in value.split(',').rev().map(str::trim).filter(|entry| !entry.is_empty()) {
    let ip = entry.parse::<IpAddr>().map(|ip| ip.to_canonical()).map_err(|_| {
        ForwardedClientIpError::InvalidIp {   // <-- returns even if outermost_trusted is Some
            header: header.clone(),
            value: entry.to_string(),
        }
    })?;
    if self.is_trusted_proxy(ip) {
        outermost_trusted = Some(ip);
        continue;
    }
    return Ok(ip);
}
```

## Reproduction

Add to `mod tests` in the same file:

```rust
#[test]
fn invalid_entry_left_of_trusted_hops_still_uses_outermost() {
    let config = xff_config(vec!["10.0.0.0/8"]);
    let outermost = IpAddr::V4(Ipv4Addr::new(10, 0, 0, 1));

    let mut headers = HeaderMap::new();
    headers.insert(
        "x-forwarded-for",
        HeaderValue::from_static("garbage, 10.0.0.1, 10.0.0.2"),
    );

    assert_eq!(config.forwarded_client_ip(&headers).unwrap(), outermost);
}
```

Run:

```
cargo test -p base-trusted-proxy invalid_entry_left_of
```

Result:

```
thread '...invalid_entry_left_of_trusted_hops_still_uses_outermost' panicked at
crates/utilities/trusted-proxy/src/trusted_proxy.rs:
called `Result::unwrap()` on an `Err` value: InvalidIp { header: "x-forwarded-for", value: "garbage" }
```

Both hops to the right of `garbage` are trusted, so per the documented contract the call should resolve to `10.0.0.1`.

## Why this looks unintended rather than deliberate strictness

The existing test `joins_header_lines_before_peeling_trusted_hops` already places an `invalid` entry in the chain and expects it to be tolerated, because right-to-left scanning resolves before reaching it:

```rust
headers.append("x-forwarded-for", HeaderValue::from_static("invalid"));
headers.append("x-forwarded-for", HeaderValue::from_static("203.0.113.10"));
headers.append("x-forwarded-for", HeaderValue::from_static("104.16.1.1"));
assert_eq!(config.client_ip(alb, &headers), client);
```

The all-trusted-hops case is the same situation (the answer is already determined before the invalid entry is reached), but it behaves differently. `all_trusted_hops_fall_back_to_outermost` covers the clean variant and passes; the variant with junk prepended is untested.

## Impact

`client_ip` swallows the error and falls back to the peer address:

```rust
Err(error) => {
    warn!(error = %error, peer = %connect_addr, "could not resolve forwarded client IP");
    connect_addr
}
```

For the two call sites, `connect_addr` is the fronting proxy:

- `crates/infra/telemetry/src/rate_limit.rs:145` — `PerIpRateLimit::enforce` keys the GCRA bucket on `client_ip`. Every request that hits this path collapses into a single bucket keyed by the proxy address, shared by all clients behind that proxy. With `DEFAULT_P2P_PROBE_REQUESTS_PER_MINUTE = 2`, that bucket is exhausted quickly and legitimate clients receive `429`.
- `crates/infra/websocket-proxy/src/server.rs:231` — connection accounting attributes the connection to the proxy rather than the client.

Reachability depends on deployment. It requires every hop to the right of the malformed entry to fall inside `trusted_proxy_cidrs`, which happens when the fronting proxy overwrites rather than appends `X-Forwarded-For`, or when clients originate from inside a trusted CIDR (the repo's own tests use broad ranges such as `10.0.0.0/8` and `104.16.0.0/13`). It also produces a `warn!` line per request on a path that is at least partly attacker-influenced.

## Suggested fix

Fall back to the already-resolved outermost trusted hop when parsing fails, rather than aborting the scan:

```rust
let Ok(ip) = entry.parse::<IpAddr>().map(|ip| ip.to_canonical()) else {
    if let Some(trusted) = outermost_trusted {
        return Ok(trusted);
    }
    return Err(ForwardedClientIpError::InvalidIp {
        header: header.clone(),
        value: entry.to_string(),
    });
};
```

This keeps `InvalidIp` for the case the existing tests cover (no trusted hop resolved yet) and satisfies the documented all-trusted-hops behavior. If the intent is instead strict rejection, the doc comment on `forwarded_client_ip` should be corrected, since it currently promises the opposite.

I am happy to open a PR with the fix and a regression test if this is assigned to me.

## Environment

- `base` @ `main`
- rustc 1.97.1, macOS (aarch64)

memosr:
Closes #4352

## What changed? Why?

`forwarded_client_ip` documents that "if every hop is a trusted proxy, the outermost (leftmost) entry is used." While scanning right to left, an entry that fails to parse returned `InvalidIp` immediately, even when `outermost_trusted` had already been resolved and the answer was therefore already determined.

This falls back to the resolved outermost trusted hop on a parse failure, and keeps `InvalidIp` for the case the existing tests cover, where no trusted hop has been resolved yet.

Impact is in `crates/infra/telemetry/src/rate_limit.rs:145`: `PerIpRateLimit::enforce` keys the GCRA bucket on `client_ip`, and a resolution failure collapses every affected request into a single bucket keyed by the proxy address.

## Note on intent

If strict rejection is the intended behavior instead, the fix belongs in the doc comment rather than the code — happy to flip it that way. #4352 has the reasoning for why this reads as unintended: the existing `joins_header_lines_before_peeling_trusted_hops` test already treats an unreachable invalid entry as tolerable.

## How has it been tested?

- `cargo test -p base-trusted-proxy` — 11 passed, 0 failed, no existing test changed
- `cargo fmt` and `cargo clippy --all-targets -- -D warnings` clean
- The new regression test fails on `main`, passes here

### DIFF
--- a/crates/utilities/trusted-proxy/src/trusted_proxy.rs
+++ b/crates/utilities/trusted-proxy/src/trusted_proxy.rs
@@ -73,12 +73,15 @@ impl TrustedProxyConfig {
                 value: String::from_utf8_lossy(value.as_bytes()).into_owned(),
             })?;
             for entry in value.split(',').rev().map(str::trim).filter(|entry| !entry.is_empty()) {
-                let ip = entry.parse::<IpAddr>().map(|ip| ip.to_canonical()).map_err(|_| {
-                    ForwardedClientIpError::InvalidIp {
+                let Ok(ip) = entry.parse::<IpAddr>().map(|ip| ip.to_canonical()) else {
+                    if let Some(trusted) = outermost_trusted {
+                        return Ok(trusted);
+                    }
+                    return Err(ForwardedClientIpError::InvalidIp {
                         header: header.clone(),
                         value: entry.to_string(),
-                    }
-                })?;
+                    });
+                };
                 if self.is_trusted_proxy(ip) {
                     outermost_trusted = Some(ip);
                     continue;
@@ -313,5 +316,16 @@ mod tests {
 
         headers.insert("x-forwarded-for", HeaderValue::from_static("::ffff:203.0.113.10"));
         assert_eq!(config.client_ip(mapped_proxy, &headers), client);
+    }
+
+    #[test]
+    fn invalid_entry_left_of_trusted_hops_still_uses_outermost() {
+        let config = xff_config(vec!["10.0.0.0/8"]);
+        let outermost = IpAddr::V4(Ipv4Addr::new(10, 0, 0, 1));
+
+        let mut headers = HeaderMap::new();
+        headers.insert("x-forwarded-for", HeaderValue::from_static("garbage, 10.0.0.1, 10.0.0.2"));
+
+        assert_eq!(config.forwarded_client_ip(&headers).unwrap(), outermost);
     }
 }


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4460, opened 2026-08-17
feat(txpool): Evict Expired Validity Txs From Pool

refcell:
## Summary

Adds the pool-side, cross-block half of beta validity-transaction expiry. The builder half (#4441) skips expired transactions during a build; this evicts them from the mempool node's txpool so they cannot live in the pool forever. Because beta advanced transactions are EIP-1559 rather than EIP-8130, they never enter the existing EIP-8130 admission guard, so this introduces a dedicated, type-agnostic BlockExpiryIndex keyed by each transaction's last valid block number, populated at admission and swept from on_canonical_state_change using the newly committed block number. Pool eviction is block-granular by design; the finer flashblock deadline remains enforced by the builder.

Stacked on #4441. Implements the second half of BASE-280 (child of BASE-258).

### DIFF
--- a/crates/execution/txpool/src/pool.rs
+++ b/crates/execution/txpool/src/pool.rs
@@ -350,10 +350,24 @@ where
         )
     }
 
-    /// Records a transaction's last valid block in the block-expiry index.
-    fn register_block_expiry(&self, hash: TxHash, last_valid_block: Option<u64>) {
+    /// Records a transaction's last valid block in the block-expiry index,
+    /// dropping any replaced transaction's stale entry in the same pass so a
+    /// fee-bump replacement does not orphan the old hash until its expiry block.
+    fn register_block_expiry(
+        &self,
+        hash: TxHash,
+        last_valid_block: Option<u64>,
+        replaced: Option<TxHash>,
+    ) {
+        if last_valid_block.is_none() && replaced.is_none() {
+            return;
+        }
+        let mut block_expiry = self.block_expiry.write();
+        if let Some(replaced) = replaced {
+            block_expiry.remove(&replaced);
+        }
         if let Some(last_valid_block) = last_valid_block {
-            self.block_expiry.write().insert(hash, last_valid_block);
+            block_expiry.insert(hash, last_valid_block);
         }
     }
 
@@ -511,7 +525,7 @@ where
             }
         };
         self.gate_protocol_admission(hash, replaced, pre_admitted)?;
-        self.register_block_expiry(hash, block_expiry_bound);
+        self.register_block_expiry(hash, block_expiry_bound, replaced);
         Ok(outcome)
     }
 
@@ -842,7 +856,7 @@ where
                     }
                 };
             self.gate_protocol_admission(hash, replaced, pre_admitted)?;
-            self.register_block_expiry(hash, block_expiry_bound);
+            self.register_block_expiry(hash, block_expiry_bound, replaced);
             return Ok(events);
         }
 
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/txpool/src/pool.rs
```diff
@@ -335,6 +341,45 @@ where
...
+        let removed = self.remove_dropped_across_pools(expired);
+        if !removed.is_empty() {
+            GuardMetrics::record_block_expiry_invalidations(removed.len());
+            debug!(
+                count = removed.len(),
+                block = block_number,
+                "validity transactions invalidated by block expiry"
+            );
+        }
+    }
```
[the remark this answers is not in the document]
refcell: Fixed in dcf1c4bcf. `expire_by_block` now calls `self.release_from_guard(&removed)` after pool removal, so the eviction path releases guard capacity directly instead of leaning on the `reconcile_guard` sweep — consistent with the other eviction paths. This matters for the (uncommon but valid) case of an EIP-8130 transaction that carries `block_number` validity predicates: such a tx is tracked by both the block-expiry index and the guard, so it needs an explicit release. `guard.release` is a no-op for the common EIP-1559 validity txs that were never admitted to the guard.

crates/execution/txpool/src/validity.rs
```diff
@@ -271,6 +271,41 @@ impl ValidityPredicate {
...
+        let mut upper: Option<U256> = None;
+        let mut impossible = false;
+        for predicate in predicates {
+            let Self::BlockNumber { op, value } = predicate else { continue };
+            let candidate = match op {
+                ValidityOperator::LessThan => match value.checked_sub(U256::from(1)) {
+                    Some(max) => max,
+                    None => {
+                        impossible = true;
+                        continue;
```
BrianBland: Let's just `return Some(0)` here instead of tracking `impossible`

crates/execution/txpool/src/pool.rs
```diff
@@ -335,6 +341,50 @@ where
...
+        let _admission_guard = self.protocol_admission_lock.lock();
+        let expired = self.block_expiry.write().drain_expired(block_number);
+        let removed = self.remove_dropped_across_pools(expired);
+        // Release any guard slots directly rather than deferring to the
+        // reconciliation sweep: an EIP-8130 transaction may carry block_number
+        // validity predicates, so an evicted tx can also hold guard capacity.
+        // `release` is a no-op for the common EIP-1559 case that is not tracked.
+        self.release_from_guard(&removed);
+        if !removed.is_empty() {
+            GuardMetrics::record_block_expiry_invalidations(removed.len());
```
BrianBland, line 393: Something I noticed while reviewing this is that `remove_dropped_across_pools` allocates a vec of all of the removed hashes, but is only ever used for recording the `len` metrics. We might want to clean this up in a future pass

crates/execution/txpool/src/validity.rs
refcell: Yes good catch

crates/execution/txpool/src/pool.rs
```diff
@@ -459,6 +511,7 @@ where
             }
         };
         self.gate_protocol_admission(hash, replaced, pre_admitted)?;
+        self.register_block_expiry(hash, block_expiry_bound);
```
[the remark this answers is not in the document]
refcell: Fixed in 1e0be4311. `register_block_expiry` now takes the `replaced` hash and removes its stale entry in the same locked pass before inserting the new bound, at both call sites (`add_transaction` and `add_transaction_and_subscribe`). This keeps the fee-bump path from orphaning the old hash until its expiry block, so `by_hash` no longer grows one entry per replacement. Added two unit tests covering replacement with and without a new bound.

BrianBland (approved):

PR #4390, opened 2026-08-12
chore: reth 2.5.0 upgrade

meyer9:

PR #4506, opened 2026-08-18
fix(tx-manager): clear the mempool replacement threshold strictly

robriks:
## Summary

`FeeCalculator::calc_threshold_value` returned `x + x / divisor` as the bumped fee. That value is not *above* the mempool's replacement threshold — it **is** the threshold. For any `x`:

```
floor(x * 110 / 100) == x + floor(x / 10)
```

Mempools require a replacement to be **strictly** greater than that, so every fee bump landed exactly one wei short and was rejected as `replacement transaction underpriced`.

A failed bump does not advance `BumpState`, so the next attempt recomputed the identical value from the identical baseline and was rejected identically. The bump could never succeed; the loop only ended when the original transaction was finally mined by other means.

## Observed in production (Base Sepolia, batcher)

10,704 rejections across two ~2-minute bursts. A 30-second slice:

```
WARN publish failed, will retry            error=replacement transaction underpriced
WARN fee bump failed, will retry next tick error=replacement transaction underpriced
INFO gas price increase computed  old_fee_cap=15297156  bumped_fee_cap=16826871
WARN publish failed, will retry            error=replacement transaction underpriced
INFO gas price increase computed  old_fee_cap=15297156  bumped_fee_cap=16826871
```

253 `publish failed` + 254 byte-identical `gas price increase computed` in 30 s. And `16826871` is exactly `floor(15297156 * 110 / 100)`. The tip legs matched too: `1000000 -> 1100000` is exactly `floor(1000000 * 110 / 100)`. Both legs sat precisely on the boundary, which is why the rejection was deterministic rather than intermittent.

Each burst stalled batch submission for ~56 s against a 20 s cadence, so this has a real latency cost and is not only log noise.

## The fix

**1. Add one wei** so the bump clears the threshold instead of sitting on it. This also subsumes the old "at least `x + 1`" special case for small `x` where `x / 10` truncated to zero, so that branch goes away.

**2. Stop a failed bump from re-firing immediately.** `publish_tx` routes its error through `process_send_error`, which re-sets the `bump_fees` flag that the send loop had just taken — so the loop re-entered the bump path on the very next iteration instead of on the resubmission timer, contradicting its own `will retry next tick` log line and driving retries at ~8/s. The loop now takes the flag again after each attempt.

This second part is defence in depth: with the threshold fixed, our own arithmetic no longer produces a rejection, but a competing sender on the same nonce, or a node configured with a larger price bump, still can — and that should cost one attempt per resubmission interval, not a hot loop.

## Blast radius

`calc_threshold_value` is shared by the batcher, proposer, challenger and prover-registrar. The change makes every fee bump one wei more expensive and makes bumps actually land. Runaway escalation is still bounded by the existing `check_fee_limits` / `fee_limit_multiplier` path, which is untouched.

## Test plan

- [x] `cargo test -p base-tx-manager` — 376 pass. 15 expectations changed, every one of which encoded the old boundary value (`calc_threshold_value` cases plus the `update_fees` cases that derive from them).
- [x] New `calc_threshold_value_strictly_clears_mempool_threshold` asserts the property directly against `floor(x * (100 + bump) / 100)`, using the two fee values observed in the incident. It fails on the previous implementation.
- [x] `cargo clippy -p base-tx-manager --all-targets` and `cargo fmt --check` clean.
- [x] `cargo check -p base-batcher-service -p base-proposer -p base-challenger` — consumers unaffected.
- [ ] **Not covered by an automated test:** the loop-pacing change in part 2. Asserting it needs a full mocked send-loop RPC script under the deterministic runtime, which is brittle relative to the value; the failure mode is noisy retries rather than incorrectness. Called out here rather than papered over.

## Follow-up

Needs a port to `main`, same as base#4442 -> base#4445.

### DIFF
--- a/crates/utilities/tx-manager/src/fees.rs
+++ b/crates/utilities/tx-manager/src/fees.rs
@@ -51,17 +51,22 @@ impl FeeCalculator {
         blob_base_fee.saturating_mul(2)
     }
 
-    /// Returns the minimum replacement value that satisfies geth's
-    /// tx-replacement rules.
+    /// Returns the minimum replacement value that clears a mempool's
+    /// tx-replacement rule.
+    ///
+    /// A mempool rejects a replacement whose fee is not **strictly** above
+    /// `x * (100 + bump_percent) / 100`. Because that division truncates, the
+    /// threshold it compares against is exactly `x + x / divisor` — so
+    /// returning that value lands *on* the boundary and is rejected. One extra
+    /// wei clears it.
     ///
     /// * **Regular transactions** (`is_blob = false`): 10 % bump →
-    ///   `x + x / 10`, but always at least `x + 1` when `x > 0`.
-    /// * **Blob transactions** (`is_blob = true`): 100 % bump →
-    ///   `x + x` (i.e. `2 × x`), but always at least `x + 1` when `x > 0`.
+    ///   `x + x / 10 + 1`.
+    /// * **Blob transactions** (`is_blob = true`): 100 % bump → `2x + 1`.
     /// * When `x == 0`, returns `0` (there is no fee to bump).
     ///
-    /// The "at least +1" rule prevents stuck bump loops where
-    /// `x / 10 == 0` for small values.
+    /// The trailing `+ 1` also subsumes the old "at least `x + 1`" rule for
+    /// small `x` where `x / 10` truncates to zero.
     ///
     /// Uses saturating arithmetic — the result is capped at [`u128::MAX`]
     /// rather than panicking on overflow.
@@ -71,16 +76,11 @@ impl FeeCalculator {
             return 0;
         }
 
-        let bump = if is_blob {
-            // 100 % bump: x + x = 2x
-            x
-        } else {
-            // 10 % bump: x / 10, but at least 1
-            let v = x / 10;
-            if v > 0 { v } else { 1 }
-        };
+        // 100 % bump for blobs, 10 % otherwise. `x + x / divisor` is exactly
+        // `floor(x * (100 + bump_percent) / 100)`, i.e. the mempool's threshold.
+        let bump = if is_blob { x } else { x / 10 };
 
-        x.saturating_add(bump)
+        x.saturating_add(bump).saturating_add(1)
     }
 
     /// Selects final `(tip, fee_cap)` values that satisfy geth's replacement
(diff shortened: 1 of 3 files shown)


robriks:
Correcting the framing in my own description, since it affects how you'd prioritise this.

The PR says the submission-cadence stalls mean "this has a real latency cost and is not only log noise". **That is wrong and I'd like to withdraw it.** Base Sepolia turned out to be in an active incident during the observation window (storage-v2 `write_storage_history` doing one sequential RocksDB last-shard `get` per unique `(address, slot)`, with `mineFlip()` bursts pushing busy batches to 18k–74k unique slots; fixed by reth#25282 in 2.4.0/2.5.0, carried by #4390, not yet tagged).

So the causality runs the other way. L1 stalled → the transaction sat unmined past the 48 s resubmission timeout → that **both** fired the bump timer into the boundary bug **and** blocked following submissions behind the same nonce. The retry loop is a parallel symptom of the stall, not its cause. And a *successful* bump would not have helped, because the bottleneck was block persistence rather than fee priority — a higher-fee replacement lands in the same stalled mempool.

**The case for merging is unchanged but rests on correctness, not latency:**

1. `calc_threshold_value` returns exactly the mempool boundary for *every* input, so **every fee bump on every service has always been rejected** — batcher, proposer, challenger, prover-registrar. This is a stuck-transaction recovery mechanism that has never once succeeded. It is harmless on a chain with no fee competition, and stops being harmless the first time a transaction is genuinely underpriced, at which point there is no recovery path and the send fails at `confirmation_timeout`.
2. During an L1 incident the loop fires ~8 write attempts/second — roughly a thousand extra `eth_sendRawTransaction` calls per 2-minute burst — at an endpoint that is already failing. I can't demonstrate it worsened the incident, but it is poor behaviour under exactly the conditions where restraint matters.
3. It removes a false diagnostic signal that masks real problems.

No urgency, and please don't merge it expecting a latency improvement.

robriks:
Superseded by base#4529, which targets `main`.

The bug is identical on `main` and matters more there: on a testnet with no fee competition a broken fee bump is invisible because transactions get mined at the minimum tip anyway, but on Ethereum the bump is the load-bearing recovery path when gas or blob fees spike. `main` also has more consumers of the crate (`proof/submission`, `zk-fork-dispute`, `basectl`, the TEE registrar) than `l3` does.

Closing this rather than carrying two PRs: `l3` periodically merges `main`, so it will pick the fix up, and on the L3 devnet the bug is harmless — it never needs a fee bump.

PR #4513, opened 2026-08-19
fix(upgrade-signal): Fail Closed On Unsupportable Live Upgrades

refcell:
## Summary

The live upgrade-signal poller used to retry a failed apply and then discard it, leaving an outdated node running on its old config so it would silently fork off the network at activation. This makes the handling fail-closed instead: a node that cannot support an upgrade nearing activation halts with a non-zero exit rather than diverge, while a malformed L1 signal only alarms loudly (halting the fleet on a governance typo would be a self-inflicted outage). The halt window is derived from the L1 poll cadence rather than a fixed 24 hours, so a node stops just before activation without front-running the outage or blocking a restart on a distant upgrade, and the startup and runtime paths across execution and consensus now behave consistently.

### DIFF
--- a/crates/execution/cli/src/upgrade_signal.rs
+++ b/crates/execution/cli/src/upgrade_signal.rs
@@ -3,9 +3,9 @@
 use base_execution_chainspec::BaseChainSpec;
 use base_node_runner::{BaseNodeExtension, BaseRpcContext, FromExtensionConfig, NodeHooks};
 use base_upgrade_signal::{
-    UpgradeSignalApplySummary, UpgradeSignalConfig, UpgradeSignalDefaults,
-    UpgradeSignalMetricLayer, UpgradeSignalMonitor, UpgradeSignalRefresher,
-    UpgradeSignalRuntimeApplier, UpgradeSignalSchedule,
+    PackedProtocolVersion, UpgradeSignalApplySummary, UpgradeSignalConfig, UpgradeSignalDefaults,
+    UpgradeSignalMetricLayer, UpgradeSignalMetrics, UpgradeSignalMonitor, UpgradeSignalPollOutcome,
+    UpgradeSignalRefresher, UpgradeSignalRuntimeApplier, UpgradeSignalSchedule,
 };
 use jsonrpsee::{RpcModule, core::RpcResult, types::ErrorObject};
 use reth_chainspec::EthChainSpec;
@@ -33,15 +33,18 @@ impl ExecutionUpgradeSignal {
         chain_spec: &mut BaseChainSpec,
     ) -> eyre::Result<()> {
         let reader = config.signal_config.reader(config.l1_rpc.clone())?;
-        let schedule = config
+        let Some(schedule) = config
             .signal_config
-            .read_required_startup_schedule(
+            .read_startup_schedule(
                 &reader,
                 "execution startup",
                 &[UpgradeSignalMetricLayer::Execution],
                 UpgradeSignalDefaults::STARTUP_SCHEDULE_RETRY_INTERVAL,
             )
-            .await?;
+            .await?
+        else {
+            return Ok(());
+        };
 
         Self::apply_schedule_to_chain_spec(chain_spec, &schedule)?;
 
@@ -66,14 +69,21 @@ impl ExecutionUpgradeSignal {
         refresher: &UpgradeSignalRefresher,
     ) -> RpcResult<UpgradeSignalApplySummary> {
         match refresher.read_schedule().await {
-            Ok(schedule) => refresher.apply(&schedule).map_err(|error| {
-                warn!(
-                    target: "upgrade_signal",
-                    error = %error,
-                    "failed to validate execution runtime upgrade signal"
-                );
-                ErrorObject::owned(-32005, "failed to validate upgrade signal", None::<()>)
-            }),
+            Ok(schedule) => match refresher.apply(&schedule) {
+                Ok(summary) => {
+                    UpgradeSignalMetrics::record_apply_success(refresher.metrics_layer, &schedule);
+                    Ok(summary)
+                }
+                Err(error) => {
+                    UpgradeSignalMetrics::record_apply_failure(refresher.metrics_layer, &schedule);
+                    warn!(
+                        target: "upgrade_signal",
+                        error = %error,
+                        "failed to validate execution runtime upgrade signal"
+                    );
+                    Err(ErrorObject::owned(-32005, "failed to validate upgrade signal", None::<()>))
+                }
+            },
             Err(error) => {
                 warn!(
                     target: "upgrade_signal",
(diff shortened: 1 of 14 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -65,6 +65,7 @@ impl ExecutionUpgradeSignal {
     ) -> RpcResult<UpgradeSignalApplySummary> {
         match refresher.read_schedule().await {
             Ok(schedule) => refresher.apply(&schedule).map_err(|error| {
+                UpgradeSignalMetrics::record_apply_failure(refresher.metrics_layer, &schedule);
```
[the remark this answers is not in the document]
refcell: Good catch — fixed. The RPC path now records `record_apply_success` on the `Ok` branch (mirroring the poller in `state.rs`), so a manual "fail → succeed" sequence via `admin_refreshUpgradeSignal` clears the sticky `apply_failed` gauge instead of relying on the live poller or a restart. The consensus RPC path (`admin.rs`) records neither success nor failure and is unaffected.

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -204,6 +241,91 @@ impl UpgradeSignalConfig {
...
+    /// This mirrors the runtime poller so a restart is not blocked by an upgrade that is still far
+    /// off: see [`Self::evaluate_startup_schedule`] for the policy. Returns `None` when the schedule
+    /// must be skipped (started with an alarm), or an error that must abort startup.
+    pub async fn read_startup_schedule(
+        &self,
+        reader: &AlloyUpgradeSignalReader,
+        log_context: &'static str,
+        metrics_layers: &[UpgradeSignalMetricLayer],
+    ) -> Result<Option<UpgradeSignalSchedule>, UpgradeSignalError> {
+        let schedule = self.read_schedule(reader, log_context, metrics_layers).await?;
```
jackchuma: Route consensus startup through the fail-closed startup policy. For `RuntimeAdmin`, a far-future protocol version the binary cannot support is meant to return `Skip` so the live poller can continue monitoring until the halt window. However, production `ConsensusNodeCommand::apply_initial_upgrade_signal` still calls `read_validated_schedule` (`crates/consensus/cli/src/node.rs:533`), and the normal in-process consensus startup does the same. Those consensus paths therefore abort immediately as soon as such a schedule is published, rather than applying the new 24-hour fail-closed policy; migrate these callers to `read_startup_schedule` and honor its `None` result

etc/systems/src/l2/in_process_follow_consensus.rs
```diff
@@ -197,17 +199,19 @@ impl InProcessFollowConsensus {
...
-                    warn!(
+                    error!(
                         target: "upgrade_signal",
-                        error = %error,
-                        "failed to auto-apply follow-mode upgrade signal update"
+                        upgrade = %upgrade_id.contract_id(),
+                        activation_timestamp,
+                        "in-process follow node halting upgrade signal monitor (fail closed)"
                     );
+                    break;
```
jackchuma, line 225: Stop the follow node on a fail-closed outcome. When the in-process follow node observes an imminent unsupported upgrade, this branch only exits the monitor task. The `FollowNode` task stored in `self.handle` keeps running and can continue following past activation, so this does not actually fail closed in the affected system-test configuration. Propagate the failure to the node task or cancel/abort it instead of just breaking the polling loop

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -121,6 +133,28 @@ impl UpgradeSignalConfig {
...
+    /// returned here: a zero version is trivially supported by the version check, so it is surfaced
+    /// as a non-fatal alarm elsewhere rather than halting the node on an L1 misconfiguration.
+    pub fn fail_closed_upgrade<'a>(
+        &self,
+        schedule: &'a UpgradeSignalSchedule,
+        now_secs: u64,
+        lead_secs: u64,
+    ) -> Option<&'a UpgradeSignal> {
+        schedule.signals.iter().find(|signal| {
+            self.validate_signal_supported_protocol_version(signal).is_err()
```
jackchuma, line 168: Does this start returning false as soon as the min version is updated onchain?

refcell: Yes it does. The intent is that the min version is only updated once that version has been rolled out to node operators which I think is a safe assumption

crates/utilities/upgrade-signal/src/config/schedule.rs
refcell: Great catch, I think you're right.

etc/systems/src/l2/in_process_follow_consensus.rs
refcell: Yep will fix this test.

crates/utilities/upgrade-signal/src/config/mod.rs
```diff
@@ -36,10 +37,35 @@ impl UpgradeSignalDefaults {
...
+    /// `safe`, and `finalized` block tags.
+    ///
+    /// The window is deliberately short. A large lead would *front-run* the outage: it halts every
+    /// outdated node long before activation, and — worse — before governance could still correct a
+    /// mistakenly scheduled upgrade by clearing or rescheduling the signal (which a running node
+    /// would pick up on its next poll and never halt). Early operator warning does not depend on
+    /// this window: the sticky `apply_failed` alarm is raised on *every* failed apply regardless of
+    /// how far off the activation is. The halt only needs enough margin to stop just ahead of
+    /// activation.
+    pub const HALT_LEAD_POLL_INTERVALS: u32 = 2;
```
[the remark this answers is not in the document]
refcell: The PR description has been updated — it no longer claims a fixed 24h. It now states the halt lead time is derived from the L1 poll cadence (`HALT_LEAD_POLL_INTERVALS × poll_interval`) rather than a fixed 24 hours, so the window tracks the configured block tag.

crates/utilities/upgrade-signal/src/state.rs
```diff
@@ -117,37 +185,180 @@ impl UpgradeSignalMonitor {
...
+    ) -> UpgradeSignalPollOutcome {
+        let apply_error = match refresher.apply(schedule) {
+            Ok(_) => {
+                self.mark_schedule_applied(schedule);
+                self.last_apply_failure = None;
+                UpgradeSignalMetrics::record_apply_success(self.metrics_layer, schedule);
+                return UpgradeSignalPollOutcome::Continue;
+            }
+            Err(apply_error) => apply_error,
+        };
```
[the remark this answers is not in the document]
refcell: Acknowledged — no change needed. The `UpgradeSignalApplySummary` (including `applied_upgrades`) is `.log()`'d inside `apply` itself, so it isn't silent; the caller has no decision that depends on the count today. Flagging for the future if the logging story moves out of `apply`.

crates/utilities/upgrade-signal/src/state.rs
```diff
@@ -72,41 +96,85 @@ impl UpgradeSignalStateUpdate {
...
+        /// Upgrade whose imminent activation forces the halt.
+        upgrade_id: BaseUpgrade,
+        /// L2 activation timestamp of that upgrade.
+        activation_timestamp: u64,
+        /// Minimum node protocol version the upgrade requires (packed).
+        minimum_protocol_version: U256,
+        /// Node protocol version this binary advertises (packed).
+        node_protocol_version: U256,
+    },
+}
```
jackchuma, line 165: `UpgradeSignalPollOutcome` is a custom two-state result that only exists to carry a halt condition already represented by `UpgradeSignalError::NodeUpgradeRequired`. The three consumers immediately turn `HaltNode` into an error, panic, or log-and-stop action. Returning `Result<(), UpgradeSignalError>` from `poll_and_apply` would use the standard type, remove the exported enum and repeated unpacking, and let each caller handle the same error directly

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -31,6 +31,16 @@ pub struct UpgradeSignalConfig {
...
 
+/// What a node should do with a startup schedule after the fail-closed policy has been applied.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum StartupScheduleAction {
+    /// Apply the schedule to the node's configuration.
+    Apply,
+    /// Do not apply the schedule; the node is starting with a loud alarm because it cannot apply an
+    /// upgrade that is still far from activation (a live poller will fail it closed later).
+    Skip,
+}
```
jackchuma, line 42: `StartupScheduleAction` is only used inside this crate, and its value is immediately converted into the `Option<UpgradeSignalSchedule>` that callers actually need. `evaluate_startup_schedule` can accept ownership of the schedule and return `Result<Option<UpgradeSignalSchedule>, UpgradeSignalError>` directly: `Some` applies it, `None` starts with the alarm. This removes an exported intermediary type without changing the policy

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -122,6 +134,42 @@ impl UpgradeSignalConfig {
...
+    pub fn fail_closed_upgrade<'a>(
+        &self,
+        schedule: &'a UpgradeSignalSchedule,
+        now_secs: u64,
+        lead_secs: u64,
+    ) -> Option<&'a UpgradeSignal> {
+        schedule.signals.iter().find(|signal| {
+            self.validate_signal_supported_protocol_version(signal).is_err()
+                && now_secs.saturating_add(lead_secs) >= signal.activation_timestamp
+        })
```
jackchuma, line 170: `lead_secs` is passed through `evaluate_startup_schedule` only to equal `self.halt_lead_time().as_secs()` at every production call site and in the current tests. Compute the lead time inside `fail_closed_upgrade` instead, keeping `now_secs` injectable for deterministic tests. This eliminates a policy parameter that callers should not be able to vary independently of the configured block tag

crates/utilities/upgrade-signal/src/state.rs
```diff
@@ -47,6 +50,27 @@ impl UpgradeSignalSchedule {
...
+            .map(|signal| {
+                format!(
+                    "{}={}",
+                    signal.upgrade_id.contract_id(),
+                    PackedProtocolVersion::new(signal.protocol_version)
+                )
+            })
+            .collect::<Vec<_>>()
+            .join(" ")
+    }
```
jackchuma, line 73: `required_protocol_versions` allocates a formatted string for every active signal solely for log fields. The existing `validation_error` already identifies the failing upgrade and includes both the required and supported protocol versions, so the extra full-schedule field duplicates operator-facing information. Remove this helper and log `error = %validation_error` alone

jackchuma (approved):

PR #4521, opened 2026-08-19
fix(proof): Propagate Real Boot Oracle Errors

refcell:
## Summary

`BootInfo::load` treated every error from the four optional preimage reads (proposer, intermediate block interval, L1 head number, and schedule L2 block number) as an absent key and substituted a default value. This meant operational failures such as timeouts, I/O errors, and a closed oracle were silently swallowed, letting the proof continue with unintended inputs and ultimately waste substantial work on a rejected result rather than surfacing the real failure.

This change adds `BootInfo::get_optional_local`, which defaults only on the explicit `PreimageOracleError::KeyNotFound` and propagates every other oracle error as `OracleProviderError::Preimage`. Tests cover that absent optional keys still default for backwards compatibility, and that a non-`KeyNotFound` failure on any of the four reads now aborts the load.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

jackchuma (commented):
Approach looks right to me and matches the finding's recommendation — defaulting only on the explicit absent-key signal and propagating real oracle failures. One substantive gap before this goes in.

**The ZK guest path doesn't report an absent key as `KeyNotFound`, so the backwards-compat default silently becomes a hard failure there.**

`PreimageStore` is the oracle the range program runs against inside the zkVM, and a map miss comes back as `InvalidPreimageKey`, not `KeyNotFound`:

```rust
// crates/proof/zk/utils/src/witness/preimage_store.rs:103
async fn get(&self, key: PreimageKey) -> PreimageOracleResult<Vec<u8>> {
    let Some(value) = self.preimage_map.get(&key) else {
        return Err(PreimageOracleError::InvalidPreimageKey);
    };
    Ok(value.clone())
}
```

That's reached via `run_range_program(oracle: Arc<PreimageStore>, ..)` -> `get_inputs_for_pipeline` -> `BootInfo::load` (`crates/proof/succinct/programs/range/utils/src/lib.rs:22-26`, `crates/proof/zk/utils/src/witness/executor.rs:40`). So after this change, a witness missing any of the four optional keys aborts the load instead of defaulting — exactly the tolerance the "optional — defaults for backwards compatibility" comments are there to provide.

Worth being precise about severity: `BootKeyValueStore` currently serves all four keys (`crates/proof/host/src/kv/boot.rs:43-51`), so a witness from a current host has them and nothing breaks today. The exposure is host/guest version skew and replaying older witnesses, which is the scenario those defaults exist for in the first place.

For the other two clients, the change is fine or an improvement:
- Nitro enclave (`crates/proof/tee/nitro-enclave/src/oracle.rs:112`) returns `KeyNotFound`, so it keeps defaulting correctly. This is the one path where the fallback genuinely works.
- Over the pipe, `OracleReader` never constructs `KeyNotFound` at all — a host-side miss kills the preimage server and the client surfaces `IOError(Closed)`. The old blanket `Err(_) => default` was papering over a dead host there, so propagating is strictly better.

I'd suggest fixing it in `PreimageStore` rather than widening the tolerated set in `boot.rs`: have `get` return `KeyNotFound` for a genuine map miss and reserve `InvalidPreimageKey` for `check_preimage`'s hash mismatch (`preimage_store.rs:89`), which is what that variant means everywhere else. Then `get_optional_local` stays a single, honest predicate and the ZK path keeps its backwards compatibility.

**Test coverage note:** `MockOracle::get` returns `KeyNotFound` on a miss, which matches the enclave oracle but not `PreimageStore`. That's why `defaults_optional_keys_when_absent` stays green while the ZK behavior changes underneath it. A case asserting the absent-key behavior of the ZK store (or a mock returning `InvalidPreimageKey`) would pin this down.

Smaller things, both fine as-is:
- Dropping `error = %e` from the three debug logs reads well now that only `KeyNotFound` can reach those branches — the error no longer carries information.
- The one red check (`Base Std Interface Tests (Cobalt)`) is a dependency resolution failure on `alloy-primitives`, unrelated to this change.

jackchuma (approved):

PR #4523, opened 2026-08-19
fix(upgrade-signal): Order Protocol Versions By Semver

refcell:
## Summary

The minimum protocol version gate compared packed protocol versions as raw U256 integers, which inverts prerelease ordering. A release candidate packs with a nonzero prerelease field in the low bits, so it compares greater than its final release, meaning a node running the final release could be rejected by a minimum that is only a release candidate of the same version. This change introduces a PackedProtocolVersion newtype that decodes the packed semver fields and orders by major.minor.patch with a prerelease sorting below its matching release, ignoring build and reserved bits. The compatibility check and packed_protocol_version helper now use it, and existing tests that relied on raw-integer ordering were corrected.

### DIFF
--- a/crates/utilities/upgrade-signal/src/config/mod.rs
+++ b/crates/utilities/upgrade-signal/src/config/mod.rs
@@ -4,6 +4,8 @@ use core::time::Duration;
 
 use alloy_primitives::U256;
 
+use crate::PackedProtocolVersion;
+
 mod args;
 pub use args::{UpgradeSignalArgs, UpgradeSignalL1RpcArgs, UpgradeSignalStartupConfig};
 
@@ -43,11 +45,13 @@ impl UpgradeSignalDefaults {
         ))
     }
 
-    /// Encodes a `major.minor.patch` version into the packed-semver `uint256` layout used by the
-    /// L1 `ProtocolVersions` contract: `major << 96 | minor << 64 | patch << 32`, with the
-    /// prerelease field left zero.
+    /// Encodes a final-release `major.minor.patch` version into the packed-semver `uint256` layout
+    /// used by the L1 `ProtocolVersions` contract, leaving the prerelease and build fields zero.
+    ///
+    /// See [`PackedProtocolVersion`] for the field layout and the ordering rules that govern
+    /// protocol-version compatibility checks.
     pub const fn packed_protocol_version(major: u32, minor: u32, patch: u32) -> U256 {
-        U256::from_limbs([(patch as u64) << 32, ((major as u64) << 32) | minor as u64, 0, 0])
+        PackedProtocolVersion::pack(major, minor, patch, 0).into_inner()
     }
 
     /// Maps a packed Cargo version to the advertised node protocol version, promoting the
--- a/crates/utilities/upgrade-signal/src/config/mod.rs
+++ b/crates/utilities/upgrade-signal/src/config/mod.rs
@@ -36,7 +36,8 @@ impl UpgradeSignalDefaults {
     ///
     /// Release branches sync the Cargo package version to the `GitHub` release tag, so release
     /// binaries advertise the release semver as their supported protocol version. Dev builds
-    /// (workspace `0.0.0`) advertise `U256::MAX` so no contract minimum rejects them.
+    /// (workspace `0.0.0`) advertise the maximum version-type-`0` version so no contract minimum
+    /// rejects them.
     pub fn node_protocol_version() -> U256 {
         Self::advertised_protocol_version(Self::packed_protocol_version(
             env!("CARGO_PKG_VERSION_MAJOR").parse::<u32>().expect("Cargo package major is numeric"),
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/utilities/upgrade-signal/src/version.rs
```diff
@@ -0,0 +1,143 @@
...
+/// * a pre-release (`pre-release != 0`) sorts *below* its matching release (`pre-release == 0`),
+///   and pre-releases of the same `major.minor.patch` sort by their pre-release counter;
+/// * `build` and the reserved/version-type high bytes are ignored.
+///
+/// A raw `U256` comparison instead sorts a pre-release *above* its release, because the pre-release
+/// field holds a larger integer than the release's zero. That inverts the intended order and can
+/// reject a node running a final release under a minimum that is a release candidate of the same
+/// version.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct PackedProtocolVersion(U256);
```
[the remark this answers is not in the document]
refcell: Good catch — this is a real `Ord`/`PartialEq` contract violation. Fixed in 525f26b62 by implementing `PartialEq` manually via `ordering_key()` (keeping `derive(Eq)`), so equality now ignores `build`/reserved exactly as `cmp` does and `a.cmp(&b) == Equal` implies `a == b`. I kept the raw value in `new()`/`into_inner()` intact rather than stripping bits, so no read-back value is silently mutated. Strengthened `build_field_does_not_affect_ordering` to also assert `plain == with_build` while their raw integers differ.

crates/utilities/upgrade-signal/src/config/mod.rs
jackchuma: This sentinel encodes "no contract minimum can reject a dev build", which depended on `U256::MAX` being the raw-integer maximum (the exact property this PR removes). `U256::MAX` decodes to `prerelease = u32::MAX`, so it never gets the `u64::MAX` promotion, and `pack(u32::MAX, u32::MAX, u32::MAX, 0)` now outranks it. Suggest `PackedProtocolVersion::pack(u32::MAX, u32::MAX, u32::MAX, 0).into_inner()` so it's the top element by construction, plus a test that the sentinel clears an arbitrary minimum. Note this also needs changing if you reject unknown version-types, since `U256::MAX` has version-type `0xFF`

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -49,8 +50,13 @@ impl UpgradeSignalConfig {
     }
 
     /// Returns true if this node supports the minimum protocol version attached to `signal`.
+    ///
+    /// Compatibility compares the packed versions by their OP-Stack semver ordering (see
```
jackchuma: Nit: should we avoid "OP-Stack" terminology?

crates/utilities/upgrade-signal/src/config/schedule.rs
jackchuma: Looks like this one was missed in the `+ U256::from(1)` cleanup. `config` here comes from `UpgradeSignalConfig::new(...)`, so `node_protocol_version` is the `U256::MAX` dev sentinel and `+ 1` wraps to `U256::ZERO`. The test passes only because the `activation_timestamp == 0` early return fires before the version check, so it would pass under any ordering. Suggest `supported_config()` with `packed_protocol_version(1, 1, 1)` to match the other two

crates/utilities/upgrade-signal/src/version.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// Ordering key that applies the semver pre-release rule.
+    ///
+    /// A final release (`prerelease == 0`) is promoted above every pre-release of the same
+    /// `major.minor.patch` by ranking it as [`u64::MAX`], which no 32-bit pre-release counter can
+    /// reach.
+    const fn ordering_key(self) -> (u32, u32, u32, u64) {
+        let prerelease_rank =
+            if self.prerelease() == 0 { u64::MAX } else { self.prerelease() as u64 };
+        (self.major(), self.minor(), self.patch(), prerelease_rank)
+    }
```
jackchuma: `ordering_key` drops the version-type byte, which flips this gate from fail-closed to fail-open. The spec says protocol versions with a different `version-type` should not be compared directly. Could we reject a non-zero version-type in `validate_signal_protocol_version`, or rank unknown types above everything so this stays fail-closed? The claim on L18 that the version-type bytes are ignored should change too; `build` is correctly ignored per spec, version-type isn't

crates/utilities/upgrade-signal/src/config/schedule.rs
refcell: yea... damn agent

jackchuma (commented):
Oh sorry one more thing before approving - I think the README is out of date now since it still references the old layout (`major << 96 | minor << 64 | patch << 32`)

jackchuma (approved):

PR #4526, opened 2026-08-19
fix(upgrade-signal): Fail Closed On Empty L1 Schedule Reads

refcell:
## Summary

A successful but empty `getSchedule()` read previously mapped to an empty schedule, which the manual admin refresh path would apply as authoritative, silently wiping every runtime upgrade override and returning success while the live poller ignored it. An empty read is now rejected with a typed `EmptySchedule` error instead of being applied, keeping empty-success distinct from a read failure.

Because a healthy append-only `ProtocolVersions` contract is never empty, an empty read means the node cannot yet see the authoritative activation schedule. The three read paths treat that distinctly:

- **Startup (`read_required_startup_schedule`)** — **fail closed**. Booting on the base configuration would risk activating forks at different times than peers reading a populated contract, forking the node — and any blocks it builds — off the network. Instead of proceeding, startup retries without an attempt limit, logging loudly at `error!` on every failure, until the contract returns a valid non-empty schedule. Only unrecoverable errors (malformed contract data, missing/unsupported protocol version) abort startup. The future stays cancellation-safe for shutdown.
- **Admin refresh** — surfaces the empty read as an error immediately (no retry), rather than clearing overrides.
- **Live poller** — logs at `debug` and skips the cycle; empty reads are not counted as `l1_read_errors_total`.

`EmptySchedule` is not part of the bounded transport-level retry predicate: at the finalized/safe tags the empty→populated transition lags by epochs, far past that budget, so retrying there only delayed startup and multiplied RPC. The unbounded startup wait is the mechanism that patiently blocks until L1 is authoritative.

### DIFF
--- a/crates/utilities/upgrade-signal/src/config/schedule.rs
+++ b/crates/utilities/upgrade-signal/src/config/schedule.rs
@@ -1,9 +1,8 @@
 use core::time::Duration;
 
 use alloy_primitives::{Address, U256};
-use backon::Retryable;
+use backon::{ConstantBuilder, Retryable};
 use base_common_genesis::UpgradeActivationSink;
-use base_retry::{DEFAULT_UNBOUNDED_INITIAL_DELAY, DEFAULT_UNBOUNDED_MAX_DELAY, RetryConfig};
 use tracing::{error, info};
 use url::Url;
 
@@ -147,6 +146,7 @@ impl UpgradeSignalConfig {
                 &reader,
                 log_context,
                 &[UpgradeSignalMetricLayer::Execution, UpgradeSignalMetricLayer::Consensus],
+                UpgradeSignalDefaults::STARTUP_SCHEDULE_RETRY_INTERVAL,
             )
             .await?;
 
@@ -223,18 +223,22 @@ impl UpgradeSignalConfig {
     ///
     /// This is the startup counterpart to the manual admin refresh, which instead surfaces an empty
     /// schedule as an error without retrying.
+    ///
+    /// `retry_interval` is the fixed delay between attempts; it is paced for legible retry logs
+    /// rather than fast recovery (see
+    /// [`UpgradeSignalDefaults::STARTUP_SCHEDULE_RETRY_INTERVAL`]).
     pub async fn read_required_startup_schedule(
         &self,
         reader: &AlloyUpgradeSignalReader,
         log_context: &'static str,
         metrics_layers: &[UpgradeSignalMetricLayer],
+        retry_interval: Duration,
     ) -> Result<UpgradeSignalSchedule, UpgradeSignalError> {
         let mut attempt = 1_u64;
-        let retry_config =
-            RetryConfig::unbounded(DEFAULT_UNBOUNDED_INITIAL_DELAY, DEFAULT_UNBOUNDED_MAX_DELAY);
+        let backoff = ConstantBuilder::default().with_delay(retry_interval).without_max_times();
 
         (|| self.read_validated_schedule(reader, log_context, metrics_layers))
-            .retry(retry_config.to_backoff_builder())
+            .retry(backoff)
             .when(|error| {
                 matches!(
                     error,
@@ -306,6 +310,7 @@ mod tests {
                 &reader,
                 "startup",
                 &[UpgradeSignalMetricLayer::Consensus],
+                Duration::from_millis(10),
             )
             .await
             .unwrap();
@@ -337,6 +342,7 @@ mod tests {
                 &reader,
                 "startup",
                 &[UpgradeSignalMetricLayer::Consensus],
+                Duration::from_millis(10),
             ),
             swap,
         );
@@ -360,6 +366,7 @@ mod tests {
                 &reader,
                 "startup",
                 &[UpgradeSignalMetricLayer::Consensus],
+                Duration::from_millis(10),
             )
             .await
             .unwrap_err();
--- a/crates/utilities/upgrade-signal/src/config/schedule.rs
+++ b/crates/utilities/upgrade-signal/src/config/schedule.rs
@@ -1,8 +1,10 @@
 use core::time::Duration;
 
 use alloy_primitives::{Address, U256};
+use backon::Retryable;
 use base_common_genesis::UpgradeActivationSink;
-use tracing::{info, warn};
+use base_retry::{DEFAULT_UNBOUNDED_INITIAL_DELAY, DEFAULT_UNBOUNDED_MAX_DELAY, RetryConfig};
+use tracing::{error, info};
 use url::Url;
 
 use super::{UpgradeSignalBlockTag, UpgradeSignalDefaults, UpgradeSignalMode};
@@ -140,16 +142,13 @@ impl UpgradeSignalConfig {
         CL::Error: std::error::Error + Send + Sync + 'static,
     {
         let reader = self.reader(l1_rpc)?;
-        let Some(schedule) = self
-            .read_optional_startup_schedule(
+        let schedule = self
+            .read_required_startup_schedule(
                 &reader,
                 log_context,
                 &[UpgradeSignalMetricLayer::Execution, UpgradeSignalMetricLayer::Consensus],
             )
-            .await?
-        else {
-            return Ok(());
-        };
+            .await?;
 
         UpgradeSignalRuntimeApplier::apply_schedule_to_sink(chain_id, &schedule, execution_sink)
             .map_err(eyre::Report::new)?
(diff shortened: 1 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -197,16 +200,51 @@ impl UpgradeSignalConfig {
...
+    /// admin refresh, which instead surfaces an empty schedule as an error.
+    pub async fn read_optional_startup_schedule(
+        &self,
+        reader: &AlloyUpgradeSignalReader,
+        log_context: &'static str,
+        metrics_layers: &[UpgradeSignalMetricLayer],
+    ) -> Result<Option<UpgradeSignalSchedule>, UpgradeSignalError> {
+        match self.read_validated_schedule(reader, log_context, metrics_layers).await {
+            Ok(schedule) => Ok(Some(schedule)),
+            Err(UpgradeSignalError::EmptySchedule) => {
```
jackchuma: Should this be gated on mode? `EmptySchedule`'s doc says an empty read means a "misconfigured, uninitialized, or mock contract" and here we respond to that by booting on the base configuration with only a warn. For a node that opted into `StartupApply` / `RuntimeAdmin`, L1 is authoritative for activation timestamps, so proceeding on genesis times risks activating forks at different times than peers that read a populated contract. This path also covers the builder, so a builder could produce blocks on a stale fork schedule. If tolerating is the deliberate availability tradeoff, can we say so explicitly in the doc comment?

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -219,6 +257,81 @@ mod tests {
...
+                    "id": 0,
+                    "result": minimum_protocol_version,
+                }));
+            })
+            .await;
+        server
+    }
+
+    #[tokio::test]
+    async fn startup_read_returns_present_schedule() {
```
jackchuma, line 303: The PR's headline behavior (`read_optional_startup_schedule` returning `Ok(None)` on empty) is untested. With the `schedule_server` helper right above, that's about five lines using an empty `uint64[]`. Also, the block + `getSchedule` + `minimumProtocolVersion` mock triple is now duplicated four times across this module and `contract.rs`; worth extracting

crates/utilities/upgrade-signal/src/contract.rs
```diff
@@ -321,6 +335,11 @@ impl AlloyUpgradeSignalReader {
     /// outright; after `max_attempts` failures the last error is returned (fail-fast). This future
     /// is cancellation-safe: dropping it during shutdown cancels an in-flight HTTP request or
     /// retry sleep.
+    ///
+    /// [`UpgradeSignalError::EmptySchedule`] is retried alongside provider errors: a freshly
+    /// deployed or just-initialized contract can briefly report an empty schedule at the finalized
+    /// tag, so a few retries let a real schedule settle before callers fall back. Decode and
+    /// protocol-version errors remain fail-fast.
```
jackchuma: This rationale doesn't hold at the default block tag. `READ_ATTEMPTS = 3` with 2s backoff capped at 10s is a ~6s total budget, but the finalized tag lags two epochs — `poll_interval()` returns 15 min for `Finalized` for exactly that reason. A just-initialized contract won't become non-empty within 6s at that tag, so this adds ~6s to startup and triples RPC calls without achieving the stated purpose. Genuinely transient flakes are already covered by the `Provider` arm. Either correct the rationale or drop `EmptySchedule` from the predicate

crates/utilities/upgrade-signal/src/contract.rs
```diff
@@ -335,7 +354,12 @@ impl AlloyUpgradeSignalReader {
 
         (|| self.read_schedule(metrics_layers))
             .retry(retry_config.to_backoff_builder())
-            .when(|error| matches!(error, UpgradeSignalError::Provider { .. }))
+            .when(|error| {
+                matches!(
+                    error,
+                    UpgradeSignalError::Provider { .. } | UpgradeSignalError::EmptySchedule
+                )
+            })
```
jackchuma: Worth noting this predicate is shared with the admin path: `UpgradeSignalRefresher::read_schedule` (`runtime/refresher.rs:48`) goes through `config::read_schedule` → `read_schedule_with_retries`, so `admin_refreshUpgradeSignal` against an empty contract now holds the RPC handler for the full retry budget before returning `-32003`. Intended?

crates/utilities/upgrade-signal/src/contract.rs
jackchuma, line 398: `EmptySchedule` now lands in this `Err` arm, so we log "failed to read" for a read that succeeded and returned nothing. The doc at line 381 ("Records `l1_read_errors_total` and returns `None` when the read fails") is also now inaccurate for the empty case, since it returns `None` without recording the metric. At the `Latest` tag the poll interval is 12s, so an unseeded devnet contract warns every 12 seconds. Suggest a separate branch with its own message at a lower level

crates/utilities/upgrade-signal/src/contract.rs
```diff
@@ -534,6 +559,120 @@ mod tests {
         mock.assert_calls_async(3).await;
     }
 
+    #[tokio::test]
+    async fn rejects_empty_schedule_read_without_recording_read_error() {
```
jackchuma: Nit: this only asserts `matches!(error, EmptySchedule)`; the "without recording read error" half of the name isn't verified. There's no metric-assertion infra in this crate, so probably just rename to `rejects_empty_schedule_read`

crates/utilities/upgrade-signal/src/config/schedule.rs
refcell: Done in 4a0576f32 — added `startup_read_tolerates_empty_schedule` covering the `Ok(None)` path via the existing helper, and extracted the shared block + `getSchedule` + `minimumProtocolVersion` mock triple into `test_utils::MockL1::schedule_server`, now reused across `schedule.rs` and `contract.rs`.

crates/utilities/upgrade-signal/src/contract.rs
refcell: Agreed — the rationale did not hold at the finalized/safe tags. Dropped `EmptySchedule` from the retry predicate in 4a0576f32 so it is surfaced immediately; genuine transient flakes remain covered by the `Provider` arm. Rewrote the doc to say so and renamed the test to `does_not_retry_empty_schedule_reads` (asserts a single call).

crates/utilities/upgrade-signal/src/contract.rs
refcell: Good catch — resolved by the same change: with `EmptySchedule` out of the retry predicate (4a0576f32), `admin_refreshUpgradeSignal` against an empty contract now returns `-32003` immediately instead of holding the RPC handler for the full retry budget.

crates/utilities/upgrade-signal/src/contract.rs
refcell: Fixed in 4a0576f32 — `read_schedule_tolerant` now has a dedicated `EmptySchedule` arm logged at `debug` ("empty schedule; skipping live apply") instead of the misleading "failed to read" warn, and the doc comment now states empty reads return `None` without recording `l1_read_errors_total`.

crates/utilities/upgrade-signal/src/contract.rs
refcell: Renamed to `rejects_empty_schedule_read` in 4a0576f32. There is no metric-assertion infra in the crate, so the "without recording read error" half was unverifiable.

crates/utilities/upgrade-signal/src/config/schedule.rs
```diff
@@ -226,6 +264,81 @@ mod tests {
...
+                &[UpgradeSignalMetricLayer::Consensus],
+            )
+            .await
+            .unwrap()
+            .expect("a non-empty schedule is present");
+
+        assert_eq!(schedule.signals.len(), 1);
+        assert_eq!(schedule.signals[0].upgrade_id, BaseUpgrade::Regolith);
+        assert_eq!(schedule.signals[0].activation_timestamp, 0);
+    }
```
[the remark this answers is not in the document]
refcell: Addressed in 4a0576f32 — added `startup_read_tolerates_empty_schedule` covering the `Ok(None)` path of `read_optional_startup_schedule`.

crates/utilities/upgrade-signal/src/config/schedule.rs
refcell: Decision: fail closed

crates/utilities/upgrade-signal/src/contract.rs
```diff
@@ -311,16 +320,28 @@ impl AlloyUpgradeSignalReader {
                 }
             };
 
-        Ok(Self::map_schedule(&timestamps, minimum_protocol_version, l1_block_number))
+        let schedule = Self::map_schedule(&timestamps, minimum_protocol_version, l1_block_number);
+        if schedule.signals.is_empty() {
+            return Err(UpgradeSignalError::EmptySchedule);
+        }
```
jackchuma, line 326: Wonder if we should add a metric here?

crates/utilities/upgrade-signal/src/contract.rs
```diff
@@ -290,6 +290,15 @@ impl AlloyUpgradeSignalReader {
...
     /// per-upgrade failures no longer exist.
+    ///
+    /// A successful read that yields no signals is rejected with
+    /// [`UpgradeSignalError::EmptySchedule`] rather than mapped to an empty schedule: applying an
+    /// empty schedule would replace every runtime override with base activations and report
+    /// success, silently diverging from the live poller, which ignores empty reads. An empty read
+    /// is not counted as an `l1_read_errors_total` failure, keeping empty success distinct from a
+    /// read failure. Callers decide how to treat it: startup retries then tolerates it (see
+    /// [`read_schedule_with_retries`](Self::read_schedule_with_retries)), while the manual admin
+    /// refresh surfaces it as an error instead of clearing overrides.
```
jackchuma, line 301: Nit: this comment is stale now

crates/utilities/upgrade-signal/src/contract.rs
```diff
@@ -311,16 +320,28 @@ impl AlloyUpgradeSignalReader {
...
     /// Used on the startup path, where a single transient L1 error should not abort node launch
     /// outright; after `max_attempts` failures the last error is returned (fail-fast). This future
     /// is cancellation-safe: dropping it during shutdown cancels an in-flight HTTP request or
     /// retry sleep.
+    ///
+    /// Only [`UpgradeSignalError::Provider`] errors are retried. An empty schedule is not a
+    /// transient flake at the finalized or safe tags: the tag lags the initializing transaction by
+    /// epochs, far longer than the retry budget, so retrying would only delay startup and multiply
+    /// RPC calls without changing the outcome. Empty reads are surfaced immediately and left for the
+    /// caller to tolerate (startup) or reject (admin refresh). Decode and protocol-version errors
```
jackchuma, line 343: Nit: stale comment

jackchuma (approved):

crates/utilities/upgrade-signal/src/contract.rs
refcell: Good call — added a metric in follow-up PR #4595 (31810b384). `read_schedule` now increments a new `empty_schedule_reads_total` counter (labelled by layer) on the empty branch, so the fail-closed condition is observable without being conflated with `l1_read_errors_total`.

crates/utilities/upgrade-signal/src/contract.rs
refcell: Fixed in #4595 (31810b384). After the fail-closed change, startup no longer tolerates empty reads — it retries indefinitely until the contract is populated. Rewrote this `read_schedule` doc to say startup fails closed (and that empty reads now increment `empty_schedule_reads_total` rather than `l1_read_errors_total`).

crates/utilities/upgrade-signal/src/contract.rs
refcell: Fixed in #4595 (31810b384). Updated the `read_schedule_with_retries` doc: empty reads are surfaced immediately, then the startup path retries them indefinitely to fail closed while the admin refresh rejects them — no longer 'tolerate (startup)'.

PR #4529, opened 2026-08-19
fix(tx-manager): clear the mempool replacement threshold strictly

robriks:
## Summary

`FeeCalculator::calc_threshold_value` returns `x + x / divisor` as the bumped fee. That is not *above* the mempool's replacement threshold — it **is** the threshold, because for every `x`:

```
floor(x * 110 / 100) == x + floor(x / 10)
```

Mempools require a replacement to be **strictly** greater, so the bump lands exactly one wei short and is rejected as `replacement transaction underpriced`. And because a failed bump does not advance `BumpState`, the next attempt recomputes the identical value from the identical baseline and is rejected identically — **the fee bump can never succeed.**

The blob path has the same shape: a 100 % bump gives `2x` against a `2x` threshold.

## Why this matters more here than on a testnet

Found on `l3` (base#4506) via a Base Sepolia devnet batcher, where 10,704 rejections appeared in two ~2-minute bursts, all with `old_fee_cap=15297156 -> bumped_fee_cap=16826871` — exactly `floor(15297156 * 110 / 100)` — and tips at `1000000 -> 1100000`, likewise exactly on the boundary. Both legs sitting on the line is why the rejection was deterministic rather than intermittent.

On that devnet the consequence was only noise, because there is no fee competition and transactions get mined at the minimum tip regardless — the bump has never been *needed*. On Ethereum that inverts: gas and blob fee markets spike, a transaction crafted at time T can be underpriced seconds later, and the fee bump is the load-bearing recovery mechanism. With it permanently broken, such a transaction can never be replaced — it sits until `confirmation_timeout` (default 300 s) and the send fails. For a batcher that is delayed DA; for a proposer or challenger it is a missed proposal or a missed dispute response.

`calc_threshold_value` is `pub`, and on this branch the crate is consumed by the batcher, proposer, challenger, `proof/submission`, `zk-fork-dispute`, `basectl`, the TEE registrar and the load tests.

## The fix

**1. Add one wei**, so the bump clears the threshold instead of sitting on it. This subsumes the old "at least `x + 1`" special case for small `x` where `x / 10` truncated to zero, so that branch goes away.

**2. Stop a failed bump re-firing immediately.** `publish_tx` routes its error through `process_send_error`, which re-sets the `bump_fees` flag the send loop had just taken — so the loop re-entered the bump path on the next iteration rather than on the resubmission timer, contradicting its own `will retry next tick` log line and driving retries at ~8/s. The loop now takes the flag again after each attempt. Defence in depth: with the threshold fixed our own arithmetic no longer produces a rejection, but a competing sender on the same nonce, or a node configured with a larger price bump, still can — and that should cost one attempt per resubmission interval, not a hot loop.

Runaway escalation stays bounded by the existing `check_fee_limits` / `fee_limit_multiplier` path, which is untouched.

## Test plan

- [x] `cargo test -p base-tx-manager` on this branch — 377 pass. 15 expectations changed, every one of which encoded the old boundary value.
- [x] New `calc_threshold_value_strictly_clears_mempool_threshold` asserts the property directly against `floor(x * (100 + bump) / 100)`, using the two fee values observed in the incident. It fails on the previous implementation.
- [x] `cargo clippy -p base-tx-manager --all-targets` and `cargo fmt --check` clean (verified on the `l3` port).
- [ ] **Not covered by an automated test:** the loop-pacing change in part 2. Asserting it needs a full mocked send-loop RPC script under the deterministic runtime, which is brittle relative to the value; the failure mode is noisy retries rather than incorrectness.

Port of base#4506 (`l3`), same pattern as base#4442 -> base#4445.

### DIFF
--- a/crates/utilities/tx-manager/src/fees.rs
+++ b/crates/utilities/tx-manager/src/fees.rs
@@ -51,14 +51,23 @@ impl FeeCalculator {
         blob_base_fee.saturating_mul(2)
     }
 
-    /// Returns the minimum replacement value that clears a mempool's
+    /// Returns the minimum replacement value that clears **every** mempool's
     /// tx-replacement rule.
     ///
-    /// A mempool rejects a replacement whose fee is not **strictly** above
-    /// `x * (100 + bump_percent) / 100`. Because that division truncates, the
-    /// threshold it compares against is exactly `x + x / divisor` — so
-    /// returning that value lands *on* the boundary and is rejected. One extra
-    /// wei clears it.
+    /// Replacement policy is client policy, not consensus, and the clients we
+    /// submit to round the threshold differently:
+    ///
+    /// * **geth** compares against `floor(x * (100 + bump) / 100)` and accepts
+    ///   equality (`core/txpool/legacypool/list.go`).
+    /// * **reth** compares against `div_ceil(x * (100 + bump), 100)` and
+    ///   requires at least that (`transaction-pool/src/validate/mod.rs`).
+    ///
+    /// `x + x / divisor` is exactly geth's floor, so it satisfies geth but is
+    /// one wei short of reth whenever `x` is not divisible by the divisor. The
+    /// trailing `+ 1` clears the stricter of the two, which is what a
+    /// replacement needs in order to be valid on either client — and to
+    /// propagate across a mixed network, since a reth peer still holding the
+    /// original rejects an under-bumped replacement and keeps the original.
     ///
     /// * **Regular transactions** (`is_blob = false`): 10 % bump →
     ///   `x + x / 10 + 1`.
--- a/crates/utilities/tx-manager/src/fees.rs
+++ b/crates/utilities/tx-manager/src/fees.rs
@@ -51,17 +51,22 @@ impl FeeCalculator {
         blob_base_fee.saturating_mul(2)
     }
 
-    /// Returns the minimum replacement value that satisfies geth's
-    /// tx-replacement rules.
+    /// Returns the minimum replacement value that clears a mempool's
+    /// tx-replacement rule.
+    ///
+    /// A mempool rejects a replacement whose fee is not **strictly** above
+    /// `x * (100 + bump_percent) / 100`. Because that division truncates, the
+    /// threshold it compares against is exactly `x + x / divisor` — so
+    /// returning that value lands *on* the boundary and is rejected. One extra
+    /// wei clears it.
     ///
     /// * **Regular transactions** (`is_blob = false`): 10 % bump →
-    ///   `x + x / 10`, but always at least `x + 1` when `x > 0`.
-    /// * **Blob transactions** (`is_blob = true`): 100 % bump →
-    ///   `x + x` (i.e. `2 × x`), but always at least `x + 1` when `x > 0`.
+    ///   `x + x / 10 + 1`.
+    /// * **Blob transactions** (`is_blob = true`): 100 % bump → `2x + 1`.
     /// * When `x == 0`, returns `0` (there is no fee to bump).
     ///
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

robriks:
**Correcting my own description — the blast radius is narrower than it claims, and reviewers should read this first.**

I asserted this bites on Ethereum and that the blob path has the same shape. Having now read both mempool implementations, **both claims are wrong.**

`calc_threshold_value` returns `x + floor(x/10)`, which is exactly `floor(x * 110 / 100)`. The two clients compare differently:

**geth** (`core/txpool/legacypool/list.go`) floors the threshold and rejects only if the replacement is *below* it:
```go
thresholdFeeCap := aFeeCap.Div(aFeeCap, b)   // floor
if tx.GasFeeCapIntCmp(thresholdFeeCap) < 0 { return false, nil }
```
Our value equals the threshold, so `< 0` is false — **geth accepts it.**

**reth** (`crates/transaction-pool/src/validate/mod.rs`) rounds the threshold *up*:
```rust
let required_bumped_fee = |existing_fee: u128| existing_fee.saturating_mul(100 + price_bump).div_ceil(100);
if maybe_replacement.max_fee_per_gas() < required_bumped_fee(self.max_fee_per_gas()) { return true }
```
When `x % 10 != 0`, `ceil > floor`, so our value is one wei short — **reth rejects it.**

So the corrected scope:

- **Only reth-based mempools.** Ethereum settlement has been fine all along, which is why this has never caused an incident despite being years old. The paths under real fee pressure settle to Ethereum, where geth accepts the value.
- **Only the 10% non-blob bump**, and only when the old fee is not divisible by 10 — about 90% of values, so it is near-total *on reth*, not intermittent.
- **Blobs are unaffected.** The 100% bump gives `ceil(200x/100) == 2x` exactly, and the old code already returned `2x`. My "same shape" claim was wrong.

The observed incident matches exactly: fee cap `15297156` is not divisible by 10, so `floor` = 16826871 and `ceil` = 16826872 — we sent 16826871 and were rejected. The tip `1000000` *is* divisible by 10, so that leg passed. I earlier said both legs sat on the boundary; only the fee cap did.

**The fix is unchanged and still correct** — `+1` clears `div_ceil` on reth and costs one wei on geth. But the justification is Base/OP-stack settlement, not Ethereum: our L3 settles to Base Sepolia, any appchain using `--l1-tx-format base` settles to Base, and any service RBF-ing on Base hits the same thing. I'll rewrite the body to match; flagging it here so the history shows what changed and why.

PR #4547, opened 2026-08-20
fix(consensus): Absorb Out-Of-Lockstep Safe-Head Updates

refcell:
## Summary

The engine notifies the derivation actor of every safe-head advance, including out-of-lockstep advances from L1 consolidation catch-up, rather than only confirmations of attributes the derivation pipeline itself submitted. When one of those updates arrived while derivation was in AwaitingL1Data or AwaitingSignal, the state machine treated it as an illegal transition and returned a fatal error that crashed the node, which we observed on a sequencer shortly after a restart while it rapidly consolidated its backlog.

This change absorbs the confirmation in those two states instead of crashing. It advances the confirmed safe head only when the head moves forward, and warns plus increments a new metric on a non-advancing head rather than regressing the safe head. Confirmations that legitimately move the safe head backwards during a reorg continue to flow through the reset path via AwaitingUpdateAfterSignal and are unaffected. New unit tests cover the forward absorb, the non-advancing ignore for equal and backwards heads, the AwaitingSignal case, and a regression guard proving the reorg backwards-head path still applies.

### DIFF
--- a/crates/consensus/service/src/actors/derivation/state_machine.rs
+++ b/crates/consensus/service/src/actors/derivation/state_machine.rs
@@ -1,7 +1,8 @@
 use base_protocol::{AttributesWithParent, L2BlockInfo};
 use derive_more::PartialEq;
 use thiserror::Error;
-use tracing::info;
+
+use crate::Metrics;
 
 /// The possible states of the [`DerivationStateMachine`] implemented by the
 /// [`crate::DerivationActor`].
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4549, opened 2026-08-20
feat(batcher): add incremental channel compression

xenoliss:
## Summary

Add channel compression primitives to `base-comp`:

- `CompressionStream` compresses a derivation channel incrementally as bytes are
  appended.
- `CompressionAlgo::compress_channel` compresses a complete channel in one call.

Brotli channels carry the `0x01` channel-version byte; zlib streams are
self-identifying.

## Testing

- `cargo test -p base-comp --all-features`

### DIFF
--- a/crates/batcher/comp/src/stream.rs
+++ b/crates/batcher/comp/src/stream.rs
@@ -109,8 +109,10 @@ impl CompressionStream {
                 // miniz `mz_compressBound`.
                 input_size.saturating_add(input_size / 16).saturating_add(67)
             }
-            CompressionBackend::Brotli(_) => BrotliEncoderMaxCompressedSize(input_size)
-                .saturating_add(CompressionAlgo::BROTLI_CHANNEL_VERSION as usize),
+            CompressionBackend::Brotli(_) => {
+                // Channel-version prefix byte.
+                BrotliEncoderMaxCompressedSize(input_size).saturating_add(1)
+            }
         }
     }
 
--- a/crates/batcher/comp/src/stream.rs
+++ b/crates/batcher/comp/src/stream.rs
@@ -1,0 +1,219 @@
+//! Incremental derivation-channel compression.
+
+use alloc::{boxed::Box, vec::Vec};
+use std::{fmt, io::Write};
+
+use brotli::{CompressorWriter, enc::BrotliEncoderMaxCompressedSize};
+use miniz_oxide::{
+    DataFormat,
+    deflate::{
+        CompressionLevel,
+        core::{CompressorOxide, TDEFLFlush, TDEFLStatus, compress_to_output},
+    },
+};
+
+use crate::{CompressionAlgo, CompressionError};
+
+/// Concrete state owned by one [`CompressionStream`].
+pub enum CompressionBackend {
+    /// Streaming zlib encoder and bytes emitted since the last transfer.
+    Zlib {
+        /// Miniz encoder state.
+        compressor: Box<CompressorOxide>,
+        /// Bytes emitted since the last transfer.
+        output: Vec<u8>,
+    },
+    /// Streaming Brotli encoder writing into its transferable output buffer.
+    Brotli(Box<CompressorWriter<Vec<u8>>>),
+}
+
+impl fmt::Debug for CompressionBackend {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Zlib { output, .. } => {
+                f.debug_struct("Zlib").field("buffered_bytes", &output.len()).finish()
+            }
+            Self::Brotli(compressor) => f
+                .debug_struct("Brotli")
+                .field("buffered_bytes", &compressor.get_ref().len())
+                .finish(),
+        }
+    }
+}
+
+/// A single incremental compressor for one derivation channel.
+pub struct CompressionStream {
+    backend: CompressionBackend,
+    /// Total bytes returned by prior [`append`](Self::append) calls.
+    output_size: usize,
+}
+
+impl CompressionStream {
+    /// Creates an empty compressor for `algorithm`.
+    pub fn new(algorithm: CompressionAlgo) -> Self {
+        let brotli = |quality| {
+            let mut output = Vec::with_capacity(4096);
+            output.push(CompressionAlgo::BROTLI_CHANNEL_VERSION);
+            CompressionBackend::Brotli(Box::new(CompressorWriter::new(output, 4096, quality, 22)))
+        };
+        let backend = match algorithm {
+            CompressionAlgo::Zlib => CompressionBackend::Zlib {
+                compressor: Box::new(CompressorOxide::with_format_and_level(
+                    DataFormat::Zlib,
+                    CompressionLevel::BestCompression,
+                )),
+                output: Vec::new(),
+            },
+            CompressionAlgo::Brotli9 => brotli(9),
+            CompressionAlgo::Brotli10 => brotli(10),
+            CompressionAlgo::Brotli11 => brotli(11),
+        };
+        Self { backend, output_size: 0 }
+    }
+
+    /// Append input and return newly emitted compressed bytes.
+    pub fn append(&mut self, input: &[u8]) -> Result<Vec<u8>, CompressionError> {
+        // miniz reports consumed input; treat the append as committed only
+        // when every byte was accepted.
+        match &mut self.backend {
+            CompressionBackend::Zlib { compressor, output } => {
+                let (status, consumed) =
+                    compress_to_output(compressor, input, TDEFLFlush::None, |bytes| {
+                        output.extend_from_slice(bytes);
+                        true
+                    });
+                if status != TDEFLStatus::Okay || consumed != input.len() {
+                    return Err(CompressionError::Zlib);
+                }
+            }
+            CompressionBackend::Brotli(compressor) => compressor.write_all(input)?,
+        }
+
+        let output = match &mut self.backend {
+            CompressionBackend::Zlib { output, .. } => core::mem::take(output),
+            CompressionBackend::Brotli(compressor) => core::mem::take(compressor.get_mut()),
+        };
+        self.output_size += output.len();
+        Ok(output)
+    }
+
+    /// Bytes already returned by [`append`](Self::append), excluding any `finish` suffix.
+    pub const fn output_size(&self) -> usize {
+        self.output_size
+    }
+
+    /// Conservative upper bound for a finished stream of `input_size` uncompressed bytes.
+    pub fn max_output_size(&self, input_size: usize) -> usize {
+        match &self.backend {
+            CompressionBackend::Zlib { .. } => {
+                // miniz `mz_compressBound`.
+                input_size.saturating_add(input_size / 16).saturating_add(67)
+            }
+            CompressionBackend::Brotli(_) => BrotliEncoderMaxCompressedSize(input_size)
+                .saturating_add(CompressionAlgo::BROTLI_CHANNEL_VERSION as usize),
+        }
+    }
+
+    /// Finishes the stream and returns compressed bytes not previously transferred.
+    pub fn finish(self) -> Result<Vec<u8>, CompressionError> {
+        match self.backend {
+            CompressionBackend::Zlib { mut compressor, mut output } => {
+                let (status, consumed) =
+                    compress_to_output(&mut compressor, &[], TDEFLFlush::Finish, |bytes| {
+                        output.extend_from_slice(bytes);
+                        true
+                    });
+                if status != TDEFLStatus::Done || consumed != 0 {
+                    return Err(CompressionError::Zlib);
+                }
+                Ok(output)
+            }
+            // Dropping the writer emits Brotli's stream trailer.
+            CompressionBackend::Brotli(compressor) => Ok((*compressor).into_inner()),
+        }
+    }
+}
+
+impl fmt::Debug for CompressionStream {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("CompressionStream")
+            .field("backend", &self.backend)
+            .field("output_size", &self.output_size())
+            .finish_non_exhaustive()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use base_common_genesis::RollupConfig;
+    use base_protocol::Brotli;
+    use miniz_oxide::inflate::decompress_to_vec_zlib;
+
+    use super::*;
+
+    const CHUNKS: [&[u8]; 3] = [b"first batch", b"second batch", b"third batch"];
+
+    fn collect(algorithm: CompressionAlgo, chunks: &[&[u8]]) -> Vec<u8> {
+        let mut compressor = CompressionStream::new(algorithm);
+        let mut compressed = Vec::new();
+        for chunk in chunks {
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/batcher/comp/src/types.rs
```diff
@@ -29,6 +48,41 @@ pub enum CompressionAlgo {
...
+
+    /// Compresses one complete derivation channel.
+    ///
+    /// Brotli output starts with [`BROTLI_CHANNEL_VERSION`](Self::BROTLI_CHANNEL_VERSION);
+    /// zlib is self-identifying.
+    pub fn compress_channel(self, input: &[u8]) -> Result<Vec<u8>, CompressionError> {
+        match self {
+            Self::Zlib => Ok(miniz_oxide::deflate::compress_to_vec_zlib(input, 9)),
+            #[cfg(feature = "std")]
+            brotli => {
```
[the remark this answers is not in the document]
xenoliss: Intentionally left as-is: this match shape is transitional. #4556 replaces the Brotli9/10/11 variants with Brotli(u8), which deletes this arm entirely

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    /// Conservative upper bound for a finished stream of `input_size` uncompressed bytes.
+    pub fn max_output_size(&self, input_size: usize) -> usize {
+        match &self.backend {
+            CompressionBackend::Zlib { .. } => {
+                // miniz `mz_compressBound`.
+                input_size.saturating_add(input_size / 16).saturating_add(67)
+            }
+            CompressionBackend::Brotli(_) => BrotliEncoderMaxCompressedSize(input_size)
+                .saturating_add(CompressionAlgo::BROTLI_CHANNEL_VERSION as usize),
```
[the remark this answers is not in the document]
xenoliss: Fixed in 46c1d45: the bound now adds a literal 1 for the prefix length; the constant is kept for the value semantics.

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,219 @@
...
+                        output.extend_from_slice(bytes);
+                        true
+                    });
+                if status != TDEFLStatus::Done || consumed != 0 {
+                    return Err(CompressionError::Zlib);
+                }
+                Ok(output)
+            }
+            // Dropping the writer emits Brotli's stream trailer.
+            CompressionBackend::Brotli(compressor) => Ok((*compressor).into_inner()),
```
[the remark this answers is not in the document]
xenoliss: Checked the brotli 8.0.4 sources. `flush `runs `BROTLI_OPERATION_FLUSH`, not FINISH, so it would insert a flush block and change the output bytes vs one-shot compression. And the sink is a `Vec`, whose writes are infallible, so there is no error to surface. Keeping `into_inner `as is.

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,221 @@
...
+    backend: CompressionBackend,
+    /// Total bytes returned by prior [`append`](Self::append) calls.
+    output_size: usize,
+}
+
+impl CompressionStream {
+    /// Creates an empty compressor for `algorithm`.
+    pub fn new(algorithm: CompressionAlgo) -> Self {
+        let brotli = |quality| {
+            let mut output = Vec::with_capacity(4096);
```
refcell: Where is the magic `4096` number from? Can we make this a constant on `CompressionStream` so we know what this is?

e.g.
```rust
impl CompressionStream {
    /// Doc comments
    pub const COMPRESSION_CAPACITY: u64 = 4096;
}
```

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,221 @@
...
+            output.push(CompressionAlgo::BROTLI_CHANNEL_VERSION);
+            CompressionBackend::Brotli(Box::new(CompressorWriter::new(output, 4096, quality, 22)))
```
refcell: Maybe same with `22`

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,221 @@
...
+                compressor: Box::new(CompressorOxide::with_format_and_level(
+                    DataFormat::Zlib,
+                    CompressionLevel::BestCompression,
+                )),
+                output: Vec::new(),
+            },
+            CompressionAlgo::Brotli9 => brotli(9),
+            CompressionAlgo::Brotli10 => brotli(10),
+            CompressionAlgo::Brotli11 => brotli(11),
+        };
```
refcell: Looks like `CompressionAlgo` is something we define. Can we just define `impl Into<CompressionBackend> for CompressionAlgo` and localize this conversion on `CompressionAlgo`?

e.g.
```rust
impl Into<CompressionBackend> for CompressionAlgo {
    fn into(algo: Self) -> CompressionBackend {
        match algorithm {
            ...
        }
    }
}
```

refcell: Even further it looks like we can also define `From<CompressionAlgo> for CompressionStream` since all this constructor accepts is a `CompressionAlgo`

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,221 @@
...
+                    return Err(CompressionError::Zlib);
+                }
+            }
+            CompressionBackend::Brotli(compressor) => compressor.write_all(input)?,
+        }
+
+        let output = match &mut self.backend {
+            CompressionBackend::Zlib { output, .. } => core::mem::take(output),
+            CompressionBackend::Brotli(compressor) => core::mem::take(compressor.get_mut()),
+        };
```
refcell: I'd prefer if these were methods on `CompressionBackend` so we're not matching on `&mut self.backend` here. We should be able to do:

```rust
self.backend.write(bytes)?;
let output = self.backend.output();
```

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,221 @@
...
+        match &self.backend {
+            CompressionBackend::Zlib { .. } => {
+                // miniz `mz_compressBound`.
+                input_size.saturating_add(input_size / 16).saturating_add(67)
+            }
+            CompressionBackend::Brotli(_) => {
+                // Channel-version prefix byte.
+                BrotliEncoderMaxCompressedSize(input_size).saturating_add(1)
+            }
+        }
```
refcell: These magic numbers should be conses on `CompressionBackend` I think and we should also be able to place this method on `CompressionBackend` impl

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+impl fmt::Debug for CompressionStream {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("CompressionStream")
+            .field("backend", &self.backend)
+            .field("output_size", &self.output_size())
+            .finish_non_exhaustive()
+    }
+}
```
refcell: I really want to get rid of these custom Debug impls... we can use `derive_more::Debug` and just `#[skip]` fields we don't want printed in the debug output.

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,221 @@
...
+                        output.extend_from_slice(bytes);
+                        true
+                    });
+                if status != TDEFLStatus::Done || consumed != 0 {
+                    return Err(CompressionError::Zlib);
+                }
+                Ok(output)
+            }
+            // Dropping the writer emits Brotli's stream trailer.
+            CompressionBackend::Brotli(compressor) => Ok((*compressor).into_inner()),
```
refcell: This as well. It seems to me that CompressionStream should mostly pass right through to `CompressionBackend`. The `match` on backend should all be methods on `CompressionBackend` and then `CompressionStream` functions just dispatch to those methods.

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,241 @@
...
+        CompressionLevel,
+        core::{CompressorOxide, TDEFLFlush, TDEFLStatus, compress_to_output},
+    },
+};
+
+use crate::{CompressionAlgo, CompressionError};
+
+/// Concrete state owned by one [`CompressionStream`].
+#[derive(derive_more::Debug)]
+pub enum CompressionBackend {
```
BrianBland, line 19: It might be worth making this generic over the output type (currently `Vec<u8>`.

We should be able to bound this to a `std::io::Write`

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,241 @@
...
+
+        match algorithm {
+            CompressionAlgo::Zlib => Self::Zlib {
+                compressor: Box::new(CompressorOxide::with_format_and_level(
+                    DataFormat::Zlib,
+                    CompressionLevel::BestCompression,
+                )),
+                output: Vec::new(),
+            },
+            CompressionAlgo::Brotli9 => brotli(9),
```
BrianBland, line 118: It's hypothetically possible that we would use Brotli8 under some circumstances, like if we're catching up from a long backlog and are cpu-constrained.
I wonder if we should just support all Brotli variants here, or if lower compression levels are mostly a footgun.

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,241 @@
...
+
+impl CompressionBackend {
+    /// Buffer size used by the Brotli writer.
+    const BROTLI_BUFFER_SIZE: usize = 4096;
+
+    /// Brotli sliding-window exponent (`2^22` bytes).
+    const BROTLI_LGWIN: u32 = 22;
+
+    /// Appends input and returns newly emitted compressed bytes.
+    pub fn append(&mut self, input: &[u8]) -> Result<Vec<u8>, CompressionError> {
```
BrianBland, line 41: If we just use the standard `write` instead of `append`, we can probably avoid extra copies and buffer allocations.

crates/batcher/comp/src/stream.rs
```diff
@@ -0,0 +1,241 @@
...
+            }
+            Self::Brotli(_) => {
+                // Channel-version prefix byte.
+                BrotliEncoderMaxCompressedSize(input_size).saturating_add(1)
+            }
+        }
+    }
+
+    /// Finishes the stream and returns compressed bytes not previously transferred.
+    pub fn finish(self) -> Result<Vec<u8>, CompressionError> {
```
BrianBland, line 78: If we go with the standard Write approach, this is mostly a `flush` and `close`.

crates/batcher/comp/src/stream.rs
xenoliss: Done in the backend-delegate commit — `CompressionBackend::BROTLI_BUFFER_SIZE` (4096) and `BROTLI_LGWIN` (22), with docs.

crates/batcher/comp/src/stream.rs
xenoliss: Same commit: `CompressionBackend::BROTLI_LGWIN = 22`.

crates/batcher/comp/src/stream.rs
xenoliss: Done as `impl From<CompressionAlgo> for CompressionBackend` (prefer `From` over `Into`) in the backend-delegate commit.

xenoliss: Done: `impl From<CompressionAlgo> for CompressionStream`. `new()` just forwards to that.

crates/batcher/comp/src/stream.rs
xenoliss: Done. `append` / `max_output_size` / `finish` live on `CompressionBackend`; `CompressionStream` only tracks `output_size` and dispatches.

crates/batcher/comp/src/stream.rs
xenoliss: Done — constants and `max_output_size` are on `CompressionBackend`.

crates/batcher/comp/src/stream.rs
xenoliss: Done: both types use `derive_more::Debug` with `#[debug(skip)]` on encoder state.

crates/batcher/comp/src/stream.rs
xenoliss: Done in the same backend-delegate commit — stream methods are thin wrappers over `CompressionBackend`.

crates/batcher/comp/src/stream.rs
xenoliss: Agreed it's a product choice, not a footgun we need to close here. #4556 replaces the Brotli9/10/11 variants with `Brotli(u8)`, so quality 8 is available there if catch-up ever wants it.

crates/batcher/comp/src/stream.rs
xenoliss: Keeping `append` on purpose: it returns the bytes emitted since the last call, which the channel encoder consumes immediately. `std::io::Write` doesn't give that back without draining an internal buffer (what we already do inside the backend).

crates/batcher/comp/src/stream.rs
xenoliss: Same as above — `finish` is the close/trailer path. A `Write::flush` on Brotli inserts a flush block (`BROTLI_OPERATION_FLUSH`), which would change the bitstream vs one-shot compression, so we don't expose that.

crates/batcher/comp/src/stream.rs
xenoliss: I think the current path is already pretty cheap. `Bytes::from(Vec)` lets the channel take ownership of the compressed buffer without copying, and the extra allocation per append is mostly because the channel owns each chunk rather than because of `append` vs `Write`. A `Write` sink would also not tell the caller how many compressed bytes were just produced, which framing and the channel limits currently rely on.

I am open to revisiting this in a separate fast-follow once the batcher is running well, maybe with something like `BytesMut` + `split_to` FIFO in Channel.

crates/batcher/comp/src/stream.rs
xenoliss: Same as the `append` comment above, open to revisiting this in a separate fast-follow once the batcher is running well.

refcell (approved):

PR #4607, opened 2026-08-21
feat(rpc): Upgrade Readiness RPC And Command

refcell:
## Summary

Adds a public, read-only `base_upgradeReadiness` JSON-RPC method (in a new `base` namespace, so external node operators can query it without enabling the privileged admin namespace) and a `basectl upgrade-readiness` subcommand that wraps it. It reports whether the node's advertised protocol version satisfies the upgrades scheduled on L1 — returning the node version, each upgrade's required minimum, activation timestamp, and per-upgrade `supported`/`would_halt` flags — so operators can confirm a node will follow an upgrade before it activates. An optional `target_version` (`--target-version` in the CLI) checks readiness against an announced version before it is even scheduled on L1, covering the gap between rolling out a release and publishing the schedule, and the subcommand exits non-zero when the node is not ready so a fleet rollout can gate on it. This is stacked on #4513 (base branch `rf/fix/upgrade-signal-apply-retry`), whose fail-closed predicates the readiness report reuses so it can never disagree with the node's actual halt behavior.

### DIFF
--- a/crates/consensus/rpc/src/base.rs
+++ b/crates/consensus/rpc/src/base.rs
@@ -1,0 +1,124 @@
+//! Implements the public `base` RPC namespace: read-only, non-admin endpoints for node operators.
+
+use core::str::FromStr;
+
+use async_trait::async_trait;
+use base_consensus_gossip::Metrics;
+use base_upgrade_signal::{
+    AlloyUpgradeSignalReader, PackedProtocolVersion, UpgradeReadiness, UpgradeSignalConfig,
+    UpgradeSignalDefaults, UpgradeSignalError, UpgradeSignalMetricLayer,
+};
+use jsonrpsee::{
+    core::RpcResult,
+    types::{ErrorCode, ErrorObject},
+};
+use tracing::warn;
+
+use crate::BaseApiServer;
+
+/// Server implementation of the public [`crate::BaseApiServer`] (`base` namespace).
+///
+/// Holds the node's upgrade-signal configuration and a contract reader so it can answer readiness
+/// queries with a fresh, authoritative L1 read (which also confirms the node can reach the contract).
+/// It is read-only and never mutates node state, so it is safe to expose on the public RPC.
+#[derive(Debug, Clone)]
+pub struct BaseRpc {
+    /// Upgrade-signal schedule read configuration (carries this node's advertised protocol version).
+    pub upgrade_signal_config: UpgradeSignalConfig,
+    /// Hardened L1 contract reader built from `upgrade_signal_config`.
+    pub upgrade_signal_reader: AlloyUpgradeSignalReader,
+}
+
+impl BaseRpc {
+    /// Creates a new `base`-namespace RPC server from the node's upgrade-signal config and reader.
+    pub const fn new(
+        upgrade_signal_config: UpgradeSignalConfig,
+        upgrade_signal_reader: AlloyUpgradeSignalReader,
+    ) -> Self {
+        Self { upgrade_signal_config, upgrade_signal_reader }
+    }
+
+    /// Parses an optional `major.minor.patch[-rc.N]` target version into a packed value.
+    ///
+    /// Returns an `invalid params` RPC error rather than a generic failure so an operator sees
+    /// exactly what was rejected.
+    fn parse_target_version(
+        target_version: Option<String>,
+    ) -> Result<Option<alloy_primitives::U256>, ErrorObject<'static>> {
+        target_version
+            .map(|version| {
+                PackedProtocolVersion::from_str(&version).map(PackedProtocolVersion::into_inner)
+            })
+            .transpose()
+            .map_err(|error| {
+                ErrorObject::owned(ErrorCode::InvalidParams.code(), error.to_string(), None::<()>)
+            })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use base_upgrade_signal::UpgradeSignalDefaults;
+
+    use super::BaseRpc;
+
+    #[test]
+    fn parses_or_rejects_target_version() {
+        assert_eq!(BaseRpc::parse_target_version(None).unwrap(), None);
+        assert_eq!(
+            BaseRpc::parse_target_version(Some("1.2.3".to_string())).unwrap(),
+            Some(UpgradeSignalDefaults::packed_protocol_version(1, 2, 3))
+        );
+
+        let error = BaseRpc::parse_target_version(Some("not-a-version".to_string())).unwrap_err();
+        assert_eq!(error.code(), jsonrpsee::types::ErrorCode::InvalidParams.code());
+    }
+}
+
+#[async_trait]
+impl BaseApiServer for BaseRpc {
+    async fn upgrade_readiness(
+        &self,
+        target_version: Option<String>,
+    ) -> RpcResult<UpgradeReadiness> {
+        Metrics::rpc_calls("base_upgradeReadiness").increment(1.0);
+
+        let target = Self::parse_target_version(target_version)?;
+        let now_secs = UpgradeSignalDefaults::now_secs();
+
+        // Fresh, authoritative read from the same contract and block tag the node uses. An empty
+        // contract is a valid pre-schedule state (the operator is likely checking a `target_version`
+        // ahead of #4), so it is reported as "nothing scheduled" rather than surfaced as an error.
+        match self
+            .upgrade_signal_config
+            .read_schedule(
+                &self.upgrade_signal_reader,
+                "upgrade readiness",
+                &[UpgradeSignalMetricLayer::Consensus],
+            )
+            .await
+        {
+            Ok(schedule) => Ok(self.upgrade_signal_config.evaluate_readiness(
+                &schedule.signals,
+                Some(schedule.l1_block_number),
+                now_secs,
+                target,
+            )),
+            Err(UpgradeSignalError::EmptySchedule) => {
+                Ok(self.upgrade_signal_config.evaluate_readiness(&[], None, now_secs, target))
+            }
+            Err(error) => {
+                warn!(
+                    target: "upgrade_signal",
+                    error = %error,
+                    "failed to read L1 upgrade schedule for readiness query"
+                );
+                Err(ErrorObject::owned(
+                    -32006,
+                    "failed to read L1 upgrade schedule",
+                    Some(error.to_string()),
+                ))
+            }
+        }
+    }
+}
--- a/crates/consensus/rpc/src/base.rs
+++ b/crates/consensus/rpc/src/base.rs
@@ -1,12 +1,16 @@
 //! Implements the public `base` RPC namespace: read-only, non-admin endpoints for node operators.
 
 use core::str::FromStr;
+use std::{
+    sync::{Arc, Mutex},
+    time::{Duration, Instant},
+};
 
 use async_trait::async_trait;
 use base_consensus_gossip::Metrics;
 use base_upgrade_signal::{
     AlloyUpgradeSignalReader, PackedProtocolVersion, UpgradeReadiness, UpgradeSignalConfig,
-    UpgradeSignalDefaults, UpgradeSignalError, UpgradeSignalMetricLayer,
+    UpgradeSignalDefaults, UpgradeSignalError, UpgradeSignalMetricLayer, UpgradeSignalSchedule,
 };
 use jsonrpsee::{
     core::RpcResult,
@@ -15,6 +19,15 @@ use jsonrpsee::{
 use tracing::warn;
 
 use crate::BaseApiServer;
+
+/// A cached L1 schedule read, used to collapse bursts of readiness queries into one L1 read.
+#[derive(Debug, Clone)]
+struct CachedSchedule {
+    /// When the read completed.
+    read_at: Instant,
+    /// The schedule read from L1; `None` means the contract had no schedule (empty).
+    schedule: Option<UpgradeSignalSchedule>,
+}
 
 /// Server implementation of the public [`crate::BaseApiServer`] (`base` namespace).
 ///
@@ -27,15 +40,30 @@ pub struct BaseRpc {
     pub upgrade_signal_config: UpgradeSignalConfig,
     /// Hardened L1 contract reader built from `upgrade_signal_config`.
     pub upgrade_signal_reader: AlloyUpgradeSignalReader,
+    /// Short-TTL cache of the last L1 schedule read, shared across all clones of this server.
+    schedule_cache: Arc<Mutex<Option<CachedSchedule>>>,
 }
 
 impl BaseRpc {
+    /// How long a schedule read is reused before the next query reads L1 again.
+    ///
+    /// This endpoint is on the public, unauthenticated RPC and each miss reads the L1 contract (with
+    /// the reader's own retries). A short TTL collapses a burst of queries into a single L1 read,
+    /// bounding the amplification into L1 traffic. It is intentionally short — a few seconds — so it
+    /// only dedups bursts and does not meaningfully stale the answer: the node itself polls L1 far
+    /// less often, and a schedule changes only on a (rare) governance action.
+    const CACHE_TTL: Duration = Duration::from_secs(5);
+
     /// Creates a new `base`-namespace RPC server from the node's upgrade-signal config and reader.
-    pub const fn new(
+    pub fn new(
         upgrade_signal_config: UpgradeSignalConfig,
         upgrade_signal_reader: AlloyUpgradeSignalReader,
     ) -> Self {
-        Self { upgrade_signal_config, upgrade_signal_reader }
+        Self {
+            upgrade_signal_config,
+            upgrade_signal_reader,
+            schedule_cache: Arc::new(Mutex::new(None)),
+        }
     }
 
     /// Parses an optional `major.minor.patch[-rc.N]` target version into a packed value.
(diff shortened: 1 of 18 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/utilities/upgrade-signal/src/version.rs
```diff
@@ -102,6 +102,61 @@ impl PackedProtocolVersion {
...
+        let major = parts.next().and_then(|p| p.parse::<u32>().ok());
+        let minor = parts.next().and_then(|p| p.parse::<u32>().ok());
+        let patch = parts.next().and_then(|p| p.parse::<u32>().ok());
+        let extra = parts.next();
+        let (Some(major), Some(minor), Some(patch), None) = (major, minor, patch, extra) else {
+            return Err(ParseProtocolVersionError::new(s));
+        };
+
+        Ok(Self::pack(major, minor, patch, prerelease))
+    }
```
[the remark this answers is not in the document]
refcell: Good catch — fixed. `from_str` now rejects `-rc.0`: a prerelease of 0 is the sentinel for a final release, so `"1.2.3-rc.0"` was silently promoted to `1.2.3` (and `to_string()` dropped the suffix). It now returns `ParseProtocolVersionError`, and `"1.2.3-rc.0"` is in the rejected-input test list.

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,124 @@
...
+                &[UpgradeSignalMetricLayer::Consensus],
+            )
+            .await
+        {
+            Ok(schedule) => Ok(self.upgrade_signal_config.evaluate_readiness(
+                &schedule.signals,
+                Some(schedule.l1_block_number),
+                now_secs,
+                target,
+            )),
```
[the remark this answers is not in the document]
refcell: Addressed. Since this is on the public, unauthenticated RPC, the handler now caches the L1 schedule read behind a short TTL (`CACHE_TTL = 5s`) shared across all clones of the server, so a burst of readiness queries collapses into a single L1 read rather than amplifying into repeated contract reads. A single legitimate call still gets the reader's full retry robustness on a cache miss. The TTL is intentionally short — it only dedups bursts and doesn't meaningfully stale the answer, since the node polls L1 far less often and a schedule changes only on a rare governance action. Read errors are never cached, so transient failures retry on the next query.

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    /// Returns the cached schedule when the entry is still within [`Self::CACHE_TTL`].
+    ///
+    /// The outer `Option` distinguishes a cache hit from a miss; the inner `Option` is the cached
+    /// value (`None` = the contract was empty). A poisoned lock degrades to a miss (a fresh read).
+    fn fresh_cached_schedule(&self) -> Option<Option<UpgradeSignalSchedule>> {
+        let guard = self.schedule_cache.lock().ok()?;
+        let cached = guard.as_ref()?;
+        (cached.read_at.elapsed() < Self::CACHE_TTL).then(|| cached.schedule.clone())
+    }
```
jackchuma: Coalesce concurrent schedule cache misses. When the cache is cold or expired, concurrent calls all observe the miss before any of them writes the result, because this lock is released before `read_schedule(...).await`. A public RPC burst can therefore launch up to the configured 1024 independent multi-request, retrying L1 reads and occupy RPC capacity; the cache only helps after the first read finishes. Use a single-flight or in-flight cache to coalesce misses

crates/consensus/service/src/service/node.rs
```diff
@@ -633,13 +633,20 @@ impl RollupNode {
         let engine_rpc_actor = rpc_builder
             .as_ref()
             .map(|_| (engine_rpc_processor, (cancellation.clone(), engine_rpc_request_rx)));
+        // Public `base` namespace, available whenever the upgrade signal is configured (any mode),
+        // built from the same config and reader the node uses so readiness reflects reality.
+        let base_rpc = self
+            .upgrade_signal_config
+            .as_ref()
+            .map(|c| BaseRpc::new(c.config.clone(), c.reader.clone()));
```
jackchuma, line 643: Report readiness from the schedule the node applies. `base_rpc` is constructed for every configured upgrade-signal mode, including default `metrics-only` and `startup-apply`. The former never applies the L1 schedule and the latter does not apply subsequent L1 changes, but `BaseRpc` independently reads the current L1 schedule and computes `ready` from version support. After a schedule or reschedule, `basectl upgrade-readiness` can therefore exit 0 for a node whose active configuration will not follow that schedule; gate this namespace to a live-applied mode or evaluate the actually applied schedule

crates/utilities/upgrade-signal/src/readiness.rs
```diff
@@ -0,0 +1,239 @@
...
+        let upgrades: Vec<UpgradeReadinessEntry> = signals
+            .iter()
+            .filter(|signal| signal.activation_timestamp > 0)
+            .map(|signal| UpgradeReadinessEntry {
+                upgrade_id: signal.upgrade_id.contract_id().to_string(),
+                required_protocol_version: PackedProtocolVersion::new(signal.protocol_version)
+                    .to_string(),
+                activation_timestamp: signal.activation_timestamp,
+                supported: self.supports_signal_protocol_version(signal),
+                would_halt: self.signal_fails_closed(signal, now_secs, lead_secs),
```
jackchuma: Reject malformed scheduled signals in readiness. When a scheduled signal has `activation_timestamp > 0` and `protocol_version == 0`, this marks it supported because zero sorts below any node version, so the later `all` calculation emits `ready: true`. `validate_schedule_protocol_versions` rejects exactly that schedule, meaning startup-apply can refuse to start and runtime-admin cannot apply it. The pre-flight CLI can therefore exit successfully for a node that cannot apply the onchain schedule; surface the validation failure or mark the report unready

crates/utilities/upgrade-signal/src/version.rs
```diff
@@ -102,6 +102,69 @@ impl PackedProtocolVersion {
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "invalid protocol version '{}', expected 'major.minor.patch' with an optional '-rc.N'",
+            self.input
+        )
+    }
+}
+
+impl std::error::Error for ParseProtocolVersionError {}
```
jackchuma: `ParseProtocolVersionError` is a public type that is only used internally and immediately converted to a string for the RPC error response. It does not carry structured information that any caller consumes.

  Suggestion: use `String` as `PackedProtocolVersion`’s `FromStr::Err`, retain the existing descriptive message, and remove the custom error type and its public re-export

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,183 @@
...
+/// queries with a fresh, authoritative L1 read (which also confirms the node can reach the contract).
+/// It is read-only and never mutates node state, so it is safe to expose on the public RPC.
+#[derive(Debug, Clone)]
+pub struct BaseRpc {
+    /// Upgrade-signal schedule read configuration (carries this node's advertised protocol version).
+    pub upgrade_signal_config: UpgradeSignalConfig,
+    /// Hardened L1 contract reader built from `upgrade_signal_config`.
+    pub upgrade_signal_reader: AlloyUpgradeSignalReader,
+    /// Short-TTL cache of the last L1 schedule read, shared across all clones of this server.
+    schedule_cache: Arc<Mutex<Option<CachedSchedule>>>,
```
jackchuma: `BaseRpc` is never cloned in this PR, and `jsonrpsee::RpcModule` already shares its context across requests. The `Arc` plus `Clone` derives therefore only support hypothetical copies of the server.

  Suggestion: store `Mutex<Option<CachedSchedule>>` directly and remove `Clone` from `BaseRpc` and `CachedSchedule`. The mutex still protects concurrent RPC requests without the extra ownership layer

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,183 @@
...
+
+        Ok(match schedule {
+            Some(schedule) => self.upgrade_signal_config.evaluate_readiness(
+                &schedule.signals,
+                Some(schedule.l1_block_number),
+                now_secs,
+                target,
+            ),
+            None => self.upgrade_signal_config.evaluate_readiness(&[], None, now_secs, target),
+        })
```
jackchuma: Both match arms call `evaluate_readiness` with identical inputs except for extracting `signals` and `l1_block_number` from the optional schedule. Keeping two full calls makes future changes to readiness inputs easier to accidentally apply to only one path.

  Suggestion: derive `(signals, l1_block_number)` from `schedule.as_ref()` first, then make one `evaluate_readiness` call

crates/utilities/upgrade-signal/src/readiness.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+/// Readiness against a caller-supplied target protocol version.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct UpgradeReadinessTarget {
+    /// The requested target minimum, rendered as `major.minor.patch` semver.
+    pub required_protocol_version: String,
+    /// Whether this node's advertised version satisfies the target.
+    pub supported: bool,
+}
```
jackchuma: When `target_version` is supplied, `ready` is defined as exactly `target.supported`. The response’s `target` object therefore repeats both the caller-provided version and the same boolean already available at the top level.

  Suggestion: remove `UpgradeReadinessTarget` and the optional `target` response field. The caller already knows which target it requested; `ready` remains the machine-readable answer

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,183 @@
...
+        let schedule = self.cached_schedule().await.map_err(|error| {
+            warn!(
+                target: "upgrade_signal",
+                error = %error,
+                "failed to read L1 upgrade schedule for readiness query"
+            );
+            ErrorObject::owned(
+                -32006,
+                "failed to read L1 upgrade schedule",
+                Some(error.to_string()),
```
jackchuma: Drop the `data` — it can carry the L1 endpoint URL and the upstream response body (`contract.rs` passes the body to `http_error`), and this is the unauthenticated namespace. `admin.rs` already redacts the equivalent failure, and the `warn!` above keeps the detail server-side

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,183 @@
...
+        target_version: Option<String>,
+    ) -> RpcResult<UpgradeReadiness> {
+        Metrics::rpc_calls("base_upgradeReadiness").increment(1.0);
+
+        let target = Self::parse_target_version(target_version)?;
+        let now_secs = UpgradeSignalDefaults::now_secs();
+
+        // An empty contract is a valid pre-schedule state (the operator is likely checking a
+        // `target_version` ahead of #4), so it is reported as "nothing scheduled" rather than as an
+        // error.
```
jackchuma: `#4` looks like a truncated PR reference

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,201 @@
...
+            .upgrade_signal_config
+            .read_schedule(
+                &self.upgrade_signal_reader,
+                "upgrade readiness",
+                &[UpgradeSignalMetricLayer::Consensus],
+            )
+            .await
+        {
+            Ok(schedule) => Some(schedule),
+            Err(UpgradeSignalError::EmptySchedule) => None,
```
jackchuma: Propagate `EmptySchedule` as an error instead of caching it as `None`. [#4526](https://github.com/base/base/pull/4526) defines an empty append-only schedule as non-authoritative; here it becomes `ready: true`, so `basectl` can exit 0 while the node retains a stale schedule

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,201 @@
...
+            Err(error) => return Err(error),
```
jackchuma: Share failed refreshes with waiters. Each queued caller currently performs another three-attempt L1 read after an error. Broadcast or briefly cache failures and test concurrent failure

crates/consensus/service/src/actors/rpc/actor.rs
```diff
@@ -139,6 +142,12 @@ where
         );
         modules.merge(rollup_rpc.into_rpc())?;
 
+        // Public `base` namespace (read-only, non-admin), enabled when the upgrade signal is
+        // configured so operators — including external ones — can query upgrade readiness.
+        if let Some(base_rpc) = self.base_rpc {
+            modules.merge(base_rpc.into_rpc())?;
```
jackchuma, line 163: Bound readiness calls across transports. HTTP middleware does not limit WS messages, and jsonrpsee enables WS even when `ws_enabled` is false. Use `.http_only()` when disabled and transport-independent limits when enabled

crates/utilities/upgrade-signal/src/config/types.rs
```diff
@@ -21,6 +32,16 @@ impl UpgradeSignalMode {
...
 
+    /// Returns true if this mode applies live L1 schedule changes after startup (and fails closed
+    /// on an unsupportable one).
+    ///
+    /// Only [`Self::RuntimeAdmin`] tracks the schedule live; the other modes observe it for metrics
+    /// but never apply or halt on a change seen after startup. Upgrade-readiness reporting uses this
+    /// to avoid claiming a node will follow the current L1 schedule when its mode never will.
+    pub const fn applies_live_schedule(self) -> bool {
+        matches!(self, Self::RuntimeAdmin)
+    }
```
jackchuma: `applies_live_schedule` is currently a one-caller alias for `allows_runtime_admin`: both return `true` only for `RuntimeAdmin`. Keeping two names for the same predicate creates a second behavior that must stay synchronized as modes evolve.

  Consider removing `applies_live_schedule` and use `self.mode.allows_runtime_admin()` in `evaluate_readiness`

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// [`CachedRead`]) so a burst during an L1 outage does not turn into one retrying read per
+    /// caller, then retried once the short error TTL expires.
+    async fn cached_schedule(&self) -> Result<Option<UpgradeSignalSchedule>, UpgradeSignalError> {
+        if let Some(cached) = self.fresh_cached_read() {
+            return cached;
+        }
+
+        // Coalesce concurrent misses: hold the refresh lock across the L1 read so a burst of
+        // queries collapses into a single in-flight read rather than one read per caller.
+        let _refresh = self.refresh_lock.lock().await;
```
jackchuma: Avoid queuing every request behind the L1 refresh. When a cold/expired cache coincides with a slow or unavailable L1 provider, every `base_upgradeReadiness` caller waits on this mutex. Each waiting handler retains one of the shared `rpc.max-concurrent` HTTP slots, so a public caller can fill the default 1024 slots with readiness requests and prevent healthz and all other RPC methods from being served until the L1 retry sequence or HTTP timeout completes. Make refresh followers fail fast or serve a bounded stale/cached result instead of queuing them behind the L1 read

crates/consensus/rpc/src/base.rs
```diff
@@ -0,0 +1,270 @@
...
+#[async_trait]
+impl BaseApiServer for BaseRpc {
+    async fn upgrade_readiness(
+        &self,
+        target_version: Option<String>,
+    ) -> RpcResult<UpgradeReadiness> {
+        Metrics::rpc_calls("base_upgradeReadiness").increment(1.0);
+
+        let target = Self::parse_target_version(target_version)?;
+        let now_secs = UpgradeSignalDefaults::now_secs();
```
jackchuma: Capture the clock after the schedule read. For a cache miss, this timestamp is taken before `cached_schedule()` can spend up to three 15-second L1 attempts plus backoff. If an activation enters the halt lead window while that I/O is in progress, the response can report `would_halt: false` even though it is returned after the node would halt; queued callers reuse similarly stale request-start timestamps. Sample `now_secs` after the awaited schedule read and immediately before evaluating readiness

crates/consensus/rpc/src/base.rs
refcell: Fixed in 242af3fe4. Refresh followers now `try_lock` and decline instead of awaiting: they serve the most recent cached value (even if past its TTL) or fail fast on a cold cache, so a burst can no longer queue behind the L1 read and occupy shared RPC slots.

crates/consensus/rpc/src/base.rs
refcell: Fixed in 242af3fe4. `now_secs` is now sampled after the awaited schedule read, immediately before `evaluate_readiness`, so a read spanning the halt-lead window can't report a stale `would_halt`. Followers no longer queue, so there are no reused stale request-start timestamps either.

crates/consensus/service/src/actors/rpc/actor.rs
```diff
@@ -81,7 +84,22 @@ pub(crate) async fn launch_rpc_server(
...
                 .expect("Critical: Failed to build GET method proxy"),
         );
-    let server = Server::builder().set_http_middleware(middleware).build(config.socket).await?;
+    // The tower HTTP middleware above (concurrency limit, timeout, load shed) only bounds HTTP
+    // requests — it does not see individual JSON-RPC calls streamed over a WebSocket connection, and
+    // jsonrpsee serves WS by default even when the WS engine module is not merged. So a WS client
+    // could otherwise stream unauthenticated `base_*` calls past those limits. Disable the WS
+    // transport entirely unless it is explicitly enabled, and cap total concurrent connections
+    // (transport-independent) as a backstop.
+    let mut server_config = ServerConfig::builder()
```
jackchuma, line 93: `max_connections` limits sockets, not RPC calls. jsonrpsee spawns each WS message independently, so one connection can still issue unbounded concurrent calls and bypass the HTTP limits. Add an RPC-layer concurrency limit that covers HTTP and WS

jackchuma (approved):
just one comment that could be addressed in a follow up if desired

PR #4614, opened 2026-08-21
feat(load-tests): Expose Position Validity Predicates

refcell:
## Summary

The load-test tooling previously exposed only balance and storage validity predicates, exercising two of the server's four predicate types and leaving the build-position predicates (block_number, flashblock_index) — and the builder expiry/eviction paths they drive — without any load-test coverage. This adds block_number and flashblock_index variants to the validity predicate config, the runtime template, and the router, mirroring the server enum. These predicates carry only an operator and value and read the build context, so they resolve to the same predicate for every transaction regardless of from/to. It also lays the groundwork for BASE-164 delayed-validity transactions, which express a fixed future validity time as a block_number lower bound. The README validity section documents the new types and unit tests cover config parsing and router resolution.

PR #4617, opened 2026-08-21
feat(load-tests): Submit Fixed-Future Validity Transactions

refcell:
## Summary

Implements BASE-164: an optional `validity.future_validity_delay` knob that makes the whole validity cohort become valid at the same future block, producing a predictable, precisely-timed sequencer load spike. At run start the tool reads the chain tip and freezes one absolute target block (`tip + ceil(delay / block_time)`, at least one block ahead) into the router before it is cloned per sender, so every validity transaction carries the same `block_number >= target` predicate and the cohort parks until that block and then activates together. This keeps the cohort simultaneous even when block_time is approximate, since one absolute target is shared. The delay threads through the config, LoadConfig, and to_load_config, adds LatestBlockExt::latest_block_number to the RPC client, and is validated to require ratio > 0 and to count its appended predicate against the maximum (a delay alone satisfies the non-empty-predicates rule). Documented in the README with unit tests for target computation, predicate injection, config parsing, and validation.

Stacked on #4614 (position predicates) since it builds on the same router and config; review after that merges. Metrics proving the spike landed at the target follow in a separate PR.

### DIFF
--- a/crates/infra/load-tests/README.md
+++ b/crates/infra/load-tests/README.md
@@ -360,6 +360,28 @@ The final summary's `by_cohort` breakdown reports confirmed transactions split
 across the `plain` and `validity_pass` cohorts, so plain traffic can be compared
 against validity traffic when the workload is enabled.
 
+##### Delayed (fixed future) validity
+
+Set `future_validity_delay` to make the whole validity cohort become valid at the
+same future block, producing a predictable, precisely-timed sequencer load spike:
+
+```yaml
+validity:
+  ratio: 1.0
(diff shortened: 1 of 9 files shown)


refcell:
Closing in favor of #4614. As discussed, the `block_number` validity predicate exposed in #4614 already provides the delayed-validity mechanism: configure a future `block_number >= X` predicate on the validity cohort and it parks until block X, then becomes valid simultaneously. The auto relative-targeting layer (compute X from now + delay at run start) added here isn't necessary for the load-testing use case; operators can pick a future block directly. README guidance for choosing X will be added to #4614.

PR #4618, opened 2026-08-21
feat(load-tests): Report Validity Spike Metrics

refcell:
## Summary

Completes BASE-164's observability: surfaces whether the delayed-validity load spike actually landed at its target block. The resolved target block is recorded on the config summary at run start, and a new `validity_spike` summary section reports, over the `validity_pass` cohort, the first block a validity transaction confirmed in, its delta from the target (`0` means it landed exactly at the target), and counts of transactions confirming at/after versus before the target. A non-zero before-target count signals the `block_number` predicate was not enforced. Adds `validity_future_delay` and `validity_target_block` to the config summary, documents the section in the README, and covers the aggregation with unit tests.

Stacked on #4617 (the delayed-validity mechanism), which resolves the target block this measures; review after that merges.

### DIFF
--- a/crates/infra/load-tests/README.md
+++ b/crates/infra/load-tests/README.md
@@ -382,6 +382,13 @@ cohort stays simultaneous even if `block_time` is approximate (that only shifts
 which wall-clock second the spike lands). The delay consumes one predicate slot,
 requires `ratio > 0`, and may be used with no other predicates configured.
 
+When a delay is set, the summary reports the resolved `validity_target_block` in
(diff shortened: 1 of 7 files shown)


refcell:
Closing along with #4617. The delayed-validity spike is achievable directly with the `block_number` validity predicate exposed in #4614 (a future `block_number >= X` parks the cohort until block X, then activates it together), so the dedicated auto-targeting + its spike metrics aren't needed. Documentation guidance for choosing the target block will be added to #4614.

PR #4447, opened 2026-08-14
fix(system-tests): fix upgrade-signal tests and speed up stack startup

BrianBland:
## Summary
- Merge-queue `ci / System Tests` has been red since Denim landed in #4377. The only assertion failures are two colocated unit tests in `base-system-tests` that assumed Cobalt was last in `CONTRACT_VARIANTS` (`schedule.last() == 42/84`, actual `0` because Denim now occupies that slot).
- Look up schedule slots by upgrade instead of first/last position so appending the next hardfork cannot break these tests again.
- Cut stack startup: default L1 slots to 1s (L2 stays 2s) and overlap live `op-deployer` with Lighthouse startup. Deploy confirms one tx per L1 slot, so this is the dominant CI cost.
- Action tests were already a required check and are green on `main`. System tests already ran on the merge queue but were **not** required, which is why broken tests were not blocking merges. `ci / System Tests` is now a required status check on `main`.

## Test plan
- [x] `just build::contracts && cargo nextest run -p base-system-tests --lib` — 3 passed
- [x] Added `ci / System Tests` to required status checks on `main`
- [ ] Merge-queue `ci / System Tests` goes green on this PR

PR #4641, opened 2026-08-24
feat(telemetry): add node telemetry v1 MVP

montycheese:
## Summary

This is the smallest end-to-end slice of node telemetry that proves the shape works: a consensus node collects a full node report, POSTs it to the existing `base-telemetry` service, and the service records it.

**It ships dark.** The reporting actor is `None` unless `--telemetry.endpoint` is set, so merging this cannot cause any node to report anywhere. Turning sending on is a separate change.

## What changed

Two new crates. `base-telemetry-types` is the wire schema and nothing else, depended on by both the client and the service so the two cannot drift. `base-telemetry-client` is collection and transport — identity, hardware, latency sampling, and a `ReportSink` seam — deliberately client-agnostic so the execution node and snapshot CLI can reuse it without a move.

`TelemetryActor` in the consensus service is modeled on `UpgradeSignalMetricsActor`: same `NodeActor` impl, same interval-plus-cancellation select loop. `define_telemetry_args!` sits beside `define_metrics_args!` in `base-cli-utils` and flattens into `BaseCli`. `base telemetry preview` prints the exact payload a node would send, so the payload is reviewable before anything is ever sent.

On the service side, `/v1/ingest` lands in the existing `base-telemetry-service` behind the rate limiter already there, recording through a `ReportRecorder` seam whose one implementation writes JSONL and emits a structured event per report.

## Design notes worth reviewing

**The actor can never return `Err`.** `spawn_and_wait!` installs a `cancellation.drop_guard()` per task, so any actor returning `Err` cancels the whole node. A telemetry failure taking down a Base node is the exact inversion of what this is for. Every recoverable error is logged and swallowed, and a test forces the sink to fail every call and asserts the actor stays up.

**This reads typed in-process state rather than scraping the Prometheus registry.** Two mutually exclusive recorders exist in this tree — `BaseCli::run` installs one, reth's `install_prometheus_recorder()` installs another and `.expect()`s — and only one can win `set_global_recorder`. They also disagree on naming, since reth's wraps everything in a `PrefixLayer::new("reth")`. A scraper would have to know which recorder won in order to know what a metric is called. Every value in this payload already has a typed source: heads from `watch::Receiver<EngineState>`, peers from `P2pRpcRequest`. Registry scraping earns its place when we want breadth across the ~40 crates using `define_metrics!`; that is a follow-up, not a prerequisite.

**The wire format is snake_case, not the repo's camelCase HTTP convention.** These become log facets, and the precedent is the transaction-event journal in `observability-events`, which is snake_case NDJSON for the same reason. It is a log record, not a browser-facing API.

**`schema_version` is an integer.** The receiver rule is "reject an unknown major," which an exact string match cannot express.

**The telemetry ID is a UUID, and the migration is silent.** Identities minted earlier were 32 undashed hex characters, which is exactly the simple UUID form, so an existing `telemetry-id` file parses unchanged and a node keeps its identity across the upgrade. There is a test pinning that.

**The payload is flat.** Everything at the top level answers *which node is this*; everything in a nested block answers *what is it reporting*. That is the split consumers actually use — the top level is what you group and filter by, the blocks are what you aggregate. A test pins the exact top-level key set so the contract cannot drift silently.

**Rate limits key on the edge IP, never on anything in the payload.**

## Testing

Unit coverage across the sampler, identity, hardware collector, reporter queue-drop behavior, and the actor's failure tolerance; ingest tests for 202, 400, 413, 429 and a JSONL round-trip; and an `etc/systems` end-to-end test.

**Draft because this has not been compiled since being rebased onto main.** It was green beforehand — clippy clean, `cargo +nightly fmt` clean, 395 lib tests plus 2 integration tests — and the rebase produced one trivial import conflict, but CI is the first real check on the merged result. `cargo metadata --locked` does pass, so the manifests and lockfile are at least internally consistent.

The system test needs a live devnet and has not been re-run since #4447 reshaped stack bring-up.

## Not in this PR

Nothing enables sending. There is no public ingress and no remote kill switch, so this is deliberately not deployable to third-party nodes as-is. The execution-layer client is also out of scope; v1 is consensus-only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2918,6 +2918,8 @@ dependencies = [
  "base-execution-evm",
  "base-node-core",
  "base-node-runner",
+ "base-telemetry-client",
+ "base-telemetry-types",
  "base-txpool-rpc",
  "base-upgrade-signal",
  "clap",
@@ -2926,6 +2928,7 @@ dependencies = [
  "reth-cli-commands",
  "reth-cli-runner",
  "serde",
+ "serde_json",
  "tokio",
  "tokio-util",
  "tracing",
@@ -3485,6 +3488,7 @@ version = "0.0.0"
 dependencies = [
  "backtrace",
  "base-metrics",
+ "base-telemetry-client",
  "clap",
  "dirs-next",
  "eyre",
@@ -3819,6 +3823,8 @@ dependencies = [
  "base-jwt",
  "base-protocol",
  "base-retry",
+ "base-telemetry-client",
+ "base-telemetry-types",
  "base-upgrade-signal",
  "clap",
  "dirs 6.0.0",
@@ -4023,6 +4029,8 @@ dependencies = [
  "base-metrics",
  "base-node-runner",
  "base-protocol",
+ "base-telemetry-client",
+ "base-telemetry-types",
  "base-upgrade-signal",
  "bytes",
  "derive_more 2.1.1",
@@ -6272,7 +6280,11 @@ dependencies = [
  "base-shadow-indexer",
  "base-shadow-indexer-db",
  "base-shadow-metrics",
+ "base-telemetry-client",
+ "base-telemetry-service",
+ "base-telemetry-types",
  "base-test-utils",
+ "base-trusted-proxy",
  "base-tx-forwarding",
  "base-tx-manager",
  "base-txpool-rpc",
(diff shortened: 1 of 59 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4645, opened 2026-08-24
test(systems): add in-process prover-service and ZK host

mw2000:
## Summary
Add `InProcessProverService` (Postgres testcontainer + JSON-RPC) and `InProcessZkHost` (DryRun only) so system tests can boot the prove stack without cluster/network credentials.

### DIFF
--- a/etc/systems/Cargo.toml
+++ b/etc/systems/Cargo.toml
@@ -105,9 +105,9 @@ async-trait.workspace = true
 tracing = { workspace = true, features = ["std"] }
 serde_json = { workspace = true, features = ["std"] }
 serde = { workspace = true, features = ["derive", "std"] }
-sqlx = { workspace = true, features = ["runtime-tokio", "postgres", "migrate", "tls-native-tls"] }
-testcontainers = { workspace = true, features = ["blocking", "host-port-exposure"] }
 testcontainers-modules = { workspace = true, features = ["postgres"] }
+testcontainers = { workspace = true, features = ["blocking", "host-port-exposure"] }
+sqlx = { workspace = true, features = ["runtime-tokio", "postgres", "migrate", "macros", "tls-native-tls"] }
 
 [features]
 default = [ "upgrade-signal" ]
--- a/etc/systems/Cargo.toml
+++ b/etc/systems/Cargo.toml
@@ -72,10 +72,10 @@ base-common-rpc-types.workspace = true
 base-common-precompiles.workspace = true
 
 # proof
-base-prover-service.workspace = true
 base-proof-zk-host.workspace = true
-base-prover-service-db.workspace = true
+base-prover-service.workspace = true
 base-proof-zk-backend.workspace = true
+base-prover-service-db.workspace = true
 base-prover-service-client.workspace = true
 base-prover-service-protocol = { workspace = true, features = ["rpc-server"] }
 
@@ -105,7 +105,7 @@ async-trait.workspace = true
 tracing = { workspace = true, features = ["std"] }
 serde_json = { workspace = true, features = ["std"] }
 serde = { workspace = true, features = ["derive", "std"] }
-sqlx = { workspace = true, features = ["runtime-tokio", "postgres", "tls-native-tls"] }
+sqlx = { workspace = true, features = ["runtime-tokio", "postgres", "migrate", "tls-native-tls"] }
 testcontainers = { workspace = true, features = ["blocking", "host-port-exposure"] }
 testcontainers-modules = { workspace = true, features = ["postgres"] }
 
--- a/etc/systems/Cargo.toml
+++ b/etc/systems/Cargo.toml
@@ -158,7 +158,6 @@ futures.workspace = true
 indicatif.workspace = true
 tokio-tungstenite.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
-testcontainers-modules = { workspace = true, features = ["postgres"] }
 
 [[bench]]
 name = "b20_zk_proving"
--- a/etc/systems/Cargo.toml
+++ b/etc/systems/Cargo.toml
@@ -71,13 +71,21 @@ base-common-network.workspace = true
 base-common-rpc-types.workspace = true
 base-common-precompiles.workspace = true
 
+# proof
+base-prover-service.workspace = true
+base-proof-zk-host.workspace = true
+base-prover-service-db.workspace = true
+base-proof-zk-backend.workspace = true
+base-prover-service-client.workspace = true
+base-prover-service-protocol = { workspace = true, features = ["rpc-server"] }
+
 # tokio
 tokio-util.workspace = true
 tokio = { workspace = true, features = ["full", "rt-multi-thread", "macros"] }
 
 # rpc
 axum = { workspace = true, features = ["http1", "tokio"] }
-jsonrpsee = { workspace = true, features = ["http-client"] }
+jsonrpsee = { workspace = true, features = ["http-client", "server"] }
 reth-rpc-layer.workspace = true
 tower.workspace = true
 
@@ -97,7 +105,9 @@ async-trait.workspace = true
 tracing = { workspace = true, features = ["std"] }
 serde_json = { workspace = true, features = ["std"] }
 serde = { workspace = true, features = ["derive", "std"] }
+sqlx = { workspace = true, features = ["runtime-tokio", "postgres", "tls-native-tls"] }
 testcontainers = { workspace = true, features = ["blocking", "host-port-exposure"] }
+testcontainers-modules = { workspace = true, features = ["postgres"] }
 
 [features]
 default = [ "upgrade-signal" ]
@@ -129,8 +139,6 @@ base-common-rpc-types-engine.workspace = true
 
 # proof
 base-optimism-rpc.workspace = true
-base-prover-service-client.workspace = true
-base-prover-service-protocol.workspace = true
 
 # execution
 base-shadow-indexer.workspace = true
@@ -149,6 +157,7 @@ anyhow.workspace = true
 futures.workspace = true
 indicatif.workspace = true
 tokio-tungstenite.workspace = true
+tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
 testcontainers-modules = { workspace = true, features = ["postgres"] }
 
 [[bench]]
--- a/etc/systems/src/prover_service.rs
+++ b/etc/systems/src/prover_service.rs
@@ -1,0 +1,159 @@
+//! In-process prover-service JSON-RPC server backed by a Postgres testcontainer.
+
+use std::{
+    fs,
+    net::SocketAddr,
+    path::{Path, PathBuf},
+    time::Duration,
+};
+
+use base_prover_service::{
+    ProverServiceServer, ServerConfig, StatusPoller, WorkerApiConfig, WorkerQueueConfig,
+};
+use base_prover_service_db::{DatabaseConfig, ProofRequestRepo};
+use base_prover_service_protocol::{ProverRequesterApiServer, ProverWorkerApiServer};
+use eyre::{Result, WrapErr, ensure};
+use jsonrpsee::server::{Server, ServerHandle};
+use sqlx::Executor;
+use testcontainers::{ContainerAsync, runners::AsyncRunner};
+use testcontainers_modules::postgres::Postgres;
+use tokio::task::JoinHandle;
+use tracing::info;
+use url::Url;
+
+/// Production default shared with the prover-service binary.
+const MAX_PROOF_RETRIES: i32 = 3;
+/// Dry-run SP1 execute can outlive the production 300s default lock.
+const DEFAULT_LOCK_DURATION_SECONDS: u32 = 1800;
+const MAX_LOCK_DURATION_SECONDS: u32 = 3600;
+const STATUS_POLLER_INTERVAL_SECS: u64 = 30;
+const STUCK_REQUEST_TIMEOUT_MINS: i32 = 10;
+
+/// In-process prover-service requester and worker JSON-RPC on one listen address.
+pub struct InProcessProverService {
+    /// Shared requester and worker JSON-RPC URL.
+    url: Url,
+    server_handle: Option<ServerHandle>,
+    status_handle: JoinHandle<()>,
+    _postgres: ContainerAsync<Postgres>,
+}
+
+impl std::fmt::Debug for InProcessProverService {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("InProcessProverService").field("url", &self.url).finish_non_exhaustive()
+    }
+}
+
+impl InProcessProverService {
+    /// Starts Postgres, applies prover-service migrations, and binds JSON-RPC on `127.0.0.1:0`.
+    pub async fn start() -> Result<Self> {
+        let postgres = Postgres::default()
+            .start()
+            .await
+            .wrap_err("failed to start prover-service Postgres testcontainer")?;
+        let port = postgres
+            .get_host_port_ipv4(5432)
+            .await
+            .wrap_err("failed to map prover-service Postgres port")?;
+        let database_url =
+            format!("postgres://postgres:postgres@127.0.0.1:{port}/postgres?sslmode=disable");
+
+        let db_config = DatabaseConfig {
+            url: database_url,
+            max_connections: 5,
+            connection_timeout: Duration::from_secs(5),
+        };
+        let pool = db_config.init_pool().await.map_err(|error| {
+            eyre::eyre!("failed to connect to prover-service Postgres: {error}")
+        })?;
+        apply_prover_migrations(&pool).await?;
+        let repo = ProofRequestRepo::new(pool);
+
+        let server_config = ServerConfig {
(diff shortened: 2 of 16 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

etc/systems/tests/common/mod.rs
```diff
@@ -13,14 +13,18 @@ pub(crate) const L1_CHAIN_ID: u64 = 1337;
 pub(crate) const L2_CHAIN_ID: u64 = 84538453;
 pub(crate) const BASE_AZUL_ACTIVATION_BLOCK: u64 = 0;
 pub(crate) const BASE_BERYL_ACTIVATION_BLOCK: u64 = 3;
+#[allow(dead_code)]
```
jackchuma: Remove dead-code lint suppressions. These new `#[allow(dead_code)]` attributes hide unused helpers in each integration-test binary instead of keeping the shared module scoped to the tests that use those helpers. Split or localize the helpers rather than suppressing the lint, as the repository explicitly forbids adding allow-lints

etc/systems/src/zk_host.rs
```diff
@@ -0,0 +1,165 @@
...
+    ///
+    /// Only [`ZkBackend::DryRun`] is supported: cluster and network need remote credentials.
+    pub async fn start(
+        stack: &SystemTestStack,
+        prover_service_url: &Url,
+        backend: ZkBackend,
+    ) -> Result<Self> {
+        if backend != ZkBackend::DryRun {
+            bail!("InProcessZkHost only supports ZkBackend::DryRun, got {backend}");
+        }
```
jackchuma: `InProcessZkHost::start` accepts any `ZkBackend`, but immediately rejects every value except `DryRun`; the sole call site also always passes `DryRun`. This exposes a configuration choice the helper cannot honor and adds a guard solely to reject it.

  Make `start` unconditionally construct the dry-run backend and remove the `backend` parameter and `bail!` branch. Add backend selection only when a second supported backend is actually usable in system tests

etc/systems/src/prover_service.rs
```diff
@@ -0,0 +1,163 @@
...
+        let status_poller = StatusPoller::new(
+            repo.clone(),
+            STATUS_POLLER_INTERVAL_SECS,
+            STUCK_REQUEST_TIMEOUT_MINS,
+            MAX_PROOF_RETRIES,
+            server_config.worker_queue,
+        );
+        let status_handle = tokio::spawn(async move {
+            status_poller.run().await;
+        });
```
jackchuma: The status poller only handles stale-request retries and expired-claim reaping. This PR’s smoke test verifies that the JSON-RPC service accepts a request and that the ZK host can start; it neither submits work nor exercises recovery behavior.

  Do not spawn and retain a background recovery task in this harness yet. Start it when a system test needs to verify retry or expired-lock behavior

etc/systems/src/prover_service.rs
```diff
@@ -0,0 +1,163 @@
...
+    Ok(())
+}
+
+fn prover_migrations_dir() -> Result<PathBuf> {
+    let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+    let dir = manifest_dir.join("../../crates/proof/prover-service/db/migrations");
+    dir.canonicalize().wrap_err_with(|| {
+        format!("prover-service migrations directory not found at {}", dir.display())
+    })
+}
```
jackchuma: `apply_prover_migrations` manually discovers SQL files, filters and sorts them, reads each file, and executes raw SQL. `sqlx` already provides a migration runner that handles ordering and tracks applied migrations.

  Replace the manual loop with `sqlx::migrate::Migrator::new(prover_migrations_dir()?).await?.run(pool).await?`; retain only the migration-directory lookup and contextual errors

etc/systems/tests/common/mod.rs
```diff
@@ -34,6 +38,33 @@ pub(crate) async fn start_beryl_system() -> Result<(SystemTestStack, RootProvide
...
+        .with_l1_chain_id(L1_CHAIN_ID)
+        .with_l2_chain_id(L2_CHAIN_ID)
+        .with_base_azul_activation_block(BASE_AZUL_ACTIVATION_BLOCK)
+        .with_base_beryl_activation_block(BASE_BERYL_ACTIVATION_BLOCK)
+        .with_base_cobalt_activation_block(BASE_COBALT_ACTIVATION_BLOCK)
+        .build()
+        .await?;
+    let provider = system.l2_builder_provider()?;
+    wait_for_block(&provider, BASE_COBALT_ACTIVATION_BLOCK + 1).await?;
+    Ok((system, provider))
```
jackchuma: `start_cobalt_stack` adds a customizable `SystemTestStackBuilder` path, but no current test supplies custom builder options. `start_cobalt_system` is its only caller, so this is an abstraction for a future use case.

  Inline the existing builder setup into `start_cobalt_system`. Reintroduce a builder-taking helper when a second test actually needs different stack options

etc/systems/Cargo.toml
jackchuma, line 1: Nit: waterfall

etc/systems/src/zk_host.rs
```diff
@@ -0,0 +1,165 @@
...
+        let chain_id = config
+            .get("chainId")
+            .and_then(serde_json::Value::as_u64)
+            .ok_or_eyre("L1 genesis config is missing chainId")?;
+
+        let dir = tempfile::tempdir().wrap_err("failed to create succinct config directory")?;
+        let l1_dir = dir.path().join("L1");
+        let l2_dir = dir.path().join("L2");
+        fs::create_dir_all(&l1_dir).wrap_err("failed to create succinct L1 config directory")?;
+        fs::create_dir_all(&l2_dir).wrap_err("failed to create succinct L2 config directory")?;
```
jackchuma: `install_succinct_chain_configs` creates an empty `L2` directory before starting the host, but `OPSuccinctDataFetcher::fetch_and_save_rollup_config` already creates its configured L2 directory before writing the rollup config.

  Keep passing the L2 path through `l2_config_dir`, but remove the redundant `fs::create_dir_all(&l2_dir)` call

etc/systems/tests/in_process_zk.rs
```diff
@@ -0,0 +1,45 @@
...
+    let service = InProcessProverService::start().await?;
+    let client = ProofRequesterClient::connect(
+        &ProverServiceClientConfig::new(service.url().as_str())
+            .with_request_timeout(Duration::from_secs(30)),
+    )
+    .wrap_err("failed to connect requester client to in-process prover-service")?;
+    let missing = client
+        .get_proof(GetProofRequest { session_id: "missing-in-process-zk-smoke".to_owned() })
+        .await;
+    ensure!(missing.is_err(), "get_proof for an unknown session must fail");
```
jackchuma: This accepts any error, so transport failures, method-not-found, or database errors all make the smoke test pass. Please assert `error.is_not_found()` before logging that JSON-RPC was accepted

etc/systems/tests/in_process_zk.rs
```diff
@@ -0,0 +1,45 @@
...
+    info!(url = %service.url(), "prover-service accepted JSON-RPC");
+
+    info!("starting in-process ZK host");
+    let _host = InProcessZkHost::start(&system, service.url(), ZkBackend::DryRun).await?;
```
jackchuma: The host is dropped immediately after being spawned, before it necessarily makes a worker RPC. Please wait for an observable successful worker poll or claim so the worker namespace and host wiring are actually tested

etc/systems/tests/common/mod.rs
mw2000: Moved Cobalt boot into `tests/common/cobalt.rs` and path-include it only from the binaries that need it, so the shared Beryl module no longer needs `#[allow(dead_code)]`.

etc/systems/src/zk_host.rs
mw2000: Removed the `backend` parameter. `start` always builds the dry-run backend.

etc/systems/src/prover_service.rs
mw2000: Dropped the status poller from this harness. We can add it back when a test actually covers retry or expired-lock recovery.

etc/systems/src/prover_service.rs
mw2000: Replaced the manual SQL loop with `sqlx::migrate::Migrator`.

etc/systems/tests/common/mod.rs
mw2000: Inlined the builder setup into `start_cobalt_system`. Will reintroduce a builder-taking helper only when a second test needs different stack options.

etc/systems/Cargo.toml
mw2000: Sorted the proof deps by line length.

etc/systems/src/zk_host.rs
mw2000: Removed the empty L2 `create_dir_all`. The fetcher still gets `l2_config_dir` and creates that directory itself.

etc/systems/tests/in_process_zk.rs
mw2000: The smoke test now requires `error.is_not_found()` for the missing session.

etc/systems/tests/in_process_zk.rs
mw2000: `InProcessZkHost::start` now waits for the first successful worker `get_next_proof` before returning, so the host has actually talked to the worker namespace.

etc/systems/tests/common/cobalt.rs
```diff
@@ -0,0 +1,45 @@
...
+use eyre::{Result, WrapErr};
+use tokio::time::{sleep, timeout};
+
+const L1_CHAIN_ID: u64 = 1337;
+const L2_CHAIN_ID: u64 = 84538453;
+const BASE_AZUL_ACTIVATION_BLOCK: u64 = 0;
+const BASE_BERYL_ACTIVATION_BLOCK: u64 = 3;
+const BASE_COBALT_ACTIVATION_BLOCK: u64 = 5;
+const BLOCK_PRODUCTION_TIMEOUT: Duration = Duration::from_secs(30);
+const BLOCK_POLL_INTERVAL: Duration = Duration::from_millis(500);
```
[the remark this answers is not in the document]
mw2000: Done. \`cobalt.rs\` now uses the shared constants and \`wait_for_block\`. Beryl-only helpers moved to \`common/beryl.rs\` and are path-included only by the binaries that call them, so we do not reintroduce \`#[allow(dead_code)]\` in the Cobalt-only smoke test.

etc/systems/src/zk_host.rs
```diff
@@ -0,0 +1,219 @@
...
+            );
+        let host = ZkHost::new(client, provers, host_config);
+        let run_cancel = cancel.clone();
+        let join = tokio::spawn(async move {
+            host.run_until_cancelled(run_cancel).await;
+        });
+        timeout(FIRST_WORKER_POLL_TIMEOUT, first_poll_rx.wait_for(|ready| *ready))
+            .await
+            .wrap_err("in-process ZK host did not complete a worker poll")?
+            .wrap_err("in-process ZK host stopped before the first worker poll")?;
```
jackchuma: Cancel the host before propagating a readiness timeout. When the first worker poll exceeds the 10-second timeout, these `?` operators return from `start` before `InProcessZkHost` is constructed. Dropping the local `JoinHandle` detaches the spawned discovery task and dropping `CancellationToken` does not cancel it, so the task keeps polling after the service/stack and its config tempdir are torn down. Cancel and abort (or otherwise await) `join` on this error path before returning

etc/systems/Cargo.toml
```diff
@@ -96,7 +104,9 @@ async-trait.workspace = true
 tracing = { workspace = true, features = ["std"] }
 serde_json = { workspace = true, features = ["std"] }
 serde = { workspace = true, features = ["derive", "std"] }
+sqlx = { workspace = true, features = ["runtime-tokio", "postgres", "migrate", "tls-native-tls"] }
 testcontainers = { workspace = true, features = ["blocking", "host-port-exposure"] }
+testcontainers-modules = { workspace = true, features = ["postgres"] }
```
jackchuma: waterfall

etc/systems/src/prover_service.rs
```diff
@@ -0,0 +1,126 @@
...
+        .wrap_err("failed to apply prover-service migrations")?;
+    Ok(())
+}
+
+fn prover_migrations_dir() -> Result<PathBuf> {
+    let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
+    let dir = manifest_dir.join("../../crates/proof/prover-service/db/migrations");
+    dir.canonicalize().wrap_err_with(|| {
+        format!("prover-service migrations directory not found at {}", dir.display())
+    })
```
jackchuma: This test harness reimplements migration-directory lookup and runtime loading solely to invoke SQLx’s migrator. The workspace already uses `sqlx::migrate!`, which embeds the migrations at compile time and removes the `Path`/`PathBuf`, `canonicalize`, and two helper functions.

  Could this be reduced to:

```rust
  sqlx::migrate!("../../crates/proof/prover-service/db/migrations")
      .run(&pool)
      .await
      .wrap_err("failed to apply prover-service migrations")?;
```

etc/systems/src/zk_host.rs
```diff
@@ -0,0 +1,224 @@
...
+        info!(chain_id, "installed succinct L1/L2 config directories");
+        Ok(dir)
+    }
+}
+
+impl Drop for InProcessZkHost {
+    fn drop(&mut self) {
+        self.cancel.cancel();
+        if let Some(join) = self.join.take() {
+            join.abort();
```
leopoldjoy, line 174: Aborting this task cannot cancel the `spawn_blocking` SP1 execution, so please isolate dry-run execution in a killable process or otherwise prevent timeout/drop from leaving the test runtime waiting indefinitely.

etc/systems/tests/common/cobalt.rs
jackchuma, line 1: This helper has a single caller (`in_process_zk.rs`) and largely repeats the existing Beryl stack setup. The extra file and `#[path]` module wiring add indirection without reuse yet. Consider inlining the Cobalt builder setup into `in_process_zk.rs` and extract a shared helper only once another Cobalt test needs it

etc/systems/src/prover_service.rs
```diff
@@ -0,0 +1,104 @@
...
+const MAX_PROOF_RETRIES: i32 = 3;
+/// Dry-run SP1 execute can outlive the production 300s default lock.
+const DEFAULT_LOCK_DURATION_SECONDS: u32 = 1800;
+const MAX_LOCK_DURATION_SECONDS: u32 = 3600;
+
+/// In-process prover-service requester and worker JSON-RPC on one listen address.
+pub struct InProcessProverService {
+    /// Shared requester and worker JSON-RPC URL.
+    url: Url,
+    server_handle: Option<ServerHandle>,
```
jackchuma, line 25: `server_handle` is wrapped in `Option` solely so `Drop` can call `take()`. `ServerHandle::stop` takes `&self`, so the optional state is unnecessary. Please store `ServerHandle` directly and call `self.server_handle.stop()` in `Drop`

etc/systems/src/zk_host.rs
```diff
@@ -0,0 +1,224 @@
...
+const DEFAULT_SEQUENCE_WINDOW: u64 = 50;
+const RANGE_CYCLE_LIMIT: u64 = 1_000_000_000_000;
+const RANGE_GAS_LIMIT: u64 = 1_000_000_000_000;
+/// Bound for the first worker `get_next_proof` after spawn.
+const FIRST_WORKER_POLL_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// In-process SP1 ZK host bound to a live [`SystemTestStack`].
+pub struct InProcessZkHost {
+    cancel: CancellationToken,
+    join: Option<JoinHandle<()>>,
```
jackchuma, line 41: `join` is optional only to move it out during cleanup. `JoinHandle::abort` takes `&self`, so this can be a plain `JoinHandle<()>`. Please remove the `Option` and call `self.join.abort()` directly in `Drop`

jackchuma (approved):

PR #4650, opened 2026-08-24
fix(consensus): validate blob encoding version byte

### LINKED ISSUE #4633: BlobData::decode() checks the wrong byte as the blob encoding-version byte
## Summary

`BlobData::decode()` in `crates/consensus/derive/src/sources/blob_data.rs` validates the blob encoding version at the wrong byte offset:

```rust
// Validate the blob encoding version
if data[VERSIONED_HASH_VERSION_KZG as usize] != BLOB_ENCODING_VERSION {
    return Err(BlobDecodingError::InvalidEncodingVersion);
}
```

`VERSIONED_HASH_VERSION_KZG` is imported from `alloy_eips::eip4844` and is `0x01` -- EIP-4844's versioned-hash prefix byte, an unrelated constant. Using it as an index here makes this check look at `data[1]`.

Per the blob encoding format, byte **0** of the first field element is the actual encoding-version byte -- the function itself confirms this two lines later:

```rust
encoded_byte[0] = data[0];
```

and the length field is read from `data[2..5]`:

```rust
let length = u32::from_be_bytes([0, data[2], data[3], data[4]]) as usize;
```

So byte 1 is unused padding (always zero), and the version check is validating that padding byte instead of the real version byte at `data[0]`.

## Failure scenario

Any blob where `data[0]` is set to an unsupported/garbage encoding version, but `data[1] == 0` (which it always is, since nothing ever writes to byte 1), passes the version check that's supposed to reject it. `decode()` then proceeds to interpret non-version-0 encoded data as if it were version 0, producing garbage output instead of returning `InvalidEncodingVersion`.

## Why this has gone unnoticed

Every existing test in the file writes the check byte at the same wrong offset, e.g.:

```rust
data[VERSIONED_HASH_VERSION_KZG as usize] = BLOB_ENCODING_VERSION;
```

so none of them exercise `data[0]`, and `test_blob_data_decode_invalid_encoding_version` uses `vec![1u8; 32]` (every byte set to `1`), which can't distinguish which index is actually being checked. The bug has been present since this crate was ported into the repo.

## Proposed fix

Check `data[0]` directly instead of `data[VERSIONED_HASH_VERSION_KZG as usize]`, and fix the corresponding test call sites that currently write to the wrong index. This is a ~1 line functional change plus 3 test updates.

Happy to open the PR for this myself -- let me know if I should be assigned per the CONTRIBUTING process.

huklaa:
## Summary

Fixes #4633.

`BlobData::decode()` was validating the blob encoding version using `VERSIONED_HASH_VERSION_KZG` as an array index, which points to `data[1]`.

The blob encoding version is stored at `data[0]`, so malformed blobs with an unsupported encoding version could incorrectly pass validation.

This change:
- validates `BLOB_ENCODING_VERSION` at `data[0]`
- removes the unrelated `VERSIONED_HASH_VERSION_KZG` import
- updates the affected tests
- adds a regression case where `data[0]` is invalid while `data[1]` remains valid

## Testing

Updated the existing `BlobData` decode tests to cover the correct encoding-version byte.

### DIFF
--- a/crates/consensus/derive/src/sources/blob_data.rs
+++ b/crates/consensus/derive/src/sources/blob_data.rs
@@ -2,7 +2,7 @@
 
 use alloc::{boxed::Box, vec};
 
-use alloy_eips::eip4844::{BYTES_PER_BLOB, Blob, VERSIONED_HASH_VERSION_KZG};
+use alloy_eips::eip4844::{BYTES_PER_BLOB, Blob};
 use alloy_primitives::Bytes;
 pub use base_protocol::BLOB_MAX_DATA_SIZE;
 
@@ -30,7 +30,7 @@ impl BlobData {
         let data = self.data.as_ref().ok_or(BlobDecodingError::MissingData)?;
 
         // Validate the blob encoding version
-        if data[VERSIONED_HASH_VERSION_KZG as usize] != BLOB_ENCODING_VERSION {
+        if data[0] != BLOB_ENCODING_VERSION {
             return Err(BlobDecodingError::InvalidEncodingVersion);
         }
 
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/2</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4655, opened 2026-08-25
fix(tx-manager): retry nonce-too-high on initial publish

xenoliss:
## Summary
- `NonceTooHigh` at initial publish is now transient: retry the same signed transaction (30 attempts, 2s delay) instead of aborting and resetting the nonce.
- Concurrent publications reserve nonces in order but can reach different RPC backends out of order; retrying lets the lower nonce propagate without serializing receipt confirmation or fee bumping.
- Drops the abort-on-`NonceTooHigh` path in `SendState` (state flag, `critical_error` branch, nonce reset/reuse policies), since it no longer surfaces as a critical error.

## Test plan
- [x] `cargo test -p base-tx-manager --all-features`
- [x] `cargo clippy -p base-tx-manager --all-features --all-targets`

### DIFF
--- a/crates/utilities/tx-manager/src/manager.rs
+++ b/crates/utilities/tx-manager/src/manager.rs
@@ -248,6 +248,12 @@ where
     /// Fixed delay between retry attempts in [`Self::prepare`].
     pub const PREPARE_RETRY_DELAY: Duration = Duration::from_secs(2);
 
+    /// Maximum retries when an RPC temporarily rejects an ordered nonce.
+    pub const PUBLISH_MAX_RETRIES: usize = 30;
+
+    /// Fixed delay between nonce-too-high publication retries.
+    pub const PUBLISH_RETRY_DELAY: Duration = Duration::from_secs(2);
+
     /// Creates a new [`SimpleTxManager`] with an injected runtime.
     ///
     /// This mirrors [`Self::new`] but lets deterministic tests control
@@ -956,8 +962,11 @@ where
     /// When `send_tx` exits with an error, the nonce manager may be reset
     /// to prevent stale-counter gaps on the next call:
     ///
-    /// * **Pre-publish errors** (no tx was ever sent): the nonce is always
-    ///   reset so the next attempt re-fetches from the chain.
+    /// * **Pre-publish errors** from `send` reset the cache. `send_async`
+    ///   recycles its pre-reserved nonce instead so concurrent reservations
+    ///   remain valid.
+    /// * **`NonceTooHigh`** is retried with the same signed transaction before
+    ///   reaching cleanup.
     /// * **`SendTimeout`**: the nonce is always reset — even after a
     ///   successful publish — because the timeout may have cancelled
     ///   `prepare()` mid-flight during a fee bump, leaking a nonce from
@@ -1016,10 +1025,7 @@ where
         }
 
         if Self::should_reset_nonce_on_send_error(&result, &send_state, nonce_override) {
-            // Three policies (see should_reset_nonce_on_send_error):
-            // • NonceTooHigh — always reset, even with nonce_override.
-            //   No tx entered the mempool; reserved_high_water protects
-            //   concurrent reservations from collision after reset.
+            // Two policies (see should_reset_nonce_on_send_error):
             // • SendTimeout — always reset. The timeout may have cancelled
             //   prepare() after it reserved a nonce but before the tx reached
             //   the mempool, leaking the nonce manager's internal counter.
@@ -1048,10 +1054,6 @@ where
     ) -> bool {
         match result {
             Ok(_) => false,
-            // Always reset on NonceTooHigh — the publish failed, no tx
-            // entered the mempool, and reserved_high_water protects
-            // concurrent reservations from collision after reset.
-            Err(TxManagerError::NonceTooHigh) => true,
             // When a nonce_override was used, the nonce was irrevocably
             // consumed at reserve_nonce() time. Resetting the nonce manager
             // would corrupt concurrent send_async reservations.
@@ -1083,12 +1085,8 @@ where
         send_state: &SendState,
     ) -> bool {
         match result {
-            // Nonce errors indicate the nonce is genuinely invalid, not a
-            // transient failure.  Returning it to the reuse pool would
-            // cause an infinite retry loop: after a reset() the chain
-            // nonce is re-fetched, but advance_nonce() pops
-            // returned_nonces first, reissuing the same invalid value.
-            Ok(_) | Err(TxManagerError::NonceTooHigh | TxManagerError::NonceTooLow) => false,
+            // A too-low nonce was already consumed and must not be reused.
+            Ok(_) | Err(TxManagerError::NonceTooLow) => false,
             Err(_) => !send_state.has_published(),
         }
     }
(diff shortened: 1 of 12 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -248,6 +248,12 @@ where
     /// Fixed delay between retry attempts in [`Self::prepare`].
     pub const PREPARE_RETRY_DELAY: Duration = Duration::from_secs(2);
 
+    /// Maximum retries when an RPC temporarily rejects an ordered nonce.
+    pub const PUBLISH_MAX_RETRIES: usize = 30;
+
+    /// Fixed delay between nonce-too-high publication retries.
+    pub const PUBLISH_RETRY_DELAY: Duration = Duration::from_secs(2);
```
refcell: Can we follow up on this pr and make these tunable via cli flags?

Also, the publish retry delay seems quite high to me. It shouldn't need `2s` for a tx to propagate. Maybe `1s` is a bit more eager

refcell (dismissed):
Approving to unblock, but worth following up on

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -1112,7 +1110,7 @@ where
         // the need for a separate suggest_gas_price_caps() call.
         let prepared =
             self.prepare_with_initial_caps(candidate, None, None, nonce_override, None).await?;
-        let tx_hash = self.publish_tx(send_state, &prepared.raw_tx, None).await?;
+        let tx_hash = self.publish_tx_with_nonce_retry(send_state, &prepared.raw_tx).await?;
```
jackchuma, line 1107: Could we protect synchronous in-flight nonces across this retry? A/B can sign nonces n/n+1; B gets `NonceTooHigh` and sleeps; A fails pre-publish and resets the cache; C refetches and publishes n; B then publishes n+1, but the cache still considers n+1 next, so D reissues it. `send_async` is protected by `reserved_high_water`, but concurrent `send()` calls—used by registrar tasks—are not. Please track the signed nonce across resets or limit this retry to pre-reserved nonces, with a regression test

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -1365,6 +1363,30 @@ where
...
+    /// Concurrent sends reserve nonces in order but may reach different RPC
+    /// backends out of order. Retrying the same signed transaction lets the
+    /// lower nonce propagate without serializing receipt confirmation or fee
+    /// bumping.
+    async fn publish_tx_with_nonce_retry(
+        &self,
+        send_state: &SendState,
+        raw_tx: &Bytes,
+    ) -> TxManagerResult<B256> {
+        (|| self.publish_tx(send_state, raw_tx, None))
```
jackchuma: Stop publication retries after manager closure. When `close()` is called after a first `NonceTooHigh` response but before the backoff expires, the next closure invocation calls `publish_tx` without checking `is_closed()`. That retry can broadcast a transaction after the manager has been shut down; the existing shutdown check is only reached after publication succeeds. Re-check closure before every retry and return `ChannelClosed` instead

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -1365,6 +1363,30 @@ where
...
+            .retry(
+                ConstantBuilder::default()
+                    .with_delay(Self::PUBLISH_RETRY_DELAY)
+                    .with_max_times(Self::PUBLISH_MAX_RETRIES),
+            )
```
jackchuma: Use the injected runtime for retry sleeps. This retry uses backon's default sleeper, which is `tokio::time::sleep`, rather than `self.runtime`. A `NonceTooHigh` retry through `from_wallet_with_runtime` under `base_runtime::deterministic::Runner` therefore attempts to use a Tokio timer without a Tokio runtime and panics, instead of advancing virtual time as the injected-runtime API promises. Supply the runtime's sleep implementation to the retry

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -1365,6 +1357,30 @@ where
...
+            .retry(
+                ConstantBuilder::default()
+                    .with_delay(self.config.publish_retry_delay)
+                    .with_max_times(self.config.publish_max_retries),
+            )
+            .when(|error| matches!(error, TxManagerError::NonceTooHigh))
+            .notify(|error, delay| {
+                warn!(error = %error, ?delay, "retrying initial transaction publication");
+            })
+            .await
```
jackchuma: Honor the mempool deadline while retrying. When `tx_not_in_mempool_timeout` expires during these retries, `critical_error_at` is not checked until this method returns. For example, with a one-second mempool timeout, the fixed two-second retry can then publish successfully and set `has_published`, permanently suppressing `MempoolDeadlineExpired`; the transaction completes despite missing the configured mempool deadline. Bound or check each retry against the send state's deadline

crates/utilities/tx-manager/src/manager.rs
xenoliss: Fixed in 8df37a3.

crates/utilities/tx-manager/src/manager.rs
xenoliss: Fixed in https://github.com/base/base/commit/8df37a3ab32d46df3d0fb0bbf58c2d47ad32c0aa.

crates/utilities/tx-manager/src/manager.rs
xenoliss: Fixed in https://github.com/base/base/commit/8df37a3ab32d46df3d0fb0bbf58c2d47ad32c0aa.

crates/utilities/tx-manager/src/manager.rs
xenoliss: Fixed in https://github.com/base/base/commit/8df37a3ab32d46df3d0fb0bbf58c2d47ad32c0aa.

crates/utilities/tx-manager/src/manager.rs
xenoliss: Already done in the configurable-retries commit on this PR:

- CLI: `--tx-manager.publish-max-retries` and `--tx-manager.publish-retry-delay` (plus `BATCHER_PUBLISH_*` env)
- default delay is `1s`, not `2s`

The batcher binary forwards both into `TxManagerConfig`.

refcell (approved):

PR #4660, opened 2026-08-25
docs(upgrade-signal): Document Unreconciled Schedule Shrinkage

refcell:
## Summary

Documents audit finding #5 (LOW) from the dynamic upgrades triage: the live upgrade-signal monitor only evaluates the signals present in each freshly read schedule, so an upgrade that disappears from a later, shorter schedule (for example after an L1 reorg on a non-finalized tag) is never reconciled and its already-committed activation stays in the runtime registry. This is a deliberate simplification rather than a bug, so the change adds explanatory doc comments on `schedule_needs_apply` and `poll_and_apply` in the upgrade-signal state module instead of altering behavior. The documentation captures why the case is unreachable on the default finalized (reorg-safe) read and why the impact is bounded on reorg-prone tags, since every non-zero activation is set at least MIN_NOTICE (1 hour) in the future while reorgs resolve and re-apply well before then. Tracking removed entries would require a full reorg-aware schedule diff for no benefit on the production path, so it is noted as possible future hardening.

### DIFF
--- a/crates/utilities/upgrade-signal/src/state.rs
+++ b/crates/utilities/upgrade-signal/src/state.rs
@@ -372,12 +372,15 @@ impl UpgradeSignalMonitor {
     ///   [`crate::UpgradeSignalBlockTag::Finalized`] read is reorg-safe, so a finalized-observed
     ///   entry can never disappear from a later read; the scenario is unreachable on the recommended
     ///   configuration.
-    /// * **On a reorg-prone tag the window is far shorter than the activation lead.** `safe` and
-    ///   `latest` can reorg, but every non-zero activation timestamp is set at least `MIN_NOTICE`
-    ///   (1 hour) in the future on L1, far longer than the shallow reorg depth and the poll cadence
-    ///   (12s at `latest`). Once the schedule change re-mines onto the canonical chain it is re-read
-    ///   and re-applied — and a restart or a manual [`UpgradeSignalRefresher`] refresh always
-    ///   reconciles from the full L1 schedule — well before the stale activation could take effect.
+    /// * **On a reorg-prone tag the exposure is bounded, but not eliminated.** `safe` and `latest`
+    ///   can reorg. If the reorged registration re-mines onto the canonical chain — the common
+    ///   case — it is re-read and re-applied well before it matters, since every non-zero activation
+    ///   timestamp is set at least `MIN_NOTICE` (1 hour) in the future on L1, far longer than the
+    ///   shallow reorg depth and the poll cadence (12s at `latest`). A reorg can instead *drop* the
+    ///   registration for good, in which case nothing triggers a re-read and the stale activation
+    ///   persists until a restart or a manual [`UpgradeSignalRefresher`] refresh reconciles from the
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/utilities/upgrade-signal/src/state.rs
```diff
@@ -348,6 +353,34 @@ impl UpgradeSignalMonitor {
...
+    /// * **A governance clear does not shrink the schedule.** `getSchedule` returns id-ordered
+    ///   timestamps that [`AlloyUpgradeSignalReader::map_schedule`] zips positionally onto the
+    ///   ladder, and a cleared upgrade keeps its slot with a `0` timestamp. That `0` reads as a
+    ///   [`UpgradeSignalStateUpdate::Changed`] and is re-applied normally. Only an L1 reorg that
+    ///   unwinds the registration of a trailing upgrade id shortens the array so an entry vanishes.
+    /// * **Production reads the finalized tag, which cannot reorg.** The default
+    ///   [`crate::UpgradeSignalBlockTag::Finalized`] read is reorg-safe, so a finalized-observed
+    ///   entry can never disappear from a later read; the scenario is unreachable on the recommended
+    ///   configuration.
+    /// * **On a reorg-prone tag the window is far shorter than the activation lead.** `safe` and
```
jackchuma: Could we soften the `MIN_NOTICE` claim? The one-hour delay does not guarantee the reorged transaction will return. If it is dropped, the monitor can retain the stale activation until a restart or manual refresh. The finalized default avoids this, but `safe` and `latest` do not

refcell: Softened it to call out that a reorg can drop the registration for good so the stale activation persists until a restart or manual refresh, and made clear the finalized default is the real guarantee rather than the one-hour MIN_NOTICE lead.

PR #4661, opened 2026-08-25
fix(runner): repair ExEx WAL id gaps before launch

jowparks:
reth treats any hole in the ExEx WAL `<id>.wal` sequence as fatal at startup, so one missing file boot-loops the node until the WAL is wiped — the mainnet shadow builder `notification 251460 not found` crash. Renumber the surviving files into a contiguous run just before reth loads the WAL, discarding nothing.

Holes recur rather than being one-off. Finalization picks files by the highest block each notification touches, which is not monotonic in the file id, so after a reorg a later low-height commit is deleted while the earlier, higher-reaching file stays. A swallowed unlink failure orphans a file the same way. The clean fix belongs upstream — `Storage::iter_notifications` hard-errors on an absent id even though `fill_block_cache` rebuilds its index from file contents and needs no dense range — but this unblocks mainnet today.

Note: this supersedes the fix discussed in Slack (matching `WalError::FileNotFound` at the ExEx `try_next()` call site). That variant is only ever constructed while `Wal::new` fills its block cache during launch, before any ExEx runs, so the run-loop branch would be unreachable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4996,6 +4996,7 @@ dependencies = [
  "reth-transaction-pool",
  "serde",
  "serde_json",
+ "tempfile",
  "tokio",
  "tokio-tungstenite 0.28.0",
  "tower 0.5.3",
--- a/crates/execution/runner/Cargo.toml
+++ b/crates/execution/runner/Cargo.toml
@@ -75,6 +75,7 @@ tracing-subscriber = { workspace = true, optional = true }
 [dev-dependencies]
 base-protocol.workspace = true
 base-node-runner = { path = ".", features = ["test-utils"] }
+tempfile.workspace = true
 serde_json.workspace = true
 tokio-tungstenite.workspace = true
 criterion = { workspace = true, features = ["async_tokio", "html_reports"] }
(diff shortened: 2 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

jowparks:
Will close in lieu of ignoring WAL error on boot instead of reording files.

PR #4663, opened 2026-08-25
fix(cli): Refresh Stale P2P Fork Filter

refcell:
## Summary

Fixes an audit finding where runtime L1 upgrade schedule updates left reth's P2P fork filter stale, partitioning running and restarted nodes at fork activation. Reth builds its ForkFilter once at startup and only advances its head, so a node that adopted a schedule at runtime kept advertising the fork id it cached at startup and rejected freshly restarted peers once the fork activated. This bumps the reth dependency to the upstream commit that adds a live `set_fork_filter` API, then rebuilds and reinstalls the runtime-aware fork filter on the running node whenever the schedule changes, so it advertises the fork id it now enforces. A regression test guards the reinstall behavior, and the reth dependency jump surfaced a small `SnapshotManifest` field addition that is handled at the affected call sites.

### DIFF
--- a/crates/execution/cli/Cargo.toml
+++ b/crates/execution/cli/Cargo.toml
@@ -67,6 +67,7 @@ tracing.workspace = true
 serde_json.workspace = true
 jsonrpsee = { workspace = true, features = ["server"] }
 reqwest = { workspace = true, features = ["json", "stream"] }
+alloy-consensus.workspace = true
 alloy-eips.workspace = true
 secp256k1.workspace = true
 tokio-stream.workspace = true
@@ -78,7 +79,9 @@ proptest = { workspace = true, optional = true }
 base-common-consensus = { workspace = true, features = ["reth"] }
 
 [dev-dependencies]
+alloy-genesis.workspace = true
 alloy-primitives.workspace = true
+reth-eth-wire-types.workspace = true
 base-common-genesis.workspace = true
 rstest.workspace = true
 tempfile.workspace = true
--- a/crates/execution/cli/Cargo.toml
+++ b/crates/execution/cli/Cargo.toml
@@ -28,6 +28,7 @@ reth-network-p2p.workspace = true
 reth-cli-commands.workspace = true
 reth-node-metrics.workspace = true
 reth-network-peers.workspace = true
+reth-ethereum-forks.workspace = true
 reth-rpc-server-types.workspace = true
 reth-tasks.workspace = true
 reth-db = { workspace = true, features = ["mdbx"] }
@@ -72,16 +73,14 @@ alloy-eips.workspace = true
 secp256k1.workspace = true
 tokio-stream.workspace = true
 clap = { workspace = true, features = ["derive", "env"] }
-tokio = { workspace = true, features = ["macros", "time"] }
+tokio = { workspace = true, features = ["macros", "time", "sync"] }
 
 # reth test-vectors
 proptest = { workspace = true, optional = true }
 base-common-consensus = { workspace = true, features = ["reth"] }
 
 [dev-dependencies]
-alloy-genesis.workspace = true
 alloy-primitives.workspace = true
-reth-eth-wire-types.workspace = true
 base-common-genesis.workspace = true
 rstest.workspace = true
 tempfile.workspace = true
--- a/crates/execution/cli/src/upgrade_signal.rs
+++ b/crates/execution/cli/src/upgrade_signal.rs
@@ -1,5 +1,7 @@
 //! Execution-node upgrade signal schedule application.
 
+use std::sync::Arc;
+
 use alloy_consensus::BlockHeader;
 use base_execution_chainspec::BaseChainSpec;
 use base_node_runner::{BaseNodeExtension, BaseRpcContext, FromExtensionConfig, NodeHooks};
@@ -16,7 +18,6 @@ use reth_network::{NetworkHandle, NetworkPrimitives};
 use reth_network_p2p::sync::NetworkSyncUpdater;
 use reth_provider::{BlockNumReader, HeaderProvider};
 use reth_rpc_server_types::RethRpcModule;
-use std::sync::Arc;
 use tokio::sync::Notify;
 use tracing::{info, warn};
 use url::Url;
--- a/crates/execution/cli/src/upgrade_signal.rs
+++ b/crates/execution/cli/src/upgrade_signal.rs
@@ -83,15 +83,18 @@ impl ExecutionUpgradeSignal {
     /// chain spec and installing it through [`RuntimeForkFilterNetwork::install_fork_filter`] keeps
     /// the node's advertised fork identity — across both the devp2p session layer and the discovery
     /// ENR — aligned with the rules it now enforces.
+    ///
+    /// Returns an error if the network install fails to refresh the advertised discovery entry, so
+    /// the caller can leave its baseline unadvanced and retry rather than record a stale install.
     pub fn install_runtime_fork_filter<Net: RuntimeForkFilterNetwork>(
         chain_spec: &BaseChainSpec,
         head: Head,
         network: &Net,
-    ) -> ForkId {
+    ) -> eyre::Result<ForkId> {
         let fork_filter = chain_spec.fork_filter(head);
         let fork_id = fork_filter.current();
-        network.install_fork_filter(fork_filter);
-        fork_id
+        network.install_fork_filter(fork_filter)?;
+        Ok(fork_id)
     }
 
     /// A fork id that folds in the entire runtime schedule, used as a change signal for runtime
@@ -113,9 +116,10 @@ impl ExecutionUpgradeSignal {
     /// schedule update lands via the auto-apply path or the manual `admin_refreshUpgradeSignal` RPC;
     /// both mutate the same runtime registry that [`Self::schedule_fork_id`] reads, so a single
     /// check covers both. `schedule_id` tracks the last installed schedule and is advanced only when
-    /// a new filter is installed, so the network message is sent once per change rather than every
-    /// block. A `None` `schedule_id` has no last-installed baseline yet and therefore always
-    /// installs, forcing the initial install after startup.
+    /// a new filter is installed successfully, so the network message is sent once per change rather
+    /// than every block, and a failed install leaves the baseline unadvanced so the next poll
+    /// retries instead of advertising a stale fork id. A `None` `schedule_id` has no last-installed
+    /// baseline yet and therefore always installs, forcing the initial install after startup.
     pub fn refresh_advertised_fork_filter<Net: RuntimeForkFilterNetwork>(
         chain_spec: &BaseChainSpec,
         head: Head,
@@ -127,9 +131,22 @@ impl ExecutionUpgradeSignal {
             return None;
         }
 
-        let fork_id = Self::install_runtime_fork_filter(chain_spec, head, network);
-        *schedule_id = Some(current);
-        Some(fork_id)
+        match Self::install_runtime_fork_filter(chain_spec, head, network) {
+            Ok(fork_id) => {
+                *schedule_id = Some(current);
+                Some(fork_id)
+            }
+            // Leave `schedule_id` unadvanced so the next poll retries; the advertised discovery fork
+            // id is still stale, so recording this schedule as installed would suppress that retry.
+            Err(error) => {
+                warn!(
+                    target: "upgrade_signal",
+                    error = %error,
+                    "failed to install runtime P2P fork filter; leaving schedule baseline unadvanced to retry on next poll"
+                );
+                None
+            }
+        }
     }
 
     /// Reconciles the advertised P2P fork filter with the current runtime schedule, reading the head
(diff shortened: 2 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -64,6 +67,83 @@ impl ExecutionUpgradeSignal {
...
+    /// chain spec and installing it through [`NetworkSyncUpdater::set_fork_filter`] keeps the node's
+    /// advertised fork identity aligned with the rules it now enforces.
+    pub fn install_runtime_fork_filter<Net: NetworkSyncUpdater>(
+        chain_spec: &BaseChainSpec,
+        head: Head,
+        network: &Net,
+    ) -> ForkId {
+        let fork_filter = chain_spec.fork_filter(head);
+        let fork_id = fork_filter.current();
+        network.set_fork_filter(fork_filter);
```
jackchuma: reth’s production path hard-codes the `eth` ENR key, but Base advertises under `opel`. Since discovery prefers the existing `opel` entry over the `eth` fallback, this updates devp2p while leaving Base’s advertised discovery fork ID stale. Please update the configured fork key and add a production-path regression; `RecordingNetwork` does not cover this integration

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -177,6 +264,12 @@ impl BaseNodeExtension for ExecutionUpgradeSignalRuntimeExtension {
                         interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
                         let mut signal = Box::pin(signal);
 
+                        // Baseline for detecting runtime schedule changes. Advanced only when the
+                        // fork filter is reinstalled, so a single check covers both the auto-apply
+                        // path and the manual admin RPC without re-installing on every block.
+                        let mut schedule_id =
+                            ExecutionUpgradeSignal::schedule_fork_id(chain_spec.as_ref());
```
jackchuma: `schedule_id` snapshots mutable registry state rather than the installed filter. RPC starts before this task, so an admin refresh can update the registry first; this then records the new schedule as already installed while the network still has the old filter, suppressing reconciliation until another schedule change. Outside that race, admin refresh can still lag by the 15-minute finalized poll interval. Please trigger immediate reconciliation after admin refresh and force an initial filter installation

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -205,6 +298,39 @@ impl BaseNodeExtension for ExecutionUpgradeSignalRuntimeExtension {
...
                                     }
+
+                                    // Adopt any runtime schedule change into the live P2P fork
+                                    // filter so this node advertises the fork id it now enforces
+                                    // instead of the one cached at startup. The cheap schedule-fork-id
+                                    // compare guards the provider read so an unchanged schedule never
+                                    // reads the head or logs a spurious head-read failure.
+                                    if ExecutionUpgradeSignal::schedule_fork_id(chain_spec.as_ref())
+                                        != schedule_id
+                                    {
```
jackchuma: Refresh the filter from the admin RPC. When `admin_refreshUpgradeSignal` applies a schedule between monitor ticks, this branch is not reached until the next L1 poll. With the default `Finalized` block tag, that can leave the node enforcing the new schedule while advertising the old ForkId for 15 minutes; a refresh that arrives before `schedule_id` is initialized can also be absorbed into the baseline and never trigger an install. Refresh the P2P filter directly from the successful admin-apply path, or notify this task immediately after that path commits

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -361,4 +487,111 @@ mod tests {
...
+    /// peers once the fork activated. The fix reinstalls the fork filter whenever the runtime
+    /// schedule changes. This exercises the exact per-poll routine the monitor loop runs and asserts
+    /// it (1) is a no-op while the schedule is unchanged, (2) installs a filter matching a freshly
+    /// restarted node the moment a fork is scheduled at runtime, and (3) is idempotent afterwards.
+    /// If the fix regresses, the node would keep the stale filter and these assertions fail.
+    #[test]
+    fn refresh_advertised_fork_filter_tracks_runtime_schedule_changes() {
+        use alloy_genesis::Genesis;
+        use base_execution_chainspec::BaseChainSpecBuilder;
+        use reth_chainspec::Chain;
```
jackchuma: Hoist test-only imports into the test module. Move these imports to the top of the `#[cfg(test)] mod tests` block. Keeping `use` items inside the test function violates the repository import-placement rule in `AGENTS.md:15`

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -64,6 +67,83 @@ impl ExecutionUpgradeSignal {
...
+    /// advertised fork identity aligned with the rules it now enforces.
+    pub fn install_runtime_fork_filter<Net: NetworkSyncUpdater>(
+        chain_spec: &BaseChainSpec,
+        head: Head,
+        network: &Net,
+    ) -> ForkId {
+        let fork_filter = chain_spec.fork_filter(head);
+        let fork_id = fork_filter.current();
+        network.set_fork_filter(fork_filter);
+        fork_id
```
jackchuma: `install_runtime_fork_filter` has a single caller and only forwards `fork_filter`, `current`, and `set_fork_filter`. Inlining those three statements into `refresh_advertised_fork_filter` keeps the full refresh operation together without losing a reusable boundary

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -361,4 +487,111 @@ mod tests {
...
+
+    impl RecordingNetwork {
+        fn installed_opt(&self) -> Option<reth_chainspec::ForkFilter> {
+            self.installed.lock().unwrap().clone()
+        }
+
+        fn installed(&self) -> reth_chainspec::ForkFilter {
+            self.installed_opt().expect("set_fork_filter was never called")
+        }
+    }
```
jackchuma, line 627: These two test-only accessors add an extra layer over a single `Mutex<Option<ForkFilter>>`, each with one caller. The test can lock `installed` directly for the no-op assertion and clone/`expect` it for the installation assertion

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -361,4 +487,111 @@ mod tests {
...
+
+        let chain_id = 9_100_100;
+        RuntimeUpgradeRegistry::clear_chain(chain_id);
+
+        let spec = BaseChainSpecBuilder::default()
+            .chain(Chain::from_id(chain_id))
+            .genesis(Genesis::default())
+            .with_fork(EthereumHardfork::Osaka, ForkCondition::Never)
+            .with_fork(BaseUpgrade::Azul, ForkCondition::Never)
+            .build();
```
jackchuma: This test builds a custom chain spec solely to choose a unique chain ID. It can reuse the existing `BaseChainSpec::devnet()` fixture, reset Osaka and Azul to `Never`, then read its chain ID. That also removes the new `alloy-genesis` dev dependency and lockfile churn

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -361,4 +487,111 @@ mod tests {
...
+
+    /// Regression test for the runtime fork-filter fix.
+    ///
+    /// Audit finding: a runtime schedule update left reth's P2P fork filter stale, so a running node
+    /// kept advertising the fork id it cached at startup and partitioned from freshly restarted
+    /// peers once the fork activated. The fix reinstalls the fork filter whenever the runtime
+    /// schedule changes. This exercises the exact per-poll routine the monitor loop runs and asserts
+    /// it (1) is a no-op while the schedule is unchanged, (2) installs a filter matching a freshly
+    /// restarted node the moment a fork is scheduled at runtime, and (3) is idempotent afterwards.
+    /// If the fix regresses, the node would keep the stale filter and these assertions fail.
```
jackchuma: This doc comment restates the regression, implementation, and each assertion already expressed by the test body. The test name plus its local comments are enough context here; removing this narrative keeps the regression test easier to scan

crates/execution/cli/src/upgrade_signal.rs
refcell: This rule is probably too strict tbh... this isn't a big deal.

crates/execution/cli/src/upgrade_signal.rs
refcell: This keeps it unit testable though and gives a canon place for installation. I'd prefer to keep this.

crates/execution/cli/src/upgrade_signal.rs
refcell: This _can lock_ assumption doesn't really matter since the only call site are tests.

crates/execution/cli/Cargo.toml
```diff
@@ -67,11 +68,12 @@ tracing.workspace = true
 serde_json.workspace = true
 jsonrpsee = { workspace = true, features = ["server"] }
 reqwest = { workspace = true, features = ["json", "stream"] }
+alloy-consensus.workspace = true
```
jackchuma, line 71: Nit: waterfall (not sure how important this codebase rule is either)

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -124,6 +270,49 @@ impl ExecutionUpgradeSignal {
...
+        if let Some(discv5) = self.discv5() {
+            // NB: reth's `encode_and_set_eip868_in_local_enr` double-encodes — it rlp-encodes the
+            // value, then the inner `enr_insert` rlp-encodes those bytes again, producing an entry
+            // `get_fork_id` cannot decode. Insert the typed entry directly so it is encoded exactly
+            // once, matching how the startup ENR (`add_value_rlp`) and discv4 (below) store it.
+            let opel = core::str::from_utf8(NetworkStackId::OPEL)
+                .expect("opel network stack id is valid utf-8");
+            discv5.with_discv5(|discv5| {
+                let _ = discv5.enr_insert(opel, &entry);
+            });
```
jackchuma, line 308: Handle `enr_insert` failure before advancing `schedule_id`. The write can fail, but its result is discarded and lines 125–132 still mark the schedule installed. Later polls then skip retries, leaving Base’s preferred `opel` fork ID stale until restart or another schedule change. Please propagate and log the failure, and advance the baseline only after `opel` succeeds

crates/execution/cli/Cargo.toml
refcell: Not important

refcell: Can also be a follow up, no need to block here

crates/execution/cli/src/upgrade_signal.rs
refcell: Fixed in a2b7e02ea. `install_fork_filter` now returns `eyre::Result<()>` and propagates the `enr_insert` failure; `refresh_advertised_fork_filter` advances `schedule_id` only on `Ok`, logging a `warn!` and leaving the baseline unadvanced on failure so the next poll retries. discv4 (infallible) is written first so it still advertises the new fork id even when the discv5 `opel` write short-circuits. Added `refresh_advertised_fork_filter_retries_after_failed_install` to cover the retry path.

crates/execution/cli/src/upgrade_signal.rs
```diff
@@ -124,6 +269,49 @@ impl ExecutionUpgradeSignal {
...
+        let entry = EnrForkIdEntry::from(fork_id);
+        if let Some(discv5) = self.discv5() {
+            // NB: reth's `encode_and_set_eip868_in_local_enr` double-encodes — it rlp-encodes the
+            // value, then the inner `enr_insert` rlp-encodes those bytes again, producing an entry
+            // `get_fork_id` cannot decode. Insert the typed entry directly so it is encoded exactly
+            // once, matching how the startup ENR (`add_value_rlp`) and discv4 (below) store it.
+            let opel = core::str::from_utf8(NetworkStackId::OPEL)
+                .expect("opel network stack id is valid utf-8");
+            discv5.with_discv5(|discv5| {
+                let _ = discv5.enr_insert(opel, &entry);
```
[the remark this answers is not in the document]
refcell: Addressed in a2b7e02ea. Rather than only logging, `install_fork_filter` now propagates the `enr_insert` failure so the schedule baseline (`schedule_id`) advances only after the `opel` write succeeds — a failure logs a `warn!` and leaves the baseline unadvanced so the next poll retries, closing the stale-ENR window this points at.

jackchuma (approved):

PR #4699, opened 2026-08-26
fix(eip8130): allow AuthorizeActor while locked when expiry outlives unlock

chunter-cb:
## Summary
- Mirror [base/eip-8130#90](https://github.com/base/eip-8130/pull/90): while locked, `AuthorizeActor` is permitted for a new add or expiry-only re-lease when the granted expiry outlives the unlock floor (`now + delay` hard-locked, `unlocksAt` pending unlock; `expiry == 0` always passes).
- Live actors keep frozen authenticator/scope/policy under lock; `RevokeActor` and delegation still revert `AccountIsLocked`; grants at or below the unlock floor revert `ExpiryDoesNotOutliveUnlock`.
- Move lock enforcement from the config authorizer into the apply path so batch signature auth still runs while per-op lock policy is enforced at apply time.

## Test plan
- [x] `cargo test -p base-execution-eip8130` (202 tests)
- [x] Hard-locked: expiry at floor → `ExpiryDoesNotOutliveUnlock`; above floor / unbounded → lands
- [x] Locked re-lease: expiry-only succeeds; scope rewrite → `AccountIsLocked`
- [x] Locked revoke → `AccountIsLocked`; expired actor id treated as new add
- [x] Config authorizer still validates locked batch signatures (apply rejects revoke)

### DIFF
--- a/crates/execution/eip8130/src/apply.rs
+++ b/crates/execution/eip8130/src/apply.rs
@@ -442,8 +442,11 @@ impl AccountChangeApplier {
                     // Applies one `AuthorizeActor` op: JIT expiry skip, locked-account
                     // policy, then `_authorizeActor`. Mirrors `Keystore._applyAuthorize`.
                     let (actor_id, config, policy_data) = Self::decode_authorize(&change.payload)?;
-                    // Replayable JIT path: drop an already-lapsed grant without reverting.
-                    if is_unsequenced && config.expiry != 0 && config.expiry <= now {
+                    // Replayable JIT path: drop an already-lapsed grant without
+                    // reverting. Uses the canonical `_isExpired` boundary (`now >
+                    // expiry`), so a grant with `expiry == now` is still live for
+                    // that second and installs rather than being dropped here.
+                    if is_unsequenced && config.is_expired(now) {
                         continue;
                     }
                     if locked {
@@ -595,7 +598,7 @@ impl AccountChangeApplier {
     /// Locked-account guard for `AuthorizeActor`. Mirrors `Keystore._applyAuthorize`'s
     /// `locked` block:
     ///
-    /// - **No live entry** ([`AccountConfigurationStorage::resolve_live_actor_config`]
+    /// - **No live entry** ([`AccountConfigurationStorage::resolve_live_actor_config_with_state`]
     ///   empty — unknown, revoked, disabled, or expired): a new add is allowed once
     ///   the grant's expiry outlives the unlock floor (`expiry == 0` always passes).
     /// - **Live entry** (slot populated and not expired): only an expiry rewrite is
@@ -615,7 +618,12 @@ impl AccountChangeApplier {
             return Err(ApplyError::ExpiryDoesNotOutliveUnlock);
         }
 
-        let current = storage.resolve_live_actor_config(account, actor_id, now)?;
+        // Resolve the current live config against the evolving in-batch `state`, not
+        // persisted storage: an earlier op's inline-self rewrite lands in `state`
+        // and is flushed only at batch end, so a storage read would miss it and let
+        // a second self identity/scope rewrite slip past the live-entry guard.
+        let current =
+            storage.resolve_live_actor_config_with_state(account, actor_id, state, now)?;
         if current.is_empty() {
             return Ok(());
         }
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/eip8130/src/apply.rs
```diff
@@ -423,20 +433,30 @@ impl AccountChangeApplier {
...
-                    // Skip an already-lapsed grant on the replayable JIT path
-                    // (per-change; live siblings still apply). A `0` clock disables
-                    // the skip via the `expiry != 0 && expiry <= 0` guard; both the
-                    // verifying and estimation pipelines pass the real block
-                    // timestamp so the skip is applied identically.
+                    // Replayable JIT path: drop an already-lapsed grant without reverting.
                     if is_unsequenced && config.expiry != 0 && config.expiry <= now {
                         continue;
                     }
+                    if locked {
```
xenoliss, line 428: This adds a moving deadline (`expiry > now + delay`) that `WatchManifest` does not track. A sequenced tx can become invalid without any watched state change and be retried on subsequent builds.

crates/execution/eip8130/src/apply.rs
```diff
@@ -569,6 +592,45 @@ impl AccountChangeApplier {
...
+        policy_data: &[u8],
+        state: &AccountState,
+        now: u64,
+    ) -> Result<(), ApplyError> {
+        let unlock_floor = state.unlock_floor(now);
+        if config.expiry != 0 && config.expiry <= unlock_floor {
+            return Err(ApplyError::ExpiryDoesNotOutliveUnlock);
+        }
+
+        let current = storage.resolve_live_actor_config(account, actor_id, now)?;
```
xenoliss: This reads persisted storage and misses inline-self updates held in `state` from earlier ops in the same batch. It should resolve self against the evolving `AccountState`; otherwise a second identity/scope rewrite bypasses `AccountIsLocked`.

crates/execution/eip8130/src/apply.rs
```diff
@@ -423,20 +433,30 @@ impl AccountChangeApplier {
...
+                    // Applies one `AuthorizeActor` op: JIT expiry skip, locked-account
+                    // policy, then `_authorizeActor`. Mirrors `Keystore._applyAuthorize`.
                     let (actor_id, config, policy_data) = Self::decode_authorize(&change.payload)?;
-                    // Skip an already-lapsed grant on the replayable JIT path
-                    // (per-change; live siblings still apply). A `0` clock disables
-                    // the skip via the `expiry != 0 && expiry <= 0` guard; both the
-                    // verifying and estimation pipelines pass the real block
-                    // timestamp so the skip is applied identically.
+                    // Replayable JIT path: drop an already-lapsed grant without reverting.
                     if is_unsequenced && config.expiry != 0 && config.expiry <= now {
```
xenoliss: This differs from Solidity’s `_isExpired`, which uses `now > expiry.` I think we should align and use `< now` for parity.

chunter-cb: Good catch — aligned with `_isExpired`. Fixed in c56387b1b: the JIT skip now uses a shared `ActorConfig::is_expired` (`expiry != 0 && now > expiry`), so a grant with `expiry == now` is still live for that second and installs rather than being dropped. `resolve_live_actor_config` uses the same helper for consistency, and the boundary test (`apply_skips_lapsed_jit_grant`) was updated to assert the `expiry == now` grant lands.

crates/execution/eip8130/src/apply.rs
chunter-cb: Fixed in c56387b1b. Added `resolve_live_actor_config_with_state`, which resolves the inline self from the evolving in-batch `AccountState` (the explicit `actor_config` and policy slots are written eagerly, so those are still read from storage). The locked guard now uses it, so an add-then-mutate of the inline self within a single locked batch is correctly caught as a live-entry identity change instead of reading as a fresh add. Added `locked_self_release_uses_evolving_state_within_batch`, which fails against the previous storage-based read.

crates/execution/eip8130/src/apply.rs
chunter-cb: Agreed this is real, and it's a consequence of moving lock enforcement to apply-time — the `now`-relative predicate no longer flows into the `WatchManifest` the authorizer builds. A few notes on scope, and a proposed fix:

- **Consensus is unaffected.** At inclusion the applier re-runs with the block's real `now`, so an op that has crossed the deadline reverts `ExpiryDoesNotOutliveUnlock` deterministically. This is purely a mempool-efficiency issue: the tx can linger and be retried across builds instead of being dropped when it lapses.
- **Only hard locks move.** The deadline exists only while the account is hard-locked, where `unlock_floor = now + delay`. A pending unlock uses a fixed `unlocks_at`, so `expiry <= unlocks_at` is time-independent and doesn't drift.
- **Bounded expiry only.** `expiry == 0` always passes, so only bounded grants have a deadline: the op is valid while `now < expiry - delay`.

Proposed fix: fold that per-op deadline (`min(expiry - delay)` across the batch's hard-locked AuthorizeActor ops) into the manifest's `effective_expiry`, the same way sender/payer actor expiries already seed `manifest_expiry` / `InvalidationKey::expiry_bucket` in the validator — so the pool evicts the tx at the boundary rather than on a watched state change. That spans the applier → validator → manifest boundary and deserves its own tests, so I'd prefer to do it as a focused follow-up rather than expand this PR (happy to land it here instead if you'd rather — let me know). I'll open a follow-up issue/PR either way.

xenoliss (approved):

PR #4180, opened 2026-07-27
feat(bench): run load tests on snapshot-backed devnets

meyer9:
## Summary

Adds an L1-free, snapshot-backed Base devnet and benchmark workflow that continues from a real Base mainnet Reth snapshot using production execution and consensus components.

The branch now covers four layers:

- snapshot-backed execution and standalone consensus;
- interactive devnet and managed benchmark launchers;
- adaptive load generation, including fresh-account state-growth workloads;
- benchmark/report output with sequencer, validator, Prometheus, and Engine API latency metrics.

`base/base` owns network launch, workload execution, and result production. The separate `base/benchmark` repository can consume the emitted report bundle without owning node orchestration.

## Topology and execution model

```text
snapshot builder EL <-> standalone L1-free sequencer CL
        |
        | unsafe blocks
        v
 follow-mode CL      <-> snapshot client EL
```

Both ELs open distinct writable copies of the same Base mainnet Reth datadir. Transactions enter through the builder's normal `eth_sendRawTransaction` and transaction-pool path. The sequencer builds real descendants of the snapshot head and the follow-mode client canonicalizes the same blocks.

Interactive `base-devnet snapshot` runs both roles concurrently. `base-bench snapshot` runs them sequentially: it measures the sequencer range first, stops sequencing while retaining the builder RPC as the canonical source, then starts the validator and replays that exact range. This isolates process-global metrics by role and requires sequencer/validator canonical hash equality.

This is intentionally an unsafe-chain benchmark environment. It does not model L1 derivation, deposits after startup, batching, L1 reorgs, or safe/finalized-head advancement. Snapshot descendants are not intended to be a restartable continuation of mainnet. Snapshot mode uses the standard payload service at both 2s and 200ms and does not publish or consume Flashblocks.

### Snapshot-backed execution and consensus

- opens caller-owned existing Reth datadirs directly for separate builder and follow-client ELs;
- adds a standalone, L1-free sequencer built from the production consensus engine/sequencer actors;
- drives the client EL with the production follow-mode CL;
- validates chain ID 8453, the boundary L1-info transaction, `SystemConfig`, sequence number, and optional expected head number/hash/timestamp;
- supports deterministic 2s and 200ms full-block cadences, including `BaseTime` activation for subsecond descendants;
- supports configurable block gas limits and persistence thresholds;
- uses EIP-1559 elasticity/denominator 1 so full benchmark blocks equal the gas target and do not strand a pre-signed queue through base-fee movement;
[...]

PR #4706, opened 2026-08-26
feat(consensus): L1-free standalone sequencer

meyer9:
## Summary

Adds an **L1-free standalone sequencer** to the consensus engine: a standalone attributes builder and origin selector that drive block production without an L1 chain. Sub-second (200ms) blocks are gated on the Denim upgrade via `BaseTimeUpdateTx` deposits.

This is **PR 1 of 3** splitting the combined PR #4180 into focused, independently-reviewable pieces.

## Stack

- **1/3 → this PR** · `base: main` · L1-free standalone sequencer (consensus)
- 2/3 · load-test measurement-blocks window
- 3/3 · snapshot-backed L1-free devnet + benchmark harness

## Changes

Two crates (4 files, +635/-4):

- **`base-consensus-node`** — `crates/consensus/service/src/standalone.rs` (new, +496): the standalone builder/origin-selector/derivation-client that produce L2 blocks without an L1.
- Plus the wiring/behavior changes to existing code described below.

### Changes to existing code (beyond `standalone.rs`)

**1. `crates/consensus/protocol/src/info/variant.rs` (base-protocol, derivation layer, +120/-4)**

- **Extracted `L1BlockInfoTx::into_deposit_tx(self, rollup_config, l2_block_time) -> Sealed<TxDeposit>`.** The logic that builds the first-in-block L1-info deposit transaction was previously inlined in the constructor that returns `(L1BlockInfoTx, Sealed<TxDeposit>)`; it is now a reusable public method, and the original path just calls `l1_info.into_deposit_tx(...)`. This lets the standalone sequencer mint the L1-info deposit for blocks it produces itself, without going through L1 derivation. Pure refactor — no behavior change on the existing derivation path.
- **New `L1BlockInfoTx::with_sequence_number(self, u64) -> Self`.** Returns the same L1 block info with a different L2 `sequence_number`, preserving every L1-origin, fee, and system-config field across all four hardfork variants (Bedrock/Ecotone/Isthmus/Jovian). L1-free sequencing keeps producing L2 blocks against a single synthetic origin, incrementing the sequence number each block; this is the helper that does that.
- **New `L1BlockInfoTx::time(&self) -> u64` enum dispatcher.** Adds a `time()` accessor alongside the existing per-variant dispatchers (`block_hash()`, `sequence_number()`, `id()`, …) so the origin selector reads `l1_info.time()` instead of hand-matching each variant. Future-proofs against new hardfork variants.
- Adds the `jovian::L1BlockInfoJovianBaseFields` import and a colocated unit test (`test_l1_block_info_with_sequence_number`) asserting the seq-number is updated while origin/batcher/fee fields are preserved.

**2. `crates/consensus/service/src/actors/engine/engine_request_processor.rs` (base-consensus-node, +22)**

- Adds a `skip_reset: bool` field to `EngineProcessor` and a `new_skip_reset(...)` constructor that flips it on; the existing `new_with_checkpoint` initializes it to `false`, so **default behavior is unchanged.**
- **Behavior change in `reset_engine_state()`:** when `skip_reset` is set and the engine already has a bootstrapped unsafe head (non-default), the reset short-circuits and returns that in-memory head instead of consulting the checkpoint reader / reconstructing an L1-derived safe head. Without an L1 there is no L1-info deposit to rebuild the safe head from, so this retains the bootstrapped L2 head rather than rewinding. Guarded entirely behind the `skip_reset` flag.

**3. `crates/consensus/service/src/lib.rs` (base-consensus-node, +6)**

- Declares the new `standalone` module and `pub use`-re-exports its public API (`StandaloneAttributesBuilder`, `StandaloneDerivationClient`, `StandaloneOriginSelector`, `StandalonePrefund`, `StandaloneSequencerNode`, `StandaloneUnsafePayloadGossipClient`). Wiring only.

## Verification

- `cargo clippy -p base-protocol -p base-consensus-node --all-targets --all-features -- -D warnings` → clean
- `cargo nextest run -p base-protocol -p base-consensus-node` → **554 passed, 0 skipped**
- Leaves `Cargo.lock` byte-identical to `main`

Split from #4180 (kept open for reference). Two intentional improvements over #4180: (1) the engine-reset flag/constructor are named `skip_reset`/`new_skip_reset` here (was `l1_free`/`new_l1_free` in #4180) to describe what the flag does rather than why it exists; (2) the origin selector uses a new `L1BlockInfoTx::time()` enum dispatcher instead of hand-matching each hardfork variant.

### DIFF
--- a/crates/consensus/protocol/src/info/variant.rs
+++ b/crates/consensus/protocol/src/info/variant.rs
@@ -16,7 +16,7 @@ use crate::{
     info::{
         L1BlockInfoBedrockBaseFields, L1BlockInfoEcotoneBaseFields as _, L1BlockInfoJovian,
         bedrock::L1BlockInfoBedrockOnlyFields as _, ecotone::L1BlockInfoEcotoneOnlyFields as _,
-        isthmus::L1BlockInfoIsthmusBaseFields as _,
+        isthmus::L1BlockInfoIsthmusBaseFields as _, jovian::L1BlockInfoJovianBaseFields as _,
     },
 };
 
@@ -202,8 +202,19 @@ impl L1BlockInfoTx {
             l2_block_time,
         )?;
 
+        let deposit_tx = l1_info.into_deposit_tx(rollup_config, l2_block_time);
+        Ok((l1_info, deposit_tx))
+    }
+
+    /// Converts this L1 block info into the deposit transaction placed first in an L2 block.
+    pub fn into_deposit_tx(
+        self,
+        rollup_config: &RollupConfig,
+        l2_block_time: u64,
+    ) -> Sealed<TxDeposit> {
+        let sequence_number = self.sequence_number();
         let source = DepositSourceDomain::L1Info(L1InfoDepositSource {
-            l1_block_hash: l1_info.block_hash(),
+            l1_block_hash: self.block_hash(),
             seq_number: sequence_number,
         });
 
@@ -215,7 +226,7 @@ impl L1BlockInfoTx {
             value: U256::ZERO,
             gas_limit: 150_000_000,
             is_system_transaction: true,
-            input: l1_info.encode_calldata(),
+            input: self.encode_calldata(),
         };
 
         // With the regolith upgrade, system transactions were deprecated, and we allocate
@@ -225,7 +236,7 @@ impl L1BlockInfoTx {
             deposit_tx.gas_limit = REGOLITH_SYSTEM_TX_GAS;
         }
 
-        Ok((l1_info, deposit_tx.seal_slow()))
+        deposit_tx.seal_slow()
     }
 
     /// Decodes the [`L1BlockInfoTx`] object from Ethereum transaction calldata.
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/2</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>
<details>
<summary><h2>⚠️ Ignored Reviews (2)</h2></summary>
<table>
<tr>
<th>Reviewer</th>
<th>Reason</th>
</tr>
<tr>
<td>0x00101010</td>
<td>Old or outdated review</td>
</tr>
<tr>
<td>0x00101010</td>
<td>Review missing MFA</td>
</tr>
</table>
</details>

crates/consensus/protocol/src/info/variant.rs
```diff
@@ -225,7 +236,7 @@ impl L1BlockInfoTx {
             deposit_tx.gas_limit = REGOLITH_SYSTEM_TX_GAS;
         }
 
-        Ok((l1_info, deposit_tx.seal_slow()))
+        deposit_tx.seal_slow()
```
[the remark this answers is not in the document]
meyer9: existing naming

crates/consensus/protocol/src/info/variant.rs
```diff
@@ -202,8 +202,19 @@ impl L1BlockInfoTx {
...
+    /// Converts this L1 block info into the deposit transaction placed first in an L2 block.
+    pub fn into_deposit_tx(
+        self,
+        rollup_config: &RollupConfig,
+        l2_block_time: u64,
+    ) -> Sealed<TxDeposit> {
+        let sequence_number = self.sequence_number();
         let source = DepositSourceDomain::L1Info(L1InfoDepositSource {
-            l1_block_hash: l1_info.block_hash(),
+            l1_block_hash: self.block_hash(),
```
meyer9, line 217: kinda confusing, but try_new_with_deposit_tx is static and into_deposit_tx receives the struct, so calling `self.` is matching previous behavior.

crates/consensus/service/src/standalone.rs
```diff
@@ -0,0 +1,489 @@
...
+                parent_beacon_block_root: self
+                    .rollup_config
+                    .is_ecotone_active(next_l2_time)
+                    .then_some(B256::ZERO),
+                slot_number: None,
+                target_gas_limit: None,
+            },
+            transactions: Some(transactions),
+            no_tx_pool: Some(false),
+            gas_limit: Some(self.system_config.gas_limit),
```
0x00101010, line 139: The standalone builder keeps the snapshot’s pre-Denim gas limit and EIP-1559 denominator. Production divides gas limit by 10 and multiplies the denominator by 10 at Denim activation. Without this, 200 ms benchmarks have 10× intended gas throughput and base-fee responsiveness. Persist the scaled config across subsequent descendants and test both activation and later blocks.

0x00101010 (changes requested):
@meyer9

One high level thought I just had is that we have
- base-builder
- base sequencer (although aimed to replace base-builder I guess)
- action harness
- now this L1-free sequencer

It feels like we need a way to make sure their logic does not diverge with each other, especially with all the block building changes we're making recently

crates/consensus/service/src/standalone.rs
meyer9: On this, elasticity is set to 1 anyway, so there are no fee increases. This won't have to sync real blocks and we'll never transition to Denim inside of this - it will either be active or not. We override gas limit and EIP-1559 params before the test starts, so it doesn't use the snapshot's params.

0x00101010 (approved):

haardikk21 (approved):

PR #4708, opened 2026-08-26
feat(devnet): snapshot-backed L1-free devnet + benchmark harness

meyer9:
## Summary

Adds the **etc/systems snapshot devnet**: an in-process standalone-consensus L2 stack booted from a chain snapshot, plus benchmark and devnet CLIs (`base-bench`, `base-devnet`), Prometheus metrics, and a snapshot devnet integration test.

This is **PR 3 of 3** (top of stack) splitting the combined PR #4180 into focused, independently-reviewable pieces.

## Stack

- 1/3 · L1-free standalone sequencer (consensus) — **merged**
- 2/3 · load-test measurement-blocks window — **merged**
- **3/3 → this PR** · `base: main` · snapshot devnet + benchmark harness

> Originally depended on both lower PRs; now that they're merged into `main`, this is a self-contained `etc/systems`-only change.

## Changes

`base-system-tests` (20 files, +3042/-183): `etc/systems/src/{benchmark_cli.rs, devnet_cli.rs, prometheus_metrics.rs, lib.rs, smoke.rs, system_config.rs, Cargo.toml, README.md}`, `src/bin/{base_bench.rs, base_devnet.rs}` (new `[[bin]]` targets), 8 `src/l2/*` files, `tests/snapshot_devnet.rs`, and a 2-line `Cargo.lock` delta (`base-common-chains`, `reth-ethereum-forks`).

## Verification

- `cargo clippy -p base-system-tests --all-targets --all-features -- -D warnings` → clean
- `cargo +nightly fmt --all -- --check` → clean · `cargo metadata --locked` → OK
- **Lossless-split proof:** the tip tree of this stack is byte-identical to #4180's squashed commit (`git diff <#4180-tree> HEAD` is empty)

## Future: unifying with the Docker devnet (`vibenet`)

> Design note — **not implemented in this PR.** Captured here so the snapshot devnet and the existing Docker devnet converge instead of drifting into two parallel worlds.

Today there are two local "devnets" with complementary strengths:

| | **Docker devnet** (`just devnet up`, aka `vibenet`) | **`base-devnet` snapshot** (this PR) |
|---|---|---|
| Topology | Multi-container: L1 (Reth + Lighthouse) + L2 builder/client/batcher + bootnode | Single in-process Rust binary; **L1-free** unsafe chain |
| State | Genesis-built devnet chain (`L2_CHAIN_ID=84538453`) | Writable clones of a **Base mainnet** snapshot (`chain_id=8453`) |
| Startup | Full genesis generation (minutes) | Boot from snapshot (seconds) |
| Builder RPC | `7545` (`L2_BUILDER_HTTP_PORT`) | `7545` with `--stable-ports` |
| Client RPC | `8545` (`L2_CLIENT_HTTP_PORT`) | `8545` with `--stable-ports` |
| Best for | Full-stack integration (L1, derivation, batcher, HA conductor) | Fast iteration + benchmarking on realistic mainnet state |

The key observation: **the endpoint contract is already the same** — both bind builder `7545` / client `8545`, and `basectl` already treats `vibenet` as a devnet alias (`DEVNET_ALIASES = ["devnet", "vibenet", "local-devnet", "local-vibenet"]`). That makes convergence mostly a matter of wiring, not rewrites. A concrete path:

1. **One entrypoint, snapshot as a mode.** Add `just devnet up mode=snapshot` (alongside the existing `mode=zenith`) that launches `base-devnet snapshot --stable-ports` instead of `docker compose up`. Same `just devnet down/status/logs/smoke/accounts` recipes, driven off the snapshot's `runtime.json` (which already emits `builder_rpc_url`, `builder_flashblocks_url`, `client_rpc_url`, `chain_id`, `block_interval_ms`, and the boundary head).

2. **Share config, don't duplicate it.** Have `base-devnet` read the same knobs `etc/docker/devnet-env` already defines — ports, prefund address/amount, block cadence (200ms/2s), and the hardfork activation schedule — so a single source of truth drives both stacks. `--prefund-*`, `--block-interval`, and `--stable-ports` already map cleanly onto the `devnet-env` values.

3. **Make `basectl` snapshot-aware.** Since `vibenet` is already a recognized alias, teach `basectl`'s devnet views (status, `upgrade-signal`, accounts) to discover the snapshot stack via `runtime.json`, so the same TUI works against both. Reconcile the chain-id difference (`84538453` genesis devnet vs `8453` mainnet snapshot) by keying selection off `runtime.json` rather than a hardcoded id.

4. **Reuse observability.** The Docker stack's Prometheus/Grafana/Jaeger services can scrape `base-devnet`'s existing Prometheus metrics endpoints, so the same dashboards light up for snapshot runs — no new monitoring stack required.

5. **Optional L1 for full-stack flows.** `base-devnet` is intentionally L1-free (no derivation/batching/safe-head). For the rare flows that need L1, an opt-in Anvil L1 (mirroring `docker-compose.anvil-l1.yml`) could be attached, keeping the L1-free path as the fast default.

Net effect: `vibenet` becomes the umbrella, with the Docker stack and the in-process snapshot stack as two interchangeable backends behind one CLI, one config file, and one set of dashboards — pick full-stack fidelity or snapshot-fast iteration without learning a second toolchain.

---

Split from #4180 (kept open for reference).


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

etc/systems/src/l2/in_process_builder.rs
```diff
@@ -40,6 +40,10 @@ use crate::{config::BUILDER, setup::BUILDER_ENODE_ID};
 pub struct InProcessBuilderConfig {
     /// L2 genesis JSON content.
     pub genesis_json: Vec<u8>,
+    /// Optional pre-built chain specification, used instead of `genesis_json` when provided.
+    pub chain_spec: Option<Arc<BaseChainSpec>>,
```
meyer9: can this be unified with genesis_json so only chain_spec is provided, but it's read from a genesis_json when applicable?

etc/systems/src/l2/in_process_builder.rs
```diff
@@ -61,6 +67,18 @@ pub struct InProcessBuilderConfig {
     /// Lets downstream consumers layer their own [`BaseNodeExtension`] onto the standard
     /// in-process builder wiring without forking this crate.
     pub extra_extensions: Vec<Box<dyn BaseNodeExtension>>,
+    /// Interval used by the payload builder.
+    pub block_time: Duration,
+    /// Whether to replace the standard payload service with the Flashblocks builder.
+    pub flashblocks_enabled: bool,
```
meyer9: is it possible to use the new MultiplexServiceBuilder?

etc/systems/src/l2/in_process_standalone_consensus.rs
```diff
@@ -0,0 +1,109 @@
+//! In-process L1-free consensus for snapshot-backed development networks.
```
meyer9: How is this different than the existing L1 free sequencer? Is this for validators? The name should reflect that.

etc/systems/src/l2/snapshot_stack.rs
```diff
@@ -0,0 +1,403 @@
...
+    builder: InProcessBuilder,
+    boundary: SnapshotBoundary,
+    standalone_consensus: Option<InProcessStandaloneConsensus>,
+    client: InProcessClient,
+    follow_consensus: Option<InProcessFollowConsensus>,
+    follow_config: Option<InProcessFollowConsensusConfig>,
+    block_interval: DevnetBlockInterval,
+}
+
+impl SnapshotL2Stack {
```
meyer9, line 49: Is there a common trait we can extract between this and the existing L2 stack?

etc/systems/src/benchmark_cli.rs
```diff
@@ -0,0 +1,690 @@
...
+    pub benchmark_run: String,
+    /// Unique run identifier; defaults to the visualizer output directory name.
+    #[arg(long)]
+    pub run_id: Option<String>,
+    /// Stable build identifier for visualizer comparisons.
+    #[arg(long, env = "BASE_BENCH_CLIENT_VERSION")]
+    pub client_version: Option<String>,
+    /// Optional benchmark funder key; generated ephemerally when omitted.
+    #[arg(long, env = "FUNDER_KEY", hide_env_values = true)]
+    pub funder_key: Option<PrivateKeySigner>,
```
meyer9: This should be generated on startup always as a deposit tx.

etc/systems/src/benchmark_cli.rs
```diff
@@ -0,0 +1,690 @@
...
+    /// Wei minted to the funder in the first local descendant.
+    #[arg(long, default_value_t = 1_000_000_000_000_000_000_000_u128)]
+    pub prefund_amount: u128,
```
meyer9: Just use a large amount of ETH maybe 1000 ETH

etc/systems/src/benchmark_cli.rs
```diff
@@ -0,0 +1,690 @@
...
+    pub block_interval: DevnetBlockInterval,
+    /// Load-test YAML. RPC and Flashblocks URLs are replaced with launched endpoints.
+    #[arg(long)]
+    pub load_test_config: PathBuf,
+    /// JSON result destination.
+    #[arg(long, default_value = "benchmark-result.json")]
+    pub output: PathBuf,
+    /// Optional directory for a base/benchmark visualizer-compatible result bundle.
+    #[arg(long)]
+    pub visualizer_output_dir: Option<PathBuf>,
```
meyer9: only have one output dir with all of the results, always required to be specified

etc/systems/src/benchmark_cli.rs
```diff
@@ -0,0 +1,690 @@
...
+pub struct SnapshotBenchmarkArgs {
+    /// Writable builder clone of a Base mainnet datadir.
+    #[arg(long, env = "BASE_SNAPSHOT_BUILDER_DATADIR")]
+    pub builder_datadir: PathBuf,
+    /// Writable client clone of the same Base mainnet datadir.
+    #[arg(long, env = "BASE_SNAPSHOT_CLIENT_DATADIR")]
+    pub client_datadir: PathBuf,
+    /// Interval between locally produced blocks.
+    #[arg(long, value_enum, default_value_t)]
+    pub block_interval: DevnetBlockInterval,
```
meyer9: Is there not a config to specify this?

etc/systems/src/l2/in_process_builder.rs
meyer9: Done — removed `genesis_json` from `InProcessBuilderConfig` and made `chain_spec: Arc<BaseChainSpec>` required. Added `InProcessBuilderConfig::chain_spec_from_genesis_json(&[u8])` so callers parse genesis at the boundary; `snapshot_stack`, `stack`, and `shadow_sequencer` were updated accordingly.

etc/systems/src/l2/in_process_builder.rs
meyer9: Done — collapsed the two payload arms into a single `MultiplexingServiceBuilder::new(builder_config).with_cutover_enabled(config.payload_builder_cutover)`. Confirmed cutover-disabled is equivalent to the old `FlashblocksServiceBuilder` path (`service_builder.rs` short-circuits to `FlashblocksServiceBuilder` when `!cutover_enabled && !basic_only`), so `flashblocks_enabled` was effectively a no-op and has been removed.

etc/systems/src/l2/in_process_standalone_consensus.rs
meyer9: Done — renamed `InProcessStandaloneConsensus`/`Config` to `InProcessStandaloneSequencer`/`Config` (it's a thin in-process runner for the standalone sequencer, not a consensus type). Updated the module doc, Debug label, and re-exports.

etc/systems/src/l2/snapshot_stack.rs
meyer9: Good question. A trait can be extracted (both stacks expose shutdown + rpc-url accessors), but there's no polymorphic consumer today: the benchmark path uses `SnapshotL2Stack` concretely and production uses `L2Stack` concretely, and inherent methods already cover both. Adding the trait now would be hollow indirection duplicating those inherent methods, so I've left it out and will extract it once a real generic use appears. (I prototyped an `L2StackLifecycle` trait and then dropped it for this reason.)

etc/systems/src/benchmark_cli.rs
meyer9: Done — the funder is now always an ephemeral `PrivateKeySigner::random()`, prefunded via deposit (`DevnetPrefund { address, amount }`). The `--funder-key`/`FUNDER_KEY` input is removed.

etc/systems/src/benchmark_cli.rs
meyer9: Done — dropped `--prefund-amount` and hardcoded `const PREFUND_AMOUNT_WEI: u128 = 1_000_000_000_000_000_000_000;` (1000 ETH).

etc/systems/src/benchmark_cli.rs
meyer9: Done — replaced `--output` and `--visualizer-output-dir` with a single required `--output-dir`. `benchmark-result.json` and the `visualizer/` bundle are written underneath it; `run_id` defaults to `<benchmark-run>-<timestamp>` unless `--run-id` is set.

etc/systems/src/benchmark_cli.rs
meyer9: Done — removed the standalone `--block-interval` flag; it's now derived from the load-test config's `block_time` (`block_interval_from_test_config`), mapping to `TwoSeconds`/`TwoHundredMilliseconds` and erroring on unsupported values.

PR #4730, opened 2026-08-27
fix(precompiles): aliased announce ABI via borrowed decode

rayyan224:
## Summary

`announce(bytes[] internalCalls, string, string, string)` copies each `bytes` into an owned `Vec`. If many array slots point at the same 64 KB blob, one call allocates ~64 MB (twice on V1). That showed up as ~94 s of executor time and ~98 GiB of allocations on a single protocol-capped transaction.

`announce` is the only B-20 call whose array has dynamically sized elements. Other arrays (`address[]`, `uint256[]`, `PausableFeature[]`, `uint64[]`) cannot share a tail this way, so the fix is announce-only.

Decode `announce` as slices into calldata instead of copying. Validation still uses the same `decode_sequence` + `type_check` as the owned path, so the accept set is unchanged. Malformed payloads fall through to the existing decoder and keep the same error bytes. No gas or fork change, so this also covers shipped Beryl.

Left out on purpose: removing the V1 double-decode, and charging extra gas on Cobalt. Neither is needed once the copies are gone.

Linear: [BOP-603](https://linear.app/coinbase/issue/BOP-603/cantina-16-aliased-owned-abi-decoding-is-performed-twice-during)

## Test plan

- [ ] `cargo test -p base-precompiles`
- [ ] Aliased-tail `announce` succeeds and re-dispatches each entry
- [ ] Nested `announce` still reverts `AnnouncementInProgress`
- [ ] Malformed offsets and non-UTF-8 `id` match the owned decoder's error bytes on V1 and V2
- [ ] Empty and large (non-aliased) arrays still work
- [ ] Borrowed vs owned accept-set stays identical


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/precompiles/src/b20_asset/dispatch.rs
```diff
@@ -475,17 +532,110 @@ impl<S: AssetAccounting, A: PolicyAccounting> B20AssetToken<S, A> {
...
     }
 }
 
+/// The version-agnostic view of a decoded `announce` that [`B20AssetToken::run_announce`] needs.
+///
+/// Abstracts over the two ways `announce` calldata is decoded — the borrowed fast-path token
+/// ([`BorrowedAnnounce`], which never copies the aliased `bytes[]`) and the owned
+/// [`IB20Asset::announceCall`] safety net — so the runner never has to know which produced it.
+/// Adding another representation later is a new impl, not a change to `run_announce`.
+pub(crate) trait AnnounceCall {
```
rayyan224: We should spilt this out, it's a single entity for announcement

rayyan224: Done in d4b8a1ec4 — extracted the announce entity (`AnnounceCall` trait, `BorrowedAnnounce` + its `decode`, and both impls) into a new `b20_asset/announce.rs`, re-exported like `AssetCall`. Kept `run_announce` and the `route` interception in `dispatch.rs` since the internal-call loop is a routing responsibility per the module doc. Pure move — all 708 tests pass unchanged.

crates/common/precompiles/src/policy/versions.rs
```diff
@@ -91,6 +91,12 @@ impl PolicyAbi {
     /// Where the frozen surface differs from canonical (`V1`), the calldata is validated against the
     /// frozen surface first so a rejection carries that version's consensus error bytes, then
     /// re-decoded into the canonical enum for routing.
+    ///
+    /// This owned decode (twice at `V1`) needs no borrowed fast path like `announce`'s (Cantina #16):
```
rayyan224: Not needed comment

crates/common/precompiles/src/common/abi/b20_abi.rs
```diff
@@ -45,6 +45,12 @@ impl B20Abi {
     ///
     /// The frozen surface decides what is dialable and owns any error bytes; the canonical surface
     /// then produces the value the dispatcher matches on.
+    ///
+    /// This owned decode (twice at `V1`) needs no borrowed fast path like `announce`'s (Cantina #16):
```
rayyan224: Not needed comment

crates/common/precompiles/src/b20_asset/versions.rs
```diff
@@ -104,6 +104,13 @@ impl AssetAbi {
     /// Where the frozen surface differs from canonical (`V1`), the calldata is validated against the
     /// frozen surface first so a rejection carries that version's consensus error bytes, then
     /// re-decoded into the canonical enum for routing.
+    ///
+    /// This performs an owned decode (twice at `V1`). That is safe here because `announce` — the only
+    /// asset selector whose array (`bytes[]`) has dynamically-sized elements, and thus the only one an
```
rayyan224: Not needed

crates/common/precompiles/src/b20_asset/dispatch.rs
```diff
@@ -103,6 +104,24 @@ impl<S: AssetAccounting, A: PolicyAccounting> B20AssetToken<S, A> {
...
         O: PrecompileCallObserver,
     {
+        // Fast-path `announce` before the generic decode: it is the only B-20 selector an aliased
+        // payload can amplify, and `BorrowedAnnounce::decode` validates it without materializing the
+        // `bytes[]` (mechanism in the `announce` module). The guard is total — a non-announce
+        // selector, an announce selector not dialable at `version`, or any decode rejection falls
+        // through to the generic decode below, which reproduces byte-identical error bytes (and
+        // cheaply, since a rejected payload never reaches owned materialization). `valid_selector`
+        // future-proofs a fork that drops `announce`.
+        if let Some(selector) = calldata.first_chunk::<4>().copied()
```
rayyan224: can w have is_announced_calldata. = this statement

crates/common/precompiles/src/b20_asset/announce.rs
```diff
@@ -0,0 +1,123 @@
...
+            .map_err(|_| ())?;
+        <<IB20Asset::announceCall as SolCall>::Parameters<'a> as SolType>::type_check(&token)
+            .map_err(|_| ())?;
+        Ok(Self(token))
+    }
+}
+
+impl AnnounceCall for BorrowedAnnounce<'_> {
+    fn id(&self) -> String {
+        String::from_utf8_lossy(self.0.1.as_slice()).into_owned()
```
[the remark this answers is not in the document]
rayyan224: Done in 848a21639 — switched `id()`/`description()`/`uri()` to `core::str::from_utf8(...).expect("type_check validated UTF-8").to_owned()`. Comment on the impl now spells out the reasoning: `type_check` guarantees valid UTF-8, so panicking on the impossible case beats a silent U+FFFD substitution if that invariant ever broke — that divergence from the owned `detokenize` path would be a consensus fork.

crates/common/precompiles/src/policy/versions.rs
rayyan224: Done in 848a21639 — removed the doc note; it documented a non-change.

crates/common/precompiles/src/common/abi/b20_abi.rs
rayyan224: Done in 848a21639 — removed the doc note.

crates/common/precompiles/src/b20_asset/versions.rs
rayyan224: Done in 848a21639 — removed the doc note on `AssetAbi::decode`. Kept the `announce_signature_is_frozen_across_versions` pin below since that one guards the fast-path's load-bearing invariant.

crates/common/precompiles/src/b20_asset/dispatch.rs
rayyan224: Done in 848a21639 — extracted as `BorrowedAnnounce::try_from_calldata(calldata, version) -> Option<Self>`. `route` now reads `if let Some(announce) = BorrowedAnnounce::try_from_calldata(calldata, version) { ... }` — one line, name says what it means.

roger-bai-coinbase (dismissed):

roger-bai-coinbase (approved):

PR #4743, opened 2026-08-27
fix(upgrade-signal): Reconcile Schedule Shrinkage Into Registry

refcell:
## Summary

The live upgrade-signal poller only applied a freshly read L1 schedule when one of its present entries had changed, so a pure schedule shrink — a previously-applied upgrade vanishing from a later, shorter schedule after an L1 reorg on a non-finalized tag — never tripped the apply gate and the stale activation lingered in the runtime registry until a restart or manual refresh. The poller now also treats a previously-applied upgrade that is absent from the current schedule as needing apply, which runs the existing apply path so `replace_overrides` trims the removed activation out of the registry. The applied baseline is cleared for vanished upgrades so the gate does not re-fire on every subsequent poll. This is the code fix for audit finding #5 that PR #4660 documented as a known limitation.

### DIFF
--- a/crates/utilities/upgrade-signal/src/state.rs
+++ b/crates/utilities/upgrade-signal/src/state.rs
@@ -818,10 +818,12 @@ mod tests {
         // An admin refresh transiently rewrites the registry to {Azul@50} at a newer L1 block,
         // both dropping Beryl (a missing key) and moving Azul's activation (a value drift) — without
         // touching this monitor's own view of what it applied.
-        assert!(refresher
-            .apply(&UpgradeSignalSchedule::new(2, vec![upgrade(BaseUpgrade::Azul, 50)]))
-            .unwrap()
-            .committed);
+        assert!(
+            refresher
+                .apply(&UpgradeSignalSchedule::new(2, vec![upgrade(BaseUpgrade::Azul, 50)]))
+                .unwrap()
+                .committed
+        );
 
         // L1 returns to the original schedule before the next poll. This monitor's own applied
         // values never moved, so a baseline-only check would see both entries as satisfied and never
--- a/crates/utilities/upgrade-signal/src/state.rs
+++ b/crates/utilities/upgrade-signal/src/state.rs
@@ -163,6 +163,14 @@ pub struct UpgradeSignalMonitor {
     pub metrics_layer: UpgradeSignalMetricLayer,
     /// Live observer state by contract-backed upgrade.
     states: BTreeMap<BaseUpgrade, UpgradeSignalState>,
+    /// Minimum protocol version of each active upgrade the last committed apply validated.
+    ///
+    /// The runtime registry stores only activation timestamps, not the minimum protocol version an
+    /// upgrade demands, so an activation-unchanged version bump leaves the registry matching. This
+    /// baseline lets [`Self::schedule_needs_apply`] still detect such a protocol-only change and
+    /// re-validate it (which may alarm or fail the node closed). Advanced only on a committed apply,
+    /// so a rejected or failed apply keeps the schedule offered for re-validation.
+    validated_protocol_versions: BTreeMap<BaseUpgrade, U256>,
     /// Contract values of the last schedule that failed to apply, used to page only on the first
     /// occurrence of a persistent failure rather than on every poll.
     last_apply_failure: Option<Vec<UpgradeSignal>>,
@@ -176,7 +184,12 @@ impl UpgradeSignalMonitor {
         for upgrade_id in BaseUpgrade::CONTRACT_VARIANTS {
             states.insert(upgrade_id, UpgradeSignalState::new());
         }
-        Self { metrics_layer, states, last_apply_failure: None }
+        Self {
+            metrics_layer,
+            states,
+            validated_protocol_versions: BTreeMap::new(),
+            last_apply_failure: None,
+        }
     }
 
     /// Tolerantly polls the reader, records live metrics, and — when `refresher` is supplied —
@@ -265,6 +278,7 @@ impl UpgradeSignalMonitor {
                 // nothing, so the schedule stays offered for a newer-block re-read.
                 if summary.committed {
                     self.last_apply_failure = None;
+                    self.validated_protocol_versions = Self::active_protocol_versions(schedule);
                     UpgradeSignalMetrics::record_apply_success(self.metrics_layer, schedule);
                 }
                 return UpgradeSignalPollOutcome::Continue;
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/utilities/upgrade-signal/src/state.rs
jackchuma: Could we only call `mark_schedule_applied` when `summary.committed` is true? `refresher.apply` returns `Ok` even when the registry rejects a lower-block schedule. A 101→100 reorg can therefore leave the full registry unchanged while clearing the monitor baseline, preventing future retries

crates/utilities/upgrade-signal/src/state.rs
```diff
@@ -347,15 +357,44 @@ impl UpgradeSignalMonitor {
...
     fn schedule_needs_apply(&self, schedule: &UpgradeSignalSchedule) -> bool {
-        schedule.signals.iter().any(|signal| {
-            self.states.get(&signal.upgrade_id).is_none_or(|state| state.needs_apply(signal))
-        })
+        let present: BTreeSet<BaseUpgrade> =
+            schedule.signals.iter().map(|signal| signal.upgrade_id).collect();
+        let removed = self
+            .states
+            .iter()
+            .any(|(upgrade_id, state)| state.is_applied() && !present.contains(upgrade_id));
```
jackchuma: Could removal detection use the actual registry or shared apply state? The admin refresh writes the same registry without updating the monitor. If admin applies `{Azul, Beryl}` and the monitor next sees its previous `{Azul}` schedule, it misses Beryl’s removal and leaves the stale activation

crates/utilities/upgrade-signal/src/state.rs
```diff
@@ -347,15 +357,44 @@ impl UpgradeSignalMonitor {
...
+    /// from the schedule, mirroring the runtime registry that `replace_overrides` has just trimmed
+    /// to exactly this schedule. Clearing the vanished entries keeps the apply gate from re-firing
+    /// on every subsequent poll.
     fn mark_schedule_applied(&mut self, schedule: &UpgradeSignalSchedule) {
+        let present: BTreeSet<BaseUpgrade> =
+            schedule.signals.iter().map(|signal| signal.upgrade_id).collect();
+        for (upgrade_id, state) in &mut self.states {
+            if !present.contains(upgrade_id) {
+                state.clear_applied();
+            }
```
jackchuma: Gate baseline clearing on a committed registry update. When `--upgrade-signal.l1-block-tag latest` reorgs from block 100 to 99, `replace_overrides` rejects the shorter schedule because the registry watermark is 100, but `refresher.apply` still returns `Ok` with `summary.committed == false`. `apply_and_evaluate` nevertheless calls this method, so this loop clears Beryl's applied marker even though its override is still in the registry; when the replacement chain reaches block 100 with the same shortened schedule, no absent marker remains to trigger another apply and the stale override persists. Reconcile these baselines only after a committed summary

crates/utilities/upgrade-signal/src/state.rs
refcell: Done in b89994ccf. `apply_and_evaluate` now only calls `mark_schedule_applied` / records success when `summary.committed` is true. A rejected lower-block schedule (`Ok` with `committed == false`) leaves the applied baseline untouched, so a newer-block re-read still reconciles. Covered by `rejected_lower_block_schedule_does_not_advance_the_baseline`.

crates/utilities/upgrade-signal/src/state.rs
refcell: Good catch — removal detection now runs against the authoritative runtime registry (`RuntimeUpgradeRegistry::overrides(chain_id)`) instead of the monitor's private baseline, so an override written by the admin refresh (or any override this poller never applied itself) is still seen as a removal. Covered by `removal_is_detected_against_registry_written_by_admin_refresh`.

crates/utilities/upgrade-signal/src/state.rs
refcell: Fixed in b89994ccf. The baseline clearing in `mark_schedule_applied` is now gated behind `summary.committed`, so a non-committed apply (e.g. the 100→99 reorg where `replace_overrides` rejects the shorter schedule) no longer clears a marker whose override is still live. Combined with registry-based removal detection, the replacement chain reaching block 100 re-trips the gate and trims the stale override.

crates/utilities/upgrade-signal/src/state.rs
jackchuma, line 1: Could we compare the full registry contents—including missing keys and activation values—against the schedule? Currently only extra registry keys are detected. For example, if the monitor applied `{Azul@42, Beryl@84}`, admin transiently applies `{Azul@50}`, and L1 returns to the original schedule before the next poll, both checks return false and the wrong registry persists indefinitely

refcell: Good call — the apply gate now compares the *full* registry contents against the schedule rather than the monitor's private baseline. `schedule_needs_apply` builds the exact override set a commit would produce and does a single `RuntimeUpgradeRegistry::overrides(chain_id) != expected` comparison, so present keys, missing keys, and drifted activation values all trip it.

Your exact scenario is covered by the new `transient_admin_drift_is_reconciled_by_value_and_missing_key` test: the monitor commits {Azul@42, Beryl@84}, an admin refresh transiently rewrites the registry to {Azul@50} (missing key + value drift), L1 returns to the original schedule, and the gate now fires and reconciles the registry back.

Since the registry is now the sole source of truth for the gate, the redundant `applied` baseline (and `mark_schedule_applied`) is removed entirely. Done in cf54022c3.

jackchuma, line 1: Could we retain a successfully validated protocol-version baseline alongside the registry comparison? `UpgradeActivationOverrides` does not store `protocol_version`, so a protocol-only update is skipped. If L1 raises the minimum version while leaving the activation timestamp unchanged, the registry still matches and the poller returns before validation. The node would never alarm or fail closed as activation approaches

refcell: Good catch — done in 7ead61b58. The monitor now retains a validated-protocol-version baseline alongside the registry comparison: `UpgradeSignalMonitor::validated_protocol_versions` records the minimum version of each *active* signal that the last committed apply validated (advanced only on `summary.committed`, so a rejected/failed apply keeps re-offering the schedule).

`schedule_needs_apply` now fires when the registry activations differ **or** an active signal's minimum version has drifted from that baseline. So a protocol-only bump — same activation timestamp, higher minimum — that leaves `UpgradeActivationOverrides` matching still trips the gate, forcing re-validation that can alarm or fail closed as the activation nears.

Covered by two new tests:
- `protocol_version_bump_with_unchanged_activation_trips_apply_gate` — a supported bump re-trips the gate and settles once re-committed.
- `unsupportable_protocol_bump_with_unchanged_activation_fails_closed` — an unsupportable bump inside the halt lead time returns `HaltNode`.

jackchuma (approved):

PR #4745, opened 2026-08-27
fix(precompiles): pin wrapper storage features to active upgrade (Cantina #17 / BOP-604)

rayyan224:
## Summary

- **Cantina finding #17 (Informational)** — `NonceManager` and `TxContext` used `#[precompile(install)]`, which expanded through the no-features arm of `base_precompile!` and hardcoded `StorageFeatures::Legacy`. Both install only when `upgrade >= Cobalt`, so a Cobalt-active fork ran Cobalt-gated wrappers under Legacy semantics.
- No behavior divergence today: neither storage layout uses Vec or bytes-like fields, which are the only consumers of the Cobalt storage flag today. Fix removes the trap before a future feature-sensitive field silently inherits the wrong semantics.
- Storage features are now pinned to the currently-active `BaseUpgrade` at every wrapper's EVM boundary, matching the Beryl-block precompiles.

## Changes

- Delete the no-features arm of `base_precompile!` so every call must pass `storage_features:` explicitly.
- `#[precompile(...)]` now requires a `storage_features = <expr>` config and emits it into the generated `base_precompile!` call.
- Hand-roll `NonceManager::install` / `precompile` and `TxContext::install` / `precompile` to take `upgrade: BaseUpgrade` and pin features via `UpgradeGatedStorageFeatures::from_upgrade(upgrade)`.
- Drop `#[precompile(args(admin_config))]` from `ActivationRegistry`; its hand-rolled `precompile_with_observer` already pins features, and the macro-generated `precompile()` was dead code.
- Thread `self.spec.upgrade()` through the Cobalt install block in `BasePrecompiles::install_with_observer` so it mirrors the Beryl block.
- Add proc-macro tests asserting the emitted wrapper contains `storage_features:` and wrapper unit tests asserting install succeeds at Cobalt and later upgrades.

## Test plan

- [x] `cargo build -p base-common-precompiles -p base-precompile-macros`
- [x] `cargo test -p base-common-precompiles -p base-precompile-macros` — 662 + 39 + 27 tests pass
- [x] `cargo clippy -p base-common-precompiles -p base-precompile-macros --all-targets -- -D warnings`
- [x] Downstream build check: `cargo build -p base-common-evm -p base-execution-txpool -p base-execution-eip8130 -p base-metering`

Linear: https://linear.app/coinbase/issue/BOP-604/cantina-17-cobalt-only-wrappers-silently-select-legacy-storage
Cantina: https://cantina.xyz/code/3a3e0d24-1cb7-474c-87ef-f0ad2350b5c4/findings/17

### DIFF
--- a/crates/common/precompile-storage/src/packing.rs
+++ b/crates/common/precompile-storage/src/packing.rs
@@ -53,10 +53,16 @@ impl StorageOps for PackedSlot {
 
     // A packed slot only ever holds `Packable` primitives; the `dynamic_storage_tail_cleanup`
     // path this feature gates never triggers here (dynamic types are `Layout::Slots(_)` and
-    // route through the full-slot `Slot`/`TransientOps` handlers). `Legacy` is the correct
-    // pin — the value is defensive, and it is declared explicitly (not defaulted) so the
-    // fork-sensitive dispatch never inherits an implicit value.
+    // route through the full-slot `Slot`/`TransientOps` handlers). If a future refactor
+    // routes a dynamic type through a packed intermediate, returning `Legacy` here would
+    // silently skip Cobalt tail cleanup — trip `debug_assert!` at test time so the mistake
+    // surfaces in CI, and keep the defensive `Legacy` return so prod never crashes on a
+    // consensus-critical path.
     fn storage_features(&self) -> StorageFeatures {
+        debug_assert!(
+            false,
+            "PackedSlot::storage_features reached; a dynamic-cleanup path is routing through a packed intermediate",
+        );
         StorageFeatures::Legacy
     }
 }
--- a/crates/common/precompile-storage/src/packing.rs
+++ b/crates/common/precompile-storage/src/packing.rs
@@ -24,7 +24,7 @@ use alloy_primitives::U256;
 
 use crate::{
     error::Result,
-    provider::{FromWord, Layout, StorableType, StorageOps},
+    provider::{FromWord, Layout, StorableType, StorageFeatures, StorageOps},
 };
 
 /// A helper struct to support packing elements into a single slot. Represents an
@@ -49,6 +49,15 @@ impl StorageOps for PackedSlot {
     fn store(&mut self, _slot: U256, value: U256) -> Result<()> {
         self.0 = value;
         Ok(())
+    }
+
+    // A packed slot only ever holds `Packable` primitives; the `dynamic_storage_tail_cleanup`
+    // path this feature gates never triggers here (dynamic types are `Layout::Slots(_)` and
+    // route through the full-slot `Slot`/`TransientOps` handlers). `Legacy` is the correct
+    // pin — the value is defensive, and it is declared explicitly (not defaulted) so the
+    // fork-sensitive dispatch never inherits an implicit value.
+    fn storage_features(&self) -> StorageFeatures {
+        StorageFeatures::Legacy
     }
 }
 
--- a/crates/common/precompiles/src/nonce/precompile.rs
+++ b/crates/common/precompiles/src/nonce/precompile.rs
@@ -21,8 +21,7 @@ impl NonceManager {
     /// Creates the EVM precompile wrapper with storage features pinned to `upgrade`.
     ///
     /// The wrapper reads `storage_features` from the currently-active fork instead of the macro
-    /// default, so a Cobalt-only precompile cannot silently execute under Legacy semantics
-    /// (Cantina finding #17).
+    /// default, so a Cobalt-only precompile cannot silently execute under Legacy semantics.
     pub fn precompile(upgrade: BaseUpgrade) -> DynPrecompile {
         let storage_features = UpgradeGatedStorageFeatures::from_upgrade(upgrade);
         base_precompile!(
--- a/crates/common/precompiles/src/nonce/precompile.rs
+++ b/crates/common/precompiles/src/nonce/precompile.rs
@@ -48,13 +48,4 @@ mod tests {
 
         assert!(precompiles.get(&NonceManagerStorage::ADDRESS).is_some());
     }
-
-    #[test]
-    fn install_accepts_upgrades_past_cobalt() {
-        let mut precompiles = PrecompilesMap::from_static(Precompiles::cancun());
-
-        NonceManager::install(&mut precompiles, BaseUpgrade::LATEST);
-
-        assert!(precompiles.get(&NonceManagerStorage::ADDRESS).is_some());
-    }
 }
--- a/crates/common/precompiles/src/nonce/precompile.rs
+++ b/crates/common/precompiles/src/nonce/precompile.rs
@@ -1,10 +1,61 @@
 //! Precompile entry point for the EIP-8130 2D nonce manager.
 
-use base_precompile_macros::precompile;
+use alloy_evm::precompiles::{DynPrecompile, PrecompilesMap};
+use base_common_genesis::BaseUpgrade;
 
-use crate::NonceManagerStorage;
+use crate::{NonceManagerStorage, UpgradeGatedStorageFeatures, macros::base_precompile};
 
 /// Entry point for the EIP-8130 2D nonce manager precompile.
-#[precompile(install)]
 #[derive(Debug, Default, Clone, Copy)]
 pub struct NonceManager;
+
+impl NonceManager {
+    /// Installs the nonce manager precompile, pinning storage features to `upgrade`.
+    pub fn install(precompiles: &mut PrecompilesMap, upgrade: BaseUpgrade) {
+        precompiles.extend_precompiles(core::iter::once((
+            NonceManagerStorage::ADDRESS,
+            Self::precompile(upgrade),
+        )));
+    }
+
+    /// Creates the EVM precompile wrapper with storage features pinned to `upgrade`.
+    ///
+    /// The wrapper reads `storage_features` from the currently-active fork instead of the macro
+    /// default, so a Cobalt-only precompile cannot silently execute under Legacy semantics
+    /// (Cantina finding #17).
+    pub fn precompile(upgrade: BaseUpgrade) -> DynPrecompile {
+        let storage_features = UpgradeGatedStorageFeatures::from_upgrade(upgrade);
+        base_precompile!(
+            "NonceManager",
+            storage_features: storage_features,
+            |ctx, calldata| { NonceManagerStorage::new(ctx).dispatch(ctx, &calldata) },
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use alloy_evm::precompiles::PrecompilesMap;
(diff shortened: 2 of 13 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

chunter-cb (approved):

cb-heimdall:
Review Error for chunter-cb @ 2026-08-28 17:23:15 UTC

crates/common/precompiles/src/tx_context/precompile.rs
```diff
@@ -1,10 +1,61 @@
...
+    pub fn install(precompiles: &mut PrecompilesMap, upgrade: BaseUpgrade) {
+        precompiles.extend_precompiles(core::iter::once((
+            TxContextStorage::ADDRESS,
+            Self::precompile(upgrade),
+        )));
+    }
+
+    /// Creates the EVM precompile wrapper with storage features pinned to `upgrade`.
+    ///
+    /// The wrapper reads `storage_features` from the currently-active fork instead of the macro
```
robriks, line 23: probably out of scope for this PR, but can `NonceManagerStorage` still silently execute under Legacy semantics? Claude seems to think so

The EIP-8130 executor builds its own provider at `eip8130.rs:969` and constructs the same storage over it at `eip8130.rs:1018`:

```
let mut provider = JournalStorageProvider::new(internals, Address::ZERO);
StorageCtx::enter(&mut provider, |sctx| {
    let mut nonce_mgr = NonceManagerStorage::new(sctx);
```

I think because `JournalStorageProvider` implements `PrecompileStorageProvider` at `journal.rs:70` but never overrides `storage_features()`, it can still silently take the trait default (`StorageFeatures::Legacy` at `precompile-storage/src/provider.rs:108-110`). if so, that's the same hardcoded `Legacy` this PR just deleted from `base_precompile!`, just relocated into the trait (with a second copy on `StorageOps` at L162-165)

if we add a `Vec` field to `NonceManagerStorage` in the future I think the the executor would silently write it under Legacy, which is the same scenario this PR fixes in the ABI dispatch.

a follow-up PR can make the same fix for the EIP8130 executor by deleting the two trait defaults so every provider must declare its features. That forces JournalStorageProvider to thread the current active upgrade, and the compiler enumerates the remaining sites

robriks (commented):
is `#[precompile]` unused after this PR? before there were three invocations, and this PR removes all three:

```
origin/main:crates/common/precompiles/src/activation/precompile.rs:14:#[precompile(args(admin_config: ActivationAdminConfig))]
origin/main:crates/common/precompiles/src/nonce/precompile.rs:8:#[precompile(install)]
origin/main:crates/common/precompiles/src/tx_context/precompile.rs:8:#[precompile(install)]
```

The changes are sound assuming we want to use this in the future. just wanted to mention that currently the attribute is still exported at `precompile-macros/src/lib.rs:49`, but seemingly nothing applies it.

rayyan224:
This will break tests since it affects the registering

crates/common/precompiles/src/tx_context/precompile.rs
rayyan224: Pulled into this PR. Deleted the `Legacy` defaults on both `PrecompileStorageProvider::storage_features` and `StorageOps::storage_features`, then threaded `StorageFeatures` through `JournalStorageProvider::new` (3 EIP-8130 sites) and the txpool validator's `StateProviderPrecompileStorage::new` (4 sites) — all pinned from the active upgrade. Filed [BOP-615](https://linear.app/coinbase/issue/BOP-615) for the remaining `EvmPrecompileStorageProvider::new()` Legacy-default footgun. Commit: 03b2a9ad2.

crates/execution/txpool/src/validator.rs
```diff
@@ -861,6 +884,17 @@ where
         }
     }
 
+    /// Persistent-storage features active at `timestamp`. Every admission-time
+    /// [`StateProviderPrecompileStorage`] construction must be pinned to this
+    /// so pool storage semantics track the fork the builder will include
+    /// under. Threading it through the constructor removes the trait-default
+    /// escape hatch that used to silently pick `Legacy` on a Cobalt-active fork.
+    fn eip8130_storage_features(&self, timestamp: u64) -> StorageFeatures {
```
robriks: Do we have a test for this? Something as simple as this would do:

```
#[test]
fn eip8130_storage_features_track_cobalt_activation() {
    assert_eq!(build_test_validator().eip8130_storage_features(0), StorageFeatures::Cobalt);
    assert_eq!(
        build_test_validator_with_spec(Arc::new(BaseChainSpec::mainnet()))
            .eip8130_storage_features(u64::MAX),
        StorageFeatures::Legacy,
    );
}
```

robriks: Doesn't need to land in this PR, I don't want to slow you down any more than I already have. Can slate it for a tiny follow-up that I'll approve quickly.

crates/common/precompiles/src/nonce/precompile.rs
```diff
@@ -1,10 +1,60 @@
...
+use crate::{NonceManagerStorage, UpgradeGatedStorageFeatures, macros::base_precompile};
 
 /// Entry point for the EIP-8130 2D nonce manager precompile.
-#[precompile(install)]
 #[derive(Debug, Default, Clone, Copy)]
 pub struct NonceManager;
+
+impl NonceManager {
+    /// Installs the nonce manager precompile, pinning storage features to `upgrade`.
+    pub fn install(precompiles: &mut PrecompilesMap, upgrade: BaseUpgrade) {
```
robriks, line 14: I think this might be a cross-repo breaking change, because `Base Std Interface Tests (Cobalt)` CI fails on this PR while passing on the last four main commits. 

If so, just want to make sure we open a PR against base-anvil to get this change in on that repo (maybe it's already taken care of and just unmerged?) and bump the pinned pair (base-anvil, base-std)

The call site is in base-anvil's `crates/evm/networks/src/lib.rs:198-199` on the Cobalt line, still calling the 1-arg form:

```
TxContext::install(precompiles); // -> install(precompiles, upgrade)
NonceManager::install(precompiles); // -> install(precompiles, upgrade)
```

crates/common/precompiles/src/nonce/precompile.rs
```diff
@@ -1,10 +1,60 @@
...
+        NonceManager::install(&mut precompiles, BaseUpgrade::Cobalt);
+
+        assert!(precompiles.get(&NonceManagerStorage::ADDRESS).is_some());
+    }
+
+    #[test]
+    fn install_accepts_upgrades_past_cobalt() {
+        let mut precompiles = PrecompilesMap::from_static(Precompiles::cancun());
+
+        NonceManager::install(&mut precompiles, BaseUpgrade::LATEST);
```
robriks: isn't `BaseUpgrade::LATEST` still Beryl? if so, I think the `install` will proceed with `StorageFeatures::Legacy` which is before Cobalt, not past it.

the above might be wrong, but either way I think we should either assert that `BaseUpgrade::LATEST` is the expected value when `install` succeeds, or just pin to Cobalt specifically rather than `LATEST`.

Claude suggests:
```
use base_precompile_storage::StorageFeatures;
#[test]
fn install_pins_cobalt_storage_features_at_and_past_cobalt() {
    for upgrade in [BaseUpgrade::Cobalt, BaseUpgrade::Denim, BaseUpgrade::Zenith] {
        assert_eq!(
            UpgradeGatedStorageFeatures::from_upgrade(upgrade),
            StorageFeatures::Cobalt,
            "wrapper for {upgrade:?} would run under Legacy",
        );
        let mut precompiles = PrecompilesMap::from_static(Precompiles::cancun());
        NonceManager::install(&mut precompiles, upgrade);
        assert!(precompiles.get(&NonceManagerStorage::ADDRESS).is_some());
    }
}
```

crates/common/precompiles/src/tx_context/precompile.rs
```diff
@@ -1,10 +1,60 @@
...
+        TxContext::install(&mut precompiles, BaseUpgrade::Cobalt);
+
+        assert!(precompiles.get(&TxContextStorage::ADDRESS).is_some());
+    }
+
+    #[test]
+    fn install_accepts_upgrades_past_cobalt() {
+        let mut precompiles = PrecompilesMap::from_static(Precompiles::cancun());
+
+        TxContext::install(&mut precompiles, BaseUpgrade::LATEST);
```
robriks: Same feedback as my comment on line 56 of `nonce/precompile.rs`

crates/common/precompiles/src/nonce/precompile.rs
rayyan224: Yep, we need to update base anvil latest

crates/common/precompile-storage/src/packing.rs
```diff
@@ -50,6 +50,15 @@ impl StorageOps for PackedSlot {
         self.0 = value;
         Ok(())
     }
+
+    // A packed slot only ever holds `Packable` primitives; the `dynamic_storage_tail_cleanup`
+    // path this feature gates never triggers here (dynamic types are `Layout::Slots(_)` and
```
rayyan224, line 55: Since this is used in prod, and we are expecting it to never go called maybe we should throw if this is touched thoughts ?

crates/execution/txpool/src/validator.rs
rayyan224: fn eip8130_storage_features(&self, timestamp: u64) -> StorageFeatures {
        UpgradeGatedStorageFeatures::from_upgrade(
            BaseSpecId::from_timestamp(self.chain_spec(), timestamp).upgrade(),
        )
    }

Hm the question actually might be, why are we using this to get the storage features, they are not this feature dependent

PR #4747, opened 2026-08-27
ci: reduce Depot Docker build contention

meyer9:
## Summary
- cancel superseded Depot Docker push workflows for the same ref
- run the devnet cache warmer after the publish build instead of concurrently with it
- follow Depot's Rust Dockerfile guidance by using cargo-chef plus sccache with BuildKit cache mounts
- replace mutable `/app/target` cache mounts with Cargo registry/git and `SCCACHE_DIR` cache mounts
- scope `cargo chef cook` to the packages needed by each Docker bake target
- give each Rust service target a stable sccache cache ID so normally concurrent image builds do not all serialize on one locked cache
- build the `base` runtime image's four Rust binaries in one Cargo invocation instead of four parallel builder stages
- align the Docker Rust image default with `rust-toolchain.toml` to avoid rustup installing 1.96.0 inside 1.95.0 images

## Reasoning
Depot build `3441khb2vq` was slow because the Docker workflow had multiple heavyweight Rust image builds overlapping in the same Depot project. The logs showed long waits around shared BuildKit work and cache mounts before relatively short Cargo build work.

Depot's Rust guidance is to keep Cargo registry/git caches and an sccache directory as BuildKit cache mounts, with cargo-chef used for dependency cooking. This PR moves the Rust services Dockerfile to that shape and removes the previous full Cargo target-directory cache mounts.

The important cache behavior is:
- `cargo chef prepare` isolates dependency metadata before source is copied into the final build stages.
- `cargo chef cook` is package-scoped per bake target, so building `base` does not prebuild unrelated workspace packages such as prover-only dependencies.
- `SCCACHE_CACHE_ID` is stable for repeated builds of the same target, which preserves warm Rust compiler artifacts across Depot builds.
- Different targets get different sccache IDs because BuildKit cache mounts use `sharing=locked`; using one global sccache cache would make unrelated image targets contend on the same lock.
- The `base` image still shares one stable cache ID across profiles, so this does not split dev/maxperf cache reuse. The expected concurrency reduction comes from workflow ordering and fewer jobs per worker.
- The composite `base-image-builder` avoids the old intra-image behavior where `base-builder`, `execution-builder`, `consensus-builder`, and `snapshotter-builder` could run concurrently just to assemble one `base` image.

## Validation
- `docker buildx bake -f etc/docker/docker-bake.hcl base --print`
- `docker buildx build --call=outline -f etc/docker/Dockerfile.rust-services --target base --build-arg PROFILE=dev --build-arg RUST_VERSION=1.96.0 .`
- `git diff --check`
- Depot smoke build `tbr33m6hvs`: `base` dev cache-only build completed in 129s after the Dockerfile fixes.
- Depot smoke build `wnwlk06vwz`: completed after introducing target-scoped sccache IDs; this was expected to warm the new `rust-services-base-sccache` namespace.

### DIFF
--- a/.depot/workflows/docker.yml
+++ b/.depot/workflows/docker.yml
@@ -7,6 +7,9 @@ on:
     branches:
       - main
   workflow_dispatch: {}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 env:
   REGISTRY_IMAGE: ghcr.io/${{ github.repository_owner }}/node
 permissions:
@@ -64,6 +67,7 @@ jobs:
             --set "${TARGET}.labels.org.opencontainers.image.revision=${{ steps.image-labels.outputs.revision }}" \
             --set "${TARGET}.labels.org.opencontainers.image.created=${{ steps.image-labels.outputs.created }}"
   warm-devnet-cache:
+    needs: build-and-push
     runs-on: depot-ubuntu-24.04
     env:
       PROFILE: dev
--- a/etc/docker/Dockerfile.rust-services
+++ b/etc/docker/Dockerfile.rust-services
@@ -1,10 +1,18 @@
-ARG RUST_VERSION=1.95.0
+# syntax=docker/dockerfile:1
+ARG RUST_VERSION=1.96.0
 FROM public.ecr.aws/docker/library/rust:${RUST_VERSION}-trixie AS rust-builder-base
 WORKDIR /app
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-      git libclang-dev libssl-dev pkg-config curl build-essential cmake && \
+      git libclang-dev libssl-dev pkg-config curl build-essential cmake \
+      protobuf-compiler libprotobuf-dev && \
     rm -rf /var/lib/apt/lists/*
+RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked \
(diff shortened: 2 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

mw2000 (approved):

PR #4748, opened 2026-08-27
fix(precompiles): type-witness the fast-path nonce increment (Cantina #18 / BOP-605)

rayyan224:
## Summary

Cantina finding #18 (informational) flagged `NonceManagerStorage::increment_nonce_from_current` as a public API that trusts caller-supplied `current` state: passing a stale value writes `current + 1` over whatever is in storage, so a future caller with a stale read could overwrite a newer nonce and desync channel progression. Today's only caller (`eip8130::authorize_and_apply`) validates immediately before consuming, so no exploit exists — the risk is refactor-time.

This PR replaces the raw-u64 fast path with a **single-use `SequenceNonceRead` witness** so a stale or fabricated value can no longer reach the increment path.

- `read_sequence_nonce(account, nonce_key)` performs the single SLOAD and returns a non-`Copy` receipt binding `(account, nonce_key, current)`.
- `increment_from_read(read)` consumes the receipt, writing `current + 1` to the exact slot the read observed and emitting `NonceIncremented`.
- `SequenceNonceRead` has private fields and no `pub` constructor other than `read_sequence_nonce`, and is not `Copy`/`Clone` — so it cannot be fabricated, cloned, or replayed. Callers must re-read to obtain a new receipt.
- The one production caller in `eip8130::authorize_and_apply` swaps to the new API.

The single-SLOAD / single-SSTORE cost is preserved (existing `loaded_increment_uses_one_sload_and_one_sstore` still asserts `sload=1, sstore=1`), and the safe `increment_nonce` wrapper now composes `read_sequence_nonce` + `increment_from_read`.

Linear: https://linear.app/coinbase/issue/BOP-605
Cantina: https://cantina.xyz/code/3a3e0d24-1cb7-474c-87ef-f0ad2350b5c4/findings/18

## Test plan

- [x] `cargo test -p base-common-precompiles --lib nonce::` — all 25 tests pass, including new regressions:
  - `stale_read_receipt_cannot_overwrite_newer_nonce` — receipt held across an out-of-band write is discarded; a fresh read is the only path forward.
  - `read_receipt_is_bound_to_its_account_and_key` — receipt exposes and is applied to its bound slot; a sibling account/channel is untouched.
  - `read_sequence_nonce_rejects_protocol_nonce` — nonce key `0` returns `InvalidNonceKey`.
- [x] `cargo test -p base-common-evm --lib eip8130` — 38 tests pass; sequence-nonce path integration verified via `two_dimensional_nonce_is_incremented`.
- [x] `cargo clippy -p base-common-precompiles -p base-common-evm --all-targets -- -D warnings` — clean.

### DIFF
--- a/crates/common/evm/src/eip8130.rs
+++ b/crates/common/evm/src/eip8130.rs
@@ -1050,19 +1050,13 @@ impl Eip8130Executor {
                     .map_err(BaseTransactionError::eip8130)?;
                     (protocol_nonce == 0, true)
                 } else {
-                    // Take a single-use read receipt so the current value we
-                    // validate against is provably the value the increment
-                    // writes over — no chance of handing the storage layer a
-                    // stale caller-supplied `current`.
                     let read = nonce_mgr
                         .read_sequence_nonce(sender, nonce_key)
                         .map_err(BaseTransactionError::eip8130)?;
                     let current_nonce = read.current();
                     NonceValidator::validate_sequence(tx, current_nonce, NonceMode::Inclusion)
                         .map_err(BaseTransactionError::eip8130)?;
-                    nonce_mgr
-                        .increment_from_read(read)
-                        .map_err(BaseTransactionError::eip8130)?;
+                    nonce_mgr.increment_from_read(read).map_err(BaseTransactionError::eip8130)?;
                     (current_nonce == 0, false)
                 };
 
--- a/crates/common/evm/src/eip8130.rs
+++ b/crates/common/evm/src/eip8130.rs
@@ -1050,13 +1050,18 @@ impl Eip8130Executor {
                     .map_err(BaseTransactionError::eip8130)?;
                     (protocol_nonce == 0, true)
                 } else {
-                    let current_nonce = nonce_mgr
-                        .get_nonce(sender, nonce_key)
+                    // Take a single-use read receipt so the current value we
+                    // validate against is provably the value the increment
+                    // writes over — no chance of handing the storage layer a
+                    // stale caller-supplied `current`.
+                    let read = nonce_mgr
+                        .read_sequence_nonce(sender, nonce_key)
                         .map_err(BaseTransactionError::eip8130)?;
+                    let current_nonce = read.current();
                     NonceValidator::validate_sequence(tx, current_nonce, NonceMode::Inclusion)
                         .map_err(BaseTransactionError::eip8130)?;
                     nonce_mgr
-                        .increment_nonce_from_current(sender, nonce_key, current_nonce)
+                        .increment_from_read(read)
                         .map_err(BaseTransactionError::eip8130)?;
                     (current_nonce == 0, false)
                 };
--- a/crates/common/evm/src/eip8130.rs
+++ b/crates/common/evm/src/eip8130.rs
@@ -1050,13 +1050,14 @@ impl Eip8130Executor {
                     .map_err(BaseTransactionError::eip8130)?;
                     (protocol_nonce == 0, true)
                 } else {
-                    let read = nonce_mgr
-                        .read_sequence_nonce(sender, nonce_key)
+                    let current_nonce = nonce_mgr
+                        .get_nonce(sender, nonce_key)
                         .map_err(BaseTransactionError::eip8130)?;
-                    let current_nonce = read.current();
                     NonceValidator::validate_sequence(tx, current_nonce, NonceMode::Inclusion)
                         .map_err(BaseTransactionError::eip8130)?;
-                    nonce_mgr.increment_from_read(read).map_err(BaseTransactionError::eip8130)?;
+                    nonce_mgr
+                        .increment_nonce(sender, nonce_key)
+                        .map_err(BaseTransactionError::eip8130)?;
                     (current_nonce == 0, false)
                 };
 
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/evm/src/eip8130.rs
```diff
@@ -1050,13 +1050,18 @@ impl Eip8130Executor {
                     .map_err(BaseTransactionError::eip8130)?;
                     (protocol_nonce == 0, true)
                 } else {
-                    let current_nonce = nonce_mgr
-                        .get_nonce(sender, nonce_key)
+                    // Take a single-use read receipt so the current value we
```
rayyan224: Remove comment

robriks (commented):
this PR's implementation is sound for the time being because `validate_sequence` is pure, so nothing writes the slot between the read and the increment; but I believe that soundness comes from call-site adjacency, not from the receipt.

`SequenceNonceRead` binds provenance: a current can no longer be fabricated, reused, or applied to the wrong slot. It doesn't bind freshness, which is the axis the finding names. 

Here's what I mean, Claude wrote a PoC to show that a receipt held across a write can still clobber:

```
// storage.rs L396-400 verbatim; L401-404 replaced with the last two lines.
let stale = mgr.read_sequence_nonce(ACCOUNT_A, nonce_key).unwrap();   // L396
assert_eq!(stale.current(), 0);                                       // L397
mgr.nonces.at_mut(&ACCOUNT_A).at_mut(&nonce_key).write(5).unwrap();   // L398
let _ = stale;                                                        // L400 — no-op: a wildcard neither moves nor drops

assert_eq!(mgr.increment_from_read(stale).unwrap(), 1);               // receipt outlived the write
assert_eq!(mgr.get_nonce(ACCOUNT_A, nonce_key).unwrap(), 1);          // 5 -> 1: channel rewound
```

That compiles and passes on this branch, so `stale_read_receipt_cannot_overwrite_newer_nonce` asserts a property the code doesn't quite possess.

it's probably fine to keep relying on the current implementation's call-site adjacency soundness, but if so we should leave a comment noting it to make sure future refactors don't step in it. the alternative (which I think would be more robust) would be to have the write derive its value from its own read  `get_nonce + increment_nonce`, deleting the fast path. tradeoff is that will cost one warm SLOAD (100 gas)

crates/common/precompiles/src/nonce/storage.rs
```diff
@@ -280,6 +287,36 @@ impl NonceManagerStorage<'_> {
...
 }
 
+/// Single-use receipt of a fresh sequence-channel nonce read.
+///
+/// Produced by [`NonceManagerStorage::read_sequence_nonce`] and consumed by
+/// [`NonceManagerStorage::increment_from_read`]. Fields are module-private and
+/// the type is neither `Copy` nor `Clone`, so a receipt can only be obtained
+/// by actually reading storage and cannot be replayed against later state.
+#[derive(Debug)]
+pub struct SequenceNonceRead {
```
robriks, line 297: going a bit deeper with claude:


> `SequenceNonceRead` has no lifetime parameter and three owned fields, so no borrow of the manager travels with the receipt. The `&self` borrow taken by `read_sequence_nonce` ends at the semicolon, leaving `nonce_mgr` fully unborrowed. This is what permits the later `&mut` self call to `increment_from_read`, and equally permits anything else that writes in between. Switching the read back to `&mut` self wouldn't change this: the borrow ends at the semicolon either way, because the return type holds no reference.

> So the receipt constrains where a current came from, but not whether it's still true. Take the test at L390 and replace L401-404 with one line:

```
let stale = mgr.read_sequence_nonce(ACCOUNT_A, nonce_key).unwrap();  // L396
assert_eq!(stale.current(), 0);                                      // L397
mgr.nonces.at_mut(&ACCOUNT_A).at_mut(&nonce_key).write(5).unwrap();  // L398
let _ = stale;                                                       // L400
assert_eq!(mgr.increment_from_read(stale).unwrap(), 1);              // 5 -> 1
```

> That compiles and passes under the existing test name. It shows two things: `let _ = stale` is a no-op, since a wildcard pattern neither moves nor drops and the receipt is still live; and `increment_from_read` never compares `read.current` against storage, it writes `current + 1` unconditionally.

What the type does buy is real: a current can't be fabricated, can't be used twice, and can't be applied to the wrong slot. It just doesn't buy freshness, which is the property named. 
```

rayyan224: What do you mean frehsness, the sequencer runs everything sequentially, so there wouldn't be any racecondition

chunter-cb (approved):

cb-heimdall:
Review Error for chunter-cb @ 2026-08-31 21:13:45 UTC

roger-bai-coinbase (approved):

PR #4761, opened 2026-08-28
feat(evm2): Base Fork Schedule

refcell:
## Summary

Gives `base-common-evm2` a **Base fork schedule** so the EVM feature set / gas schedule is chosen by the active Base upgrade instead of the stock Ethereum spec. Stacked on #4759.

- Adds `BaseSpecId`, wrapping `base_common_genesis::BaseUpgrade` and mapping each upgrade to the governing evm2 `SpecId` (mirrors the revm-side `into_eth_spec` table; duplicated rather than reused because that mapping targets revm's `SpecId`).
- Wires it as `BaseEvmTypes::SpecId`. The stock `BaseEvmConfigSelector` already generically maps `T::SpecId` via `Into<SpecId>`, so no custom selector is needed.
- The active upgrade stays recoverable at runtime via `Evm::config_spec_id` — so even though several upgrades share one evm2 spec (Ecotone & Fjord → CANCUN), the **L1 data fee is now fork-aware** (formula selected by the active `BaseUpgrade`) instead of hardcoding Ecotone.

Timestamp → upgrade resolution happens at the node boundary (future PR); evm2 selects purely from the `SpecId` passed to `Evm::new`.

## Testing

The differential parity harness now **sweeps every fixture across Regolith/Ecotone/Isthmus (MERGE/CANCUN/PRAGUE)** — 7 fixtures × 3 forks = 21 differential executions, all matching between evm2 and revm, including the fork-sensitive `SSTORE` case. Plus unit tests for the mapping table. All tests + `clippy -D warnings` + `+nightly fmt` clean; non-test build stays revm-free.

## Scope note — fee completeness deferred
The roadmap grouped "fork schedule **& fee completeness**" together, but the fee half (Isthmus **operator fee**, and pricing the L1 fee over the **full EIP-2718 tx** rather than calldata) has a real prerequisite: the standard-tx envelope must carry its enveloped bytes (evm2's `TxEnvelope` doesn't expose them). That's cleaner as its own PR, so this one is scoped to the fork schedule.

### DIFF
--- a/crates/common/evm2/src/handler.rs
+++ b/crates/common/evm2/src/handler.rs
@@ -2,7 +2,6 @@
 
 use alloy_consensus::Transaction;
 use alloy_primitives::{Address, U256};
-use base_common_genesis::BaseUpgrade;
 use evm2::{
     Evm, TxResult,
     ethereum::{charge_upfront, default_settle_gas},
@@ -27,11 +26,12 @@ impl BaseTxHandlerHooks {
     fn l1_fee(host: &mut Evm<'_, BaseEvmTypes>, envelope: &BaseTxEnvelope) -> U256 {
         match envelope {
             // TODO: price over the full L1-posted (EIP-2718) transaction, not just
-            // the calldata, to match the revm integration. The evm2 spec schedule
-            // is also still Ecotone-only; the Base fork schedule is layered on in
-            // follow-up work.
+            // the calldata, to match the revm integration. This requires the
+            // envelope to carry the enveloped tx bytes (evm2's `TxEnvelope` does not
+            // expose them), tracked as fee-completeness follow-up work.
             BaseTxEnvelope::Standard(tx) => {
-                host.block_env().ext.calculate_tx_l1_cost(tx.input(), BaseUpgrade::Ecotone)
+                let upgrade = host.config_spec_id().upgrade();
+                host.block_env().ext.calculate_tx_l1_cost(tx.input(), upgrade)
             }
             // Deposits are funded on L1 and exempt from the L2 L1-data fee.
             BaseTxEnvelope::Deposit(_) => U256::ZERO,
--- a/crates/common/evm2/src/spec.rs
+++ b/crates/common/evm2/src/spec.rs
@@ -1,0 +1,67 @@
+//! Base fork schedule for the EVM2 execution type family.
+
+use base_common_genesis::BaseUpgrade;
+use evm2::SpecId;
+
+/// A Base fork identifier for EVM2 execution.
+///
+/// Wraps a Base network [`BaseUpgrade`] and maps it — via [`From<BaseSpecId>`] for [`SpecId`] —
+/// to the EVM2 spec whose feature set and gas schedule govern execution at that upgrade. Used
+/// as [`BaseEvmTypes::SpecId`](crate::BaseEvmTypes), it keeps the active upgrade recoverable at
+/// runtime (`Evm::config_spec_id`) for fork-dependent fee logic, even when several upgrades
+/// share one EVM2 spec (e.g. Ecotone and Fjord both map to [`SpecId::CANCUN`]).
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct BaseSpecId(BaseUpgrade);
+
+impl BaseSpecId {
+    /// Wraps a Base network upgrade.
+    pub const fn new(upgrade: BaseUpgrade) -> Self {
+        Self(upgrade)
+    }
+
+    /// Returns the wrapped Base network upgrade.
+    pub const fn upgrade(self) -> BaseUpgrade {
+        self.0
+    }
+}
+
+impl From<BaseUpgrade> for BaseSpecId {
+    fn from(upgrade: BaseUpgrade) -> Self {
+        Self(upgrade)
+    }
+}
+
+impl From<BaseSpecId> for SpecId {
+    /// Maps a Base upgrade to the EVM2 spec that governs execution at that upgrade.
+    ///
+    /// Mirrors the revm-side mapping (`base-common-chains`' `BaseUpgrade::into_eth_spec`),
+    /// duplicated here rather than reused because that mapping targets revm's `SpecId` and
+    /// pulls in revm, which this crate deliberately avoids.
+    fn from(spec: BaseSpecId) -> Self {
+        match spec.0 {
+            BaseUpgrade::Bedrock | BaseUpgrade::Regolith => Self::MERGE,
+            BaseUpgrade::Canyon | BaseUpgrade::Delta => Self::SHANGHAI,
+            BaseUpgrade::Ecotone
+            | BaseUpgrade::Fjord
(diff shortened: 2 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/common/evm2/src/spec.rs
```diff
@@ -0,0 +1,80 @@
...
+        // Matches the revm-side into_eth_spec table.
+        assert_eq!(SpecId::from(BaseSpecId::new(BaseUpgrade::Bedrock)), SpecId::MERGE);
+        assert_eq!(SpecId::from(BaseSpecId::new(BaseUpgrade::Regolith)), SpecId::MERGE);
+        assert_eq!(SpecId::from(BaseSpecId::new(BaseUpgrade::Canyon)), SpecId::SHANGHAI);
+        assert_eq!(SpecId::from(BaseSpecId::new(BaseUpgrade::Ecotone)), SpecId::CANCUN);
+        assert_eq!(SpecId::from(BaseSpecId::new(BaseUpgrade::Fjord)), SpecId::CANCUN);
+        assert_eq!(SpecId::from(BaseSpecId::new(BaseUpgrade::Isthmus)), SpecId::PRAGUE);
+        assert_eq!(SpecId::from(BaseSpecId::new(BaseUpgrade::Jovian)), SpecId::PRAGUE);
+        assert_eq!(SpecId::from(BaseSpecId::new(BaseUpgrade::Cobalt)), SpecId::OSAKA);
+    }
```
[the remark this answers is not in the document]
refcell: Removed the mapping test rather than expanding it. As written it just restated the From match arm-for-arm, which the repo's no-tautological-tests guideline discourages. The mapping is instead validated behaviorally by the differential parity harnesses (deposit + standard-fee), which sweep the forks and would fail on any mapping error via gas/state divergence; the revm-side exhaustive table test still guards the canonical source.

crates/common/evm2/src/spec.rs
```diff
@@ -0,0 +1,67 @@
...
+            BaseUpgrade::Ecotone
+            | BaseUpgrade::Fjord
+            | BaseUpgrade::Granite
+            | BaseUpgrade::Holocene
+            | BaseUpgrade::PectraBlobSchedule => Self::CANCUN,
+            BaseUpgrade::Isthmus | BaseUpgrade::Jovian => Self::PRAGUE,
+            // Azul, Beryl, Cobalt, Denim, Zenith, and newer upgrades inherit the latest known
+            // Ethereum spec until explicitly mapped.
+            _ => Self::OSAKA,
+        }
```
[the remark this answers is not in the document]
refcell: The revm-side into_eth_spec actually uses the same wildcard — crates/common/chains/src/upgrade.rs:63 is `_ => SpecId::OSAKA`, with the comment "Azul, Beryl, Cobalt, Denim, Zenith, and newer Base upgrades inherit the latest known Ethereum spec until explicitly mapped." So the two tables are structurally identical here by design, and this mapping mirrors it deliberately.

Making only the evm2 side exhaustive would diverge from the reference: a newly-added variant would hit a compile error here and force an OSAKA-or-other decision, while the revm side would keep silently defaulting it to OSAKA — so the two mappings could disagree until both are updated. Keeping the wildcard on both keeps them in lockstep (new upgrades default to the latest spec on both engines until someone explicitly maps them in both places). Leaving as-is to match the reference.

crates/common/evm2/src/handler.rs
```diff
@@ -27,11 +26,12 @@ impl BaseTxHandlerHooks {
...
             // TODO: price over the full L1-posted (EIP-2718) transaction, not just
-            // the calldata, to match the revm integration. The evm2 spec schedule
-            // is also still Ecotone-only; the Base fork schedule is layered on in
-            // follow-up work.
+            // the calldata, to match the revm integration. This requires the
+            // envelope to carry the enveloped tx bytes (evm2's `TxEnvelope` does not
+            // expose them), tracked as fee-completeness follow-up work.
             BaseTxEnvelope::Standard(tx) => {
-                host.block_env().ext.calculate_tx_l1_cost(tx.input(), BaseUpgrade::Ecotone)
+                let upgrade = host.config_spec_id().upgrade();
```
xenoliss, line 33: is it expected that nothing in this PR hits this path? the new parity tests are all deposits.

refcell: Right — expected, and deliberate. Every new fixture is a deposit, so the `Standard` arm isn't exercised here.

Two reasons it's not covered yet:

1. The parity harness is the deposit harness from #4759; the fork sweep just rides on those deposit fixtures. Deposits return `U256::ZERO` from `l1_fee`, so they exercise the fork schedule (`config_spec_id`, per-fork gas schedule) end to end but never reach this specific arm.
2. A *differential* standard-tx test can't be meaningful until fee-completeness lands: evm2 currently prices the L1 fee over calldata only (`tx.input()`), while the revm reference prices over the full EIP-2718 envelope. Until the envelope carries its enveloped bytes (the prerequisite in the scope note / the TODO right above), a standard-tx parity run would diverge on the L1 fee by construction — it wouldn't be a real parity assertion.

What this PR actually changes on this line is narrow: swapping the hardcoded `BaseUpgrade::Ecotone` for `host.config_spec_id().upgrade()` so the fee formula is fork-aware.

If you'd like that guarded now without waiting on full-envelope parity, I can add a focused (non-differential) test: run the same standard tx through evm2 at **Ecotone vs Fjord** — both map to `CANCUN`, but Fjord swaps the L1 data-fee formula (calldata `data_gas` → FastLZ `estimatedSize`) — and assert the two fees differ / match hand-computed values. That directly pins the "recover the exact upgrade from `BaseSpecId` even when the SpecId collides" behavior, which is the whole reason the wrapper exists. Want me to add it?

refcell: Yes tests are in a later pr!

xenoliss (approved):

PR #4769, opened 2026-08-28
feat(proof): populate and advertise artifact hashes (2/3)

0xth4nh:
> Part 2 of 3, splitting #4725. Stack: #4768 → **this** → 3/3.
> **Base branch is `feat/artifact-identity` (#4768), not `main`.** Review the top commit only; GitHub retargets this to `main` once #4768 merges.
> No runtime behaviour changes in this PR.

## What changed?

Producers now stamp the artifact each job requires, and workers advertise the artifact they can execute. The prover service still ignores both fields, so nothing routes yet.

**Proposer**
- New **required** `--tee-image-hash` / `BASE_PROPOSER_TEE_IMAGE_HASH`. Carried into `ProofRequest.image_hash` and into the TEE session-ID preimage.

**Challenger**
- Reads TEE and ZK commitments from each individual game proxy via `proof_artifacts()`, so a historical game keeps requesting the artifacts it was created against rather than whatever the factory currently points at.
- Both TEE and ZK session IDs now commit to the artifact hash.

**Workers**
- Nitro host advertises its local enclave image hash.
- ZK host advertises the composite hash derived from its embedded verification keys.

### Two things worth a closer look

**Session IDs now include the artifact hash.** This is deliberate — it stops a proof produced under one artifact generation from being reused for a game expecting another. It also means every derived session ID changes on deploy. In-flight sessions keyed by the old ID will not be found; see the ordering note below.

**The Nitro host retries instead of exiting.** Reading the enclave image hash happens before polling starts. The enclave is frequently still booting when the host starts, and returning there would drop the worker from the pool permanently rather than for the few seconds the enclave needs. It now retries on the configured poll interval and only gives up on cancellation.

## Why?

Exact artifact routing needs the request to say which artifact it requires and the worker to say which one it has. Splitting this from the enforcement in 3/3 means the config rollout — particularly the new required proposer env var — gets its own deploy window, before any behaviour depends on it.

## How to test?

```sh
cargo test -p base-proposer -p base-challenger -p base-proof-worker \
           -p base-proof-tee-nitro-host -p base-proof-zk-host --lib

cargo test -p base-challenger --test driver
```

Targeted checks for the behaviour this PR adds:

```sh
# Session IDs must separate artifact generations
cargo test -p base-challenger --lib challenger_session_ids_separate_artifact_hashes
cargo test -p base-proposer   --lib tee_session_id_changes_with_image_hash

# Workers put their advertised hash on every claim request
cargo test -p base-proof-worker --lib config_builds_nitro_claim_request
```

Manual check that the proposer refuses to start without the new variable:

```sh
unset BASE_PROPOSER_TEE_IMAGE_HASH
cargo run -p base-proposer -- --game-type 1 ...   # expect a clap error naming --tee-image-hash
```

## Deployment

⚠️ **`BASE_PROPOSER_TEE_IMAGE_HASH` is required and has no default — the proposer will not start without it.** Set it to the `TEE_IMAGE_HASH()` of the currently deployed `AggregateVerifier` implementation before rolling this out. See the Artifact Pinning section added to `crates/proof/proposer/README.md`.

Because session IDs change, drain or allow in-flight proposer/challenger sessions to complete across the restart rather than expecting them to resume.

Still a no-op server-side: the prover service does not read `zk_artifact_hash` or `supported_artifact_hash` until 3/3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5577,6 +5577,7 @@ dependencies = [
 name = "base-proof-worker"
 version = "0.0.0"
 dependencies = [
+ "alloy-primitives",
  "async-trait",
  "backon",
  "base-prover-service-client",
@@ -6823,7 +6824,7 @@ dependencies = [
  "bitflags 2.13.1",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -6843,7 +6844,7 @@ dependencies = [
  "bitflags 2.13.1",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
+ "itertools 0.13.0",
  "proc-macro2",
  "quote",
  "regex",
@@ -9989,8 +9990,8 @@ dependencies = [
  "libc",
  "log",
  "rustversion",
- "windows-link 0.1.3",
- "windows-result 0.3.4",
+ "windows-link 0.2.1",
+ "windows-result 0.4.1",
 ]
 
 [[package]]
@@ -10895,7 +10896,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.5",
  "system-configuration",
  "tokio",
  "tower-layer",
@@ -10926,7 +10927,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0b1e4910d3a9137442723dfb772c32dc10674c4181ca078d2fd227cd5dce9db0"
 dependencies = [
  "bincode 1.3.3",
- "derive_more 1.0.0",
+ "derive_more 2.1.1",
  "iai-callgrind-macros",
  "iai-callgrind-runner",
 ]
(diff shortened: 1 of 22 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4773, opened 2026-08-28
fix(system-tests): stabilize in-process stack setup

haardikk21:
## Summary

- align the in-process builder payload retention window with the production builder so delayed sealing does not expire transaction-bearing payloads
- run the isolated Lighthouse beacon node as a supernode so Fulu blob reconstruction has all required data columns
- make the L1 reorg scenario use its intended slot duration and tolerate the safe label being temporarily unavailable during startup
- pause active sequencing while a late shadow derives its historical prefix, then resume after the shadow has joined
- build the setup image under the tag consumed by system tests

## Context

The recurring receipt timeouts were not caused by transaction fees or unavailable RPC nodes. The builder accepted and executed the transactions, but L1/batcher pressure delayed sealing past the in-process builder's 500 ms payload leeway. The subsequent payload lookup returned `Unknown payload`, and consensus replaced it with an empty canonical block.

The remaining failures exposed independent test-stack assumptions around PeerDAS blob custody, L1 slot timing, and late-shadow unsafe gossip.

### DIFF
--- a/etc/systems/src/l2/in_process_consensus.rs
+++ b/etc/systems/src/l2/in_process_consensus.rs
@@ -355,6 +355,15 @@ impl InProcessConsensus {
         Ok(())
     }
 
+    /// Stops the sequencer via the `admin_stopSequencer` RPC.
+    pub async fn stop_sequencer(&self) -> Result<()> {
+        let client = HttpClientBuilder::default()
+            .build(self.rpc_url().as_str())
+            .wrap_err("Failed to build RPC client")?;
+        client.admin_stop_sequencer().await.wrap_err("Failed to stop sequencer via RPC")?;
+        Ok(())
+    }
+
     /// Returns the RPC URL for this consensus node (localhost).
     pub fn rpc_url(&self) -> Url {
         Url::parse(&format!("http://{}:{}", self.rpc_addr.ip(), self.rpc_addr.port()))
--- a/etc/systems/src/l2/in_process_consensus.rs
+++ b/etc/systems/src/l2/in_process_consensus.rs
@@ -30,7 +30,7 @@ use base_upgrade_signal::{
     UpgradeSignalRuntimeApplier,
 };
 use eyre::{Result, WrapErr};
-use jsonrpsee::http_client::HttpClientBuilder;
+use jsonrpsee::http_client::{HttpClient, HttpClientBuilder};
 use tempfile::TempDir;
 use tokio::{
     task::JoinHandle,
@@ -95,6 +95,7 @@ pub struct InProcessConsensusConfig {
 /// A running in-process consensus node.
 pub struct InProcessConsensus {
     rpc_addr: SocketAddr,
+    rpc_client: HttpClient,
     p2p_tcp_port: u16,
     peer_id: String,
     _checkpoint_dir: TempDir,
@@ -277,8 +278,13 @@ impl InProcessConsensus {
             }
         }
 
+        let rpc_client = HttpClientBuilder::default()
+            .build(format!("http://{rpc_addr}"))
+            .wrap_err("Failed to build RPC client")?;
+
         Ok(Self {
             rpc_addr,
+            rpc_client,
             p2p_tcp_port,
             peer_id,
             _checkpoint_dir: checkpoint_dir,
@@ -291,13 +297,9 @@ impl InProcessConsensus {
     /// Retries the connection attempt because the P2P layer may not be fully ready immediately
     /// after the RPC endpoint becomes reachable.
     pub async fn connect_peer(&self, multiaddr: &str) -> Result<()> {
-        let client = HttpClientBuilder::default()
-            .build(self.rpc_url().as_str())
-            .wrap_err("Failed to build RPC client")?;
-
         let mut last_err = None;
         for attempt in 0..20 {
-            match client.opp2p_connect_peer(multiaddr.to_string()).await {
+            match self.rpc_client.opp2p_connect_peer(multiaddr.to_string()).await {
                 Ok(()) => {
                     info!(attempts = attempt + 1, "peer connected");
                     return Ok(());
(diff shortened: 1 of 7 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

etc/systems/src/l2/stack.rs
```diff
@@ -474,26 +478,35 @@ impl L2Stack {
...
+                            < shadow_start_safe_height
+                        {
                             sleep(Duration::from_millis(100)).await;
                         }
                         Ok::<_, eyre::Error>(())
                     })
                     .await
                     .wrap_err("Timed out waiting for delayed shadow safe catch-up")??;
                     batcher.as_ref().expect("delayed batcher started").stop();
+                    builder_consensus.start_sequencer().await.wrap_err(
```
meyer9, line 507: Does the sequencer start in the case that there is no shadow sequencer enabled?

etc/systems/tests/l1_reorg_recovery.rs
```diff
@@ -169,10 +170,10 @@ async fn wait_for_reorgable_origin(
     timeout(ORIGIN_TIMEOUT, async {
         loop {
             let status = rpc.l2_builder_sync_status().await?;
-            let safe = l1_provider
-                .get_block_by_number(BlockNumberOrTag::Safe)
-                .await?
-                .ok_or_eyre("L1 safe block is missing")?;
+            let Some(safe) = l1_provider.get_block_by_number(BlockNumberOrTag::Safe).await? else {
+                sleep(POLL_INTERVAL).await;
+                continue;
```
meyer9, line 175: should we add a timeout here so it doesn't go forever?

meyer9 (dismissed):
nice finds! just a few small questions/comments

cb-heimdall:
Review Error for meyer9 @ 2026-08-28 19:37:45 UTC

etc/systems/src/l2/in_process_consensus.rs
```diff
@@ -355,6 +355,15 @@ impl InProcessConsensus {
         Ok(())
     }
 
+    /// Stops the sequencer via the `admin_stopSequencer` RPC.
+    pub async fn stop_sequencer(&self) -> Result<()> {
+        let client = HttpClientBuilder::default()
+            .build(self.rpc_url().as_str())
+            .wrap_err("Failed to build RPC client")?;
```
refcell: Quite redundant that we're instantiating a new client in every method here

etc/systems/tests/l1_reorg_recovery.rs
```diff
@@ -169,10 +170,10 @@ async fn wait_for_reorgable_origin(
...
+            };
```
refcell, line 176: This should probably just be a provider with retry and backoff middleware policies

etc/systems/src/l2/stack.rs
haardikk21: yes, the stop+start is just a no-op in that case

etc/systems/tests/l1_reorg_recovery.rs
haardikk21: good point, will do

etc/systems/src/l2/in_process_consensus.rs
haardikk21: will fix

haardikk21: Fixed in b1d0ae9af. `InProcessConsensus` now creates one `HttpClient` after RPC readiness and reuses it for peer connection and sequencer admin calls.

etc/systems/src/l2/stack.rs
haardikk21: Yes. When no shadow sequencer is configured, `delayed_shadow` is false and the final `if !delayed_shadow` branch starts the active sequencer. Correction to the earlier reply: the stop/start cycle only runs for delayed-shadow startup; it is not what handles the no-shadow case.

etc/systems/tests/l1_reorg_recovery.rs
haardikk21: This loop is already bounded by the surrounding `timeout(ORIGIN_TIMEOUT, async { ... })`; on expiry it returns the contextual `sequencer did not reach an L1 origin deep enough to reorg` error. No additional nested timeout is needed.

etc/systems/tests/l1_reorg_recovery.rs
haardikk21: Retry/backoff middleware would retry transport/RPC errors, but the startup condition here is a successful `Ok(None)` response while the safe label is not yet available. The explicit polling loop handles that condition and is bounded by the surrounding `ORIGIN_TIMEOUT`; actual RPC errors remain surfaced immediately.

BrianBland (approved):

meyer9 (approved):

PR #4779, opened 2026-08-30
perf(protocol): reuse brotli decompression scratch across channels

jackchuma:
## Summary

`Brotli::decompress` allocated and zeroed ~52 `MiB` of scratch memory on **every call**, then dropped it. Post-Fjord, brotli is the channel compression algorithm, so this ran once per channel — roughly once per L1 block — for the whole of derivation, in both the node and the fault-proof/ZK paths.

The scratch is large because the `brotli` crate is handed a stack allocator over three fixed pools rather than a general-purpose heap allocator, so each pool has to be sized for the large-window worst case up front: a 32 `MiB` ring buffer, 4 `MiB` of `u32` tables and 16 `MiB` of Huffman tables. **None of it depends on the channel being decompressed**, so it is now allocated once and reused.

`Brotli` becomes a stateful decompressor that owns the pools and allocates them lazily on first use. A `BatchReader` lives for exactly one channel and is therefore the wrong owner, so the decompressor is held by `ChannelReader` — which lives for the whole pipeline — and threaded down as `&mut`.

Worth noting how much of that 52 `MiB` was pure waste: at the batcher's `lgwin: 22` the decoder's ring buffer is 4 `MiB`, and `alloc_no_stdlib` re-zeroes only the sub-slice it actually carves out. So the pools were sized for a worst case the batcher never produces.

Fixes F1 in the internal performance-opportunities review.

## Measured improvement

Criterion, `cargo bench -p base-protocol --bench decompress`. "Before" is a run of the same benchmark against the parent commit in a separate worktree, adapted to the old API.

| Benchmark | Before | After | Change |
|---|---|---|---|
| `channel/small` (12 `KiB` mainnet channel) | 1.6272 ms | **59.049 µs** | **27.6x faster** (−96.4%) |
| `channel/large` (1.3 `MiB` channel) | 6.2844 ms | **4.7231 ms** | **1.33x faster** (−24.8%) |

Throughput on the small channel goes from 7.33 `MiB`/s to 201.98 `MiB`/s. The small channel is the telling case: the payload is ~12 `KiB`, so before this change **over 96% of the time was spent on memory the decoder never needed**.

Allocations per `decompress` call, measured with a counting `GlobalAlloc`:

| | Allocations | Bytes |
|---|---|---|
| Before (fresh decompressor per channel) | 5 | 54,550,910 |
| After (reused decompressor) | **2** | **24,958** |

A 99.95% reduction in bytes allocated per call; the two remaining allocations are the output buffer and its one growth step.

### Benchmark methodology

The benchmark carries a third arm, `channel/small_cold_scratch`, which constructs a fresh decompressor per iteration and so reproduces the old behaviour inside the new binary. It measures **1.6369 ms**, within 0.6% of the 1.6272 ms cross-commit baseline. The two independent measurements of "before" agreeing that closely is the main evidence that the speedup is real and not a measurement artifact. That arm is kept permanently as a guard: if the other arms ever converge on it, scratch reuse has regressed.

Both benches are registered in CI's advisory benchmark runs (`run_bench_subset.sh`, `run_iai_subset.sh`) and a matching `decompress_iai` iai-callgrind bench is included for deterministic instruction counts.

## Scope of the win — please read before inferring a sync speedup

The headline 27x is a decompression-level number and **should not be read as a node-sync number**. `BatchReader::decompress` short-circuits once `self.decompressed` is populated, so decompression happens once per *channel* and is then amortized across every L2 block that channel carries. Measured on the real fixtures in this repo, timing the whole per-channel path a derivation stage runs (decompress, then decode every batch out of the channel):

| Channel | L2 blocks | Before | After | Saved | Decompression share |
|---|---|---|---|---|---|
| Mainnet fixture (1 single batch) | 1 | 1.801 ms | 0.062 ms | 1.739 ms | 96.6% |
| Span-batched (`batch.hex` re-compressed) | 143 | 9.753 ms | 8.196 ms | 1.557 ms | 16.0% |

Per L2 block that is **10.9 µs** saved in the span-batched case, which is how Base actually batches today, against a post-change per-block channel cost of 57.3 µs. Set against L1/blob fetching and the engine round-trip that dominate a syncing node, **this change is not expected to move sync-to-tip time measurably**, and no claim to that effect is being made here. The saving is per-channel and constant, so it scales with channel count, not with chain size.

Where it does carry weight:

- **ZK range proofs.** A range proof covers `block_interval = 512` L2 blocks, and the pipeline — hence `ChannelReader`, hence `Brotli` — is constructed exactly once per proof and survives `FlushChannel`/reset, so the scratch is reused across every channel in the span. In the zkVM those 52 `MiB` are ~52 `MiB` of *proven stores* per channel, and the 4 Mi `HuffmanCode` pool misses Rust's zero-fill specialization so it is written element by element. The binding SP1 limits are `range_cycle_limit`/`range_gas_limit`, so this is cycle headroom. **Not yet quantified in cycles** — `b20_zk_proving` can measure it but needs a live local L2/rollup/prover-service stack and real SP1 ELFs.
- **Allocator and memory-bandwidth pressure**, removed from the per-channel path. The 1.56 ms measured here is the direct cost; touching 52 `MiB` also evicts cache for neighbouring work, which this benchmark cannot see.

## Trade-off

A pipeline that decompresses at least one brotli channel now holds ~52 `MiB` of scratch for its lifetime instead of transiently. This is bounded and intentional: one allocation per pipeline, and the peak was already being hit before — just repeatedly, with the allocator returning it to the OS in between.

The case to sanity-check is the TEE prover, where each `Server::prove` builds its own pipeline: at mainnet's 8 parallel proofs that is ~416 `MiB` held persistently rather than transiently, against a 100 `GiB` enclave. Comfortable, but it is a real change in residency.

`Brotli::new()` is `const` and allocates nothing, so holders that never see a brotli channel (zlib-only paths, and every test using the zlib fixtures) pay nothing.

## Pre-existing observability bug found along the way (not fixed here)

`ChannelReader::next_batch` calls `decompress` on every batch, and it is a no-op after the first, so `pipeline_batch_decompress_duration_seconds` is dominated by nanosecond-scale no-op samples. Its `_sum` is valid but every percentile, mean and `_count`-derived figure is not, and there is no channels-processed counter to normalize against. Worth a follow-up if anyone wants to measure decompression's real share in production.

## Test plan

- [x] `cargo test -p base-protocol -p base-consensus-derive -p base-comp --lib` — 471 tests pass
- [x] `cargo check --workspace --all-targets` — clean
- [x] `cargo clippy -p base-protocol -p base-consensus-derive -p base-comp --all-targets` — no warnings
- [x] `cargo +nightly fmt --all --check` — clean
- [x] `no_std` compatibility verified via `check-no-std.sh` (`riscv32imac-unknown-none-elf`)
- [x] Before/after benchmarks recorded above, cross-validated against an in-binary control arm
- [ ] ZK cycle delta on a 512-block range proof (needs the local prover stack)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #4781, opened 2026-08-31
feat(challenger-e2e): assert every dispute path on the fork

0xth4nh:
Follow-up to #4785, which landed the harness and the positive case. This PR adds the dispute paths.

## What changed?

Extends `ChallengerE2e::run()` from "the challenger leaves valid games alone" to "the challenger disputes every classifier path we can stage honestly, and nothing else".

- **Path 1 `InvalidTeeProposal`.** Corrupts game A's last intermediate output root via `Checkpoint::patch`, then waits for the challenger to either clear `teeProver()` or set `zkProver()` + `counteredByIntermediateRootIndexPlusOne()`. Both are correct — the challenger tries TEE first and falls back to ZK — so insisting on `nullify` would fail the run whenever the TEE prover is briefly down.
- **Game A settles.** One quiet window on A, whichever way Path 1 landed. After a ZK challenge this is **Path 2 skip**: a legitimate challenge of a wrong TEE root must stand, and a challenger that "defends" it fails here. After a TEE nullify there is no challenge to leave standing, so the same window instead proves **idempotence** — the challenger must not re-dispute a game it has already nullified. Checking both branches is what makes idempotence covered on every run rather than on the roughly half that land as ZK challenges.
- **Path 4 `InvalidDualProposal`, then whichever path Path 4 leaves behind.** Before the challenger is released, key A requests a real SNARK of game B's *canonical* roots from the live prover-service and submits `verifyProposalProof`, giving B the dual-proof shape (`zkProver != 0`, `counteredIndex == 0`). B is then patched. A dual-proof game takes two disputes to clear, and the challenger is free to drop either proof first — so which assertion runs second is decided by what the first one observed, not fixed in advance. TEE first (`tee=0`, `zk!=0`) leaves a ZK-only game the next scan nullifies as **Path 3 `InvalidZkProposal`**. ZK first is the supported **TEE-fallback** case, where the TEE request or submission failed; that leaves a TEE-only game the next scan disputes as **Path 1**, by nullify or by challenge. Demanding the TEE proof go first would sit out the whole `CHALLENGER_E2E_DISPUTE_TIMEOUT` on a challenger doing exactly what it is supposed to.
- **Nonce baselines are sampled before the patch, not after.** `Checkpoint::patch` mutates the game and then reads it back several times to self-verify. A challenger that scans during those readbacks would already have spent the nonce the baseline is meant to precede, and its correct dispute would read as nobody's.
- **No collateral damage.** Every other game in the lookback window is snapshotted before the challenger boots and re-read at the end; all must still show the same `(teeProver, zkProver, counteredIndex)`. Every assertion above is scoped to A or B, so without this a challenger misconfigured on `game_type`, or one that starts disputing indiscriminately after its first dispute, would pass the run.

Every dispute is attributed by a nonce delta on the challenger key. Key A signs setup only; key B is the only key permitted to dispute.

**Path 2 *dispute* is deliberately not staged.** A fraudulent ZK challenge against a correct TEE root requires a wrong-root proof the real verifier accepts, which the real prover cannot produce. That half stays in the mock driver tests.

## Why?

Patching an existing game rather than creating one is what keeps this honest. The games were created and verified on the real chain before the fork point, so the `TEEVerifier`, the `TEEProverRegistry` and the game bytecode are all real — the challenger's dispute proof is genuinely verified onchain rather than waved through by a stub.

#4785 proves the challenger does not fire when it shouldn't. This PR proves it fires when it should, on every path the classifier can reach, and only on the games it was given. Together they are the gate: a zeronet challenger image cannot be promoted to sepolia unless the same release image has done both against a throwaway fork of the L1 it will run on.

## How to test?

`base-test-utils` `sol!`-includes Foundry artifacts, so the contracts have to exist before cargo will even compile the workspace — that is the `_build-contracts` step CI runs ahead of clippy:

```
(cd crates/utilities/test-utils/contracts && forge soldeer install && forge build)
BASE_SUCCINCT_ELF_STUB=1 cargo clippy -p base-challenger-e2e -p base-challenger-e2e-bin \
  --all-features --all-targets -- -D warnings
cargo test -p base-challenger-e2e
```

End to end needs a challenger pod; the topology lives in `protocols/base-proofs` under `chart/challenger-e2e`. Locally, with the same env as #4785 plus a reachable prover-service on `BASE_CHALLENGER_ZK_RPC_URL`:

```
cargo run -p base-challenger-e2e-bin
```

Expected log order — the Path 4 SNARK is a real proof, so budget `CHALLENGER_E2E_DISPUTE_TIMEOUT` (default 45m) for it:

```
spawning L1 fork
selected game (x2)
snapshotted games the challenger must not touch
requesting SNARK of canonical roots to stage Path 4
staged Path 4 dual-proof game; counteredIndex is 0
released the challenger onto the fork
the challenger left every valid game alone
corrupted intermediate output root; waiting for the challenger to dispute
the challenger disputed the corrupted game   outcome=nullified via TEE proof | challenged via ZK proof
observing the settle window ... the settle claim held
corrupted dual-proof game; waiting for Path 4
```

then whichever branch the challenger took — TEE proof dropped first:

```
Path 4: TEE proof nullified, ZK proof remains
Path 3: ZK proof nullified
the challenger touched no game it was not given
```

or the ZK fallback, when the TEE request or submission failed:

```
Path 4: ZK fallback nullified, TEE proof remains
the challenger disputed the corrupted game   outcome=nullified via TEE proof | challenged via ZK proof
Path 1: the remaining TEE proof was disputed
the challenger touched no game it was not given
```

To watch the collateral-damage check bite, widen `CHALLENGER_E2E_GAME_LOOKBACK` and point the challenger at a different `BASE_CHALLENGER_GAME_TYPE` than the driver: the run fails with `the challenger moved game 0x…, which this test never corrupted`.

To watch the settle window bite, stop the run after Path 1 and manually re-`nullify` game A — `Path 2 skip: a legitimate ZK challenge must stand — game 0x… moved from … to …`.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,665 @@
...
+        );
+
+        for metric in [
+            "base_challenger_games_invalid_total",
+            "base_challenger_nullify_tx_submitted_total",
+            "base_challenger_challenge_tx_submitted_total",
+        ] {
+            let delta = after.sum(metric) - before.sum(metric);
+            ensure!(
+                delta == 0.0,
```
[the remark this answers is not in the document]
0xth4nh: Keeping `==` here, and not parsing as `u64` — both suggestions would make this worse:

- **`f64::EPSILON` is not a tolerance.** It is the ULP at 1.0 (2.22e-16), not a scale-relative bound. Any real counter increment is >= 1.0, so `delta.abs() < f64::EPSILON` returns exactly what `delta == 0.0` returns; it adds no robustness and hides that the comparison is exact.
- **The comparison genuinely is exact.** Prometheus counters here are integers far below 2^53, so each is exactly representable in f64, and the difference of two exactly-representable integers whose result is also below 2^53 is computed exactly. There is no precision loss to guard against.
- **`u64` in `Scrape` would break the scraper.** The text-exposition format is float64 by spec — histogram `_sum` series are legitimately fractional, and the challenger exports three histograms (`validation_latency_seconds`, `nullify_tx_latency_seconds`, `challenge_tx_latency_seconds`). Narrowing the parser to `u64` would fail on them.

`clippy::float_cmp` is pedantic and not in the workspace's `clippy::all`, so it did not fire — but the reason to keep this is that the comparison is exact, not that the lint is off.

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,757 @@
...
+            }
+        };
+
+        let before = Self::read_game_state(verifier, game.address).await?;
+        info!(
+            game = %game.address,
+            window = ?config.quiet_window,
+            state = ?before,
+            "observing {claim}"
+        );
```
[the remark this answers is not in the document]
0xth4nh: Taken. The claim is a field now, not part of the message:

```rust
info!(game = %game.address, window = ?config.quiet_window, state = ?before, claim, "observing the settle window");
...
info!(game = %game.address, claim, "the settle claim held");
```

Note this PR is being split in two — the code this thread points at moves to the follow-up PR in the stack, and the fix travels with it.

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -335,6 +464,275 @@ impl ChallengerE2e {
...
+
+        let mut nonce = provider.get_transaction_count(challenger.address()).await?;
+
+        Self::poll_until(
+            config,
+            config.dispute_timeout,
+            "the challenger to TEE-nullify the dual-proof game",
+            || async {
+                Ok((verifier.tee_prover(game.address).await? == Address::ZERO).then_some(()))
+            },
```
jackchuma: Handle the supported dual-proof ZK fallback. When the TEE request or submission for an `InvalidDualProposal` fails, the challenger intentionally falls back to a ZK `nullify` (covered by `test_step_dual_proof_invalid_tee_request_failure_falls_back_to_zk_nullify` in `crates/proof/challenge/tests/driver.rs`). That supported path does not guarantee `teeProver()` becomes zero first, so this predicate can wait for the entire `dispute_timeout` and fail a correctly behaving challenger. Either require a reliably available TEE prover for this scenario or accept and assert the documented ZK-fallback transition

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -335,6 +464,275 @@ impl ChallengerE2e {
...
+        verifier: &AggregateVerifierContractClient,
+        provider: &RootProvider,
+        driver: &PrivateKeySigner,
+        challenger: &PrivateKeySigner,
+        game: Candidate,
+    ) -> Result<Path1Outcome> {
+        let fork_config = Self::fork_config(config, fork_url, driver, game);
+        let checkpoint = Checkpoint::patch(&fork_config, verifier)
+            .await
+            .context("failed to corrupt an intermediate output root on the fork")?;
```
jackchuma, line 487: Capture the nonce before mutating the game. `Checkpoint::patch` mutates the game before returning and then performs several readbacks. A fast challenger can observe the mutation and submit its transaction during those readbacks, before `await_dispute` samples `nonce_before`; the state predicate then succeeds but the unchanged sampled nonce makes `assert_challenger_acted` reject the valid action. Read the challenger nonce before patching and pass that baseline into the waiter; `run_path4_then_3` has the same post-patch sampling order

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Confirmed, and worse than a slow path — it is a state the assertion can never reach. `nullify` carries the proof type in the first calldata byte, so the ZK fallback clears `zkProver` and leaves `(tee≠0, zk=0, ci=0)`, which the scanner classifies as `InvalidTeeProposal`. Waiting on `teeProver() == 0` would burn the full `dispute_timeout` on a challenger doing exactly what `test_step_dual_proof_invalid_tee_request_failure_falls_back_to_zk_nullify` documents.

Took the second option in 70468df — requiring a TEE prover would reintroduce the flake that Path 1 already accepts both outcomes to avoid. The step now waits for *either* proof to disappear and branches on which one did: TEE first is Path 3 as before, ZK first reuses `await_dispute`, so the remaining TEE-only game is accepted by nullify or by challenge. Both branches still assert a nonce delta, so it is two attributable disputes either way.

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Confirmed. `AnvilPatch::apply` mutates through the cheatcodes and then re-reads `intermediate_output_roots` and `game_info` to verify itself, so there is a real window between the game becoming invalid and the baseline being sampled — and the failure mode is the bad one: a correct dispute reported as "something other than the challenger disputed it".

Fixed in 70468df for both call sites. `run_path1` and `run_path4_then_3` sample the nonce before `Checkpoint::patch`, and `await_dispute` now takes the baseline as a parameter rather than sampling its own.

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -335,6 +465,297 @@ impl ChallengerE2e {
...
+            window = ?config.quiet_window,
+            state = ?before,
+            claim,
+            "observing the settle window"
+        );
+        tokio::time::sleep(config.quiet_window).await;
+
+        let after = Self::read_game_state(verifier, game.address).await?;
+        ensure!(
+            after == before,
```
jackchuma, line 559: Reject repeated reverted disputes during the settle window. When the challenger incorrectly keeps submitting a dispute after Path 1 has already completed, those calls can revert without changing any of the three fields in `GameState`. This equality check therefore passes even though the challenger is still spending transactions to re-challenge a legitimate ZK challenge or re-nullify an already-nullified game. Capture the challenger nonce (or the submitted-transaction counters) before the quiet window and require it to remain unchanged

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -335,6 +465,297 @@ impl ChallengerE2e {
...
+            config,
+            config.dispute_timeout,
+            "the challenger to dispute the corrupted game",
+            || async {
+                if verifier.tee_prover(game).await? == Address::ZERO {
+                    return Ok(Some(Path1Outcome::TeeNullify));
+                }
+                let countered = verifier.countered_index(game).await? != 0;
+                if countered && verifier.zk_prover(game).await? != Address::ZERO {
+                    return Ok(Some(Path1Outcome::ZkChallenge));
```
jackchuma, line 788: Verify the challenged checkpoint index. For the ZK fallback, any nonzero `countered_index` is treated as success, although this run always corrupts the last checkpoint (`game.root_count - 1`) and should therefore produce `countered_index == game.root_count`. A challenger that submits an accepted proof against a different checkpoint would pass this E2E path without disputing the corrupted root; carry the expected index into this assertion and also verify that `zk_prover` is the challenger address

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -335,6 +465,297 @@ impl ChallengerE2e {
...
+        // game has become, and so which path must clear the remainder.
+        let tee_first = Self::poll_until(
+            config,
+            config.dispute_timeout,
+            "the challenger to nullify one of the dual-proof game's two proofs",
+            || async {
+                if verifier.tee_prover(game.address).await? == Address::ZERO {
+                    return Ok(Some(true));
+                }
+                Ok((verifier.zk_prover(game.address).await? == Address::ZERO).then_some(false))
```
jackchuma: Handle both dual-proof transitions in one observation. If the challenger polls faster than `CHALLENGER_E2E_POLL_INTERVAL` and both proofs are immediately available, it can clear both proofs before this first poll runs. `tee_prover == 0` then selects the TEE-first branch, but the nonce sampled at line 660 already includes both transactions; the subsequent ZK poll returns immediately and the second nonce assertion incorrectly requires a third transaction. Read both prover fields as one state transition and accept the already-cleared second proof with a two-transaction nonce delta

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Fixed in 324a16b1a: the expected index is carried in and checked, along with the challenger's identity.

`fork_config` sets `invalid_index: Some(candidate.root_count - 1)` and `counteredByIntermediateRootIndexPlusOne` is 1-based, so the expected value is `checkpoint.index + 1` — taken from the `Checkpoint` the patch returned rather than re-derived from `root_count`, so it stays correct if the corrupted index ever stops being the last one.

`await_dispute` gained an `expected_countered: u64` parameter, and both call sites pass `checkpoint.index + 1`:

```rust
if matches!(outcome, Path1Outcome::ZkChallenge) {
    let countered = verifier.countered_index(game).await?;
    let zk_prover = verifier.zk_prover(game).await?;
    ensure!(countered == expected_countered, "...");
    ensure!(zk_prover == challenger.address(), "...");
}
```

Deliberately outside the poll loop rather than folded into the predicate: `poll_until` swallows a predicate error and retries it (`warn!(error = %error, "poll failed; retrying")`), so an `ensure!` in there would surface as a 45m timeout instead of naming the wrong index. The message uses `saturating_sub(1)` to report 0-based indices, since `ensure!` only evaluates its format arguments on the failure path and a re-read of zero would otherwise underflow.

`zkProver()` is documented as "the address that provided a ZK proof", and `bond.rs:202` matches it against the challenger's own `claim_addresses`, so equality with key B is the right check here.

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Fixed in 324a16b1a: the settle window now freezes B's nonce as well as the game state.

Confirmed — `GameState` is `(tee_prover, zk_prover, countered_index)`, and a reverted `nullify` or `challenge` moves none of them, so the equality check passed while the challenger burned transactions.

Took the nonce rather than the submitted-tx counters: it needs no metrics scrape, and it is strictly wider — it catches a misbehaving dispute the challenger never got as far as counting, and any other transaction it had no business sending.

Requiring it to be *unchanged* rather than bounded is safe for the whole window: game B is staged but still valid at this point (it is not patched until `run_path4_then_3`), the bystanders were never touched, and game A is settled — after a TEE nullify it reads `(0, 0, 0)`, which `Scanner::classify` does not match at all, and after a ZK challenge it is a legitimate `FraudulentZkChallenge` the challenger is supposed to skip. So there is nothing on the fork it may act on. A tx-manager fee bump replaces at the same nonce, so resubmission does not trip this either.

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Fixed in 324a16b1a: both fields are read in one observation, and attribution moved to a single end-of-step delta.

Confirmed, and the nonce half is the sharper edge — the first `assert_challenger_acted` had already consumed both transactions into its return value, so the second one demanded a third that was never coming. It would have burned the full `dispute_timeout` and then failed a correct challenger.

```rust
let (tee_cleared, zk_cleared) = Self::poll_until(..., || async {
    let tee = verifier.tee_prover(game.address).await? == Address::ZERO;
    let zk = verifier.zk_prover(game.address).await? == Address::ZERO;
    Ok((tee || zk).then_some((tee, zk)))
}).await?;
```

Three branches now: both cleared is logged and done, TEE-only falls through to the Path 3 poll, ZK-only to `await_dispute` for the remaining TEE proof.

The per-step `assert_challenger_acted` calls are gone in favour of one check against the baseline taken before `Checkpoint::patch`:

```rust
ensure!(nonce_after >= nonce + 2, "...");
```

Per-step deltas cannot be made race-free here — even in the TEE-only branch the nonce is sampled after the state read, so the challenger can slip the second nullify into that gap and reproduce the same off-by-one. The invariant that actually matters is "two attributable disputes cleared this game", and that holds regardless of how the observations interleave.

This builds on 70468df from your earlier comment on the same function; that fix handled either proof going first but still assumed the driver would see the intermediate state.

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -335,6 +466,370 @@ impl ChallengerE2e {
...
+        snapshot: &[(Address, GameState)],
+    ) -> Result<()> {
+        for (game, before) in snapshot {
+            let after = Self::read_game_state(verifier, *game).await?;
+            ensure!(
+                after == *before,
+                "the challenger moved game {game}, which this test never corrupted: \
+                 {before:?} -> {after:?}"
+            );
+        }
```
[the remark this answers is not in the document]
0xth4nh: Fixed in 4cab61a35, but not by taking the suggestion — the asymmetry is deliberate, and I kept it. The diagnostic complaint underneath it is fair, so that is what changed.

The two calls are answering different questions:

- `snapshot_bystanders` decides **what to watch**. A game whose prover fields do not read is a different verifier shape, so the challenger cannot move it through `(teeProver, zkProver, counteredIndex)` at all. Leaving it out costs nothing.
- `assert_bystanders_untouched` decides **whether the run passes**. Every game in the snapshot already read cleanly once, so it is the right shape by construction — a read that fails here is the RPC, not a shape mismatch. The premise "readable at snapshot time but becomes unreadable at assertion time" is exactly the case where the quoted rationale does not apply.

So `continue` here would silently drop a game from the only assertion that catches a challenger disputing indiscriminately, and report the drop as a pass. This is the last gate before a sepolia promote; it should fail closed.

What was right in the finding is that a raw RPC error is a bad way to fail. That part is fixed:

```rust
let after = Self::read_game_state(verifier, *game).await.with_context(|| {
    format!(
        "failed to re-read bystander game {game}; it read cleanly when snapshotted, \
         so the collateral-damage check could not be completed"
    )
})?;
```

Now it is distinguishable from "the challenger moved game X" at a glance, which was the real problem.

The README wording is what invited the suggestion — it said unreadable games are "skipped, not fatal" under the step 7 heading, without saying the skip happens at snapshot time. Reworded in the same commit, along with the equivalent line in the design doc, and the function now carries a one-line note that the leniency does not carry over.

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -335,6 +466,382 @@ impl ChallengerE2e {
...
+    ) -> Result<()> {
+        for (game, before) in snapshot {
+            // Not skipped on a read failure, unlike the snapshot pass. Every
+            // game in here already read cleanly once, so it is the shape this
+            // check watches; a read that fails now is the RPC, and continuing
+            // past it would drop a game from the only assertion that catches a
+            // challenger disputing indiscriminately. Fail closed and say why.
+            let after = Self::read_game_state(verifier, *game).await.with_context(|| {
+                format!(
+                    "failed to re-read bystander game {game}; it read cleanly when snapshotted,                      so the collateral-damage check could not be completed"
```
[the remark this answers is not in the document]
0xth4nh: Fixed in 28aa9120c. Correct, and the source is worse than described — the backslash was not merely missing, the two lines had already been joined into one physical line with the indentation baked in:

```rust
"failed to re-read bystander game {game}; it read cleanly when snapshotted,                      so the collateral-damage check could not be completed"
```

That is why `rustfmt --check` and clippy both stayed quiet: to the compiler it was a single well-formed literal, just one with 22 spaces in the middle. My own tooling ate the continuation when I wrote 4cab61a3 — the neighbouring `ensure!` in the same function kept its `\` because it predates that edit.

Verified the fix by compiling the literal standalone rather than by eye:

```
[failed to re-read bystander game 0xabc; it read cleanly when snapshotted, so the collateral-damage check could not be completed]
```

Also swept the whole file for string literals containing a run of three or more interior spaces, since a silent join is invisible to fmt and clippy. This was the only one — the three messages added in 324a16b1a all kept their continuations.

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -335,6 +466,383 @@ impl ChallengerE2e {
...
+            info!(game = %game.address, "Path 3: ZK proof nullified");
+        } else {
+            info!(game = %game.address, "Path 4: ZK fallback nullified, TEE proof remains");
+            let outcome = Self::await_dispute(
+                config,
+                verifier,
+                provider,
+                game.address,
+                challenger,
+                nonce,
```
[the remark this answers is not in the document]
0xth4nh: Fixed in d74e031be: the ZK-fallback branch now passes `nonce + 1` instead of `nonce`.

Took the first of your two options deliberately. A fresh `get_transaction_count` here would reintroduce exactly the race that the end-of-step delta was added to remove — the read lands after the state observation, so the challenger can slip the TEE dispute into that gap and the "second" dispute gets a baseline that already includes it. `nonce + 1` is derived from the state read instead: this branch is only reached because `zkProver == 0`, and only the challenger clears it, so at least one transaction is already spent. That's a lower bound the challenger cannot outrun.

The inner check is now equivalent to the final `>= nonce + 2` in this branch, which is fine — it fails earlier and names the TEE dispute specifically rather than the pair.

crates/infra/challenger-e2e/Cargo.toml
```diff
@@ -21,6 +21,11 @@ alloy-provider.workspace = true
 alloy-primitives.workspace = true
 alloy-signer-local.workspace = true
 alloy-node-bindings.workspace = true
+base-proof-rpc.workspace = true
 base-proof-contracts.workspace = true
+base-proof-submission.workspace = true
+base-tx-manager.workspace = true
+base-zk-fork-dispute.workspace = true
```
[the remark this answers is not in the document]
0xth4nh: Fixed in d74e031be, though not with the suggested ordering — it has the same problem it reports. `base-proof-submission` is 38 chars and `base-zk-fork-dispute` is 37, so putting submission first still breaks the cascade.

Measured lengths, sorted:

```toml
base-proof-rpc.workspace = true                # 31
base-tx-manager.workspace = true               # 32
base-proof-contracts.workspace = true          # 37
base-zk-fork-dispute.workspace = true          # 37
base-proof-submission.workspace = true         # 38
base-prover-service-protocol.workspace = true  # 45
```

`eyre`/`url` (21, 20) and `clap`/`tokio` (57, 49) are also descending, but both pairs predate this PR and aren't in its diff, so I left them.

jackchuma (approved):

PR #4785, opened 2026-08-31
feat(challenger-e2e): drive a real challenger against an ephemeral L1 fork

0xth4nh:
Stacked on #4780. Split out of #4781 so each half is reviewable on its own; this is the harness plus the positive case, #4781 is now only the dispute paths.

## What changed?

`ChallengerE2e::run()` and the `base-challenger-e2e` binary that wraps it for `CronJob` execution.

- **Fork.** Spawns a pod-local Anvil forked from `BASE_CHALLENGER_L1_ETH_RPC`, funds two throwaway keys generated per run — A (driver, signs setup only) and B (challenger, the only key permitted to dispute).
- **Game selection.** Walks the newest `CHALLENGER_E2E_GAME_LOOKBACK` factory indices for two in-progress, uncountered, TEE-only games of the configured game type with at least one intermediate output root. Bails with a diagnostic if fewer than two exist.
- **Handshake.** Writes `/shared/challenger.env` via a rename (so the sidecar can never source a partial file) with the fork URL and B's key. The challenger sidecar blocks on that file appearing.
- **Positive case.** Waits for `base_challenger_up == 1` and a completed scan, then holds `CHALLENGER_E2E_QUIET_WINDOW` and asserts `games_scanned_total` advances while `games_invalid_total`, `nullify_tx_submitted_total` and `challenge_tx_submitted_total` stay flat.

Plus `poll_until` (used by every wait in both halves) and `annotate_timeout`, which appends the challenger's failure counters to a timeout — otherwise a timeout is indistinguishable from "nothing happened".

## Why?

The challenger is a bot that spends money and nullifies proofs. Unit tests cover the classifier against mocks; nothing covers the deployed binary against a real chain. The gate this feeds is: a zeronet challenger image cannot be promoted to sepolia unless it has been shown to behave correctly on a throwaway fork of the L1 it will actually run against.

Forking rather than deploying a fresh chain is what makes it honest. The games were created and verified on the real chain before the fork point, so the `TEEVerifier`, the `TEEProverRegistry` and the game bytecode are all real.

The positive case is the half that catches the expensive failure mode. A challenger that disputes valid games burns gas and nullifies honest proposals; there is no assertion anywhere else in the repo that the deployed binary leaves a valid chain alone.

## How to test?

```
BASE_SUCCINCT_ELF_STUB=1 cargo clippy -p base-challenger-e2e -p base-challenger-e2e-bin \
  --all-features --all-targets -- -D warnings
cargo test -p base-challenger-e2e
```

To run it end to end you need a challenger pod; the topology lives in `protocols/base-proofs` under `chart/challenger-e2e`. Locally, against a chain you have RPC for:

```
anvil --version            # must be on PATH
export BASE_CHALLENGER_L1_ETH_RPC=... BASE_CHALLENGER_L2_ETH_RPC=... \
       BASE_CHALLENGER_ZK_RPC_URL=... \
       BASE_CHALLENGER_DISPUTE_GAME_FACTORY_ADDR=... BASE_CHALLENGER_GAME_TYPE=...
cargo run -p base-challenger-e2e-bin
```

Expect it to log `selected game` twice, then block on the handshake until a challenger is reachable at `CHALLENGER_E2E_CHALLENGER_METRICS_URL`.

To see the positive case fail as intended, point `CHALLENGER_E2E_CHALLENGER_METRICS_URL` at a challenger that is actively disputing: `games_invalid_total advanced by N while every game on the fork was valid`.

### DIFF
--- a/bin/challenger-e2e/README.md
+++ b/bin/challenger-e2e/README.md
@@ -3,6 +3,9 @@
 Challenger behavioural end-to-end test binary for Base.
 
 Parses CLI arguments and delegates to `base_challenger_e2e::ChallengerE2e::run()`,
-which forks the target L1 into a pod-local Anvil, corrupts dispute games the
-live challenger has already accepted, and asserts that the challenger sidecar
-disputes them. Intended for K8s Job execution.
+which forks the target L1 into a pod-local Anvil, releases the challenger
+sidecar onto that fork, and asserts the challenger scans it without disputing
+anything while every game on it is still valid. Intended for K8s Job execution.
+
+Corrupting games and asserting the dispute paths is layered on top of the same
+harness; see the `base-challenger-e2e` crate README for the current scope.
--- a/bin/challenger-e2e/README.md
+++ b/bin/challenger-e2e/README.md
@@ -1,0 +1,8 @@
+# `base-challenger-e2e-bin`
+
+Challenger behavioural end-to-end test binary for Base.
+
+Parses CLI arguments and delegates to `base_challenger_e2e::ChallengerE2e::run()`,
+which forks the target L1 into a pod-local Anvil, corrupts dispute games the
+live challenger has already accepted, and asserts that the challenger sidecar
+disputes them. Intended for K8s Job execution.
--- a/crates/infra/challenger-e2e/src/challenger_e2e.rs
+++ b/crates/infra/challenger-e2e/src/challenger_e2e.rs
@@ -7,8 +7,9 @@ use alloy_primitives::{Address, U256, hex};
 use alloy_provider::{Provider, RootProvider};
 use alloy_signer_local::PrivateKeySigner;
 use base_proof_contracts::{
-    AggregateVerifierClient, AggregateVerifierContractClient, DisputeGameFactoryClient,
-    DisputeGameFactoryContractClient, GameStatus,
+    AggregateVerifierClient, AggregateVerifierContractClient, AnchorStateRegistryClient,
+    AnchorStateRegistryContractClient, DisputeGameFactoryClient, DisputeGameFactoryContractClient,
+    GameStatus,
 };
 use clap::Parser;
 use eyre::{Context, Result, bail, ensure, eyre};
@@ -37,9 +38,17 @@ const DISPUTE_COUNTERS: [&str; 3] = [
 /// Count series of the challenger's validation-latency histogram.
 ///
 /// Recorded once per call to the validator's `validate_output_roots`, which is
-/// reached per candidate game, so this is the only metric that shows the
-/// challenger got past the factory and looked at a game.
+/// reached once per candidate game.
 const VALIDATIONS: &str = "base_challenger_validation_latency_seconds_count";
+
+/// Failed validations, one per failed [`VALIDATIONS`] call.
+///
+/// `validate_output_roots` records its latency from a drop guard, so the
+/// histogram counts attempts rather than successes. It increments this counter
+/// exactly once on the way out of a failure — the `?` returns on the first
+/// error it observes — so the difference of the two is the number of games the
+/// challenger actually validated.
+const VALIDATION_ERRORS: &str = "base_challenger_validation_errors_total";
 
 /// Behavioural end-to-end test of the challenger.
 ///
@@ -70,10 +79,15 @@ impl ChallengerE2e {
             provider.clone(),
         );
         let verifier = AggregateVerifierContractClient::new(provider.clone());
+        let anchor_registry = AnchorStateRegistryContractClient::new(
+            config.anchor_state_registry_addr,
+            provider.clone(),
+        );
 
         // Chosen before the challenger boots, so the quiet window below is
         // measured against a fork that already contains the target games.
-        let (game_a, game_b) = Self::select_games(&config, &factory, &verifier).await?;
+        let (game_a, game_b) =
+            Self::select_games(&config, &factory, &verifier, &anchor_registry).await?;
 
         Self::release_challenger(&fork_url, &challenger)?;
         Self::await_first_scan(&config).await?;
@@ -130,6 +144,7 @@ impl ChallengerE2e {
         config: &Config,
         factory: &DisputeGameFactoryContractClient,
         verifier: &AggregateVerifierContractClient,
+        anchor_registry: &AnchorStateRegistryContractClient,
     ) -> Result<(Address, Address)> {
         let game_count = factory.game_count().await?;
         if game_count == 0 {
@@ -137,9 +152,20 @@ impl ChallengerE2e {
         }
         let floor = game_count.saturating_sub(config.game_lookback);
 
+        // The challenger scans from one past the anchor game's factory index,
+        // so anything at or before it is invisible to the challenger no matter
+        // what state it is in. Walking newest-first means stopping at the
+        // anchor is the whole lower bound.
+        let anchor_game = anchor_registry.anchor_snapshot().await?.anchor_game;
+
         let mut selected = Vec::with_capacity(2);
         for index in (floor..game_count).rev() {
             let game = factory.game_at_index(index).await?;
+            // ZERO is the starting anchor, where the challenger scans from 0.
+            if anchor_game != Address::ZERO && game.proxy == anchor_game {
+                info!(anchor_game = %anchor_game, factory_index = index, "reached the anchor");
+                break;
+            }
             if game.game_type != config.game_type {
                 continue;
             }
@@ -182,8 +208,9 @@ impl ChallengerE2e {
             return Ok((*game_a, *game_b));
         }
         bail!(
-            "need two in-progress, uncountered games of type {} in the newest {} factory \
-             indices, found {}; the fork source may be behind or the proposer may be stalled",
+            "need two in-progress, uncountered games of type {} above the anchor in the newest \
+             {} factory indices, found {}; the fork source may be behind, the proposer may be \
+             stalled, or the anchor may have advanced past them",
             config.game_type,
             game_count - floor,
             selected.len()
@@ -269,16 +296,20 @@ impl ChallengerE2e {
 
         // `games_scanned_total` counts attempted factory indices and is
         // incremented even when every game query fails, so it cannot show that
-        // any game was actually looked at. The validation histogram is only
-        // touched from inside `validate_output_roots`, which is reached per
-        // candidate game, so its count is the positive signal.
-        let validated = after.sum(VALIDATIONS) - before.sum(VALIDATIONS);
+        // any game was actually looked at. The validation histogram is closer
+        // but still counts attempts, because its latency is recorded from a
+        // drop guard that fires on the error path too. Subtracting the error
+        // counter leaves the validations that actually computed a root.
+        let attempted = after.sum(VALIDATIONS) - before.sum(VALIDATIONS);
+        let failed = after.sum(VALIDATION_ERRORS) - before.sum(VALIDATION_ERRORS);
+        let validated = attempted - failed;
         let scanned = after.sum("base_challenger_games_scanned_total")
             - before.sum("base_challenger_games_scanned_total");
         ensure!(
             validated > 0.0,
-            "the challenger validated no game in {:?} ({scanned} indices scanned); it is \
-             reaching the factory but not the games, so a quiet window proves nothing",
+            "the challenger completed no validation in {:?} ({attempted} attempted, {failed} \
+             failed, {scanned} indices scanned); a quiet window over games it never managed to \
+             check proves nothing",
             config.quiet_window
         );
 
@@ -290,14 +321,11 @@ impl ChallengerE2e {
             );
         }
 
-        // Validation errors are usually the L2 RPC rather than the challenger,
-        // so they are reported rather than fatal. A storm severe enough to
-        // matter fails the assertion above instead, because a game whose roots
-        // cannot be computed is never validated.
-        let errors = after.sum("base_challenger_validation_errors_total")
-            - before.sum("base_challenger_validation_errors_total");
-        if errors > 0.0 {
-            warn!(validation_errors = errors, "the challenger reported validation errors");
+        // Some failures are tolerable — they are usually the L2 RPC rather than
+        // the challenger — so they are reported rather than fatal. A storm that
+        // swallows every game fails the assertion above instead.
+        if failed > 0.0 {
+            warn!(validation_errors = failed, "the challenger reported validation errors");
         }
 
         info!(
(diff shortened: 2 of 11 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

bin/challenger-e2e/README.md
```diff
@@ -0,0 +1,8 @@
+# `base-challenger-e2e-bin`
+
+Challenger behavioural end-to-end test binary for Base.
+
+Parses CLI arguments and delegates to `base_challenger_e2e::ChallengerE2e::run()`,
```
jackchuma, line 5: This PR does not corrupt or dispute games; that behavior is in #4781. Please update the README to describe the current quiet-case binary

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,336 @@
...
+        )
+        .await
+    }
+
+    /// Positive case: a challenger that disputes valid games fails here.
+    ///
+    /// Nothing on the fork has been corrupted yet, so every game the
+    /// challenger can see is one it must leave alone.
+    async fn assert_quiet_on_valid_games(config: &Config) -> Result<()> {
+        let before = Scrape::fetch(&config.challenger_metrics_url).await?;
```
jackchuma, line 275: This baseline is taken after `await_first_scan()`. The scanner increments `games_scanned_total` before candidates are validated, so first-pass invalid counters can already be absorbed into `before`. Please require the cumulative failure counters to be zero on the first completed-scan scrape

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,336 @@
...
+        info!(window = ?config.quiet_window, "observing the challenger against an unmodified fork");
+        tokio::time::sleep(config.quiet_window).await;
+        let after = Scrape::fetch(&config.challenger_metrics_url).await?;
+
+        let scanned = after.sum("base_challenger_games_scanned_total")
```
jackchuma, line 305: `games_scanned_total` counts attempted indices, even when every game query or validation fails. This can return success without validating either selected game. Please require a positive successful-validation signal

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,336 @@
...
+        info!(
+            game_a = %game_a,
+            game_b = %game_b,
+            "the fork holds two disputable games and the challenger left both alone"
+        );
+        Ok(())
+    }
+
+    fn spawn_fork(config: &Config) -> Result<AnvilInstance> {
+        info!(fork_source = %config.l1_eth_rpc, port = config.anvil_port, "spawning L1 fork");
```
jackchuma: Please avoid logging the full L1 RPC URL; credentials commonly appear in userinfo, paths, or query parameters. Log only the origin or host/port

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,336 @@
...
+        }
+
+        info!(games_scanned = scanned, "the challenger left every valid game alone");
+        Ok(())
+    }
+
+    /// Appends the challenger's failure counters to a dispute timeout, which is
+    /// otherwise indistinguishable from "nothing happened".
+    async fn annotate_timeout(error: eyre::Report, config: &Config) -> eyre::Report {
+        let Ok(scrape) = Scrape::fetch(&config.challenger_metrics_url).await else {
```
jackchuma, line 341: `Scrape::fetch()` has no request timeout. The timeout annotation and quiet-window scrapes can therefore hang indefinitely on a stalled endpoint

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,336 @@
...
+    /// Hands the fork and a funded key to the challenger sidecar, which is
+    /// blocked on this file appearing.
+    ///
+    /// Written via a rename so the sidecar can never source a partial file.
+    fn release_challenger(fork_url: &Url, signer: &PrivateKeySigner) -> Result<()> {
+        // Sourced after /envmapper/mapping.env, so these override the
+        // config-service values for the run.
+        let contents = format!(
+            "export BASE_CHALLENGER_L1_ETH_RPC={fork_url}\n\
+             export BASE_CHALLENGER_PRIVATE_KEY={}\n",
```
jackchuma, line 235: Setting `BASE_CHALLENGER_PRIVATE_KEY` does not clear inherited `BASE_CHALLENGER_SIGNER_ENDPOINT` and `BASE_CHALLENGER_SIGNER_ADDRESS`. Those modes conflict, preventing startup under a production-like mapping. Please unset both variables here or test that the chart omits them

bin/challenger-e2e/README.md
0xth4nh: Fixed. The bin README now describes the quiet-case binary and points at the crate README for scope.

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Confirmed and fixed. `scanner.rs:217` increments `games_scanned_total` by the whole scanned range after the scan completes, and the validator only runs on the candidates it returns, so a dispute can land before this baseline is read and the delta absorbs it.

The three dispute counters are now required to be zero outright on the first post-scan scrape, not just flat over the window:

```rust
for metric in DISPUTE_COUNTERS {
    let total = before.sum(metric);
    ensure!(total == 0.0, "{metric} is already {total} on the first completed scan; ...");
}
```

One consequence worth naming: this also fails if the fork source carries a genuinely invalid game the real challenger would legitimately dispute. That is the right outcome — such a fork is not a clean baseline for a "disputes nothing" assertion — so the error message says so explicitly rather than reading as a test bug.

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Confirmed and fixed. `games_scanned_total` is `game_count - scan_start`, incremented unconditionally, and `evaluate_game` errors are logged and skipped — so it advances even if every query fails.

Progress is now asserted on `base_challenger_validation_latency_seconds_count`. That histogram is only touched by `timed!` inside `validate_output_roots`, which is reached per candidate game, so a nonzero delta means the challenger actually got past the factory and validated games. `games_scanned_total` is still read, but only to report the index count in the failure message.

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Fixed — host only now:

```rust
info!(fork_source = config.l1_eth_rpc.host_str().unwrap_or("<no host>"), port = config.anvil_port, "spawning L1 fork");
```

Agreed on the reasoning; the provider URLs this runs against carry the API key in the path.

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Confirmed and fixed where `Scrape::fetch` lives. It used `reqwest::get`, whose default client has no request timeout.

Worst case was exactly as you describe in `annotate_timeout`: it scrapes *after* a dispute has already timed out, so a stalled endpoint turned a clean timeout error into a hang. Now one `LazyLock<Client>` with a 10s timeout, shared across the several hundred scrapes a run makes. Deliberately not `unwrap_or_default()` on the builder — the default client has no timeout, so a silent fallback would reintroduce the hang.

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Confirmed and fixed. `private_key` is `conflicts_with = "signer_endpoint"` in the shared `SignerCli` macro (`crates/utilities/tx-manager/src/macros.rs:283`), and clap counts an env-sourced value as present, so an inherited mapping would have stopped the challenger from starting rather than degrading.

The handshake file is `source`d, so it now clears them before exporting the key:

```
unset BASE_CHALLENGER_SIGNER_ENDPOINT
unset BASE_CHALLENGER_SIGNER_ADDRESS
export BASE_CHALLENGER_L1_ETH_RPC=...
export BASE_CHALLENGER_PRIVATE_KEY=...
```

Chose the unset over relying on the chart because the plan was a separate config-service protocol without `signer-endpoint`, which makes a startup-blocking conflict depend on ops discipline. This way the driver is correct against a production-like mapping too.

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,388 @@
...
+        info!(window = ?config.quiet_window, "observing the challenger against an unmodified fork");
+        tokio::time::sleep(config.quiet_window).await;
+        let after = Scrape::fetch(&config.challenger_metrics_url).await?;
+
+        // `games_scanned_total` counts attempted factory indices and is
+        // incremented even when every game query fails, so it cannot show that
+        // any game was actually looked at. The validation histogram is only
+        // touched from inside `validate_output_roots`, which is reached per
+        // candidate game, so its count is the positive signal.
+        let validated = after.sum(VALIDATIONS) - before.sum(VALIDATIONS);
```
jackchuma: This still isn’t a successful-validation signal. `timed!` records on drop, including failed L2 RPC validations. If every validation fails, this count advances, dispute counters stay flat, and the E2E passes. Please require at least one validation not accounted for by `validation_errors_total`, or add a dedicated success counter

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,388 @@
...
+    /// real assertion: the games the challenger must ignore are the same ones
+    /// it will later be asked to dispute. Scanning newest-first also keeps the
+    /// range recent, which matters because the L2 RPC is a live node and may
+    /// have pruned the state behind an older game.
+    async fn select_games(
+        config: &Config,
+        factory: &DisputeGameFactoryContractClient,
+        verifier: &AggregateVerifierContractClient,
+    ) -> Result<(Address, Address)> {
+        let game_count = factory.game_count().await?;
```
jackchuma, line 148: `select_games()` ignores the scanner’s anchor lower bound. It can select an in-progress game at or before the current anchor, which the challenger intentionally never scans. Please resolve the anchor’s factory index and select only later games

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: You are right, and my earlier justification was wrong. `timed!` expands to `DropTimer::new(...)`, so the histogram records on the `?` path too — the count is attempts, not successes.

Fixed in 9f55b5c: the assertion is now on `validation_latency_seconds_count` minus `validation_errors_total`. `validate_output_roots` records latency exactly once per call and increments the error counter exactly once per failed call (the `?` returns on the first error it sees), so the difference is the number of games it actually got a root for. The failure message reports all three numbers (attempted / failed / indices scanned) so a total-failure run says so instead of reading as "no games". Errors below that threshold are still a warning, since they are usually the L2 RPC.

crates/infra/challenger-e2e/src/challenger_e2e.rs
0xth4nh: Confirmed — `scan_start_index` returns `anchor_index + 1`, so a candidate at or before the anchor is one the challenger never looks at, and the dispute waits in the follow-up PR would have burned the full 45m timeout on a game nobody was watching.

Fixed in 9f55b5c. `select_games` already walks newest-first, so it now reads `anchor_snapshot().anchor_game` once up front and breaks the walk when it reaches that proxy — the anchor is the lower bound by construction. `Address::ZERO` is excluded because that is the starting anchor, where the scanner starts at 0. This adds one required var, `BASE_CHALLENGER_ANCHOR_STATE_REGISTRY_ADDR`, already in the shared challenger mapping.

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,416 @@
...
+        // Two distinct accounts: A (driver) signs setup only, B is the
+        // challenger. Both are generated per run and never leave the pod.
+        let driver = PrivateKeySigner::random();
+        let challenger = PrivateKeySigner::random();
+
+        // Held until the end of run(); the fork dies with this binding.
+        let anvil = Self::spawn_fork(&config)?;
+        let fork_url = anvil.endpoint_url();
+        let provider: RootProvider = RootProvider::new_http(fork_url.clone());
+        Self::fund(&provider, &[driver.address(), challenger.address()]).await?;
```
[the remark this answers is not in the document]
0xth4nh: Fair — the positive case signs nothing, so the second key was a funded account with no purpose. Dropped in 46323fa rather than commented: it comes back in the follow-up alongside the `verifyProposalProof` call that spends it, which is where a reviewer can actually see what it is for.

crates/infra/challenger-e2e/src/challenger_e2e.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        // Cumulative, not a delta. `games_scanned_total` is incremented for the
+        // whole scanned range before any candidate is validated, so a scan that
+        // has already completed can have disputed something before this
+        // baseline was taken; a delta comparison would absorb it into `before`
+        // and pass.
+        for metric in DISPUTE_COUNTERS {
+            let total = before.sum(metric);
+            ensure!(
+                total == 0.0,
```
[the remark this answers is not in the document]
0xth4nh: Agreed, and worth stating explicitly: the check is deliberately absolute rather than a delta against `await_first_scan`, so anything disputed in that gap is caught rather than absorbed into a baseline. A fork source that genuinely carries an invalid game should fail this run — the harness has no way to tell that apart from the challenger misbehaving, and the error message says so.

jackchuma (approved):

PR #4812, opened 2026-08-31
fix(derive): preserve singular batch parent hashes

0x00101010:
## Problem

A direct singular batch encodes the hash of the L2 parent it was built on. That hash is a consensus-critical fork binding: derivation must reject the batch if it does not match the current safe head.

Before Denim, `BatchValidator` replaced every singular batch's encoded parent hash with the current safe-head hash before calling `SingleBatch::check_batch`. This made the existing parent check tautological.

The bug is observable when an L2 reorg leaves a stale channel on the parent chain and derivation restarts from an agreed safe head in the middle of that channel:

1. Batches before the agreed head are classified as `Past` and skipped without flushing the channel.
2. The stale batch at the next expected timestamp still points to its original, reorged L2 parent.
3. The validator overwrites that parent with the current safe head, allowing the stale suffix to attach to a different fork.

These are not necessarily duplicate batch submissions: they can be validly posted data from an old unsafe branch that remains on L1 after an L2 reorg. A one-second block time does not prevent the issue because the timestamp identifies a height, not the fork at that height.

PR #4783 fixed the related post-Denim case, where several 200 ms blocks can share a whole-second timestamp, but deliberately retained the parent overwrite before Denim. As a result, pre-Denim chains—including one-second L3s—remain affected.

## Solution

Preserve the encoded parent hash for direct singular batches at every fork. The existing `SingleBatch::check_batch` then rejects a stale-fork batch with `ParentHashMismatch`, and `BatchValidator` flushes the invalid channel.

Span-derived singular batches require different handling: their inner parent relationships are implicit after the enclosing span has been validated. This PR therefore assigns the current L2 parent only when a singular batch is popped from `BatchStream`'s span buffer. Direct singular batches bypass that buffer and retain their wire-encoded parent.

The Denim ancestry logic from #4783 remains unchanged. At the next expected timestamp, it still distinguishes a batch built on a canonical ancestor—which can be skipped—from a batch with an unknown or stale-fork parent—which drops and flushes the channel.

## Validation

- Added a one-second pre-Denim regression test covering a `Past` stale-channel prefix followed by a next-timestamp parent mismatch; the prefix is skipped and the stale suffix flushes the channel.
- Verified direct singular batches pass through `BatchStream` without parent mutation.
- Verified singular batches hydrated from a real span receive the current derived parent.
- `cargo test -p base-consensus-derive -- --test-threads=1` (216 passed)
- `cargo clippy -p base-consensus-derive --all-targets --all-features -- -D warnings`
- `cargo +nightly fmt --all -- --check`

### DIFF
--- a/crates/consensus/derive/src/stages/batch/batch_stream.rs
+++ b/crates/consensus/derive/src/stages/batch/batch_stream.rs
@@ -71,7 +71,7 @@ where
         Ok(self.config.is_holocene_active(origin.timestamp))
     }
 
-    /// Gets a [`SingleBatch`] from the in-memory buffer.
+    /// Gets a span-derived [`SingleBatch`] from the in-memory buffer and assigns its parent hash.
     pub fn get_single_batch(
         &mut self,
         parent: L2BlockInfo,
@@ -79,8 +79,12 @@ where
     ) -> Result<Option<SingleBatch>, SpanBatchError> {
         trace!(target: "batch_span", buffer_len = self.buffer.len(), "Attempting to get a SingleBatch from buffer");
 
+        let parent_hash = parent.block_info.hash;
         self.try_hydrate_buffer(parent, l1_origins)?;
-        Ok(self.buffer.pop_front())
+        Ok(self.buffer.pop_front().map(|mut batch| {
+            batch.parent_hash = parent_hash;
+            batch
+        }))
     }
 
     /// Hydrates the buffer with single batches derived from the span batch, if there is one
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

meyer9 (approved):

dguenther (approved):
lgtm! I'll do a pass on comments in the action tests, I think there might be some to clean up

cb-heimdall:
Review Error for dguenther @ 2026-08-31 21:51:29 UTC

PR #4843, opened 2026-09-01
feat(eip8130): reject seconds-denominated validity bounds at mempool ingress

chunter-cb:
## Summary

EIP-8130 `valid_after`/`valid_before` are Unix **milliseconds**, evaluated on-chain against `block.timestamp * 1000`. A bound accidentally supplied in **seconds** (e.g. `1_700_000_000` instead of `1_700_000_000_000`) is ~1000× too small. Today the mempool would reject it with a misleading reason — `"transaction is not yet valid"` / `"transaction validity window has elapsed"` — and worse, a seconds-valued `valid_after` would be treated as *already active* at execution rather than delaying activation.

This adds a **unit guard at mempool ingress**: a non-zero bound below `TIMESTAMP_MS_THRESHOLD` (`10^11`, the seconds/ms split from the EIP's Timestamp Normalization section) is rejected up front with an actionable reason:

> `valid_after/valid_before must be Unix milliseconds, not seconds`

## Scope: pool-only, non-consensus

Block validity is **unchanged** — the execution path in `common/evm/src/eip8130.rs` is untouched, so no fork coordination is required. A seconds-valued tx that bypassed the pool would still fail its window check at execution (it is long-"expired" against `block.timestamp * 1000`), so no honest block producer could include it. This guard purely upgrades the *mempool rejection reason* from confusing to actionable, and closes the `valid_after`-in-seconds footgun.

This is **deliberately stricter than the EIP**, which normalizes seconds vs. milliseconds *per value* (accepting both). Full acceptance of both denominations is a separate, **consensus-level** change (it must also normalize in the execution/block-validity path); this PR is the safe interim step.

## Changes

- **`constants.rs`** — `Eip8130Constants::TIMESTAMP_MS_THRESHOLD` (`100_000_000_000`) and a `timestamp_is_seconds(value)` helper.
- **`signed.rs`** — new `Eip8130TimestampError::TimestampNotMilliseconds`, checked first (field-format, `now`-independent) in `validate_timestamp`.
- **`validator.rs`** — maps the new error to its ingress rejection reason; adds a `BaseL1BlockInfo::set_timestamp` seam so the integration harness can seed a realistic clock.
- **Tests** — new unit + ingress coverage (seconds `valid_before`/`valid_after`, threshold boundary, ordering vs. nonce-free structural checks). Existing timestamp/nonce-free fixtures migrated from toy values (`now_ms = 0`) to realistic millisecond-scale timestamps, since sub-`10^11` bounds are now (correctly) treated as seconds.

## Follow-up (not in this PR)

Full spec-conformant seconds↔ms normalization across the consensus/execution path, if we decide to accept both denominations rather than require milliseconds.

### DIFF
--- a/crates/common/consensus/src/transaction/eip8130/constants.rs
+++ b/crates/common/consensus/src/transaction/eip8130/constants.rs
@@ -278,6 +278,30 @@ impl Eip8130Constants {
     /// `MAX_CODE_SIZE` limit. EIP-8130 places runtime code directly, so the
     /// mempool rejects oversized code before execution.
     pub const MAX_CODE_SIZE: usize = 24_576;
+
+    /// Heuristic unit boundary distinguishing **seconds** from **milliseconds**
+    /// for the transaction validity-window bounds (`valid_after`/`valid_before`),
+    /// per EIP-8130 "Timestamp Normalization": a non-zero bound below this is a
+    /// seconds value, at or above it is milliseconds. A Unix timestamp in seconds
+    /// does not reach `10^11` until the year 5138, while a millisecond Unix
+    /// timestamp has exceeded `10^11` since 1973, so the split is unambiguous for
+    /// every realistic value.
+    ///
+    /// This node currently admits **millisecond**-denominated bounds only and
+    /// uses the threshold to reject seconds-denominated bounds at ingress with an
+    /// actionable error (see [`Self::timestamp_is_seconds`]) instead of silently
+    /// misreading them as long-expired. Full per-value seconds/ms normalization
+    /// (accepting both, per the spec) is a separate consensus-level change.
+    pub const TIMESTAMP_MS_THRESHOLD: u64 = 100_000_000_000;
+
+    /// Whether `value` is a non-zero validity-window bound that appears to be
+    /// denominated in **seconds** rather than milliseconds — i.e. below
+    /// [`Self::TIMESTAMP_MS_THRESHOLD`]. A `0` bound means "no bound" and returns
+    /// `false`.
+    #[must_use]
+    pub const fn timestamp_is_seconds(value: u64) -> bool {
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

chunter-cb:
Superseded by #5040, which implements the proper fix: full seconds↔milliseconds normalization of `valid_after`/`valid_before` across both the mempool and the consensus execution path (accept both denominations), rather than rejecting seconds at ingress. Since EIP-8130 is enshrined at Cobalt and Cobalt isn't live yet, this is launch behavior and needs no separate fork gate.

PR #4848, opened 2026-09-02
chore(reth): bump to v2.5.2 with parallel history reads

meyer9:
Backports the Reth `v2.5.2` pin and parallel sharded-history-read change to the `v1.3.0` release line.

Pins `meyer9/reth` at `966744f917711b2a4553653bd407653d7577e633`.

Tests:
- `cargo +nightly check -p base --locked`


meyer9:
Superseded by #4851, which targets releases/v1.3.1.

PR #4871, opened 2026-09-02
fix(prover): publish standalone requester RPC on loopback

jackchuma:
Closes CHAIN-4913.

## Summary

`etc/docker/docker-compose.prover.yml` published the prover-service requester RPC with short-syntax and no host IP, so Docker bound it to every host interface. That API is unauthenticated, `zk_backend` is caller-controlled and includes the paid `Network` variant, and fresh session IDs sidestep the per-session retry cap — so for a third-party operator following [STANDALONE_PROVING.md](https://github.com/base/base/blob/main/docs/guides/STANDALONE_PROVING.md) on a host reachable from an untrusted network, any caller could queue paid jobs against their deposited PROVE, or call `deleteProofRequest` / `deleteProofsByTeeSigner` / `listProofs` to list and delete proofs they already paid for.

Base's own prover deployment is not publicly reachable and is unaffected; this is a shipped-default fix for operators running our guide.

**The fix** is `"127.0.0.1:${PROVER_SERVICE_RPC_PORT:-9000}:9000"`. `RPC_LISTEN_ADDR=0.0.0.0:9000` is deliberately unchanged — the in-container listener has to bind a wildcard to receive forwarded traffic.

**Keeping that binding effective** takes three declarative checks, because a loopback publish can be routed past at the daemon level:

- The Compose network pins `gateway_mode_ipv4/ipv6: nat`, so a daemon default of `routed` or `nat-unprotected` cannot leave the container address reachable from the LAN.
- `just prover up` requires Docker Engine >= 28.3.3. [moby#41872](https://github.com/moby/moby/issues/41872) let a LAN peer reach a loopback-published port with a forged packet until 28.0, and [CVE-2025-54388](https://github.com/moby/moby/security/advisories/GHSA-x4rx-4gw3-53p4) drops the protecting rules on a firewalld reload in 28.2.0 through 28.3.2.
- `just prover up` refuses a reused Docker network whose options (non-`nat` gateway mode, or `trusted_host_interfaces`) would permit direct routing.

**Also**: `PROVER_SERVICE_RPC_PORT` now interpolates into the host-port position, so `just prover up` rejects values that are not a bare port number instead of failing with an opaque Compose parse error. A value like `0.0.0.0:9000` was valid before this change.

Requester authentication, per-principal backend authorization, and cumulative PROVE spending caps were recommended in the report but are out of scope — that is real design work, and hard to justify for a stack operators run locally.

## Dropped from an earlier revision

An earlier push of this branch also ran a veth-based probe on every `just prover up` that empirically tested whether a synthetic remote host could reach a loopback-published port. It has been removed. It required `NET_ADMIN` plus the host network and PID namespaces on the operator's machine, refused to run under rootless and userns-remap daemons — configurations the CVE advisory lists as *unaffected* and names as its workaround — was skipped entirely on Docker Desktop, and could not catch a firewalld reload occurring after startup, which is the residual risk the version gate actually covers. The three checks above cover the same ground declaratively.

## Test plan

- [x] `docker compose config` renders `host_ip: 127.0.0.1`, `published: "9000"`, `target: 9000`, and `gateway_mode_ipv4/ipv6: nat` on the default network.
- [x] With `PROVER_SERVICE_RPC_PORT=19000`, the override lands in the host-port position and still binds `127.0.0.1`.
- [x] `RPC_LISTEN_ADDR` / `WORKER_RPC_LISTEN_ADDR` still render as `0.0.0.0:9000` / `0.0.0.0:9001`; neither the worker RPC nor Postgres is published to the host.
- [x] Version gate exercised against a stub daemon: `27.5.1` and `28.3.2` rejected, `28.3.3` (exact boundary) and `29.1.0` accepted.
- [x] Reused-network check exercised against a stub daemon: rejects `nat-unprotected` gateway mode and a non-empty `trusted_host_interfaces`, accepts clean `nat`.
- [x] `PROVER_SERVICE_RPC_PORT=0.0.0.0:9000` fails with the actionable error; a bare port passes and proceeds to the existing ELF check.
- [x] `basectl --prover-rpc http://localhost:9000` still connects: `localhost` resolves to `::1` first on macOS, but clients fall back to `127.0.0.1` (verified against a v4-only listener, 2ms connect).
- [x] The zk-host path is untouched — it reaches the service over the Compose network as `http://base-prover-service:9001`, not through the host-published port.
- [ ] Full `just prover up <network>` end to end with a funded requester key (needs built SP1 ELFs, live RPC endpoints, and PROVE — not runnable here).

## Notes for reviewers

- Running Compose directly rather than through `just prover up` skips the version and reused-network checks. The Compose-level `nat` pinning still applies. This is stated in the guide.
- With no Docker daemon running, the version gate reports "requires Docker Engine 28.3.3 or later" rather than "daemon unreachable". Worth a follow-up if you find that confusing; I left it rather than widen this PR.
- `etc/docker/docker-compose.anvil-l1.yml` publishes the same requester port on a wildcard for the devnet `nitro-local` profile. Left alone: that stack has no funded requester key, so the spending risk does not apply, and every other devnet port there is published the same way.

### DIFF
--- a/bin/prover/service/README.md
+++ b/bin/prover/service/README.md
@@ -14,4 +14,18 @@ API.
 
 The unauthenticated requester and worker APIs are internal service endpoints.
 Restrict them to trusted components with private-network controls; never expose
-them publicly. Wildcard binds support container networking.
+them publicly. Wildcard binds support container networking: the in-container
+listener must accept forwarded traffic, so the deployment — not
+`RPC_LISTEN_ADDR` — decides who can reach it.
+
+The requester API is the more sensitive of the two. It accepts a
+caller-controlled `zk_backend`, including the paid `network` variant, so a
+reachable port lets any caller spend the deployment's requester funds, with no
+cumulative cap. It also exposes `deleteProofRequest`, `deleteProofsByTeeSigner`,
+and `listProofs`, so a reachable port allows deleting proofs already paid for.
+
+The standalone Compose stack therefore publishes the requester port on
+`127.0.0.1` only (see `etc/docker/docker-compose.prover.yml`). Reaching it from
+another host requires putting your own authenticated proxy in front of the
+loopback port, not republishing the port on a public interface. See
+[docs/guides/STANDALONE_PROVING.md](../../../docs/guides/STANDALONE_PROVING.md).
--- a/bin/prover/service/README.md
+++ b/bin/prover/service/README.md
@@ -30,4 +30,4 @@ bridge network rather than `routed` or `nat-unprotected`, and `just prover up`
 rejects reused networks with direct-routing options. Reaching it from another
 host requires putting your own authenticated proxy in front of the loopback
 port, not republishing the port on a public interface. See
-[docs/guides/STANDALONE_PROVING.md](../../../docs/guides/STANDALONE_PROVING.md).
+[`docs/guides/STANDALONE_PROVING.md`](../../../docs/guides/STANDALONE_PROVING.md).
--- a/bin/prover/service/README.md
+++ b/bin/prover/service/README.md
@@ -25,7 +25,9 @@ cumulative cap. It also exposes `deleteProofRequest`, `deleteProofsByTeeSigner`,
 and `listProofs`, so a reachable port allows deleting proofs already paid for.
 
 The standalone Compose stack therefore publishes the requester port on
-`127.0.0.1` only (see `etc/docker/docker-compose.prover.yml`). Reaching it from
-another host requires putting your own authenticated proxy in front of the
-loopback port, not republishing the port on a public interface. See
+`127.0.0.1` only (see `etc/docker/docker-compose.prover.yml`). It uses a NAT
+bridge network rather than `routed` or `nat-unprotected`, and `just prover up`
+rejects reused networks with direct-routing options. Reaching it from another
+host requires putting your own authenticated proxy in front of the loopback
+port, not republishing the port on a public interface. See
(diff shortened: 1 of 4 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

leopoldjoy (approved):

PR #4876, opened 2026-09-02
feat(profiling): add opt-in CPU profiling crate installable on any node

jowparks:
Adds a `base-execution-profiling` crate providing an opt-in, disabled-by-default CPU profiling HTTP endpoint (gzipped pprof) that installs on any node built via `base-node-runner`, gated behind `--enable-profiling`. Wired into the standalone builder, standard `base-reth`/`base rpc`/`base follow` nodes, and the unified `base sequencer` (the shadow-builder path now that deployments run the unified binary). Zero cost when disabled is locked by a benchmark gate and a discriminating socket test.

## Local verification

- `cargo test -p base-execution-profiling`: 19 passed.
- Exercised the real HTTP server with a 3-second, 101 Hz capture: `200 OK`, `application/octet-stream`, and `profile.pb.gz` attachment.
- `gzip -t` and `go tool pprof -top/-raw` parsed the response successfully as a CPU profile with `samples/count` and `cpu/nanoseconds` sample types (283 samples over 3 seconds).
- The response is the standard gzipped pprof protobuf accepted by Datadog profiling intake. The Datadog uploader or Agent still needs to attach deployment metadata such as `service`, `env`, and `version`; no live Datadog upload was performed in this local test.

Generated with Toshi


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/builder/profiling/src/server.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    /// Serves profiling requests until the cancellation token is triggered.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error when the TCP listener cannot bind or the HTTP server exits unexpectedly.
+    pub async fn serve(self) -> eyre::Result<()> {
+        // Compose port class controls exposure; binding 0.0.0.0 inside the container keeps host
+        // capture reachable, while 127.0.0.1 would trap the endpoint inside the container.
+        let address = format!("0.0.0.0:{}", self.port);
```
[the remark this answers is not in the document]
jowparks: Not running on k8s, running in bare metal. Keeping it this way

crates/builder/profiling/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+axum = { workspace = true, features = ["http1", "tokio", "query"] }
+
+# compression
+flate2 = { workspace = true }
+
+# profiling
+pprof = { workspace = true, features = ["flamegraph", "protobuf-codec"] }
+
+[dev-dependencies]
+tower = { workspace = true, features = ["util"] }
```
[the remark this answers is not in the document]
jowparks: Already addressed in ab2514096 — the `flamegraph` feature was dropped, so the pprof dependency now enables only `["protobuf-codec"]` and `inferno` is gone from Cargo.lock. The remaining gzipped pprof protobuf output is exactly what Datadog ingests. The inline suggestion here is anchored to a stale line (it now targets the `tower` dev-dependency), so it should not be committed.

crates/builder/profiling/src/profiler.rs
```diff
@@ -0,0 +1,255 @@
...
+        drop(guard);
+        drop(capture_permit);
+
+        let profile = report.pprof().map_err(ProfilerError::Pprof)?;
+        let protobuf = profile
+            .write_to_bytes()
+            .map_err(|error| ProfilerError::ProtobufEncode { message: error.to_string() })?;
+        let mut encoder = GzEncoder::new(Vec::new(), Compression::default());
+        encoder.write_all(&protobuf).map_err(ProfilerError::Gzip)?;
+        let bytes = encoder.finish().map_err(ProfilerError::Gzip)?;
```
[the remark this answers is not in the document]
jowparks: Fixed in c50ce380c — the report build, pprof/protobuf serialization, and gzip now run inside tokio::task::spawn_blocking so the CPU-heavy finalization no longer blocks a runtime worker on the shared builder executor. ProfilerGuard proved Send; the blocker was the borrowed tokio MutexGuard lifetime, so capture now uses try_lock_owned() and moves the owned guard into the blocking task.

crates/builder/profiling/src/profiler.rs
```diff
@@ -0,0 +1,262 @@
...
+            let profile = report.pprof().map_err(ProfilerError::Pprof)?;
+            let protobuf = profile
+                .write_to_bytes()
+                .map_err(|error| ProfilerError::ProtobufEncode { message: error.to_string() })?;
+            let mut encoder = GzEncoder::new(Vec::new(), Compression::default());
+            encoder.write_all(&protobuf).map_err(ProfilerError::Gzip)?;
+            encoder.finish().map_err(ProfilerError::Gzip)
+        })
+        .await
+        .map_err(|error| ProfilerError::ProtobufEncode { message: error.to_string() })??;
```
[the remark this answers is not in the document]
jowparks: Fixed in f4c8bcc28 — added a dedicated `ProfilerError::TaskJoin` variant and mapped the spawn_blocking JoinError to it, so a panic/shutdown no longer surfaces as a misleading protobuf-encode error.

bin/builder/src/main.rs
```diff
@@ -78,3 +81,15 @@ fn main() {
...
+mod tests {
+    use clap::CommandFactory;
+
+    use super::*;
+
+    #[test]
+    fn builder_cli_argument_ids_are_unique() {
+        BuilderCli::command().debug_assert();
+    }
+}
```
wbj-cb: what is this?

BrianBland (commented):
Can we make the profiling extensions agnostic to the specific node variant? This could be useful on non-builders as well

crates/execution/cli/src/standard_node.rs
```diff
@@ -503,6 +504,71 @@ impl TryFrom<&ShadowIndexerArgs> for ShadowIndexerConfig {
     }
 }
 
+const DEFAULT_PROFILING_PORT: u16 = 6061;
+
+/// CLI arguments for a node's opt-in CPU profiling HTTP server.
```
jowparks, line 520: 🔴 bug: `ProfilingArgs` is never flattened into `StandardNodeArgs` or installed by `StandardBaseRethNode::runner`. `base-reth`/`base rpc` cannot enable profiling, while `base sequencer` accepts these flags through `BuilderArgs` but silently ignores them. Wire intended binaries or narrow the “any node” claim.

crates/execution/profiling/src/server.rs
```diff
@@ -0,0 +1,353 @@
...
+        duration: Duration,
+        frequency: Option<u32>,
+    ) -> impl Future<Output = Result<Vec<u8>, ProfilerError>> + Send {
+        Self::capture(self, duration, frequency)
+    }
+}
+
+fn profile_router<P: ProfileCapture>(profiler: P) -> Router {
+    Router::new()
+        .route("/debug/pprof/profile", get(capture_profile::<P>))
```
jowparks: 🟡 risk: Axum GET routes also execute for HEAD, so `HEAD /debug/pprof/profile` performs the full capture and gzip work before discarding the response body. Add an explicit cheap HEAD rejection and a regression test.

crates/execution/profiling/src/extension.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        let hooks = Box::new(extension).apply(hooks);
+
+        NODE_STARTED_HOOK_REGISTRATIONS.with(|count| assert_eq!(count.get(), 0));
+        assert_eq!(Arc::strong_count(&marker), 2);
+        drop(hooks);
+        assert_eq!(Arc::strong_count(&marker), 1);
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
```
jowparks, line 155: 🟡 risk: this test never executes the extension hook; the enabled branch manually starts a separate `ProfilingServer`. A regression where the hook stops spawning the server would still pass. Exercise the applied hook through the runner test harness.

crates/execution/profiling/src/server.rs
```diff
@@ -0,0 +1,353 @@
...
+        .with_state(ProfileState { profiler })
+}
+
+async fn capture_profile<P: ProfileCapture>(
+    State(state): State<ProfileState<P>>,
+    Query(query): Query<ProfileQuery>,
+) -> Result<Response, ProfilingServerError> {
+    let default_seconds = 30.min(state.profiler.max_capture_seconds());
+    let duration = Duration::from_secs(query.seconds.unwrap_or(default_seconds));
```
jowparks, line 156: 🔴 bug: `seconds=0` is accepted, allowing rapid creation and finalization of the large fixed pprof collector. Reject durations below one second, and keep the capture permit until encoding and gzip complete so finalizers cannot overlap.

crates/execution/profiling/src/profiler.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// # Errors
+    ///
+    /// Returns [`ProfilerError::Busy`] when another capture is active, validation errors for an
+    /// unsupported duration or frequency, or an encoding/capture error from the profiling path.
+    pub async fn capture(
+        &self,
+        duration: Duration,
+        frequency: Option<u32>,
+    ) -> Result<Vec<u8>, ProfilerError> {
+        // The preallocated pprof collector costs 200 MB+ of RSS while its guard is live. Reject
```
jowparks: 🟡 accuracy: this memory description is incorrect for pinned pprof 0.15 defaults. The collector is roughly 65 MiB on a 64-bit target and a replacement remains resident after the first capture; reset can transiently hold both collectors. Correct the comment and operator documentation.

crates/execution/profiling/README.md
```diff
@@ -0,0 +1,13 @@
+# `base-execution-profiling`
+
+Opt-in CPU profiling for a running Base node.
+
+Runs a sampling profiler in-process and exposes captured profiles over a small HTTP server,
```
jowparks, line 5: 🟡 completeness: shipped release binaries use the symbol-stripped `maxperf` profile. Document the `release-symbols`/`profiling` build requirement or external symbolization workflow; otherwise production profiles can contain mostly unresolved addresses.

jowparks:
PR description is stale against current head: crate is now `base-execution-profiling`, flamegraph support was removed, and no profiling-specific benchmark gate exists. Update description before merge.

crates/execution/cli/src/standard_node.rs
jowparks: Fixed in 0ff23e492. `ProfilingArgs` is now flattened into `StandardNodeArgs` (and `ExecutionNodeArgs`) and `StandardBaseRethNode::runner` installs `ProfilingExtension` after the shadow indexer, so `base-reth`/`base rpc` can enable profiling too. Added parse tests `test_standard_node_args_parses_profiling_flags` and `test_standard_node_args_profiling_disabled_by_default`. Verified no duplicate clap flags via the existing `builder_cli_argument_ids_are_unique` test — the builder keeps its own `BuilderArgs.profiling`, so there is no double install.

crates/execution/profiling/src/server.rs
jowparks: Fixed in 0ff23e492. The route now registers an explicit HEAD handler (`get(capture_profile).head(reject_head)`) that returns `405 Method Not Allowed` with an `Allow: GET` header without running a capture. Added test `profile_route_rejects_head_without_capturing` asserting the 405, the `Allow` header, and that no capture ran.

crates/execution/profiling/src/server.rs
jowparks: Fixed the `seconds=0` half in 0ff23e492: added `ProfilerError::DurationTooShort` with a `MIN_CAPTURE` of 1s, rejected before any pprof guard is built, mapped to `400` at the server. Tests `capture_rejects_zero_duration` and `capture_rejects_sub_second_duration` cover it. On the permit: it is intentionally released right after `report.build()` consumes the guard, which is what frees the resident pprof collector; the remaining `pprof()`/protobuf/gzip work operates on the already-built report, not the collector, so holding the permit longer would only serialize concurrent captures without protecting the allocation the limit is about.

crates/execution/profiling/src/profiler.rs
jowparks: Fixed in 0ff23e492. Dropped the fabricated "200 MB+" figure; the comment now states that pprof keeps its sample collector resident for the lifetime of the guard and that we bound the capture window to stop an unbounded request from holding that allocation. Avoided quoting a specific byte figure since it is target- and version-dependent.

crates/execution/profiling/README.md
jowparks: Fixed in 0ff23e492. Added a Symbols section to the crate README documenting that `release`/`maxperf` set `strip = "symbols"` (so captures resolve to raw addresses), and how to get symbolized output: build with `--profile profiling` (line-table debug info, no LTO, frame pointers in Docker via docker-bake.hcl — the intended on-node profile) or `--profile release-symbols` (release-identical codegen with symbols restored), or symbolize afterwards against a matching unstripped binary. Also added a Usage section.

crates/execution/profiling/src/extension.rs
jowparks: Not fixing this one. Driving the registered node-started closure requires a real `BaseFullNode` (the hook signature is `FnOnce(BaseFullNode)`), which is not constructible in a unit test — the only peer that uses this hook, `TxForwardingExtension`, has no hook test at all for the same reason. Making the test invoke the actual closure would mean extracting the server spawn into a method taking a `reth_tasks::TaskExecutor` and adding `reth-tasks` as a production dependency to a crate that today only depends on `base-node-runner`, which isn't worth it for a test-only gain. The existing tests already assert the load-bearing behavior: `enabled_extension_registers_one_node_started_hook` (exactly one hook registered when enabled), `disabled_extension_preserves_existing_hooks_without_registering_one`, and `profiling_extension_opens_socket_only_when_enabled`. Open to revisiting if a node test harness lands that makes exercising the closure cheap.

jowparks:
Good catch — done in 68ceeea3f. `base sequencer` was parsing the builder profiling flags but never installing `ProfilingExtension`, so the endpoint was silently unavailable on the unified binary. Wired it into the sequencer runner to match standalone `base-builder`, which is what matters now that mainnet shadow builder (and zeronet) run `base sequencer` via the unified bin.

Coverage now:
- `base-builder` (standalone) — already installed
- `base rpc` / `base follow` / standard `base-reth` — via `ExecutionNodeArgs` -> runner
- `base sequencer` (unified) — this change

Added parse tests (`parses_profiling_args_and_builds_config`, `profiling_defaults_to_disabled`) and documented the flags in the `base sequencer` README section. PR title/summary updated.

PR #4894, opened 2026-09-03
fix(system-tests): keep builder API registered alongside validity ingress

wbj-cb:
Stack 2/3, based on #4893.

Four `tx_forwarding` tests fail on `main` — every one that calls `start_validity_system()` and relies on forwarding. Three time out waiting for a receipt, two never see the transaction go pending.

### Cause

Reth `extend_rpc_modules` is one slot, not a list:

```rust
fn set_extend_rpc_modules<F>(&mut self, hook: F) -> &mut Self {
    self.extend_rpc_modules = Box::new(hook);   // replaces; does not chain
}
```

`InProcessBuilder::launch` called it twice: directly, to register `BuilderApiImpl` (`base_insertValidatedTransaction`), and again via `NodeHooks::apply_to`. `apply_to` only claims the slot when some extension registered an RPC module, and none did — until #4819 began installing `SendRawTransactionValidityExtension` whenever validity transactions are enabled. The builder API was then silently discarded on exactly those stacks:

```
batch item rejected error=ErrorObject { code: MethodNotFound, message: "Method not found" }
```

One missing method covers both symptoms. Nothing reaches the builder, so satisfied predicates never get a receipt and unsatisfied ones never go pending.

Registering the builder API through `NodeHooks` puts both in one closure.

### Scope

Harness only. `bin/builder` and `bin/base` install every extension via `install_ext`, so they issue one `extend_rpc_modules` call and were never affected. No production node lost the method.

### Testing

Whole binary, serially: `9 passed; 0 failed` (609s).

CI on the combined branch, against the same suite: 8 failures to 3, removing exactly the five `tx_forwarding` ones.

| Run | Result |
|---|---|
| before | 90 run, 82 passed, 8 failed |
| after | 90 run, 87 passed, 3 failed |

The remaining 3 are pre-existing SIGSEGVs in `activation_registry`, `eip8130`, and `in_process_zk` — unrelated to this change, failing identically before it. #4895 adds diagnostics for them.

PR #4893, opened 2026-09-03
fix(system-tests): shut down the stack in the direct builder ingress test

wbj-cb:
Stack 1/3.

`test_validity_transaction_submitted_directly_to_builder_is_included` never calls `system.shutdown()`. Its assertions pass, and then the test hangs dropping the runtime:

```
tx_forwarding.rs:430  (Ok(()))
  core::ptr::drop_in_place<tokio::runtime::Runtime>
    tokio::runtime::blocking::pool::BlockingPool::shutdown
      tokio::runtime::blocking::shutdown::Receiver::wait
```

`DestinationReader::run` is a blocking thread looping on `while !self.cancel.is_cancelled()`, and only `system.shutdown()` cancels that token. The blocking pool never drains, so dropping the test runtime blocks forever and nextest eventually terminates the test.

That is why CI reports this one with no error message, while the other `tx_forwarding` failures carry a real deadline error.

Add the missing shutdown, matching every other test in the file and what #4816 did for them.

### Testing

Passes in 66.7s and the process exits (70s wall). Before, it hung indefinitely at teardown.

PR #4895, opened 2026-09-03
fix(system-tests): give the system-test threads a larger stack

wbj-cb:
Stack 3/3, based on #4894.

Three Cobalt system tests abort with SIGSEGV on Linux CI:

- `activation_registry::test_activation_registry_cobalt_admin_rotation`
- `eip8130::eip8130_transaction_is_mined`
- `in_process_zk::in_process_prover_and_zk_host_start`

They are unrelated to #4893 / #4894 and failed identically before them.

### Cause

A temporary `SA_SIGINFO` handler reported the fault:

```
fault address: 0x7fa6167fd718 (si_code=2)
```

`si_code=2` is `SEGV_ACCERR` — a mapped page the process may not touch, i.e. a `PROT_NONE` stack guard page. A null or wild pointer would report `SEGV_MAPERR` instead. The address sits in the thread-stack mmap region, and the backtrace runs through reth's node launch on the libtest thread:

```
reth_network::discovery::Discovery::new::{{closure}}
BaseNetworkBuilder::build_network -> ComponentsBuilder::build_components
EngineNodeLauncher::launch_node
InProcessBuilder::start -> L2Stack::start -> SystemTestStackBuilder::build
cobalt::start_cobalt_system
std::sys::thread::unix::Thread::new::thread_start     (libtest thread)
```

Reth's launch chain is a deeply nested set of async state machines. Unoptimized, it overflows the 2 MiB default that libtest's per-test thread inherits. Only the Cobalt tests trip it, because the extra fork widens an already-marginal frame — `activation_registry` is the clincher, with 8 Beryl tests passing and its 1 Cobalt test dying in the same binary. Mainnet is a release build and was never affected.

### Fix

Set `RUST_MIN_STACK` on both `just devnet tests` and `tests-ci`, so local Linux runs get it too. Production already does the equivalent: `bin/prover/zk-host` and `bin/prover/nitro-host` both call `with_thread_stack_size(8 * 1024 * 1024)`.

Setting it in `.config/nextest.toml` under `[env]` does **not** work — nextest accepts the key and silently never passes it to the test process. Reading the variable inside a test gave `Err(NotPresent)` that way and `Ok("33554432")` from the command line. That trap cost a full CI cycle and is called out in a comment.

### Evidence

Full system-test suite, green:

**https://depot.dev/orgs/m722kg9kwb/workflows/p0cj9bh91s?job=8zqgft1s51&repo=base%2Fbase**

```
90 tests run, 90 passed, 0 skipped, 0 failed
```

That run was commit `188b56221`, which additionally flipped `run_system_tests` on so the suite would execute on a PR. **That flag and all diagnostic scaffolding have since been reverted** — this PR now contains only the two-line Justfile change.

### DIFF
--- a/etc/docker/Justfile
+++ b/etc/docker/Justfile
@@ -172,13 +172,17 @@ pull-images:
     docker pull ghcr.io/paradigmxyz/reth:v1.11.4
     docker pull sigp/lighthouse:v8.0.1
 
+# Reth's unoptimized node launch overflows the 2 MiB stack libtest gives each test thread,
+# which SIGSEGVs on Linux. Do not move to nextest.toml `[env]`; nextest silently drops it.
+system_test_min_stack := "33554432"
+
 # Runs system tests (requires Docker)
 tests: _install-nextest _build-contracts
-    cargo nextest run -p base-system-tests
+    RUST_MIN_STACK={{ system_test_min_stack }} cargo nextest run -p base-system-tests
 
 # Runs system tests with ci profile for minimal disk usage
 tests-ci: _install-nextest _build-contracts
-    cargo nextest run -P ci --locked -p base-system-tests --cargo-profile ci --no-fail-fast
+    RUST_MIN_STACK={{ system_test_min_stack }} cargo nextest run -P ci --locked -p base-system-tests --cargo-profile ci --no-fail-fast
 
 # Runs system test benchmarks
 bench name:
--- a/etc/systems/src/l2/in_process_builder.rs
+++ b/etc/systems/src/l2/in_process_builder.rs
@@ -193,6 +193,20 @@ impl InProcessBuilder {
             ))
             .apply(hooks);
         }
+        // Reth's `extend_rpc_modules` is a single-slot hook that silently replaces whatever was
+        // registered before it, and `NodeHooks::apply_to` claims that slot for every extension
+        // RPC module. Registering the builder API here instead keeps both in one closure.
+        let hooks = hooks.add_rpc_module(move |ctx| {
+            let api =
+                BuilderApiImpl::<_, base_execution_txpool::TransactionValidity>::with_extensions(
+                    ctx.pool().clone(),
+                    accept_validity_transactions,
+                    DEFAULT_MAX_VALIDITY_PREDICATES,
+                );
+            ctx.modules.merge_configured(api.into_rpc())?;
+            Ok(())
+        });
+
         let node_builder = NodeBuilder::new(node_config.clone())
             .with_database(db)
             .with_launch_context(runtime.clone())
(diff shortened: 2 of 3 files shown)


wbj-cb:
The handler worked — we have a crash site, and it is **startup, not teardown**:

```
 1: base_cli_utils::sigsegv::unix_impl::print_stack_trace
 2: <signal trampoline>
 3: reth_network::discovery::Discovery::new::{{closure}}
 4: reth_network::manager::NetworkManager<N>::new::{{closure}}
 6: BaseNetworkBuilder::build_network::{{closure}}
 7: ComponentsBuilder::build_components::{{closure}}
 9: EngineNodeLauncher::launch_node::{{closure}}
12: InProcessBuilder::start  ->  13: L2Stack::start  ->  14: SystemTestStackBuilder::build
15: cobalt::start_cobalt_stack  ->  17: test_activation_registry_cobalt_admin_rotation
44: std::sys::thread::unix::Thread::new::thread_start          (libtest thread)
```

All three tests share this path, which is why three unrelated workloads die in a 66-72s band: it is the shared Cobalt stack boot, not test-specific logic.

`Discovery::new` is plain safe Rust — UDP binds, discv4/discv5, DNS discovery — with no raw-pointer work, and the backtrace shows no recursion cycle. That points at stack exhaustion of one very large unoptimized async state machine rather than a bad dereference. `bin/prover/zk-host` and `bin/prover/nitro-host` already call `with_thread_stack_size(8 * 1024 * 1024)` for the same reason; tests never got the equivalent.

Pushed `RUST_MIN_STACK=32MiB` as the discriminator. Aborts disappear means stack exhaustion and that setting is the fix, promoted to its own mergeable PR. Aborts persist means a real bad pointer and we go to gdb for the fault address.

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

meyer9 (approved):

PR #4873, opened 2026-09-02
feat(shadow-indexer-db): mirror the hash columns as hex text

wbj-cb:
Shadow block hashes do not survive the ETL into Snowflake: they arrive as Python `memoryview` reprs like `<memory at 0x74b086814b80>` rather than hashes, which breaks every downstream query that joins on them. This PR starts storing the hashes as text so they land intact. Reads are unchanged, so nothing downstream needs a coordinated deploy.

- Adds `hash_hex` and `canonical_hash_hex` to `shadow_blocks`, populated by the writer alongside the existing byte columns.
- Uses `0x`-prefixed lowercase hex, matching the shadow-metrics API and every other EVM hash in Snowflake, so the values join without normalization.
- Adds CHECK constraints that reject any other spelling at write time.
- Leaves reads on the existing byte columns, so `shadow-metrics` is unaffected by this deploy.
- Fills rows already stored via an online backfill run after this ships; indexing the new columns and dropping the old ones follow in #4910.

PR #4910, opened 2026-09-03
feat(shadow-indexer-db): drop the BYTEA hash columns for their text mirrors

wbj-cb:
> Stacked on #4873.

Shadow block hashes currently exist twice in `shadow_blocks`, as bytes and as text, which is the intermediate state that let the text columns land without downtime. This PR removes the byte columns so hashes are stored as text only, completing the change. It ships after the backfill from #4873 has finished.

- Drops the BYTEA `hash` and `canonical_hash` columns and renames the text columns onto their names.
- Rebuilds the three dependent indexes against the text columns first, including the partial index Postgres would otherwise drop silently along with `canonical_hash`.
- Changes `ShadowBlockRow`, `ShadowCanonicalRef` and `ShadowSummaryRow` to carry `String` instead of `Vec<u8>`.
- Normalizes incoming path hashes in `shadow-metrics` before lookup, since matching is now string equality.
- Requires the backfill and constraint validation to have run first, and the builder to deploy before `shadow-metrics`.

### Backfill script used
```bash
#!/usr/bin/env bash
set -euo pipefail

export PGPASSWORD="$SHADOW_INDEXER_DB_PASSWORD"

BATCH=${BATCH:-5000}
MAX_RETRIES=${MAX_RETRIES:-5}
SLEEP=${SLEEP:-0.1}

lo=$(psql -XAtc "SELECT COALESCE(min(number), 0) - 1 FROM shadow_blocks")
hi=$(psql -XAtc "SELECT COALESCE(max(number), 0) FROM shadow_blocks")

echo "backfilling ($lo, $hi] in batches of $BATCH"

while [ "$lo" -lt "$hi" ]; do
  next=$(( lo + BATCH ))
  [ "$next" -gt "$hi" ] && next=$hi

  attempt=1
  until psql -XAtq -v ON_ERROR_STOP=1 -c "
    SET lock_timeout = '2s';
    SET statement_timeout = '60s';
    UPDATE shadow_blocks
       SET hash_hex = '0x' || encode(hash, 'hex'),
           canonical_hash_hex = CASE WHEN canonical_hash IS NULL THEN NULL
                                     ELSE '0x' || encode(canonical_hash, 'hex') END
     WHERE number > $lo AND number <= $next
       AND (hash_hex IS NULL
            OR (canonical_hash IS NOT NULL AND canonical_hash_hex IS NULL));
  "
  do
    if [ "$attempt" -ge "$MAX_RETRIES" ]; then
      echo "batch ($lo, $next] failed after $attempt attempts; aborting at lo=$lo" >&2
      exit 1
    fi
    echo "batch ($lo, $next] failed (attempt $attempt), retrying" >&2
    attempt=$(( attempt + 1 ))
    sleep $(( attempt * 2 ))
  done

  lo=$next
  echo "progress $lo / $hi"
  sleep "$SLEEP"
done

echo "done"
```

### Verification

All rows have been backfilled
```
shadow_metrics=> SELECT COUNT(*) from shadow_blocks WHERE hash_hex IS NULL;
 count
-------
     0
(1 row)
```

```
shadow_metrics=> select count(*) from shadow_blocks where canonical_hash is not null and canonical_hash_hex is null;
 count
-------
     0
(1 row)
```

### DIFF
--- a/crates/execution/shadow-indexer-db/migrations/0015_drop_bytea_hash_columns.sql
+++ b/crates/execution/shadow-indexer-db/migrations/0015_drop_bytea_hash_columns.sql
@@ -1,0 +1,71 @@
+-- Completes the BYTEA -> TEXT move: drops the byte columns and renames the hex
+-- mirrors onto their names, so `shadow_blocks.hash` and
+-- `shadow_blocks.canonical_hash` are the text the ETL can actually read.
+--
+-- PRECONDITION: the backfill has finished and
+-- `shadow_blocks_hash_hex_not_null` has been VALIDATEd out of band. Without
+-- that, the SET NOT NULL below scans the heap under ACCESS EXCLUSIVE instead of
+-- reading the constraint, and fails outright if any row still holds NULL. That
+-- failure is deliberate -- it is the gate that stops this landing on a database
+-- that has not been prepared -- but it panics the builder at startup, so run the
+-- backfill first.
+--
+-- ORDERING: deploy the builder before shadow-metrics. The builder runs
+-- migrations at startup and then reads the new names; shadow-metrics runs no
+-- migrations and reads whatever is there. Between the two deploys its by-hash
+-- endpoints will error. That window is unavoidable -- no rename preserves both
+-- names at once -- so keep it short and roll them together.
+--
+-- Every statement here is catalog-only. Nothing scans, nothing rewrites.
+--
+-- The timeout is about acquiring the lock, not holding it. Each statement here
+-- finishes in under a millisecond, but every one needs ACCESS EXCLUSIVE, and
+-- that has to wait out any open reader -- including the Snowflake ETL, whose
+-- SELECTs run long enough that they already collided with migration 0007. At
+-- 5s a single unlucky overlap fails the migration, which panics the builder at
+-- startup and crashloops it.
+--
+-- Waiting is not free either: once this is queued for ACCESS EXCLUSIVE, new
+-- queries queue behind it, so the timeout is also the worst-case stall for
+-- everything touching the table. Five minutes is a real stall -- roughly 150
+-- blocks of shadow writes at a two-second cadence, which the writer's 1024-slot
+-- channel absorbs -- accepted because this runs once, on a canary, and the
+-- alternative is a builder that crashloops until someone notices.
+SET LOCAL lock_timeout = '5min';
+
+-- Raised above lock_timeout on purpose. statement_timeout covers the lock wait
+-- too, so if the server default were lower it, and not lock_timeout, would
+-- decide how long this waits -- silently capping the five minutes above. Every
+-- statement here is catalog-only and finishes in under a millisecond, so the
+-- only thing this ceiling can cut short is the wait itself.
+SET LOCAL statement_timeout = '6min';
+
+-- Constant time: proven from the validated constraint rather than the heap.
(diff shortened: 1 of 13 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/execution/shadow-indexer-db/src/models.rs
```diff
@@ -28,8 +28,8 @@ pub struct ShadowBlockRow {
 pub struct ShadowCanonicalRef {
     /// Block number.
     pub number: i64,
-    /// Raw block hash.
-    pub hash: Vec<u8>,
+    /// Block hash, as `0x`-prefixed lowercase hex. See [`crate::ShadowHash`].
+    pub hash: String,
```
haardikk21, line 32: i wonder if we use the alloy `B256` type here instead of `String` - which should be natively supplied to you, we can get rid of a lot of encoding/decoding extra stuff we have to do?

crates/execution/shadow-indexer-db/migrations/0015_drop_bytea_hash_columns.sql
```diff
@@ -0,0 +1,51 @@
...
+--
+-- ORDERING: deploy the builder before shadow-metrics. The builder runs
+-- migrations at startup and then reads the new names; shadow-metrics runs no
+-- migrations and reads whatever is there. Between the two deploys its by-hash
+-- endpoints will error. That window is unavoidable -- no rename preserves both
+-- names at once -- so keep it short and roll them together.
+--
+-- Every statement here is catalog-only. Nothing scans, nothing rewrites.
+
+SET LOCAL lock_timeout = '5s';
```
haardikk21: this 5s timeout has previously caused issues on the mainnet DB, recommend increasing it significantly higher or removing it if we can (no timeout?)

crates/execution/shadow-indexer-db/src/models.rs
wbj-cb: I can do this and fold it as the already existing `ShadowHash` newtype. Natively, B256 maps to ByteA storage types not text. but we need text for ETL

wbj-cb: Broke this out into a followup PR since its a larger, rust only code change https://github.com/base/base/pull/5002

haardikk21 (approved):

PR #4913, opened 2026-09-03
fix(prefetch): bound queued request memory

BrianBland:
## Summary

- replace B20 transfer hint `Vec` construction with a fixed five-slot stack buffer
- cap queued prefetch requests globally and divide channel capacity from that cap, so worker count cannot multiply queue memory
- preserve exact hint-slot coverage with updated golden coverage and add a global-admission unit test

## Testing

- `cargo test -p base-precompile-storage -p base-execution-state-prefetch`
- `cargo test -p base-common-precompiles --features test-utils --test b20_asset_v1_golden golden_transfer_hint_slots_match_sload_footprint`
- `git diff --check`

Generated with Toshi

### DIFF
--- a/crates/common/precompile-storage/src/prefetch.rs
+++ b/crates/common/precompile-storage/src/prefetch.rs
@@ -16,7 +16,7 @@
 //! installed — tests, tools, and `no_std` proof environments — hints are a
 //! no-op atomic load.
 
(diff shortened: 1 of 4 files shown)


BrianBland:
Superseded by an upstream-head stacked PR from `base:fix/prefetch-queue-bound`.

PR #4930, opened 2026-09-05
feat(proposer): resolve proposal intervals per game for Denim

0xth4nh:
Section 2 of the [Denim proof config switchover](https://bdocs.cbhq.net/d/denim-proof-config-switchover/v/latest/) design.

Depends on base/contracts#431, which adds `intervalsForStartingBlock(uint256)` to `AggregateVerifier`. This PR adds the Rust binding for it, so it must land after the contract is deployed.

## What changed?

**`crates/proof/contracts`**
- Added `intervalsForStartingBlock(uint256)` to the `IAggregateVerifier` `sol!` interface and `AggregateVerifierClient::read_intervals_for_starting_block`, which validates the pair (`block_interval >= 2`, non-zero intermediate, divisible) on read.

**`crates/proof/proposer`**
- Replaced the `ProposalIntervals` helper with `Intervals` (the resolved pair, plus its `intermediate_block_numbers`) and `IntervalResolver`, which reads the factory's current `gameImpls` entry and then `intervalsForStartingBlock(starting_block)` on every lookup.
- Removed `block_interval` / `intermediate_block_interval` from `DriverConfig`, `ProofDispatcherConfig`, `ProofRecoveryConfig`, `ProofSubmitter` and `ProofCollector`. All five now hold an `Arc<IntervalResolver>` and resolve from the starting block of the game in hand.
- `service.rs` no longer reads the two intervals at startup; it builds the resolver instead. The `gameImpls != 0` fail-fast check and the `init_bonds` read are kept.
- `ProofSubmitter::submit` takes `starting_block_number` from the caller (the collector's cursor) instead of computing `target_block - block_interval`, and rejects a target block that does not sit exactly one resolved interval after that starting block.

## Why?

Denim changes `BLOCK_INTERVAL` 600 → 6,000 and `INTERMEDIATE_BLOCK_INTERVAL` 30 → 300 at a fixed L2 block. The proposer read both from the implementation exactly once at startup and copied them into four config structs that were never refreshed, so from the boundary onward:

- the recovery forward walk builds the next game's UUID with the pre-Denim interval, `games()` returns `Address::ZERO`, and it logs "No game found at expected block, will propose from here" — re-proposing over games that already exist;
- `ProofSubmitter` derives `starting_block = target_block - block_interval` from the same stale value, so the committed range and the `extraData` in the game UUID are both wrong;
- `ProofDispatcher` puts the stale `intermediate_block_interval` in every `ProofRequest`.

Resolving per game also means an implementation upgrade is picked up without a proposer restart, since `gameImpls` is re-read each lookup.

Note on the doc's guidance to follow `basectl` (`crates/infra/basectl/src/rpc/games.rs:393-409`) and read the interval from the **game proxy** rather than the factory's mutable `gameImpls` entry: that applies to inspecting a game that already exists. The proposer's reads are all *predictive* — it computes the UUID of a game that has not been created yet — so there is no proxy to read from. `intervalsForStartingBlock` on the implementation is the equivalent: it returns exactly the pair the constructor will apply to that starting block, and for pre-Denim starting blocks the new implementation returns the same values the old one had.

`IntervalResolver` deliberately has no cache — two `eth_call`s per resolved target, against a 12s poll and the rollup output fetches already on the same path. There is a `ponytail:` comment naming the upgrade path if the L1 read budget ever becomes the constraint.

## How to test?

```
cd crates/proof
cargo test -p base-proposer -p base-proof-contracts
cargo clippy -p base-proof-contracts -p base-proposer -p base-challenger --all-targets -- -D warnings
```

The two new tests are the ones that matter:

1. `cargo test -p base-proposer test_recovery_forward_walk_crosses_fast_activation_block`

   Builds a five-game chain whose starting blocks straddle a Denim activation at block 300 — 100-block games at 0/100/200, 200-block games at 300/500 — and runs the same fixture through two resolvers. With the Denim-aware resolver the walk reaches block 700 and parent `proxy_addr(4)`. With a fixed pre-Denim resolver (the old behaviour) the same chain stalls at block 300, which is the bug.

2. `cargo test -p base-proposer submit_rejects_target_block_that_contradicts_resolved_interval`

   `submit(.., starting_block = 25, target_block = 100, ..)` against a 100-block resolver fails with "is not 100 blocks after starting block 25" and creates no game.

Manual check against a chain with the upgraded verifier deployed but Denim not yet scheduled — the proposer should behave identically to today, since `intervalsForStartingBlock` returns the pre-Denim pair for every starting block until the `ProtocolVersions` schedule has a Denim entry:

```
cargo run -p base-proposer -- --l1-eth-rpc <l1> --rollup-rpc <rollup> --prover-rpc <prover> \
  --dispute-game-factory-addr <factory> --anchor-state-registry-addr <asr> --game-type <type> --dry-run
```

Expect the startup log to no longer print `block_interval` / `intermediate_block_interval` (they are per game now), and the recovery walk to reach the same tip it does on `main`.

## Follow-ups (not in this PR)

- Challenger (`crates/proof/challenge`): `AnchorUpdater` builds the next game's UUID from the same startup-cached pair. Separate PR, based on this one — it reuses `read_intervals_for_starting_block`. `GameScanner::resolve_intermediate_block_interval` also needs work: the doc lists it as already correct, but it caches by *implementation address*, so once one implementation serves both sides of the boundary it returns a single value for every game.
- Nitro TEE / ZK gating, per sections 4-5 of the doc.

---

### Update — back to Denim

The 200ms cadence switch has moved back to **Denim**, `ProtocolVersions` index 13. base/contracts#434 reverts the index in `AggregateVerifier` (12 → 13, semver 0.3.0); this branch follows it in prose and fixture names.

Nothing functional moved in either direction. This service resolves the interval pair from the verifier per game and never names a fork itself, so the earlier Cobalt retarget and this one are both comment-and-identifier changes. The mock builders and test names introduced along the way are cadence-neutral (`fast_activation_block`, `with_fast_intervals`, `..._crosses_fast_activation_block`) and did not need renaming for either move.

Index 13 sits one past the end of the current Base mainnet schedule, where index 12 was in range but zero. `_firstFastBlock()` reports "unscheduled" for both, so the deploy-time requirement is unchanged: the activation has to be registered on `ProtocolVersions` before the fork or the verifier stays on the slow cadence silently.

The branch name still says `denim` — which is now correct again. It was never renamed, since that would close and reopen the PR.

### DIFF
--- a/crates/proof/contracts/src/aggregate_verifier.rs
+++ b/crates/proof/contracts/src/aggregate_verifier.rs
@@ -253,6 +253,7 @@ pub trait AggregateVerifierClient: Send + Sync {
     /// Cobalt switches the verifier to a shorter cadence at a fixed L2 block, so
     /// the pair is a function of the game's starting block. Callers must resolve
     /// it per game rather than reading `BLOCK_INTERVAL` once at startup.
+    /// Legacy implementations fall back to their fixed interval getters.
     async fn read_intervals_for_starting_block(
         &self,
         impl_address: Address,
@@ -488,10 +489,25 @@ impl AggregateVerifierClient for AggregateVerifierContractClient {
     ) -> Result<(u64, u64), ContractError> {
         let contract =
             IAggregateVerifier::IAggregateVerifierInstance::new(impl_address, &self.provider);
-        let result = contract_call!(
+        let result = match contract_call!(
             contract.intervalsForStartingBlock(U256::from(starting_block)).call(),
             "intervalsForStartingBlock failed"
-        )?;
+        ) {
+            Ok(result) => result,
+            Err(error) if error.is_missing_method() => {
+                let (block_interval, intermediate_block_interval) = futures::try_join!(
+                    self.read_block_interval(impl_address),
+                    self.read_intermediate_block_interval(impl_address),
+                )?;
+                if !block_interval.is_multiple_of(intermediate_block_interval) {
+                    return Err(ContractError::validation(format!(
+                        "BLOCK_INTERVAL ({block_interval}) is not divisible by INTERMEDIATE_BLOCK_INTERVAL ({intermediate_block_interval})"
+                    )));
+                }
+                return Ok((block_interval, intermediate_block_interval));
+            }
+            Err(error) => return Err(error),
+        };
 
         let block_interval: u64 = result
             ._0
--- a/crates/proof/contracts/src/aggregate_verifier.rs
+++ b/crates/proof/contracts/src/aggregate_verifier.rs
@@ -250,7 +250,7 @@ pub trait AggregateVerifierClient: Send + Sync {
     /// `AggregateVerifier` implementation applies to a game whose range starts
     /// at `starting_block`.
     ///
-    /// Denim switches the verifier to a shorter cadence at a fixed L2 block, so
+    /// Cobalt switches the verifier to a shorter cadence at a fixed L2 block, so
     /// the pair is a function of the game's starting block. Callers must resolve
     /// it per game rather than reading `BLOCK_INTERVAL` once at startup.
     async fn read_intervals_for_starting_block(
@@ -512,7 +512,7 @@ impl AggregateVerifierClient for AggregateVerifierContractClient {
         }
         if !block_interval.is_multiple_of(intermediate_block_interval) {
             return Err(ContractError::validation(format!(
-                "BLOCK_INTERVAL ({block_interval}) is not divisible by                  INTERMEDIATE_BLOCK_INTERVAL ({intermediate_block_interval})"
+                "BLOCK_INTERVAL ({block_interval}) is not divisible by INTERMEDIATE_BLOCK_INTERVAL ({intermediate_block_interval})"
             )));
         }
 
--- a/crates/proof/contracts/src/aggregate_verifier.rs
+++ b/crates/proof/contracts/src/aggregate_verifier.rs
@@ -60,6 +60,13 @@ sol! {
         /// Returns the intermediate block interval for intermediate output root checkpoints.
         function INTERMEDIATE_BLOCK_INTERVAL() external view returns (uint256);
 
+        /// Returns the `(blockInterval, intermediateBlockInterval)` pair the verifier
+        /// applies to a game whose range starts at `startingBlock`.
+        function intervalsForStartingBlock(uint256 startingBlock)
+            external
+            view
+            returns (uint256, uint256);
+
         /// Returns the game type.
         function gameType() external view returns (uint32);
 
@@ -238,6 +245,19 @@ pub trait AggregateVerifierClient: Send + Sync {
         &self,
         impl_address: Address,
     ) -> Result<u64, ContractError>;
+
+    /// Reads the `(block_interval, intermediate_block_interval)` pair the
+    /// `AggregateVerifier` implementation applies to a game whose range starts
+    /// at `starting_block`.
+    ///
+    /// Denim switches the verifier to a shorter cadence at a fixed L2 block, so
+    /// the pair is a function of the game's starting block. Callers must resolve
+    /// it per game rather than reading `BLOCK_INTERVAL` once at startup.
+    async fn read_intervals_for_starting_block(
+        &self,
+        impl_address: Address,
+        starting_block: u64,
+    ) -> Result<(u64, u64), ContractError>;
 
     /// Returns the intermediate output roots for the given game.
     ///
@@ -459,6 +479,44 @@ impl AggregateVerifierClient for AggregateVerifierContractClient {
         }
 
         Ok(interval)
+    }
+
+    async fn read_intervals_for_starting_block(
+        &self,
+        impl_address: Address,
+        starting_block: u64,
+    ) -> Result<(u64, u64), ContractError> {
+        let contract =
+            IAggregateVerifier::IAggregateVerifierInstance::new(impl_address, &self.provider);
+        let result = contract_call!(
+            contract.intervalsForStartingBlock(U256::from(starting_block)).call(),
+            "intervalsForStartingBlock failed"
+        )?;
+
+        let block_interval: u64 = result
+            ._0
+            .try_into()
+            .map_err(|_| ContractError::validation("BLOCK_INTERVAL overflows u64"))?;
+        let intermediate_block_interval: u64 = result
+            ._1
+            .try_into()
+            .map_err(|_| ContractError::validation("INTERMEDIATE_BLOCK_INTERVAL overflows u64"))?;
+
+        if block_interval < 2 {
+            return Err(ContractError::validation(
+                "BLOCK_INTERVAL must be at least 2 (single-block proposals are not supported)",
+            ));
+        }
+        if intermediate_block_interval == 0 {
+            return Err(ContractError::validation("INTERMEDIATE_BLOCK_INTERVAL cannot be 0"));
+        }
+        if !block_interval.is_multiple_of(intermediate_block_interval) {
+            return Err(ContractError::validation(format!(
+                "BLOCK_INTERVAL ({block_interval}) is not divisible by                  INTERMEDIATE_BLOCK_INTERVAL ({intermediate_block_interval})"
+            )));
+        }
+
+        Ok((block_interval, intermediate_block_interval))
     }
 
     async fn intermediate_output_roots(
--- a/crates/proof/contracts/src/aggregate_verifier.rs
+++ b/crates/proof/contracts/src/aggregate_verifier.rs
@@ -250,7 +250,7 @@ pub trait AggregateVerifierClient: Send + Sync {
     /// `AggregateVerifier` implementation applies to a game whose range starts
     /// at `starting_block`.
     ///
-    /// Cobalt switches the verifier to a shorter cadence at a fixed L2 block, so
+    /// Denim switches the verifier to a shorter cadence at a fixed L2 block, so
     /// the pair is a function of the game's starting block. Callers must resolve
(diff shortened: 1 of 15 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proposer/src/proof_recovery.rs
```diff
@@ -143,8 +151,14 @@ impl ProofRecovery {
...
         if let Some(cached) = usable_cache
             && cached.game_count == count
-            && ProofTarget::next_block(cached.state.l2_block_number, self.config.block_interval)
-                .is_none_or(|next_block| next_block > finalized_head)
+            && ProofTarget::next_block(
+                cached.state.l2_block_number,
+                self.intervals
+                    .for_starting_block(cached.state.l2_block_number)
+                    .await?
+                    .block_interval,
```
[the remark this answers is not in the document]
0xth4nh: Leaving the `?` here. Falling through on error would reach `forward_walk`, which resolves the intervals for the same starting block as its first step, so a transient RPC failure ends in the same `Err` either way — the warn-and-continue variant just spends an extra `game_count`/`anchor_snapshot` round trip before failing. `try_recover_and_plan` already turns that `Err` into a warn and a retry next tick, so the cached state is not lost, only deferred one poll interval.

The pattern at lines 81-89 differs because it is on the `Option`-returning path, where there is no `Err` to propagate.

crates/proof/contracts/src/aggregate_verifier.rs
```diff
@@ -239,6 +246,19 @@ pub trait AggregateVerifierClient: Send + Sync {
...
+    /// at `starting_block`.
+    ///
+    /// Cobalt switches the verifier to a shorter cadence at a fixed L2 block, so
+    /// the pair is a function of the game's starting block. Callers must resolve
+    /// it per game rather than reading `BLOCK_INTERVAL` once at startup.
+    async fn read_intervals_for_starting_block(
+        &self,
+        impl_address: Address,
+        starting_block: u64,
+    ) -> Result<(u64, u64), ContractError>;
```
jackchuma, line 261: Resolve challenger intervals for each game. The new start-dependent API is only consumed by `base-proposer`. `GameScanner::evaluate_game` still calls `read_intermediate_block_interval` through a cache keyed only by the current implementation address, and `ChallengerService` still passes startup `BLOCK_INTERVAL` values to `AnchorUpdater`. For any Cobalt-era game whose starting block selects a different interval pair, the scanner maps roots to incorrect checkpoint blocks and produces invalid proof requests, while the anchor updater builds the wrong UUID and cannot advance the anchor. Resolve the pair using each candidate or anchor starting block before validating or looking up games

crates/proof/proposer/src/service.rs
```diff
@@ -126,25 +127,8 @@ impl ProposerService {
...
-            return Err(eyre::eyre!(
-                "BLOCK_INTERVAL ({block_interval}) must be at least 2; single-block proposals are not supported"
-            ));
-        }
-        if block_interval % intermediate_block_interval != 0 {
-            return Err(eyre::eyre!(
-                "BLOCK_INTERVAL ({block_interval}) is not divisible by INTERMEDIATE_BLOCK_INTERVAL ({intermediate_block_interval})"
-            ));
-        }
+        let init_bond = factory_client.init_bonds(config.game_type).await?;
```
jackchuma, line 130: Validate the interval resolver before readiness. Startup now checks only `gameImpls` and `initBonds`, then starts the background pipeline and marks the service ready before making any `intervalsForStartingBlock` call. If the implementation lacks the new selector or returns an invalid pair, both recovery loops will fail the resolver on every poll, leaving a healthy/running service that never dispatches or submits proofs; the previous startup reads failed immediately in this situation. Resolve and validate a representative starting block during service initialization so this remains a startup error

jackchuma (commented):
Just a few minor comments. Also, any code that relies on a new function in AggregateVerifier should probably have a fallback set up if the function does not exist (so we can deploy this before the contract upgrade)

crates/proof/contracts/src/aggregate_verifier.rs
0xth4nh: Covered in #4931, which resolves scanner intervals from each game proxy and updates AnchorUpdater to resolve by the candidate starting block. I kept it separate so this PR stays scoped to proposer behavior; the two changes will be deployed together before Cobalt.

crates/proof/proposer/src/service.rs
0xth4nh: Fixed in 20938a981. Startup now resolves the interval pair for the current anchor block before the service can become ready. The contract client also falls back to the legacy fixed interval getters when intervalsForStartingBlock is unavailable, so this binary can be deployed before the verifier upgrade.

crates/proof/proposer/src/proof_submitter.rs
```diff
@@ -117,20 +119,25 @@ impl ProofSubmitter {
...
         }
 
+        // The intervals are a function of the starting block, not of process
+        // config: the verifier switches cadence at the Cobalt activation block.
+        let intervals = self
+            .intervals
+            .for_starting_block(starting_block_number)
+            .await
+            .map_err(SubmitAction::Failed)?;
+        if starting_block_number.saturating_add(intervals.block_interval) != target_block {
```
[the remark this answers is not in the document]
0xth4nh: Fixed in 0685a46ea. The guard now uses checked_add and rejects overflow.

crates/proof/contracts/src/aggregate_verifier.rs
```diff
@@ -461,6 +481,44 @@ impl AggregateVerifierClient for AggregateVerifierContractClient {
...
+            return Err(ContractError::validation(
+                "BLOCK_INTERVAL must be at least 2 (single-block proposals are not supported)",
+            ));
+        }
+        if intermediate_block_interval == 0 {
+            return Err(ContractError::validation("INTERMEDIATE_BLOCK_INTERVAL cannot be 0"));
+        }
+        if !block_interval.is_multiple_of(intermediate_block_interval) {
+            return Err(ContractError::validation(format!(
+                "BLOCK_INTERVAL ({block_interval}) is not divisible by                  INTERMEDIATE_BLOCK_INTERVAL ({intermediate_block_interval})"
```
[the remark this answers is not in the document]
0xth4nh: Fixed in the current head; the validation error is now a single-line string with no embedded formatting whitespace.

0xth4nh:
All of the referenced inline findings are already addressed:

- The `aggregate_verifier.rs` error string no longer contains formatting whitespace.
- The `proof_recovery.rs` `?` behavior was reviewed and intentionally retained; the inline thread explains why it retries identically without losing the cached state.
- The submitter guard uses `checked_add` and rejects overflow.

No remaining changes are needed for those findings.

crates/proof/contracts/src/aggregate_verifier.rs
```diff
@@ -239,6 +246,20 @@ pub trait AggregateVerifierClient: Send + Sync {
...
+    ///
+    /// Cobalt switches the verifier to a shorter cadence at a fixed L2 block, so
+    /// the pair is a function of the game's starting block. Callers must resolve
+    /// it per game rather than reading `BLOCK_INTERVAL` once at startup.
+    /// Legacy implementations fall back to their fixed interval getters.
+    async fn read_intervals_for_starting_block(
+        &self,
+        impl_address: Address,
+        starting_block: u64,
+    ) -> Result<(u64, u64), ContractError>;
```
jackchuma: Use start-aware cadence for basectl proof requests. `GamesClient::game_details` still treats `INTERMEDIATE_BLOCK_INTERVAL()` on a game proxy as the committed stride (`crates/infra/basectl/src/rpc/games.rs:393-402`), but this change makes the committed pair depend on the game's starting block. `ProofProposeRequest` then uses that value for the prover request and rejects any differing explicit override (`crates/infra/basectl/src/rpc/prover.rs:144-180`), so `basectl proofs propose` becomes unusable for games in the non-default cadence regime. Query `read_intervals_for_starting_block(address, starting_block)` and populate the returned intermediate interval in `GameDetails`

0xth4nh: Fixed in 58a14f1a6. `game_details` now calls `read_intervals_for_starting_block(game_proxy, starting_block)` and uses the returned intermediate interval, while preserving the legacy-getter fallback. I also updated the AggregateVerifier game-type probe to use the same start-aware API so the new implementation is not misclassified after the old getter is removed.

jackchuma (dismissed):

jackchuma (approved):

PR #4931, opened 2026-09-05
feat(challenger): resolve proposal intervals per game for Denim

0xth4nh:
#4930 has merged, so this diff is now against `main`. Depends on base/contracts#431, which adds `intervalsForStartingBlock(uint256)` to `AggregateVerifier`.

Implements section 3 of [the proof config switchover doc](https://bdocs.cbhq.net/d/denim-proof-config-switchover/v/latest/md).

## What changed?

**`crates/proof/contracts`**
- New `resolve_intervals(factory, verifier, game_type, starting_block)` helper: reads `gameImpls`, rejects a zero implementation, then calls `intervalsForStartingBlock` on it. Reading the implementation on every call means a governance `setImplementation` is picked up without a restart.
- It queries the *implementation* rather than a game proxy because its callers resolve intervals for games that do not exist yet — the anchor's successor, the proposer's next proposal. For a game that **does** exist, callers should call `read_intervals_for_starting_block` on the game's own proxy instead; see the `GameScanner` bullet below.
- `read_intervals_for_starting_block` therefore takes either address, and its parameter is named `verifier_address` rather than `impl_address`. Verifiers deployed before `intervalsForStartingBlock` existed hit the missing-method fallback to `BLOCK_INTERVAL()` / `INTERMEDIATE_BLOCK_INTERVAL()`, which is what keeps in-flight pre-upgrade games working.

**`crates/proof/challenge`**
- `AnchorUpdater` no longer takes `block_interval` / `intermediate_block_interval`. `next_game` resolves them from the anchor block — which is exactly where the next game's range starts — before building the lookup key.
- `GameScanner`: dropped `interval_cache` and `resolve_intermediate_block_interval`. `evaluate_game` now fetches `starting_block_number` first and resolves the interval from it — **through the game's own proxy, not the factory's current implementation**. A game is a CWIA clone, so it delegates to the implementation baked into its bytecode at creation; reading through it returns the pair the game was actually created with, and an implementation upgrade cannot retroactively change how an already-created game's checkpoints are interpreted. The starting block is already on the same code path, so this is one extra `eth_call` next to the seven the scanner already makes per game.
- `ChallengerService::new` still fails fast at startup, but on a resolution rather than on raw getters: it reads the anchor block from `AnchorStateRegistry`, then calls `resolve_intervals` for it. The `gameImpls != 0` check stays; the divisibility and non-zero checks moved into `read_intervals_for_starting_block`, where they now run on every resolution rather than once. The resolved pair is logged and discarded — nothing caches it. **Note:** this makes startup depend on `anchor_snapshot()` succeeding, which it did not before.
- `MockAggregateVerifier` gained `with_intervals` / `with_fast_intervals` so tests can model the cadence switch.

**`crates/proof/proposer`**
- `IntervalResolver::for_starting_block` delegates to the shared helper instead of duplicating the `gameImpls` lookup.

## Why?

The intervals are no longer process-wide constants. Post-Denim the verifier applies `BLOCK_INTERVAL = 6,000` / `INTERMEDIATE_BLOCK_INTERVAL = 300` to games starting at or after the activation block, and the pre-Denim `600` / `30` to everything before it. One implementation serves both sides, so anything that reads the pair once and keeps it is wrong for half the games it sees.

For `AnchorUpdater` the failure is silent: it computes the UUID of the game that should follow the anchor and looks it up in the factory. With a stale interval the UUID simply does not match anything, `games()` returns `address(0)`, and the updater logs `next anchor game not found` at debug and keeps going. The anchor stops advancing, which degrades the challenger's scan-start optimization (every tick rescans from a stale anchor) and the proposer's anchor fallback.

`GameScanner` was listed in the doc as already correct because its cache is keyed on implementation address and a `setImplementation` changes that address. That reasoning holds for an implementation swap but not for Denim: the switch happens *inside* one implementation, so the cache key never changes and every game gets whichever interval was resolved first. A wrong intermediate interval there produces a `CheckpointCountMismatch` per game (the Immunefi #74652 failure mode) rather than a silent stall — loud, but it stops the challenger from validating anything past the boundary.

## How to test?

```
cd crates/proof
cargo test -p base-challenger -p base-proposer -p base-proof-contracts
cargo clippy -p base-proof-contracts -p base-proposer -p base-challenger --all-targets -- -D warnings
```

Two new tests carry the regression, both written so they fail against the old behaviour rather than merely passing against the new one:

- `cargo test -p base-challenger --lib anchor::tests::poll_finds_next_game_built_with_fast_intervals` — anchor sits at the activation block; the next game spans the Denim interval and carries two intermediate roots. The same fixture is polled twice: once against a verifier that switches cadence (anchor advances, one `setAnchorState` tx) and once against a verifier that never switches, i.e. the pair the old startup-cached updater would have used (no tx — the UUID does not match). The pre-Denim target block is present in the mock L2 provider so the second arm fails on the UUID, not on a missing output root.
- `cargo test -p base-challenger --lib scanner::tests::test_scan_resolves_intervals_per_game_across_cadence_activation` — two games straddling the activation block; asserts they resolve `5` and `2` respectively. The old per-implementation cache returns one value for both.

`test_scan_happy_path` now asserts one interval read per candidate, keyed by each candidate's own proxy address, instead of one read per implementation — that assertion is both the cache removal and the read-through-the-proxy change.

Full run on this branch: 73 + 22 challenger, 59 proposer, 45 contracts, 0 failures.

## Known limitation

`AnchorUpdater::next_game` resolves from the factory's *current* implementation, but the successor game may already exist and have been created with an older pair. Across a `setImplementation` that changes the pair the keyed lookup never matches and the anchor stalls at `next anchor game not found` — and unlike the old startup-cached version, a restart no longer clears it. Denim itself does not trigger this: the pair is a function of the starting block, not of which implementation is current. Closing it means scanning the factory from the anchor index when the keyed lookup misses; left as a follow-up.

## Follow-ups

Nitro TEE and ZK (doc sections 4 and 5) are untouched and still need the same treatment.

---

### Update — back to Denim

The 200ms cadence switch has moved back to **Denim**, `ProtocolVersions` index 13. base/contracts#434 reverts the index in `AggregateVerifier` (12 → 13); the semver stays at 0.2.0 since that version has not shipped anywhere yet. This branch follows it in prose and fixture names.

Nothing functional moved in either direction. This service resolves the interval pair from the verifier per game and never names a fork itself, so the earlier Cobalt retarget and this one are both comment-and-identifier changes. The mock builders and test names introduced along the way are cadence-neutral (`fast_activation_block`, `with_fast_intervals`, `..._crosses_fast_activation_block`) and did not need renaming for either move.

Index 13 sits one past the end of the current Base mainnet schedule, where index 12 was in range but zero. `_firstFastBlock()` reports "unscheduled" for both, so the deploy-time requirement is unchanged: the activation has to be registered on `ProtocolVersions` before the fork or the verifier stays on the slow cadence silently.

The branch name still says `denim` — which is now correct again. It was never renamed, since that would close and reopen the PR.

### DIFF
--- a/crates/proof/challenge/src/anchor.rs
+++ b/crates/proof/challenge/src/anchor.rs
@@ -6,7 +6,7 @@ use alloy_primitives::Address;
 use base_proof_contracts::{
     AggregateVerifierClient, AnchorRoot, AnchorSnapshot, AnchorStateRegistryClient,
     DisputeGameFactoryClient, GameStatus, encode_set_anchor_state_calldata, game_lookup_blocks,
-    game_lookup_key,
+    game_lookup_key, resolve_intervals,
 };
 use base_proof_rpc::L2Provider;
 use base_tx_manager::TxManager;
@@ -23,8 +23,6 @@ pub struct AnchorUpdater {
     cached_next_game: Option<(AnchorSnapshot, Address)>,
     anchor_state_registry_address: Address,
     game_type: u32,
-    block_interval: u64,
-    intermediate_block_interval: u64,
 }
 
 impl std::fmt::Debug for AnchorUpdater {
@@ -32,8 +30,6 @@ impl std::fmt::Debug for AnchorUpdater {
         f.debug_struct("AnchorUpdater")
             .field("anchor_state_registry_address", &self.anchor_state_registry_address)
             .field("game_type", &self.game_type)
-            .field("block_interval", &self.block_interval)
-            .field("intermediate_block_interval", &self.intermediate_block_interval)
             .finish_non_exhaustive()
     }
 }
@@ -46,8 +42,6 @@ impl AnchorUpdater {
         l2_provider: Arc<dyn L2Provider>,
         anchor_state_registry_address: Address,
         game_type: u32,
-        block_interval: u64,
-        intermediate_block_interval: u64,
     ) -> Self {
         Self {
             factory_client,
@@ -56,8 +50,6 @@ impl AnchorUpdater {
             cached_next_game: None,
             anchor_state_registry_address,
             game_type,
-            block_interval,
-            intermediate_block_interval,
         }
     }
 
@@ -79,7 +71,8 @@ impl AnchorUpdater {
             Some((cached_anchor, address)) if cached_anchor == anchor => address,
             _ => {
                 self.cached_next_game = None;
-                let Some(address) = self.next_game(anchor.anchor_root, anchor.anchor_game).await
+                let Some(address) =
+                    self.next_game(verifier_client, anchor.anchor_root, anchor.anchor_game).await
                 else {
                     return;
                 };
@@ -114,11 +107,34 @@ impl AnchorUpdater {
         }
     }
 
-    async fn next_game(&self, anchor_root: AnchorRoot, anchor_game: Address) -> Option<Address> {
+    async fn next_game(
+        &self,
+        verifier_client: &dyn AggregateVerifierClient,
+        anchor_root: AnchorRoot,
+        anchor_game: Address,
+    ) -> Option<Address> {
+        // Resolved from the anchor block, which is where the next game's range starts:
+        // the verifier switches to a shorter cadence at the Denim activation block, so a
+        // pair cached at startup silently stops matching any game past the boundary.
+        let (block_interval, intermediate_block_interval) = match resolve_intervals(
+            self.factory_client.as_ref(),
+            verifier_client,
+            self.game_type,
+            anchor_root.l2_block_number,
+        )
+        .await
+        {
+            Ok(intervals) => intervals,
+            Err(e) => {
+                warn!(error = %e, "failed to resolve anchor update intervals");
+                return None;
+            }
+        };
+
         let blocks = match game_lookup_blocks(
             anchor_root.l2_block_number,
-            self.block_interval,
-            self.intermediate_block_interval,
+            block_interval,
+            intermediate_block_interval,
         ) {
             Ok(blocks) => blocks,
             Err(e) => {
@@ -154,8 +170,8 @@ impl AnchorUpdater {
         let key = match game_lookup_key(
             anchor_root.l2_block_number,
             parent,
-            self.block_interval,
-            self.intermediate_block_interval,
+            block_interval,
+            intermediate_block_interval,
             &intermediate_roots,
         ) {
             Ok(key) => key,
@@ -309,14 +325,18 @@ mod tests {
 
     use super::*;
     use crate::test_utils::{
-        MockAggregateVerifier, MockAnchorStateRegistry, MockDisputeGameFactory, MockL2Provider,
-        MockTxManager, addr, build_test_header_and_account, mock_state, receipt_with_status,
+        MockAggregateVerifier, MockAnchorStateRegistry, MockDisputeGameFactory, MockGameState,
+        MockL2Provider, MockTxManager, addr, build_test_header_and_account, mock_state,
+        receipt_with_status,
     };
 
     const ASR_ADDRESS: Address = Address::new([0xAA; 20]);
     const GAME_TYPE: u32 = 1;
     const BLOCK_INTERVAL: u64 = 100;
     const INTERMEDIATE_BLOCK_INTERVAL: u64 = 100;
+    const DENIM_ACTIVATION_BLOCK: u64 = 200;
+    const DENIM_BLOCK_INTERVAL: u64 = 50;
+    const DENIM_INTERMEDIATE_BLOCK_INTERVAL: u64 = 25;
 
     fn insert_l2_block(l2: &mut MockL2Provider, block: u64) -> B256 {
         let storage_hash = B256::repeat_byte(block as u8);
@@ -338,6 +358,12 @@ mod tests {
                 .unwrap()
                 .extra_data;
         factory.insert_uuid_game(GAME_TYPE, output_root, extra_data, game);
+    }
+
+    /// The mock verifier resolves the intervals the anchor tests build their games with.
+    fn verifier(games: HashMap<Address, MockGameState>) -> MockAggregateVerifier {
+        MockAggregateVerifier::new(games)
+            .with_intervals(BLOCK_INTERVAL, INTERMEDIATE_BLOCK_INTERVAL)
     }
 
     fn tx_success(tx_hash: B256) -> base_tx_manager::SendResponse {
@@ -362,8 +388,6 @@ mod tests {
             l2 as Arc<dyn L2Provider>,
             ASR_ADDRESS,
             GAME_TYPE,
-            BLOCK_INTERVAL,
-            INTERMEDIATE_BLOCK_INTERVAL,
         )
     }
 
@@ -379,7 +403,7 @@ mod tests {
         let mut state = mock_state(GameStatus::DefenderWins, Address::ZERO, 100);
         state.anchor_state_registry = ASR_ADDRESS;
 
-        let verifier = MockAggregateVerifier::new(HashMap::from([(game, state)]));
+        let verifier = verifier(HashMap::from([(game, state)]));
         let (submitter, tx_manager) = submitter(vec![tx_success(tx_hash)]);
         let mut updater = updater(factory, anchor_registry, Arc::new(l2));
 
@@ -401,7 +425,7 @@ mod tests {
         insert_next_game(&factory, ASR_ADDRESS, output_root, game);
         let state = mock_state(GameStatus::InProgress, Address::ZERO, 100);
 
-        let verifier = MockAggregateVerifier::new(HashMap::from([(game, state)]));
+        let verifier = verifier(HashMap::from([(game, state)]));
         let (submitter, tx_manager) = submitter(vec![]);
         let mut updater = updater(factory, anchor_registry, Arc::new(l2));
 
@@ -421,7 +445,7 @@ mod tests {
         insert_next_game(&factory, ASR_ADDRESS, output_root, game);
         let state = mock_state(GameStatus::InProgress, Address::ZERO, 100);
 
-        let verifier = MockAggregateVerifier::new(HashMap::from([(game, state.clone())]));
+        let verifier = verifier(HashMap::from([(game, state.clone())]));
         let (submitter, tx_manager) = submitter(vec![tx_success(tx_hash)]);
         let mut updater = updater(Arc::clone(&factory), anchor_registry, Arc::new(l2));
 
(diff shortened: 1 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/contracts/src/aggregate_verifier.rs
```diff
@@ -15,8 +15,38 @@ use async_trait::async_trait;
...
+/// Denim switches the verifier to a shorter cadence at a fixed L2 block, so the pair is
+/// a function of the starting block and must be resolved per game rather than read once
+/// at startup.
+///
+/// The implementation address is read from the factory on every call so a governance
+/// `setImplementation` is picked up without a restart. The implementation — not the game
+/// proxy — is queried on purpose: callers also resolve intervals for games that do not
+/// exist yet, and games created before the Denim-aware implementation was deployed run
+/// bytecode that has no `intervalsForStartingBlock`. For pre-Denim starting blocks the
+/// Denim-aware implementation returns the same pair the old one had.
```
0x00101010: This justification is stale after df1f055f2. The scanner now calls `read_intervals_for_starting_block(factory.proxy, …)` on exactly the games this paragraph says lack `intervalsForStartingBlock`, and it works because #4930's `is_missing_method()` fallback drops to `BLOCK_INTERVAL()` / `INTERMEDIATE_BLOCK_INTERVAL()`.

Suggest keeping only the first reason (the anchor's successor and the proposer's next game have no proxy yet) and pointing existing games at the proxy path:

```suggestion
/// The implementation address is read from the factory on every call so a governance
/// `setImplementation` is picked up without a restart. The implementation — not a game
/// proxy — is queried because callers resolve intervals for games that do not exist yet
/// (the anchor's successor, the proposer's next proposal). For an existing game, call
/// `read_intervals_for_starting_block` on its proxy instead so the pair it was created
/// with is used even after an implementation upgrade; legacy proxies without
/// `intervalsForStartingBlock` fall back to their fixed interval getters.
```

The PR body's `crates/proof/contracts` and `GameScanner` bullets have the same drift — the body never says the scanner reads through the proxy, which is the more important design point of the two.

crates/proof/challenge/src/scanner.rs
```diff
@@ -393,18 +380,19 @@ impl GameScanner {
...
-            },
-            self.resolve_intermediate_block_interval(factory.game_type),
+        let (info, starting_block_number, l1_head) = tokio::try_join!(
+            self.verifier_client.game_info(factory.proxy),
+            self.verifier_client.starting_block_number(factory.proxy),
+            self.verifier_client.l1_head(factory.proxy),
         )?;
 
+        // Read through the game proxy so implementation upgrades cannot change
+        // how an already-created game's checkpoints are interpreted.
```
0x00101010: Worth stating the dependency explicitly: this only works for games created before the Denim-aware implementation because `read_intervals_for_starting_block` falls back to `BLOCK_INTERVAL()` / `INTERMEDIATE_BLOCK_INTERVAL()` when the selector is missing. If that fallback is ever tightened (e.g. to stop treating every empty revert as missing-method), every in-flight pre-upgrade game starts hitting the `game query failed` warn path.

crates/proof/challenge/src/scanner.rs
```diff
@@ -654,9 +606,11 @@ mod tests {
         assert_eq!(candidates[3].index, 4);
         assert_eq!(candidates[3].factory.game_type, 1);
         assert_eq!(candidates[3].info.l2_block_number, 400);
+        // Existing games resolve through their own proxies, not the factory's current
+        // implementation.
         assert_eq!(
             verifier.intermediate_block_interval_reads.lock().unwrap().as_slice(),
-            &[Address::repeat_byte(0x11)],
+            &[addr(0), addr(1), addr(3), addr(4)],
```
0x00101010: nit: `scan` runs `evaluate_game` through `buffer_unordered(32)`; this ordering is deterministic today only because the mocks never yield. Sorting before comparing keeps the intent — one read per candidate, through its own proxy — if a mock ever awaits:

```rust
let mut reads = verifier.intermediate_block_interval_reads.lock().unwrap().clone();
reads.sort();
assert_eq!(reads, [addr(0), addr(1), addr(3), addr(4)]);
```

crates/proof/challenge/src/anchor.rs
```diff
@@ -114,11 +107,34 @@ impl AnchorUpdater {
...
-    async fn next_game(&self, anchor_root: AnchorRoot, anchor_game: Address) -> Option<Address> {
+    async fn next_game(
+        &self,
+        verifier_client: &dyn AggregateVerifierClient,
+        anchor_root: AnchorRoot,
+        anchor_game: Address,
+    ) -> Option<Address> {
+        // Resolved from the anchor block, which is where the next game's range starts:
+        // the verifier switches to a shorter cadence at the Denim activation block, so a
+        // pair cached at startup silently stops matching any game past the boundary.
```
0x00101010, line 118: One limitation worth a sentence here: this resolves from the factory's *current* implementation, but the game being looked up already exists and was built with whatever pair applied at creation. Across a `setImplementation` that changes the pair, the UUID never matches and the anchor stalls at `next anchor game not found` — and unlike the startup-cached version, a restart no longer clears it. Not a regression in kind and Denim itself does not trigger it, but the comment reads as if the anchor block alone guarantees the right pair. A factory scan from the anchor index when the keyed lookup misses would close it; fine as a follow-up.

crates/proof/challenge/src/service.rs
```diff
@@ -131,6 +101,15 @@ impl ChallengerService {
...
             "AnchorStateRegistry client initialized"
         );
+        let anchor_block =
+            anchor_registry_client.anchor_snapshot().await?.anchor_root.l2_block_number;
+        let (block_interval, intermediate_block_interval) =
+            resolve_intervals(&factory_client, &verifier_client, config.game_type, anchor_block)
+                .await?;
+        info!(
+            anchor_block,
+            block_interval, intermediate_block_interval, "Validated proposal interval resolver"
```
0x00101010: nit: "Validated proposal interval resolver" describes the intent, not the event. Something like `"resolved proposal intervals at anchor block"` says what happened. Also note the startup fail-fast now depends on `anchor_snapshot()` succeeding (line 105), which the PR body does not mention.

0x00101010 (commented):
Verified locally on the branch head: `cargo test -p base-challenger -p base-proposer -p base-proof-contracts` → 73 + 22, 59, 45 passed, 0 failed; clippy with `-D warnings` on all targets is clean.

The core change is right: intervals are resolved where the starting block is known (anchor block in `AnchorUpdater::next_game`, `startingBlockNumber()` in `GameScanner::evaluate_game`, caller-supplied in the proposer), and both regression tests genuinely discriminate old from new behaviour. Per-game scanner failures are warned and skipped, so a bad interval read degrades one game, not the tick.

No blocking issues. Inline comments are doc fixes and a test nit — the `resolve_intervals` doc and the PR body still describe the pre-`df1f055f2` design where the scanner read from the implementation, and the next reader could "correct" the scanner back to it.

crates/proof/contracts/src/aggregate_verifier.rs
0xth4nh: Fixed in 7b02749b5: took the suggestion. The `resolve_intervals` doc now keeps only the "games that do not exist yet" reason and points existing games at the proxy path. I also carried it into the trait doc for `read_intervals_for_starting_block`, since that is where a caller decides which address to pass, and renamed the parameter `impl_address` → `verifier_address` — it takes an implementation *or* a game proxy, and the old name said otherwise.

The PR body had the same drift and is now updated: the `crates/proof/contracts` bullets carry the corrected justification, and the `GameScanner` bullet spells out the read-through-the-proxy design and why CWIA makes it work.

crates/proof/challenge/src/scanner.rs
0xth4nh: Fixed in 7b02749b5: spelled out both halves at the call site — that a game is a CWIA clone (delegates to the implementation baked in at creation, not an upgradeable proxy), and that pre-upgrade games only work because `read_intervals_for_starting_block` falls back to `BLOCK_INTERVAL()` / `INTERMEDIATE_BLOCK_INTERVAL()` on missing-method, so tightening that fallback pushes every in-flight pre-upgrade game onto the warn path below.

crates/proof/challenge/src/scanner.rs
0xth4nh: Fixed in 7b02749b5: applied the suggestion verbatim.

crates/proof/challenge/src/anchor.rs
0xth4nh: Fixed in 7b02749b5: recorded the limitation in the comment — current implementation vs. the pair the successor was created with, the `next anchor game not found` stall across a `setImplementation` that changes the pair, that a restart no longer clears it, that Denim itself does not trigger it, and the factory-scan-from-anchor-index fix. Left as a follow-up as you suggested.

crates/proof/challenge/src/service.rs
0xth4nh: Fixed in 7b02749b5: log message is now `resolved proposal intervals at anchor block`.

The `anchor_snapshot()` dependency is real — startup now fails if that call fails, where before it did not. Added to the PR body; the `ChallengerService::new` bullet also no longer claims the service stopped reading intervals at startup, which it does not.

crates/proof/contracts/src/aggregate_verifier.rs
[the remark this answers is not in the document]
0xth4nh: Disagree — this finding is stale rather than wrong. It was written on 2026-09-09 against the Cobalt-targeted revision; the team moved the 200ms cadence switch back to Denim and the branch was retargeted the same day (47053acfd). The contract now reads `ProtocolVersions` index 13 (Denim), so the prose saying "Denim" matches what the code targets. `anchor.rs`, `scanner.rs` and `test_utils.rs` are Denim-worded for the same reason. Nothing changed for this one.

0x00101010 (approved):

PR #4932, opened 2026-09-06
chore(genesis): close out the block-time audit follow-ups

0xth4nh:
Last piece of the Denim proof-config switchover workstream in this repo. Follows #4930 (proposer) and #4931 (challenger); those two carry the behavioural change, this one closes the two loose ends the block-time audit left behind.

## What changed?

**1. Deleted `RollupConfig::block_number_lower_bound_from_timestamp`.**

```rust
pub const fn block_number_lower_bound_from_timestamp(&self, timestamp: u64) -> u64 {
    timestamp.saturating_sub(self.genesis.l2_time).saturating_div(self.block_time)
}
```

Dead public API — zero production call sites, only its own unit test. Deleted the test with it.

**2. Rewrote §4 "Add activation methods to `RollupConfig`" in `docs/guides/UPGRADES.md`.**

The section still taught hand-written predicates:

```rust
pub fn is_first_base_azul_block(&self, timestamp: u64) -> bool {
    self.is_base_azul_active(timestamp)
        && !self.is_base_azul_active(timestamp.saturating_sub(self.block_time))
}
```

Neither the shape nor the arithmetic matches the code any more. It now documents the `rollup_fork_methods!` macro, shows the standalone and cascading entry forms against real entries (`Base Azul`, `Isthmus implies Jovian`), and states why the generated `is_first_X_block` takes an explicit parent timestamp.

## Why?

Different reasons for the two.

The dead function is a trap, not a bug. It looks like the inverse of `l2_block_timestamp()` and is not: it assumes the legacy 2s cadence, so for any post-Denim block it returns roughly `denim_activation_block + (t - denim_ts) / 2` instead of the real number. It cannot be fixed in place either — a seconds-granular timestamp maps to about five 200ms blocks, so a `u64` return cannot express the answer. Renaming it to advertise the assumption was the alternative; deleting is cheaper and nothing calls it.

The doc is the one that actually costs something. `block_time` **stays `2` after Denim** — the 200ms cadence is layered on top of the legacy schedule rather than replacing it — so `timestamp - block_time` walks back ten blocks, not one, on a post-Denim chain. `is_first_X_block` written that way returns true for every block in that window. That is exactly the log-only defect that exists today at `crates/consensus/service/.../engine_request_processor.rs:349`, and the guide would reintroduce it at every future hardfork. Blast radius is every fork after this one, which is why it is worth the diff.

## How to test?

```
cargo clippy -p base-common-genesis --all-targets -- -D warnings
cargo test -p base-common-genesis
```

98 tests pass, clippy clean. The deletion is compiler-verified: `block_number_lower_bound_from_timestamp` was `pub`, so any surviving caller anywhere in the workspace fails the build. Confirm the tree is clean:

```
git grep -n block_number_lower_bound_from_timestamp   # expect no output
```

For the doc, check every symbol it now names actually exists:

```
git grep -n 'is_base_azul_active\|implies is_jovian_active' crates/common/genesis/src/rollup.rs
git grep -n 'macro_rules! rollup_fork_methods' crates/common/genesis/src/rollup.rs
```

`crates/common/genesis/src/rollup.rs:188` is the line that makes the point — `pub fn $first(&self, timestamp: u64, parent_timestamp: u64)`, parent passed in rather than derived.

## Not in scope

The audit's other findings are outside the proof system and belong to other owners — gossip peer scoring (`crates/consensus/peers/src/score.rs`), bundle metering's hardcoded `BLOCK_TIME = 2`, proofs-history retention expressed in 2s blocks, the `engine_request_processor.rs:349` synthetic parent timestamp, and `ingress-rpc`'s `block_time_milliseconds` default of 2000. Being routed separately; none are consensus-critical.

### DIFF
--- a/crates/common/genesis/src/rollup.rs
+++ b/crates/common/genesis/src/rollup.rs
@@ -437,15 +437,6 @@ impl RollupConfig {
         (full_millis.saturating_div(1_000), (full_millis % 1_000) as u16)
     }
 
-    /// Computes the lower-bound block number for a timestamp, relative to the L2 genesis time and
-    /// the block time.
-    ///
-    /// This uses floor division, so multiple blocks can share the same seconds-denominated
-    /// timestamp while still mapping to the same lower bound.
-    pub const fn block_number_lower_bound_from_timestamp(&self, timestamp: u64) -> u64 {
-        timestamp.saturating_sub(self.genesis.l2_time).saturating_div(self.block_time)
-    }
-
     /// Checks the scalar value in Ecotone.
     pub fn check_ecotone_l1_system_config_scalar(scalar: [u8; 32]) -> Result<(), &'static str> {
         let version_byte = scalar[0];
@@ -1286,17 +1277,5 @@ mod tests {
     fn l2_block_full_millis_saturates() {
         let saturating = rollup_config_with_denim(u64::MAX, u64::MAX, None);
         assert_eq!(saturating.l2_block_timestamp_millis(1), u64::MAX);
-    }
-
-    #[test]
-    fn test_compute_block_number_lower_bound_from_time() {
-        let cfg = RollupConfig {
-            genesis: ChainGenesis { l2_time: 10, ..Default::default() },
-            block_time: 2,
-            ..Default::default()
-        };
-
-        assert_eq!(cfg.block_number_lower_bound_from_timestamp(20), 5);
-        assert_eq!(cfg.block_number_lower_bound_from_timestamp(30), 10);
     }
 }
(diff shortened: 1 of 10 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

0x00101010 (approved):

Issue #16, opened 2025-04-18
Will there be flashblocks state integration with revm?


ScriptLike:
Hi,

Was going through github and was pleasantly surprised to find this reth exex. I recently saw on CT that Anika said that Base is targeting May for flashblocks release. Will there be state integration in this repo either through revm or another method before release? I was planning to create a exex for both state and flashblocks rpc support but if you guys have plans to support both features pre flashblocks mainnet release (gives users sometime to implement and test on testnet), I would be happy to contribute in anyway. 

Thanks!

danyalprout:
Hey! We would love to have execution of the transactions done per flashblock. It would enable more RPC's (like eth_call) and potentially be a performance improvement.

@cody-wang-cb was planning to look at this, but if you want to contribute please do!

ScriptLike:
Hey @danyalprout and @cody-wang-cb glad to hear executions done per flashblocks is in the roadmap. If the timeline for state integration to support RPC's like eth_call is post mainnet release I would love to submit a PR in the upcoming weekend or two. Else if there are already plans in the work, maybe @cody-wang-cb could post a spec of how he plans to approach state integration and we could bounce some ideas off each other?

cody-wang-cb:
Hey @ScriptLike ! Great to hear we have other folks looking into contributing to the flashblock RPC :) 
I haven't started to look into it so feel free to open a new issue/PR to explore your design. Also this isn't going to be part of our initial launch for mainnet, but something we for sure want to integrate whenever it's ready.

ScriptLike:
Hey @cody-wang-cb @danyalprout I just got some time to take a look and sync up a base sepolia node. Have a draft ready and will test this week some simulations for eth_call like endpoint

Quick question, I noticed that my reth node seems to sync blocks around 2 blocks slower than the flashblocks data's block number. I have --engine.persistence-threshold 0 and --engine.memory-block-buffer-target 0 set. Is my peering/configuration just bad or is there another setting I am missing

Btw is there also a discord or telegram group specifically for flashblocks feature development?

ScriptLike:
Hey @cody-wang-cb, @danyalprout , @jowparks is there a reason why sometimes there are duplicate txns being emitted? Many times at block n index 10 txn A is emitted but then at block n+1 index 1 the same txn A is emitted again.

Futhermore sometimes the flashblocks websocket doesn't seem to emit all transactions for a future block and sometimes it will just skip blocks. 

Are the above behaviors known? 

[...]

PR #4954, opened 2026-09-08
fix(precompiles): tri-state, hardfork-gated announce decode (Cantina #16 full remediation)

rayyan224:
## Summary

PR #4730 (commit `d6069d2ac052`) fixed BOP-603 / Cantina #16: a valid aliased `bytes[]` payload could force `announce` into an O(aliases·tail_bytes) owned decode. A post-merge review found the fix incomplete on two levels, both addressed here across two commits.

**Commit 1 — close the regression.** The borrowed fast path added in #4730 called `SolType::type_check(&token)` and discarded the `Result` via `.ok()?`. alloy 1.6.1's `type_check` builds its `Err` on failure by re-ABI-encoding the entire announce tuple, including every `bytes[]` alias, then hex-encoding the result — work thrown away one line later. `type_check` is just `if valid_token(token) { Ok(()) } else { Err(<expensive>) }`, so swapping to a direct `valid_token` check keeps every accept/reject decision identical while skipping that wasted construction. Verified empirically: reverted the change mid-review, reran the same test payload against the old code, confirmed identical output before restoring the fix.

**Commit 2 — the underlying gap.** Even after commit 1, a *recognized* malformed `announce` (structurally valid, but `valid_token` rejects it — e.g. invalid UTF-8 in `id`/`description`/`uri`) still fell through to the same owned decoder that valid payloads used to hit before #4730, paying its O(aliases·tail_bytes) diagnostic construction. That's pre-existing cost, not something #4730 introduced, but it's the deeper issue the review's five-point remediation was actually asking to close.

`decode_if_announce` (renamed from `try_from_calldata`) now returns a tri-state result instead of a plain `Option`:
- not an `announce` call at all → falls through to the generic decode, unchanged
- decodes cleanly → runs the call, unchanged
- recognized but malformed → the version decides inline whether that can be answered cheaply

V1 (Beryl, live on mainnet) can't take the cheap path — its revert bytes are already consensus-frozen — so it keeps falling through to the same expensive diagnostic it always has, byte-for-byte. V2 (Cobalt, confirmed not scheduled on any real network) short-circuits with a fixed-size `AbiDecodeFailed` before the owned decoder ever runs.

## Why not just bump alloy?

alloy 1.7.2 already fixes the underlying diagnostic cost generally (`type_check_fail_token` no longer re-encodes the token at all). Deliberately not used here: it would change revert bytes for every `type_check` failure across every B-20 surface, which is a live consensus break for V1 unless V1 is dual-linked to a frozen 1.6.1 decoder. That's real, but bigger, riskier, and cross-cutting — tracked separately against an existing related finding (Cantina #21 / BOP-628), not folded into this fix.

## Test plan

- Restructured `announce_dispatch_matches_owned_abi_oracle`: rows that fail via selector/length/decode_sequence issues still assert V1 == V2 == oracle, unchanged. The non-UTF-8 rows now assert V1 stays byte-identical to the oracle while V2 diverges to the new bounded error — proving the short-circuit actually fired, not just that behavior changed.
- Added golden-pin tests in both `b20_asset_v1_golden.rs` and `b20_asset_v2_golden.rs` (mirroring the real V1-unmetered/V2-metered permit-gas precedent) against the full production storage harness.
- `cargo test -p base-common-precompiles --features test-utils` — 683 lib tests + all integration suites pass.
- `cargo clippy -p base-common-precompiles --lib --tests` — clean.
- `cargo +nightly fmt --all -- --check` — clean.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

roger-bai-coinbase (approved):

PR #4970, opened 2026-09-08
fix(consensus): reject non-canonical transaction encodings

meyer9:
## Summary

Ports the relevant protections from ethereum-optimism/optimism#22778 to Base's transaction, payload, and flashblock handling.

- Reject a deposit transaction RLP body when its required `0x7e` EIP-2718 type byte is absent.
- Require payload and attributes transactions to decode and re-encode byte-for-byte before accepting them.
- Compare attributes transaction bytes directly during consolidation, rather than accepting different byte encodings that decode to the same transaction.
- Keep attributes transaction counts independent of decoding success.

## Testing

- `cargo test -p base-common-consensus -p base-common-rpc-types-engine -p base-protocol -p base-consensus-engine -p base-execution-payload-builder --lib`
- `cargo check -p base-flashblocks --lib`
- `cargo check -p base-common-consensus --no-default-features`

PR #4971, opened 2026-09-08
fix(consensus): reject non-canonical transaction encodings

meyer9:
## Summary

Backports the canonical EIP-2718 transaction-encoding protections from #4970 to `releases/v1.3.2`.

Ports the relevant protections from ethereum-optimism/optimism#22778 to Base's transaction, payload, and flashblock handling.

- Reject a deposit transaction RLP body when its required `0x7e` EIP-2718 type byte is absent.
- Require payload and attributes transactions to decode and re-encode byte-for-byte before accepting them.
- Compare attributes transaction bytes directly during consolidation, rather than accepting different byte encodings that decode to the same transaction.
- Keep attributes transaction counts independent of decoding success.

## Testing

- `cargo test -p base-common-consensus -p base-common-rpc-types-engine -p base-protocol -p base-consensus-engine -p base-execution-payload-builder --lib`
- `cargo check -p base-flashblocks --lib`


meyer9:
Superseded by #4973, whose source branch is hosted in base/base.

meyer9:
Closing in favor of #4973.

PR #4996, opened 2026-09-09
fix(flashblocks): measure pending staleness from the snapshot tip

### LINKED ISSUE #4916: Pending snapshot is dropped every few blocks on a healthy node: staleness measured from the inherited earliest header
Hi — first, thank you for `#4793`. We run a Base full node dedicated to low-latency flashblock
consumption, and the queue-lag problem that PR describes is real for us too. While measuring the
fix on mainnet we ran into a side effect we think is unintended, and we would like to offer both
the measurement and a possible correction. We may well be missing context, so please treat this
as a proposal rather than a bug report.

We are running `#4793` cherry-picked onto `v1.3.2-rc.2`, i.e. the commit as it sits on `main`
today — we picked it up deliberately, so this is not a report about a shipped release. As far as
we can tell it has not reached any release branch yet (`releases/v1.3.0`, `v1.3.1` and `v1.3.2`
are all behind it), which is why we are raising it now rather than after it ships.

### What we measure

Our node processes every flashblock of every block — `flashblocks_in_block` reports **11 at every
quantile** (sum/count = 13281/1208 = 11.00). But subscribers receive the complete set for only
part of the blocks. We compare our own `newFlashblocks` stream against the public one block by
block, and count a block as lost when we deliver fewer flashblocks than the sequencer sent:

| | blocks compared | blocks delivered incomplete |
|---|---|---|
| before the change | 149 | **24 (16.1 %)** |
| after the change, three windows | 148 + 149 + 298 | **0 (0.0 %)** |
| after the change, **7 h continuous, no restart** | **12 584** | **3 (0.024 %)** |
| **A/B test — reverted**: same binary, the two lines below put back | 149 | **25 (16.8 %)** |
| **A/B test — re-applied**, immediately after | 149 | **0 (0.0 %)** |

The first three rows do not establish causation on their own: a restart, a different uptime and
different traffic all sit between them. So we built a binary identical to the running one except
for the two lines below, and measured both states back to back, each after a restart with the
same 151 s warm-up. The loss returned to 16.8 % and disappeared again when the change was put
back. In those two windows the sequencer sent essentially the same number of frames (1647 and
1649) while subscribers received **1403** and **1649** respectively.

We also want to be precise about the 7 h row, because the short windows alone would have let us
claim zero. They do not prove it: at the rate that row measures — 1 block in 4 195 — the expected
count in 595 blocks is **0.14**, so "0 out of 595" was consistent with a problem still being there.
Only the long window is large enough to say anything.

The incomplete blocks arrive with 1–3 of 11 flashblocks and occur in a **strict period of
6 blocks**. From the revert window:
`50868145, 50868151, 50868157, 50868163, 50868169, 50868175, 50868181, 50868187`.

With the change applied, `pending_drop_stale` increments **26 times in 6.99 h** across 12 584
blocks (0.06/min, sampled every minute, no restart in the window).

**The 3 remaining blocks in the 7 h window have a different cause, which we traced separately.**
Each is preceded by a `NonceTooHigh` failure on flashblock index 1 — 26 ms before it in the one
event we captured with millisecond timing on both ends. We then instrumented both ends of the
stream: over 6.6 h we compared 11 501 blocks frame-by-frame against the canonical chain, and in
**20** of them the flashblock stream carried only a **truncated prefix** of the block — transaction hashes matching in order, 2 to 82
transactions missing from the tail, and `index 0` of the next block confirming the truncation via
`prev_flashblock_id`. That is **#693**, and we are reopening it separately with the measurement.
**It is unrelated to the bound discussed here**, and it is not something a node can fix: those
frames were never sent.

### The mechanism, as we read it

`PendingBlocksBuilder::from_previous()` carries the earliest header verbatim when the snapshot
rolls onto the next block:

```rust
let headers = vec![pending_blocks.earliest_header.clone(), pending_blocks.latest_header.clone()];
```

This is deliberate — the comment in `ReconciliationStrategy::Continue` says so: *"We do not retain
specific flashblocks here to avoid losing track of our 'earliest' pending block number"*.

The consequence is that `earliest_block_number()` never advances while the snapshot keeps pace with
the chain, so `best - earliest` grows by one per block even when the snapshot is perfectly fresh
(`latest == best + 1`). It measures how **wide** the snapshot is, not how far **behind** it is.
At the default `max_depth` of 3, `is_anchored_near()` therefore returns false every few blocks on
a node that is exactly at the tip.

That matters most in `publish_pending_blocks()`, because — as its own comment says — *"Every build
path funnels through here"*. A false negative there does not discard a stored snapshot; it refuses
to publish the one just built, and `apply_flashblock` then swaps the stored snapshot to `None`.
Every later flashblock of that block hits `MissingFirstFlashblock` and returns silently: the cache
can only chain onto a buffered predecessor, and only an index-0 payload carries the `base` needed
to rebuild. So the remainder of the block never reaches subscribers.

We think this is also why the PR's expectation — *"Missing pending is fine; the next base flashblock
on the current tip rebuilds it"* — holds later than intended: recovery does happen, but only at the
**next** index 0, which is one full block away.

Worth noting that `CanonicalBlockReconciler::reconcile()` uses the same `canonical - earliest`
measure, but its `DepthLimitExceeded` branch **retains the post-canonical flashblocks and rebuilds**
rather than wiping — so the same measure is already handled non-destructively one layer down.

### How we found it

We had the loss measured but were reading it as our own problem, and were close to reverting
`#4793` altogether. What redirected us was the PR description itself: the note that
`pending_drop_stale` *"ticked once at startup and then didn't climb"*. On our node it climbed
steadily. That one sentence pointed at the bound faster than any of our own measurements did.

### Proposed correction

Measure staleness from the snapshot's tip rather than from its inherited anchor, in the two places
`#4793` added:

```rust
// processor.rs :: is_anchored_near()
-        best.saturating_sub(pending_blocks.earliest_block_number()) <= self.max_depth
+        best.saturating_sub(pending_blocks.latest_block_number()) <= self.max_depth

// state.rs :: drop_pending_behind()
-        let earliest_pending_block = stale.earliest_block_number();
-        if canonical_block_number.saturating_sub(earliest_pending_block)
+        let latest_pending_block = stale.latest_block_number();
+        if canonical_block_number.saturating_sub(latest_pending_block)
```

This keeps the guard doing what it was introduced for: when the processor genuinely stalls, `latest`
stops advancing and the bound fires. `extends_canonical_tip()` independently requires `latest > best`,
so a snapshot that stopped extending the tip is still rejected. Snapshot width stays bounded by
`reconcile()`'s `DepthLimitExceeded` branch, which rebuilds instead of wiping.

### On the intent behind the bound

We think the concern behind `is_anchored_near()` is sound: `canonical_block_number()` returns
`earliest_header.number - 1`, and as the comment says, consumers execute against it. A snapshot
anchored far behind does make that base an old block.

What we believe makes it safe to measure from the tip instead is that `from_previous()` clones the
full flashblock vector, so `bundle_state` covers every block from `earliest` to `latest`. A wide
snapshot is a *further base plus a thicker overlay*, not stale state — the cost is size, not
correctness. And size is already bounded by `reconcile()`'s `DepthLimitExceeded` branch, which
rebuilds rather than wipes. So the proposal is less "remove a guard" than "let the branch that
already handles width handle it, and let the tip check handle staleness".

We also checked `base-metering`'s collector: it reads `earliest_block_number()` to detect a
snapshot regression and, when one is detected, as the point to clear the cache from. We do not
change that accessor, so metering behaves identically either way.

### Why the existing tests did not catch it

Every fixture builds a snapshot from a **single** block, where `earliest == latest` — including the
three tests added in `#4793`, whose helper is documented as *"Builds a snapshot whose earliest
pending block is `block_number`"*. The divergence only appears once a snapshot spans several blocks,
which is what `from_previous()` produces on a live chain. We added two tests covering that shape —
one fails before the change, one guards against weakening the bound:

- `canonical_notification_keeps_pending_that_still_tracks_the_tip` — a snapshot spanning more than
  `max_depth` blocks whose latest block is the child of the canonical tip must survive.
- `canonical_notification_drops_pending_that_stopped_tracking_the_tip` — a snapshot whose tip fell
  behind must still be dropped.

The crate's suite passes with the change and both new tests: `119 passed; 0 failed`.

### What else changed for us

Measured in the same two A/B windows described above, so these are like-for-like:

| | reverted | with the change |
|---|---|---|
| flashblock delivery gaps > 500 ms | 26 / 1402 | **3 / 1648** |
| largest gap between flashblocks | 2191 ms | **628 ms** |
| `eth_getLogs` p99 (pending tag) | 27 ms | **20 ms** |

Happy to open a PR with the change and the two tests if that is useful, and equally happy to be told
we have misread something — we only have one node's worth of evidence. Thanks again for the work on
this; the PR description is what let us find it.

mw2000:
## Summary

Pending snapshots were being dropped on a healthy node because staleness was measured from the inherited earliest header, which never advances as the chain moves. That made a live snapshot look older every block, so the node threw away the block it had just built and subscribers missed the rest of it. This measures staleness from the snapshot tip instead, and leaves snapshot width to the existing reconciler rebuild.

Fixes #4916

### DIFF
--- a/crates/execution/flashblocks/README.md
+++ b/crates/execution/flashblocks/README.md
@@ -24,20 +24,21 @@ node's real tip whenever applying updates is slower than receiving them.
 
 ### The invariant
 
-A published snapshot either tracks flashblocks anchored near the node's current canonical tip, or
+A published snapshot either still tracks flashblocks on the node's current canonical tip, or
 there is no snapshot at all.
 
 This matters because consumers execute against `PendingBlocks::canonical_block_number`, the
-canonical block the overlay is layered on. While that block stays close to the tip, callers read
-the node's in-memory canonical state. An overlay stranded on an old anchor forces them onto
-historical state instead, which is slow enough to starve the processor that produced the overlay
-and keep it stranded.
+canonical block the overlay is layered on. A snapshot whose tip has stopped advancing leaves
+that base stranded. Width from an inherited earliest header is a thicker overlay on an older
+base, not a stalled tip; `CanonicalBlockReconciler` rebuilds when that width exceeds
+`max_pending_blocks_depth`.
 
-`max_pending_blocks_depth` (CLI `--max-pending-blocks-depth`, default 3) sets the bound. It is
-measured from the earliest pending block, matching the depth `CanonicalBlockReconciler` already
-uses, so the reconciler never builds a snapshot that the tip checks then discard. The anchor is
-the block below the earliest pending block, so it may sit up to `max_pending_blocks_depth + 1`
-blocks behind the tip.
+`max_pending_blocks_depth` (CLI `--max-pending-blocks-depth`, default 3) sets the bound. Staleness
+is measured from the snapshot tip. `PendingBlocksBuilder::from_previous` keeps the inherited
+earliest header, so a healthy snapshot can grow wider than this bound while `latest` stays on
+the child of the tip. That width is bounded by `CanonicalBlockReconciler`'s `DepthLimitExceeded`
+rebuild, which retains post-canonical flashblocks instead of wiping. Measuring staleness from
+earliest would drop a live snapshot every few blocks and lose the rest of the current block.
 
 Bounding the distance rather than requiring the anchor to equal the tip is deliberate. When
 flashblocks for the next block arrive before the processor has applied the current canonical
@@ -56,10 +57,10 @@ a single expensive update. They use the notified block height directly instead o
 provider.
 
 - A canonical notification records the new height and immediately drops the published snapshot if
-  it is anchored more than `max_pending_blocks_depth` behind it. Because this runs at chain speed,
+  its tip is more than `max_pending_blocks_depth` behind it. Because this runs at chain speed,
   a snapshot cannot stay readable through a long apply. The drop is a compare-and-swap against the
   snapshot that was judged, so a snapshot the processor published concurrently, which is
-  necessarily anchored on a later tip, is left alone.
+  necessarily tracking a later tip, is left alone.
 - A flashblock for a block at or below the last notified canonical height is dropped before it is
   queued, so the queue never accumulates work that could not produce a publishable snapshot.
 
@@ -70,14 +71,15 @@ that lowers the tip does not suppress flashblocks built on the replacement chain
 queued update, because a lagging queue reports a stale height and makes a guard evaluate against a
 chain position the node left long ago.
 
-- Before an update is dispatched, a snapshot anchored more than `max_pending_blocks_depth` blocks
-  behind the tip is dropped. This covers advances the notification path did not report, such as the
-  gap between a canonical notification and the block becoming visible through the provider.
+- Before an update is dispatched, a snapshot whose tip is more than `max_pending_blocks_depth`
+  blocks behind the canonical tip is dropped. This covers advances the notification path did not
+  report, such as the gap between a canonical notification and the block becoming visible through
+  the provider.
 - Flashblocks whose block the node has already canonicalized are skipped before execution. This
   catches payloads that were fresh when queued and went stale while waiting, and cached payloads
   replayed after a canonical block arrives.
 - Every build path publishes through `publish_pending_blocks`, which re-reads the tip and refuses
-  to publish a snapshot that is anchored too far back or no longer extends past the tip. Because it
+  to publish a snapshot whose tip is too far back or no longer extends past the tip. Because it
   is the single funnel, this holds for the reorg and depth-limit rebuilds as well as for ordinary
   sequential appends.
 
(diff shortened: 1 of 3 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

haardikk21 (approved):

PR #5002, opened 2026-09-10
refactor(shadow-indexer-db): make block hashes a typed ShadowHash

wbj-cb:
> Stacked on #4910.

Block hashes are a `String` on the Rust side, which is wider than the thing it holds: any string can be stored, the canonical rendering is a convention held up by call sites, and the only real guard against a mis-spelled hash is a CHECK constraint in the database. This replaces that with a newtype over `B256` that binds as TEXT, so the format contract lives in the type system instead. No schema change and no behaviour change — the column stays TEXT and the JSON the API emits is byte-identical.

- Adds `ShadowHash(B256)` with its own `sqlx::Type`/`Encode`/`Decode`/`PgHasArrayType` impls against TEXT, and uses it for `ShadowBlockRow`, `ShadowCanonicalRef` and `ShadowSummaryRow`.
- Deliberately does **not** use alloy's own sqlx support, which maps `FixedBytes` through `Vec<u8>` onto BYTEA; features are additive, so a bare `B256` would silently start writing bytes the day any crate enables `alloy-primitives/sqlx`, breaking the ETL again with no compile error.
- Collapses hash rendering to one place, which is what the writer and reader depend on agreeing about, since lookups are string equality.
- Simplifies the two call sites that were doing it by hand: the ExEx stores `block.hash()` directly, and `parse_block_id` normalizes by parsing into the type.
- Makes a wrong-width hash unrepresentable rather than a runtime constraint violation, which is how the retention fixture failed earlier.

### DIFF
--- a/crates/execution/shadow-indexer-db/src/hash.rs
+++ b/crates/execution/shadow-indexer-db/src/hash.rs
@@ -1,38 +1,138 @@
-//! Hex formatting for the string mirrors of the BYTEA hash columns.
+//! The block hash type stored by `shadow_blocks`.
 
-use alloy_primitives::hex;
+use std::{fmt, str::FromStr};
 
-/// Formats block hashes for the `hash_hex` and `canonical_hash_hex` columns.
+use alloy_primitives::{B256, hex};
+use sqlx::{
+    Postgres,
+    encode::IsNull,
+    error::BoxDynError,
+    postgres::{PgArgumentBuffer, PgHasArrayType, PgTypeInfo, PgValueRef},
+};
+
+/// A block hash, held as bytes and stored as `0x`-prefixed lowercase hex text.
 ///
-/// The writer stores these and the reader looks rows up by string equality, so the two must agree
-/// on one spelling of a hash or a lookup silently misses instead of failing. That spelling is
-/// `0x`-prefixed lowercase hex: it is what the shadow-metrics JSON API already emits, it is how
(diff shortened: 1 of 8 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #5004, opened 2026-09-10
fix(registrar): fail closed on indeterminate CRL checks (CHAIN-5253)

jackchuma:
Closes [CHAIN-5253](https://linear.app/coinbase/issue/CHAIN-5253/fix-cantina-issue-fail-open-aws-crl-errors-let-newly-revoked).

## Problem

AWS Nitro CRL fetch and parse failures were logged and discarded. `check_chain_against_crls` returned only a list of confirmed revocations, so `SignerManager::check_crls` read an empty list as "clean" and let registration proceed. During a CRL outage, a signer whose CA chain had just been revoked — but whose revocation was not yet persisted in the onchain `CertManager` — could pass the registrar and reach `TEEProverRegistry.registerSigner`.

The periodic recheck did not bound the exposure. A later successful CRL check persisted the certificate revocation but never removed the signer, and discovery adds reachable signer addresses to `active_signers` before CRL validation runs, so orphan cleanup protected it as well. `isValidSigner` only checks registration plus image hash, so the signer stayed valid.

## Changes

**Indeterminate status is preserved.** `crl.rs` now exposes a `CrlSource` trait returning a `CrlChainStatus` that reports confirmed revocations and indeterminate checks in separate lists. A certificate with no CRL distribution point has no applicable CRL and stays clean; anything that fails to fetch or parse lands in `indeterminate` on its own, so a partial failure in a multi-certificate chain can never collapse into an empty (clean) result. `CrlChecker` is the HTTP implementation and owns the fetch timeout and redirect policy that `SignerManager` used to configure inline.

**Registration fails closed.** `SignerManager::check_crls` returns the new `RegistrarError::IndeterminateRevocation` when any applicable CRL could not be completed, before the already-registered check and before any certificate-cache or `registerSigner` transaction.

**Confirmed revocations deregister the signer.** `check_revocation_state` wraps both the onchain sentinel check and the CRL check; when either confirms a revocation it submits `deregisterSigner` rather than waiting for the signer to disappear from discovery. This is best effort and retried on the next cycle. `deregisterSigner` is idempotent onchain, so if the registry read itself is failing we submit anyway instead of leaving a revoked signer valid.

**Metrics distinguish the three outcomes.** New `crl_check_outcome_total{outcome=clean|revoked|indeterminate}` for the per-chain terminal status, plus `crl_indeterminate_certs_detected` mirroring the existing per-certificate `crl_revocations_detected`.

**Removed the legacy `CertManager` module.** `CertManager::check_and_revoke_crls` was a second fail-open CRL implementation, orphaned by 09104a6fd (cut over to hinted registration) and reachable only from its own tests. It revoked against the `NitroEnclaveVerifier` path digest rather than the `CertManager` issuer/serial identity the live flow uses. Leaving it in place would have kept a fail-open copy of exactly the behavior this PR fixes. Its only remaining users — `CertCrlInfo::from_chain`, the `path_digest` field, `onchain_revocation_check_errors`, and `testdata/minimal_attestation.hex` — went with it. Happy to restore it if there is a plan to re-wire it.

## Tests

`crl.rs` covers clean chains, an unfetchable distribution point classified as indeterminate alongside a clean certificate in the same chain, an unparseable chain surfacing as an error rather than an empty status, and CRL serial membership against a fixture CRL. `signer_manager.rs` uses a `mockall` `CrlSource` to cover clean, confirmed-revoked, partial-failure, unparseable-chain, already-registered, and not-registered paths, and asserts each terminal CRL status lands on its own metric label.

## Follow-ups for review

- **Operational consequence:** with `CRL_NITRO_VERIFIER_ADDRESS` set, an AWS CRL outage now halts new registrations rather than silently admitting them. The ticket lists last-known-good CRL caching with a strict freshness window as optional; I did not build it. Worth confirming this trade-off is what we want before rollout.
- **Should CRL checking stay optional?** The ticket asks us to confirm. I left the flag in place and added a startup warning when it is unset, but did not make it mandatory since that would be a deployment-config change.
- `--crl-nitro-verifier-address` is now only read for its presence; the address itself is unused since the hinted cut over. Renaming it to a plain boolean would be clearer but breaks existing deploy configs, so I left it alone.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #5009, opened 2026-09-10
fix(txpool): bound validity predicate deserialization

BrianBland:
## Summary

Fixes an unauthenticated DoS in `base_sendRawTransactionValidity` (CHAIN-5249), with regression tests (CHAIN-5251).

The RPC options and the builder wire `TransactionValidity` deserialized the entire `validity` array before enforcing the 64-predicate limit, so a single request could force the node to parse and allocate tens of thousands of predicates that were ultimately rejected (~14 MiB per request, multiplicative under concurrency). See the vibenet report: [Validity Transaction JSON Decoding DoS](https://bdocs.cbhq.net/d/validity-transactions-json-dos-report).

## Changes

- Added `deserialize_bounded_predicates` (`crates/execution/txpool/src/validity.rs`): a serde sequence visitor that aborts when predicate 65 is encountered, capping allocation at 65 predicates regardless of request body size. The preallocation hint is also capped so a large caller-reported length cannot itself drive a large allocation.
- Applied it to both deserialization sites:
  - `SendRawTransactionValidityOptions.validity` (mempool ingress RPC)
  - `TransactionValidity.validity` (builder wire path, `base_insertValidatedTransaction`)
- Added `#[serde(deny_unknown_fields)]` to `SendRawTransactionValidityOptions` so callers cannot pad the request body with ignored fields.
- Clamped the `--experimental-validity-max-predicates` CLI argument (execution and builder) to `1..=64` so a node cannot be configured above the wire ceiling.

## Why the deserializer bound is the constant, not the configured max

The per-node maximum cannot reach the deserializer: jsonrpsee deserializes params via a static `Deserialize` impl with no runtime context. So the bound must be the compile-time constant `DEFAULT_MAX_VALIDITY_PREDICATES` (64).

To make that constant provably correct rather than a silent behavior change, the CLI argument is now clamped to `1..=64`. A node's configured maximum can therefore never exceed the ceiling the deserializer enforces, and the configured value still applies as a lower limit afterward via `ValidityPredicate::validate_batch`. Previously the argument had no upper bound, so a value above 64 would have silently rejected valid 65-to-max batches mid-deserialization; that value is now rejected at startup instead.

## Tests

Deserialization (8 tests): decoding stops at predicate 65 including a 100k-predicate (DoS-shaped) body; a batch of exactly 64 still decodes; unknown fields on the RPC options are rejected; the bounded visitor applies on the bare sequence and the wire struct, not just the RPC options.

CLI clamp (6 tests, execution + builder): a max above 64 is rejected at startup, zero is rejected, and exactly 64 is accepted.

`cargo test` on `base-execution-txpool`, `base-txpool-rpc`, `base-execution-cli`, `base-builder-cli`, and `base-builder-core --test rpc` is green; `cargo clippy --all-targets` on the changed crates is clean; `cargo +nightly fmt --check` is clean.

## Notes

The report also recommends a lower request-body cap for this endpoint and per-IP rate limits (recs #3, #4); those are infra-level knobs deliberately left out of this PR.

Generated with Toshi

### DIFF
--- a/crates/builder/cli/src/args.rs
+++ b/crates/builder/cli/src/args.rs
@@ -194,9 +194,15 @@ pub struct Args {
     pub enable_experimental_validity_transactions: bool,
 
     /// Maximum validity predicates accepted per experimental transaction.
+    ///
+    /// Capped at [`DEFAULT_MAX_VALIDITY_PREDICATES`], the fixed wire ceiling the
+    /// request deserializer enforces. Values above it can never be honored and
+    /// are rejected at startup rather than silently truncated.
     #[arg(
         long = "builder.experimental-validity-max-predicates",
         default_value_t = DEFAULT_MAX_VALIDITY_PREDICATES,
+        value_parser = clap::builder::RangedU64ValueParser::<usize>::new()
+            .range(1..=DEFAULT_MAX_VALIDITY_PREDICATES as u64),
         requires = "enable_experimental_validity_transactions"
     )]
     pub experimental_validity_max_predicates: usize,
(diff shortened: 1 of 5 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

roger-bai-coinbase (approved):

haardikk21 (approved):

PR #5020, opened 2026-09-11
fix(upgrade-signal): refuse retroactive runtime schedule changes

jackchuma:
## Summary

Addresses audit finding #10 (*"A permitted emergency schedule change can reach nodes after the old upgrade has activated"*), and specifically the auditor's follow-up that the contract-side freeze window added in [base/contracts#424](https://github.com/base/contracts/pull/424) "reduces risk but doesn't prevent stale schedules from being applied. Needs node-side stale-update rejection."

A runtime activation override is not a forward-only setting. `BaseChainSpec::fork()` consults `RuntimeUpgradeRegistry` live for *any* timestamp, so moving or clearing an activation that has already elapsed changes the fork membership of blocks the node has already built or validated. Before this change, `UpgradeSignalRefresher::apply()` validated protocol versions and nothing else — the only freshness gate was an L1 block watermark in `RuntimeUpgradeRegistry::replace_overrides`, which orders reads against each other but says nothing about where the node is.

The result was that a late-observed schedule change was adopted silently, and the damage surfaced far from its cause: as a mismatched state root during replay, reorg handling, or proving.

## Why the freeze window alone can't close it

The freeze window bounds when governance may *author* a change. It does nothing about *observation latency*, which is what actually causes a late apply. Four paths reach the unguarded code, and only the first is what the finding describes:

- an L1 finality stall longer than the freeze window, on the default `finalized` block tag
- an L1 RPC outage spanning an activation — the first successful read afterwards is `Initialized`, and `requires_apply()` is true for it
- an operator calling `admin_refreshUpgradeSignal` after activation
- an L1 reorg to an equal-or-higher block with different contract state, which the `<`-only watermark accepts

## What this does

`UpgradeSignalRefresher::reject_retroactive_change` refuses a schedule that would change the fork rules of a block at or before `now_secs`. An activation change flips the window `[min(current, incoming), max(current, incoming))`, so it is refused exactly when that window has already opened. A cleared activation is unbounded (`u64::MAX`), and the boundary itself counts as settled because a block bearing that timestamp may already exist.

The check lives in `apply()` rather than in the poller, so the live poll path and `admin_refreshUpgradeSignal` on both the consensus and execution sides are covered in one place. Protocol versions are still validated first, so an unsupportable schedule keeps surfacing as `UnsupportedProtocolVersion` and can still fail the node closed.

On refusal: the registry keeps its last coherent schedule, a dedicated `retroactive_rejections_total` counter is raised alongside the existing sticky `apply_failed` gauge, and the poller alarms at `error` on first occurrence then drops to `debug`.

**It never halts.** The divergence predates the read — the node already built on the superseded rules — so halting every node that observes a late governance change would convert a detectable disagreement into an outage. This is the same tradeoff `apply_and_evaluate` already makes for malformed signals.

## Scope and known limits

Worth flagging for review, since it bounds what this fixes. Two changes are deliberately not refused, because the registry cannot see what the node would fall back to:

- **an upgrade with no current override** — the effective activation lives in the startup chain spec or rollup config, which this crate does not read. This is the first live apply after a `startup-apply` boot.
- **an upgrade dropped by a shorter schedule** — the commit removes the override, reverting the fork check to that same unreadable startup value. Refusing here would regress the shrink reconciliation added in #4743.

Both are covered by tests that document the behavior rather than assert it away.

This also uses wall clock rather than the node's real L2 head, which the finding notes can differ under permitted timestamp drift. Replacing wall clock with the head and the registry with the effective fork schedule closes both limits and the drift gap together; `ExecutionUpgradeSignal::current_head` already reads exactly that head for the P2P fork filter, so the plumbing exists on the execution side. I'd rather do that as a follow-up than grow this PR.

## Priority

This is a prerequisite for enabling `runtime-admin` anywhere, not an urgent production fix. The default mode is `metrics-only` and no deployment config sets the mode, so current exposure is zero.

## Testing

Eight new unit tests, covering the finding's own scenario (a clear arriving after the node passed the activation), delay and back-dating of an elapsed activation, the `now_secs` boundary, and the three cases that must keep applying — a still-future reschedule, an unchanged elapsed activation re-offered on every poll, and a first apply with no registry baseline. Plus a poller-level test that a retroactive change continues rather than halts and keeps re-alarming.

`cargo clippy --all-targets` and `cargo test` clean for `base-upgrade-signal`, `base-common-genesis`, `base-execution-cli`, `base-consensus-rpc`, and `base-execution-chainspec`.

### DIFF
--- a/crates/common/genesis/src/chain/upgrade.rs
+++ b/crates/common/genesis/src/chain/upgrade.rs
@@ -332,6 +332,15 @@ impl UpgradeActivation {
     }
 }
 
+impl Display for UpgradeActivation {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Self::Never => f.write_str("not scheduled"),
+            Self::Timestamp(timestamp) => write!(f, "{timestamp}"),
+        }
+    }
+}
+
 /// A target that can receive contract-backed upgrade activation updates.
 ///
 /// Implemented by every schedule destination (rollup config, execution chain spec, runtime
--- a/crates/execution/cli/src/upgrade_signal.rs
+++ b/crates/execution/cli/src/upgrade_signal.rs
@@ -184,7 +184,7 @@ impl ExecutionUpgradeSignal {
         refresher: &UpgradeSignalRefresher,
     ) -> RpcResult<UpgradeSignalApplySummary> {
         match refresher.read_schedule().await {
-            Ok(schedule) => match refresher.apply(&schedule) {
+            Ok(schedule) => match refresher.apply(&schedule, UpgradeSignalDefaults::now_secs()) {
                 Ok(summary) => {
                     UpgradeSignalMetrics::record_apply_success(refresher.metrics_layer, &schedule);
                     Ok(summary)
@@ -589,8 +589,9 @@ mod tests {
         let chain_id = 9_100_004;
         RuntimeUpgradeRegistry::clear_chain(chain_id);
 
-        let summary =
-            runtime_refresher(chain_id).apply(&versioned_schedule(BaseUpgrade::Azul, 42)).unwrap();
+        let summary = runtime_refresher(chain_id)
+            .apply(&versioned_schedule(BaseUpgrade::Azul, 42), 0)
+            .unwrap();
 
         assert_eq!(summary.applied_upgrades, 1);
         assert_eq!(
--- a/crates/utilities/upgrade-signal/README.md
+++ b/crates/utilities/upgrade-signal/README.md
@@ -61,6 +61,29 @@ refresh without mutation. Entries that a specific activation sink does not suppo
 ignored, and contract schedule entries for upgrades newer than the binary knows are logged and
 ignored while mapping the contract schedule.
 
+## Retroactive Schedule Changes
+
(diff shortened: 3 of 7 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #5025, opened 2026-09-11
feat(builder): record build-loop state-read latency

wbj-cb:
A sequencer both builds blocks and validates them, and reth's state-provider instrumentation is wired only into the validation path. Enabling `--engine.state-provider-metrics` there reports the cost of checking a finished block, not the cost of assembling one — so time spent blocked on state reads during building was never measurable.
## What changed

- Both payload builders (flashblocks and native) wrap their state provider in reth's `InstrumentedStateProvider`, reporting under `sync.state_provider` with a `builder` source label to distinguish it from the engine's `engine` label.
- Gated on reth's existing `--engine.state-provider-metrics` rather than a new flag, so the builder and the validation path are instrumented by one switch. Off by default.
- The wrapper is applied outside `CachedStateProvider`, so the recorded latency is the build loop's total per-read cost including cache hits. Inside, a working cache would show no improvement.
- Adds `state_provider_metrics` to `BuilderConfig` and `BaseBuilderConfig`, resolved from the node config when the payload service is built.

## Notes for reviewers

**No production effect as merged.** `--engine.state-provider-metrics` is currently off across the fleet, so this emits nothing until it is enabled. The gate is evaluated once per payload build, not per read.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/builder/cli/src/args.rs
```diff
@@ -404,6 +404,9 @@ impl Args {
         );
 
         Ok(BuilderConfig {
+            // Resolved from reth's --engine.state-provider-metrics when the payload service is
+            // built; this literal only supplies the default.
```
wbj-cb: unnecessary comment

crates/builder/core/src/flashblocks/payload.rs
```diff
@@ -296,6 +297,11 @@ where
                 Some(CachedStateMetrics::zeroed(CachedStateMetricsSource::Builder)),
             ));
         }
+        // Outermost so the recorded latency is the build loop's total cost per state read,
+        // cache hits included -- that total is what a warm shared cache is meant to reduce.
```
wbj-cb: remove comment

crates/execution/payload/src/config.rs
```diff
@@ -409,6 +430,31 @@ impl GasLimitConfig {
     }
 }
 
+/// Wraps a builder's state provider in the decorators the build loop needs.
+pub fn wrap_builder_state_provider<S>(
```
meyer9, line 434: Nice helper fn to make constructing these easier and enforcing instrumentation is after-cache latency.

Wdyt about making this also a StateProvider wrapper like `BuilderStateProvider::new(provider, cache, instrument)`?

wbj-cb: I like it!

meyer9 (approved):

PR #5045, opened 2026-09-15
perf(consensus): Streamline Reset Block Lookups

refcell:
## Summary

Reset forkchoice walks now read block headers and first deposits instead of fetching every transaction body. The authenticated Engine endpoint exposes only the two required read methods, and older endpoints retain a full-block fallback. Genesis/deposit validation and error behavior are preserved, with regression tests covering the client and real IPC endpoint. No startup speedup multiplier is claimed.

### Historical lookup microbenchmark (not end-to-end reset timing)

The reset walk only needs the block header and first deposit. These measurements compare the two RPC access patterns used by the old and new paths:

- **Full block:** `eth_getBlockByHash(hash, true)`.
- **Header + deposit:** `eth_getHeaderByHash(hash)`, then `eth_getTransactionByBlockHashAndIndex(hash, "0x0")`, sequentially, pinned to the same block hash.

| Devnet block | Transactions | Full block median | Header + deposit median | Lookup speedup | JSON response bytes: full → minimal |
| --- | ---: | ---: | ---: | ---: | ---: |
| 3,100 | 5,699 | 32.64 ms | 0.98 ms | 33.4× | 4,261,108 → 2,788 |
| 3,200 | 5,690 | 33.64 ms | 0.71 ms | 47.2× | 4,254,431 → 2,788 |
| 4,400 | 5,742 | 33.74 ms | 1.14 ms | 29.7× | 4,304,852 → 2,790 |
| 6,100 | 12,502 | 78.65 ms | 0.91 ms | 86.4× | 9,375,561 → 2,792 |

Ratios are computed from unrounded medians. Response sizes count JSON bodies only; the minimal size is the sum of both responses.

**Methodology:** Four historical blocks from an ETH-transfer-loaded local devnet, measured on the same profiling binary at baseline revision [`fdca4477bec6`](https://github.com/base/base/commit/fdca4477bec63d0097e670651c65aef9dd7efb64), before this PR. Each block had 12 rounds alternating full/minimal order; the first two rounds were discarded, leaving 10 samples per variant. Requests reused one public loopback HTTP connection with `Accept-Encoding: identity`. Timings include server processing, local transfer, and Python JSON parsing; the minimal path also checks matching header hash/parent/number/timestamp and first-deposit type/input/hash. These are warm-cache repeated reads on a shared 32-logical-CPU host, not a cold-start benchmark. The capacity-test devnet allowed unusually large blocks and unsafe backlog by disabling DA throttling; it is not a representative production-startup workload.

**What this means for reset:** The parent walk is serial, so avoiding serialization, transfer, and decoding of thousands of unrelated transactions can reduce the cost of each historical lookup. It does not reduce the number of blocks walked or eliminate other startup work. Older endpoints taking the full-block fallback do not get this optimization.

**Not measured:** An end-to-end reset/startup A/B with this patch, or time to the first produced block. The real consensus client uses authenticated IPC and Rust deserialization, whereas this experiment used public HTTP and Python. The table demonstrates the narrower lookup opportunity; **it does not establish a 30–86× startup speedup or quantify the integrated PR's reset speedup**.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #5100, opened 2026-09-17
fix(proof): reject unsupported chain IDs in the proof boot path

jackchuma:
### What changed? Why?

The proof guest now rejects L2 chain IDs outside the compiled Base chain set before it reads node-served rollup or L1 configuration. This removes the unknown-chain fallback, where the guest could otherwise execute against configuration supplied by the node.

Built-in Base `ChainConfig` entries serve as the allowlist used by the guest and host. The host validates the configured chain before starting the server and before building an in-process witness, so a misconfigured worker fails early. Local devnet remains supported and continues to use live node-served configuration because it is mutable by design.

### Notes to reviewers

The guest-side validation remains the security boundary; host-side validation only fails misconfigured workers earlier. This does not change public inputs, verifier contracts, or deployed `rollupConfigHash` values.

Boot tests that exercise mutable derivation configuration now use the supported devnet path. The new coverage verifies that unsupported chain IDs are rejected and that every supported chain has the compiled boot dependencies it needs.

### How has it been tested?

- `cargo test -p base-proof -p base-proof-host -p base-proof-zk-utils -p base-proof-tee-nitro-enclave -p base-proof-client`
- `cargo clippy --all-targets` across the affected proof, host, ZK, TEE, and Nitro host crates
- `cargo check -p base-proof --no-default-features`
- `cargo +nightly fmt --all --check`


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/proof/proof/src/boot.rs
```diff
@@ -1080,89 +1089,77 @@ mod tests {
...
         oracle.insert(L2_CLAIM_KEY, B256::repeat_byte(0x33).to_vec());
         oracle.insert(L2_CLAIM_BLOCK_NUMBER_KEY, 40_308_263u64.to_be_bytes().to_vec());
-        oracle.insert(L2_CHAIN_ID_KEY, ORACLE_CHAIN_ID.to_be_bytes().to_vec());
-        oracle.insert(
-            L2_ROLLUP_CONFIG_KEY,
-            serde_json::to_vec(&rollup_config_value).expect("rollup config should serialize"),
-        );
+        // The oracle config agrees with the committed chain ID, so the chain ID binding check
+        // passes and only the supported-chain gate can reject this boot.
+        oracle.insert_rollup_config(UNSUPPORTED_CHAIN_ID, &rollup_config);
```
leopoldjoy, line 1109: Nit: could we omit `L2_ROLLUP_CONFIG_KEY` or make its read fail here so this test proves unsupported chains are rejected before node-served configuration is read?

crates/proof/proof/src/boot.rs
```diff
@@ -1080,89 +1089,77 @@ mod tests {
...
-            err,
-            OracleProviderError::MissingActivationAdminAddress { chain_id: ORACLE_CHAIN_ID }
-        ));
+    /// `load` resolves the activation admin and the L1 config from compiled data and treats a gap in
+    /// either as fatal. Neither gap exists for the chains that ship today, so this pins the
+    /// invariant here rather than letting a newly added chain fail inside the enclave.
+    #[test]
+    fn every_supported_chain_compiles_its_boot_dependencies() {
+        for chain_config in BaseChainConfig::all() {
+            assert!(
```
leopoldjoy, line 1151: Nit: could we require an activation admin only when Beryl is scheduled, matching `BootInfo::load`, rather than preventing a valid pre-Beryl chain from being added?

crates/proof/proof/src/errors.rs
```diff
@@ -110,6 +110,16 @@ pub enum OracleProviderError {
...
+    /// A supported L2 chain resolved an L1 chain without a built-in configuration.
+    ///
+    /// Only the mutable local devnet may take its L1 configuration from the oracle. For every other
+    /// supported chain the L1 chain ID is compiled in, so a missing built-in L1 config means the
+    /// chain configuration itself is inconsistent and the proof must not execute.
+    #[error("Missing built-in L1 chain config for L1 chain ID: {l1_chain_id}")]
+    MissingL1ChainConfig {
+        /// The L1 chain ID that has no built-in configuration.
+        l1_chain_id: u64,
+    },
```
mw2000, line 122: I don't think we'll ever run into this now that we've tightened up what chains we can prove, but good to have as defense in depth

crates/proof/proof/src/boot.rs
mw2000: Since we're talking about removing older hardfork gates, it should be okay to ignore this logic since its more cleaner?

mw2000 (approved):

PR #5159, opened 2026-09-18
fix(tx-manager): keep polling a mined tx past the confirmation timeout

xenoliss:
## Summary

- Keep polling a mined transaction past `confirmation_timeout` until it reaches `num_confirmations`. The receipt poller gave up at the deadline even after recording the transaction as mined. From that point the send loop skips fee bumps (`is_waiting_for_confirmation`), reports no critical error and spawns no new poller, so with `tx_send_timeout = 0` the send never resolved although the transaction had landed. With `num_confirmations > 1` this happens when a transaction is mined late in its poller's lifetime or the L1 tip stalls.
- The deadline now only stops pollers of transactions that are not mined (replaced or dropped hashes). Add `SendState::is_mined` for the per-hash check.
- Stop a poller as soon as its receipt receiver is dropped, which happens when the send loop returns or is cancelled. A mined hash no longer has a deadline, so its poller would otherwise keep querying L1 for a send nobody waits on. Pollers of replaced hashes now also stop right away instead of at the deadline.
- `wait_mined_returns_none_on_timeout` now polls a hash that is never mined instead of a mined, unconfirmable one.

## Testing

- `cargo test -p base-tx-manager --all-features --lib`
- `cargo clippy -p base-tx-manager --all-features --all-targets -- -D warnings`


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

crates/utilities/tx-manager/src/manager.rs
```diff
@@ -1625,7 +1626,10 @@ where
                 }
             }
 
-            if runtime.now() >= deadline {
+            // Give up at the deadline only if the transaction is not mined. Once it is mined
+            // the send loop stops fee bumping and relies on this poller alone, so keep polling
+            // until the confirmation depth is reached.
+            if runtime.now() >= deadline && !send_state.is_mined(tx_hash) {
```
jackchuma, line 1643: Stop polling when the receipt consumer disappears. If `tx_send_timeout` cancels the send, the caller cancels it, or the receipt receiver is otherwise dropped after this hash has been observed mined but before it reaches the required depth, this condition suppresses every subsequent deadline. The detached poller cannot observe the closed receipt channel until it eventually tries to send a confirmed receipt, so it continues issuing RPC requests indefinitely while the manager remains open; repeated timed-out sends can therefore accumulate tasks and RPC load. Pass per-send cancellation or receiver closure into the polling loop so mined transactions stop being watched once nobody can consume the result

jackchuma (approved):

